PR #3, opened 2024-01-25
♻️ ✅ orient the objects and add some tests

hyusap:


api/src/crud.py
```diff
@@ -48,20 +58,26 @@ def create_session(db: Session, user_id: str, session: schemas.SessionCreate) ->
...
+
 def update_session(db: Session, user_id: str, session_id: int, metadata: dict) -> bool:
     # stmt = select(models.Session).where(models.Session.id == session_id).where(models.Session.user_id == user_id)
     # honcho_session = db.scalars(stmt).one_or_none()
     honcho_session = get_session(db, session_id=session_id, user_id=user_id)
     # honcho_session = db.get(models.Session, session_id)
     if honcho_session is None:
         raise ValueError("Session not found or does not belong to user")
-    honcho_session.metadata = metadata
+    honcho_session.session_data = json.dumps(metadata)
```
VVoruganti, line 69: Makes sense since we are treating the model like a string. Do you think we should be exposing the interface with a dictionary or json object then? Can also change the underlying model to instead use `json` or `jsonb`

sdk/honcho/client.py
```diff
@@ -59,54 +69,68 @@ def create_session(self, user_id: str, location_id: str = "default", session_dat
...
+
+class Session:
+    def __init__(
+        self,
+        client: Client,
+        id: int,
+        user_id: str,
+        location_id: str,
+        session_data: dict | str,
+        is_active: bool,
```
VVoruganti, line 97: `is_active` should probably be a read only value or something that is asserted by checking against the api. 

If a developer manually sets the session to active via the object then attempts to add a message it will cause an unhandled error. Can either make it self corrected in the catch statement for the error or let it crash. 

Either way should throw an error

sdk/tests/test_honcho.py
```diff
@@ -0,0 +1,55 @@
...
+    assert created_session.update({"foo": "bar"})
+    retrieved_session = client.get_session("ayush", created_session.id)
+    assert retrieved_session.session_data == {"foo": "bar"}
+
+
+def test_session_deletion():
+    client = Client("http://localhost:8000")
+    created_session = client.create_session("ayush")
+    created_session.delete()
+    retrieved_session = client.get_session("ayush", created_session.id)
```
VVoruganti, line 38: should also `assert` that `created_session.is_active` is `False`

sdk/tests/test_honcho.py
```diff
@@ -0,0 +1,55 @@
+from honcho import Client
+from uuid import uuid1
+
+
+def test_session_creation_retrieval():
+    client = Client("http://localhost:8000")
+    created_session = client.create_session("ayush")
```
VVoruganti, line 7: Don't think you should put your name directly into the code

VVoruganti (approved):
Like the restructure and general clean up of the codebase. There are a few comments left on things we can think about going forward but overall good to move forward with this

api/src/crud.py
hyusap: is json and jsonb database-agnostic?

PR #7, opened 2024-01-31
Hosted demo

VVoruganti:
Several changes that help with making a public demo of the honcho server. 

* DEV-118 - implements application level scoping
* DEV-112 - a default connection string point to https://demo.honcho.dev
* DEV-124 - basic rate limiting of 10/minute


sdk/tests/test_honcho.py
```diff
@@ -1,20 +1,20 @@
 from honcho import Client
 from uuid import uuid1
-
+import pytest
 
 def test_session_creation_retrieval():
-    client = Client("http://localhost:8000")
+    client = Client("test", "http://localhost:8000")
```
hyusap: app id should also be randomly generated for tests

api/src/main.py
```diff
@@ -22,11 +37,12 @@ def get_db():
 # Session Routes
 ########################################################
 
-@app.get("/users/{user_id}/sessions", response_model=list[schemas.Session])
-def get_sessions(user_id: str, location_id: Optional[str] = None, db: Session = Depends(get_db)):
+@router.get("/sessions", response_model=list[schemas.Session])
+def get_sessions(request: Request, app_id: str, user_id: str, location_id: Optional[str] = None, db: Session = Depends(get_db)):
```
hyusap, line 41: where is the request object being used?

sdk/tests/test_honcho.py
hyusap, line 1: add security tests, verify that users can't view other apps sessions without the uuid

api/src/main.py
VVoruganti: It is needed by the rate limiter

PR #8, opened 2024-01-31
✨ asyncification

hyusap:
uses httpx


VVoruganti (changes requested):
This makes it so the client can only be async. 

I was thinking instead exposing a `Client` and an `AsyncClient` with the only difference being that the methods are async vs sync. 

I can imagine some cases where you would not want async such as when you are not calling from an async method and cannot await.

VVoruganti:
What we can do is get rid of the requests dependency altogether and just use httpx, but depending on the client type it uses a `httpx.Client` vs a `httpx.AsyncClient`

sdk/honcho/__init__.py
```diff
@@ -1,2 +1,3 @@
-from .client import Client
+from .client import AsyncClient
+from .syncclient import Client
```
VVoruganti: Nitpick would prefer the name to be `sync_client` just to stick to snake case with python. I know it's technically a class but just not a fan of it for whatever reason. 

Honestly this file could be renamed to `core` with some of the being added elsewhere

sdk/honcho/client.py
```diff
@@ -168,20 +170,20 @@ def update(self, session_data: Dict):
...
         """Delete a session by marking it as inactive"""
         url = f"{self.base_url}/users/{self.user_id}/sessions/{self.id}"
-        response = requests.delete(url)
+        response = await self.client.delete(url)
         self._is_active = False
 
 
-class Message:
-    def __init__(self, session: Session, id: int, is_user: bool, content: str):
+class AsyncMessage:
```
VVoruganti: Can understand with the other two classes because there are specific methods for that, but this is a purely dataclass. Think we should separate it out to a different file and import it in. 

For that matter it doesn't need to have a reference to the session object.

VVoruganti (changes requested):

PR #17, opened 2024-02-14
add test actions and coverage

hyusap:


.github/workflows/run_coverage.yml
```diff
@@ -0,0 +1,51 @@
...
+          python scripts/syncronizer.py
+      - name: Start Server
+        run: |
+          cd api
+          poetry install --no-root
+          poetry run uvicorn src.main:app &
+          sleep 5
+          cd ..
+        env:
+          DATABASE_TYPE: sqlite
```
VVoruganti, line 28: With the addition of the pgvector stuff sqlite may not work. There is a docker-compose in `main` that has a postgresql instance with the pgvector extension in it. May need to do some kind of docker-compose type testing as well for testing

VVoruganti (changes requested):
Change tests to use postgres and I think we are good

PR #29, opened 2024-02-22
add actions again?

hyusap:


.github/workflows/run_coverage.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+      - name: Install poetry
+        run: |
+          pip install poetry
+      - name: Syncify Client
```
VVoruganti: Shouldn't run this - should check with the current state sync the syncified code won't be committed here

.github/workflows/run_coverage.yml
```diff
@@ -0,0 +1,57 @@
...
+        env:
+          DATABASE_TYPE: postgres
+          CONNECTION_URI: postgresql://testuser:testpwd@localhost:5432/honcho
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+      - name: Run Tests
+        run: |
+          cd sdk
+          poetry install
+          poetry run coverage run -m pytest
+          poetry run coverage xml -o coverage.xml
```
VVoruganti: Don't necessarily need to resolve this if it's not possible but would like a more detailed breakdown of the report if possible. The coverage docs give this as an example: https://nedbatchelder.com/files/sample_coverage_html/index.html

VVoruganti (changes requested):

VVoruganti (approved):

PR #32, opened 2024-02-28
Monitoring

VVoruganti:
Added integrations for sentry and open telemetry for the server. 

Closes DEV-204
Closes DEV-198


hyusap (commented):
Should we refactor this to make it optional and only trigger if env variables exist? Thinking for self-hosters, observability wouldn't be first priority, and this would require them to spin up two external services?

VVoruganti:
> Should we refactor this to make it optional and only trigger if env variables exist? Thinking for self-hosters, observability wouldn't be first priority, and this would require them to spin up two external services?

Added some higher level environment variables to completely enable or disable sentry and opentelemetry

PR #63, opened 2024-07-24
Test Suite

VVoruganti:
Add testing and test harness to the honcho API

* Fixes deriver bug
* Fixes various API bugs
* Remove extraneous database hits

fixes dev-310


src/crud.py
hyusap, line 1: file is getting kinda large, should we break this up?

src/deriver.py
hyusap, line 1: can we use some library for the ansi escapes for readability. rich, ansicolor, or colorama should all work. rich if we want more stylizied logging in the future

src/routers/collections.py
hyusap, line 1: is it possible for collection.name to be none or was the old error serving as a type checK?

tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    async def override_get_db():
+        yield db_session
+
+    app.dependency_overrides[get_db] = override_get_db
+    with TestClient(app) as c:
+        yield c
+
+
+@pytest_asyncio.fixture(scope="function")
+async def test_data(db_session):
```
hyusap: shouldn't name non testing function with test_ prefix i dont think it will cause errors but might want to rename to example or something

tests/routes/test_apps.py
```diff
@@ -0,0 +1,60 @@
...
+    response = client.post("/apps", json={"name": name, "metadata": {"key": "value"}})
+    print(response)
+    assert response.status_code == 200
+    data = response.json()
+    assert data["name"] == name
+    assert data["metadata"] == {"key": "value"}
+    assert "id" in data
+
+
+def test_get_or_create_app(client):
```
hyusap, line 20: we should also verify that app will get got if already exists

tests/routes/test_apps.py
```diff
@@ -0,0 +1,60 @@
...
+    assert data["name"] == test_app.name
+    assert data["id"] == str(test_app.id)
+
+
+def test_get_app_by_name(client, test_data):
+    test_app, _ = test_data
+    response = client.get(f"/apps/name/{test_app.name}")
+    assert response.status_code == 200
+    data = response.json()
+    assert data["name"] == test_app.name
```
hyusap, line 60: i like the ergonomics of the example data

tests/routes/test_documents.py
```diff
@@ -0,0 +1,69 @@
...
+    response = client.post(
+        f"/apps/{test_app.id}/users/{test_user.id}/collections/{data['id']}/documents",
+        json={"content": "test_text", "metadata": {}},
+    )
+    assert response.status_code == 200
+    data = response.json()
+    assert data["content"] == "test_text"
+    assert data["metadata"] == {}
+    assert "id" in data
+
```
hyusap, line 20: we have CUD do we need to test R

tests/routes/test_messages.py
```diff
@@ -0,0 +1,80 @@
...
+            "metadata": {"message_key": "message_value"},
+        },
+    )
+    assert response.status_code == 200
+    data = response.json()
+    assert data["content"] == "Test message"
+    assert data["is_user"] is True
+    assert data["metadata"] == {"message_key": "message_value"}
+    assert "id" in data
+
```
hyusap, line 30: testing all the 200s but we should also test the failure modes here, missing data, etc.

tests/routes/test_metamessages.py
```diff
@@ -0,0 +1,101 @@
...
+    response = client.get(
+        f"/apps/{test_app.id}/users/{test_user.id}/sessions/{test_session.id}/metamessages/{test_metamessage.id}?message_id={test_message.id}"
+    )
+    assert response.status_code == 200
+    data = response.json()
+    assert data["message_id"] == str(test_message.id)
+    assert data["content"] == "Test Metamessage"
+    assert data["metadata"] == {}
+    assert data["metamessage_type"] == "test_type"
+
```
hyusap, line 70: CRU do we need D?

hyusap (dismissed):
mostly pedantic stuff. lots of edge cases to think about here. some comments apply to more than one file. good stuff overall!

src/crud.py
VVoruganti: Yeah was thinking about this a bit too. I'm not sure if it makes sense to break up the `crud.py` file or just merge all of the database logic into each of the routes files. Breaking up too much might just make things more unwieldy if we have separate files for the routes and the individual resource crud logic. 

What are your thoughts on that?

Either way I think out of scope for this PR.

tests/routes/test_metamessages.py
VVoruganti: This is another debate we've thought about in terms of what is the right way to think about Honcho. If we're moving towards a world of user owned data then we don't want situations where apps can delete User's data or rewrite history, but the question is if we should be enforcing that now or introduce a change later. 

I can see some arguments in terms of allowing developers to undo mistakes, but the thought process was they could setup staging and production apps within Honcho while they are developing.

src/routers/collections.py
VVoruganti: Yeah for the update it should be possible for it to be blank if they only want to change the metadata and not the name. The previous error was just incorrect. It would basically force an update to a collection to require changing the name/including the previous name.

src/deriver.py
VVoruganti: Agree, can hold off until other changes.`rich` looks to be the most feature rich/used

tests/routes/test_messages.py
VVoruganti: Agree with this. Just going to get this through the door and get the rest of the edge cases in next week's release

VVoruganti:
## [0.0.10] — 2024-07-23

### Added

* Test cases for Storage API
* Sentry tracing and profiling
* Additional Error handling

### Changed

* Document API uses same embedding endpoint as deriver
* CRUD operations use one less database call by removing extra refresh
* Use database for timestampz rather than API
* Pydantic schemas to use modern syntax

### Fixed

* Deriver queue resolution

vintrocode (approved):

PR #71, opened 2024-10-16
Switch from UUIDv4 to NanoID

VVoruganti:
# Overview
There have been several articles discussing the performance detriments that uuid4 present when they are used as the primary key. These include being difficult to index, slowing down insert operations, and not being URL friendly. 

There have also been concerns about API routes becoming very large due to the nested objects in Honcho. To address these concerns we are using a `int8` auto-incrementing value for the primary key of the database tables and then exposing a `public_id` value that maps to a [NanoID](https://github.com/ai/nanoid) 

The end user has no concept of the internal ID at all and it is purely used for internal tracking and performance

## Sources

- https://supabase.com/blog/choosing-a-postgres-primary-key
- https://planetscale.com/blog/why-we-chose-nanoids-for-planetscales-api

# Tracking
Fixes dev-402
Contains changes for dev-435


bLopata (approved):
🔥

PR #84, opened 2024-12-27
Fix Occasional Long Operations

VVoruganti:
Through some debugging it looks like we were not handling connection pool logic properly causing a large amount of stale connections to exist on the database. 

My hypothesis is that occasionally the database would saturate causing it to take up to 15 minutes for a connection to be available to complete even simple database operations like `get_or_create_app`

I used some of the below queries on Postgres to check for connections. 

Check for idle connections older than an hour
```psql
SELECT pid, 
       usename,
       application_name,
       state,
       state_change,
       query
FROM pg_stat_activity 
WHERE 
    state = 'idle' 
    AND state_change < current_timestamp - interval '1 hour'
    AND pid <> pg_backend_pid()
    AND usename = 'postgres';
```

Check for all idle connections

```psql
SELECT 
    datname, 
    usename, 
    application_name,
    state, 
    age(clock_timestamp(), state_change) as state_duration,
    age(clock_timestamp(), query_start) as query_duration,
    query
FROM pg_stat_activity 
WHERE state = 'idle'
ORDER BY state_duration DESC;
```

Stats on database connection capacity 

```psql
SELECT current_setting('max_connections') as max_connections,
       count(*) as current_connections,
       sum(CASE WHEN state = 'active' THEN 1 ELSE 0 END) as active_connections
FROM pg_stat_activity;
```

Originally before the changes I would see upwards of 45 connections, now it is closer to 14. I also do not see any idle connections anymore.


bLopata (approved):
Great stuff. I don't imagine this would require any doc updates, but happy to help with any of that as needed.

PR #85, opened 2024-12-29
Native Support for non-default Postgres Schema

VVoruganti:
https://alembic.sqlalchemy.org/en/latest/cookbook.html#rudimental-schema-level-multi-tenancy-for-postgresql-mysql-other-databases

Commit search path based on cookbook example. Now don't need to commit the whole migration.

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -8,13 +8,15 @@ from alembic import context
 from dotenv import load_dotenv
 from sqlalchemy import engine_from_config, pool, text
 
+# Import your models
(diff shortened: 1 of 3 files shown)


bLopata (approved):
Looks complete. Do we need to update docs to guide users on schema-level multi-tenancy?

VVoruganti:
> Looks complete. Do we need to update docs to guide users on schema-level multi-tenancy?

I don't think so. Schema management should be handled automatically by setting the environment variable you want.

PR #86, opened 2024-12-29
Handle Graceful Shutdown in Deriver

VVoruganti:
Add logic to track sessions and tasks each deriver instance is handling and listen to SIGINT and SIGTERM events. Will 

- Process any currently processing messages
- Delete records of currently managed sessions from `active_queue_sessions`


bLopata (approved):
👍

PR #91, opened 2025-03-20
WIP: add scoped API keys

dr-frmr:


src/main.py
```diff
@@ -30,8 +34,24 @@
 )
 logger = logging.getLogger(__name__)
 
-# Sentry Setup
 
+# JWT Setup
+async def setup_admin_jwt():
+    async for db in get_db():
```
VVoruganti: Should be a `with` statement instead of a `for`

src/models.py
```diff
@@ -236,3 +236,17 @@ class ActiveQueueSession(Base):
...
+
+class Key(Base):
+    __tablename__ = "keys"
+    key: Mapped[str] = mapped_column(TEXT, primary_key=True, index=True, unique=True)
+    revoked: Mapped[bool] = mapped_column(Boolean, default=False)
+    created_at: Mapped[datetime.datetime] = mapped_column(
+        DateTime(timezone=True), index=True, default=func.now()
+    )
+
+    __table_args__ = (CheckConstraint("length(key) <= 1024", name="key_length"),)
```
VVoruganti: Are there any other constraints we can add on this to ensure it is a valid JWT, something similar to how we enforce the `public_id` schema to be nanoids.

src/security.py
```diff
@@ -1,28 +1,230 @@
...
+
+    Names shortened to minimize token size.
+    """
+
+    t: str = datetime.datetime.now().isoformat()
+    ad: Optional[bool] = None
+    ap: Optional[str] = None
+    us: Optional[str] = None
+    se: Optional[str] = None
+    co: Optional[str] = None
```
VVoruganti, line 69: Can we add comments for what each of these correspond too

src/security.py
```diff
@@ -1,28 +1,230 @@
...
+    print(f"\n    Generated secret key: {AUTH_JWT_SECRET}")
+
+
+def rotate_jwt_secret(new_secret: str | None = None) -> str:
+    """
+    Rotate the JWT secret and return the new admin JWT.
+    If no new secret is provided, generate a new one.
+    """
+    global AUTH_JWT_SECRET
+    AUTH_JWT_SECRET = new_secret if new_secret else secrets.token_hex(32)
```
VVoruganti: This value won't actually persist anywhere so if the project shuts down then it won't be able to recover

src/security.py
```diff
@@ -1,28 +1,230 @@
...
+
+    # Clear the cache when rotating secrets
+    clear_api_key_cache()
+
```
VVoruganti: Correct me if I'm wrong but rotating the secret also invalidates all of the existing keys so there should probably be some work done to clear the table.

tests/routes/test_keys.py
```diff
@@ -0,0 +1,173 @@
...
+    new_key = response.json()["key"]
+    original_auth = auth_client.headers["Authorization"]
+    auth_client.headers["Authorization"] = f"Bearer {new_key}"
+
+    # Try to use the new key to access a protected endpoint
+    keys_response = auth_client.get("/v1/keys/")
+    assert keys_response.status_code == 200
+
+    # Restore original auth
+    auth_client.headers["Authorization"] = original_auth
```
VVoruganti: Is there a reason to use the `original_auth` here, the original secrete is gone now so wouldn't the original not work anymore?

src/main.py
```diff
@@ -30,8 +34,24 @@
...
+        token = create_admin_jwt()
+
+        # if admin key is not already in the database, save it
+        key = await crud.get_key(db, token)
+        if key:
+            logger.info("Admin key already exists in database")
+        else:
+            logger.info("Creating new admin key in database")
+            await crud.create_key(db, token)
```
VVoruganti: If we keep starting and stopping the API without saving the JWT to secrets then we'll flood the database with different keys that are invalid. Wondering if there is a way to get around this. 

Thinking it might not make sense to auto-generate this for the user and instead require the developer to set it up initially, that may break the key rotation though

src/security.py
```diff
@@ -1,28 +1,230 @@
...
+        # If not in cache, check the database
+        key = await crud.get_key(db, token)
+        if key:
+            # Cache the result
+            cache_api_key(token, key.revoked)
+            if key.revoked:
+                raise AuthenticationException("Key is revoked")
+        else:
+            # Key not found in DB, cache as not revoked
+            cache_api_key(token, False)
```
VVoruganti: If the key isn't found how do we know that it's valid? Wouldn't this present the possibility for arbitrary keys to bypass the check.

VVoruganti: Instead would move this to the end after we've successfully decoded the key

src/routers/keys.py
```diff
@@ -0,0 +1,107 @@
...
+        raise DisabledException()
+
+    new_jwt = rotate_jwt_secret(new_secret)
+
+    key = await crud.create_key(db, new_jwt)
+
+    return {
+        "key": new_jwt,
+        "created_at": key.created_at,
+    }
```
VVoruganti: Again need to add logic that will revoke all existing keys in this case

src/routers/apps.py
```diff
@@ -4,42 +4,69 @@
...
 )
 
 
-@router.get("/{app_id}", response_model=schemas.App)
+@router.get(
+    "/",
+    response_model=schemas.App,
+    # include_in_schema=False,  XX can use this if desired to skip docs
+)
+async def get_app_from_token(jwt_params: JWTParams = Depends(auth), db=db):
```
VVoruganti: use `require_auth` everywhere instead of `auth`

VVoruganti (changes requested):
Mostly nitpicks, but the main large change to make is the automatic creation of the keys on startup. 

We also need to make sure we're revoking existing keys when we rotate secrets.

tests/routes/test_keys.py
dr-frmr: Yeah, this test should finish by verifying that old auth doesn't work anymore -- will add

PR #92, opened 2025-03-25
Hybrid long-term memory

danibalcells:
Significant change in the way the dialectic and deriver work.

Before:
- Deriver does the heavy lifting. Generates a new user representation asynchronously for each user message that comes in.
- Dialectic pulls latest user representation as additional context to answer queries.
- Problem: poor management of long-term context. Stores all long-term context as natural language in the user rep, and relies on the inference that updates the user rep to preserve everything useful from the past and add new information.

Now:
- Deriver only extracts facts from incoming messages and stores them in an embedding document store.
- Dialectic generates a user representation on the fly that's tailored to the query.
- Two steps in parallel:
  - Short-term: get current chat history and run TOM inference on it.
  - Long-term: based on dialectic query, generate 3-5 queries for the document store. Retrieve 30-50 facts.
- User representation is the concatenation of the TOM inference and the long-term facts

## Summary by CodeRabbit

- **New Features**
  - Enabled non-blocking chat interactions with enhanced fact extraction and duplicate management for faster, smoother conversations.
  - Added new environment variables for integration with various AI and cloud services.
  - Introduced functions for retrieving chat history and managing user representations based on conversation context.
  - Added a utility function for parsing XML-like content from strings.
  - Introduced new functions for Theory of Mind inference and long-term user representation management.

- **Enhancements**
  - Expanded configuration options to support integrations with external AI and cloud services.
  - Improved logging for better monitoring and debugging capabilities, including detailed logs in queue management and processing functions.
  - Streamlined message processing workflow for enhanced efficiency and clarity.
  - Introduced a unified interface for interacting with multiple language model APIs.

- **Bug Fixes**
  - Resolved issues related to duplicate document handling in CRUD operations.
  
- **Chores**
  - Updated logging configurations for testing environments to enhance clarity and control during test executions.


src/crud.py
```diff
@@ -739,6 +738,34 @@ async def create_collection(
...
+    return honcho_collection
+
+
+async def get_or_create_user_protected_collection(
+    db: AsyncSession,
+    app_id: str,
+    user_id: str,
+) -> models.Collection:
+    honcho_collection = await get_collection_by_name(db, app_id, user_id, "honcho")
+    return honcho_collection
```
VVoruganti: This isn't doing any get or create logic. It will just try to get it and may return None. Would also not fall back to using the protected creation method

src/utils/model_client.py
```diff
@@ -0,0 +1,334 @@
...
+                )
+            elif provider == ModelProvider.CEREBRAS:
+                self.api_key = api_key or os.getenv("CEREBRAS_API_KEY")
+                if not self.api_key:
+                    raise ValueError("Cerebras API key is required")
+                self.base_url = base_url or os.getenv("CEREBRAS_BASE_URL", "https://api.cerebras.ai/v1")
+                self.openai_client = AsyncOpenAI(
+                    api_key=self.api_key,
+                    base_url=self.base_url
+                )
```
VVoruganti: Since these are all using OpenAI compatible clients I think we can either keep the hardcoded value (the environment variable should never change it?) or make this a generic that takes a common environment variable called something like `OPENAI_COMPATIBLE_API_URL`

src/utils/model_client.py
```diff
@@ -0,0 +1,334 @@
...
+        # Handle system prompt with caching if enabled
+        if system:
+            if use_caching:
+                params["system"] = [
+                    {
+                        "type": "text",
+                        "text": system,
+                        "cache_control": {"type": "ephemeral"}
+                    }
+                ]
```
VVoruganti, line 169: We talked about this IRL a bit, but it only caches if the prompt is over 1000 tokens, should we have a quick check to see if it is roughly 1000 words

src/utils/model_client.py
```diff
@@ -0,0 +1,334 @@
...
+    ModelProvider.OPENROUTER: "meta-llama/Llama-3.3-70B-Instruct",
+    ModelProvider.CEREBRAS: "llama-3.3-70b",
+}
+
+class Message(Protocol):
+    """Protocol for a message that works with any provider."""
+    role: str
+    content: str
+
+class ModelClient:
```
VVoruganti, line 50: I like this idea of standardizing how we do inference, but let's add unit tests to make sure we don't get out of drift with the standards of how the inference provider's change their APIs

src/main.py
```diff
@@ -92,3 +92,8 @@ async def lifespan(app: FastAPI):
 app.include_router(metamessages.router_user_level, prefix="/v1")
 app.include_router(collections.router, prefix="/v1")
 app.include_router(documents.router, prefix="/v1")
+
+# healthcheck respond with 200 OK
+@app.get("/health")
+async def healthcheck():
+    return "OK"
```
VVoruganti: I would prefer to just use the `/docs` endpoint for this functionality as it's built-in. This will also get a bit tricky with the API key work that ishappening

src/deriver/tom/single_prompt.py
```diff
@@ -0,0 +1,199 @@
...
+            input=messages, model=DEF_MODEL
+        )
+        
+        # Generate the response with caching enabled
+        try:
+            response = await client.generate(
+                messages=messages,
+                system=TOM_SYSTEM_PROMPT,
+                max_tokens=1000,
+                temperature=0,
```
VVoruganti, line 140: temperature 0 seems strange to me, but if that's intention ignore

src/deriver/tom/__init__.py
```diff
@@ -0,0 +1,33 @@
+from .conversational import get_tom_inference_conversational, get_user_representation_conversational
+from .single_prompt import get_tom_inference_single_prompt, get_user_representation_single_prompt
+from .long_term import get_user_representation_long_term
```
VVoruganti, line 3: Let's either add a separate README under tom or at the top level README explaining the differences between these and the different configurations you can run Honcho with

src/deriver/tom/embeddings.py
VVoruganti, line 1: This doesn't look like it's used outside of this file, is this just for local testing? If so please remove

VVoruganti: More a comment on the `LocalEmbeddingStore`

src/deriver/tom/long_term.py
```diff
@@ -0,0 +1,214 @@
...
+# Constants for fact extraction
+FACT_EXTRACTION_PROVIDER = ModelProvider.ANTHROPIC
+FACT_EXTRACTION_MODEL = "claude-3-5-haiku-20241022"
+
+USER_REPRESENTATION_PROVIDER = ModelProvider.CEREBRAS
+USER_REPRESENTATION_MODEL = "llama-3.3-70b"
+
+MAX_FACT_DISTANCE = 0.85
+
+@ai_track("Tom Inference")
```
VVoruganti: This isn't the theory of mind inference anymore: change it to User Representation

src/deriver/queue.py
VVoruganti, line 1: Can you use logger debug statements for this?

src/agent.py
VVoruganti, line 1: Is this new agent architecture compatible with the previous versions of the deriver?

VVoruganti (changes requested):
Have not gotten a chance to dive into the `agent` behavior, but some comments and nitpicks on the code around it. 

Also noticing merge conflicts that we should resolve first

src/crud.py
danibalcells: weird, probably undid some changes before saving or something. added it

src/agent.py
danibalcells: Not compatible with previous versions of the deriver. The previous deriver could technically run without crashing, but it would create user representations that aren't used downstream, and it wouldn't store new facts that are expected downstream by the new dialectic.

PR #96, opened 2025-04-08
Consolidate Get Methods

VVoruganti:


VVoruganti:
@coderabbitai review

src/routers/apps.py
```diff
@@ -17,21 +17,35 @@
     tags=["apps"],
 )
 
-jwt_params = Depends(require_auth(app_id="app_id"))
+# jwt_params = Depends(require_auth(app_id="app_id"))
```
dr-frmr: can be deleted now it seems

tests/routes/test_scoped_api.py
```diff
@@ -335,7 +343,7 @@ def test_get_session_by_id_with_auth(auth_client, sample_data):
 
     # Test with app and user scoped JWT
     response = auth_client.get(
-        f"/v1/apps/{test_app.public_id}/users/{test_user.public_id}/sessions/{session_id}"
+        f"/v1/apps/{test_app.public_id}/users/{test_user.public_id}/sessions?session_id={session_id}"
     )
```
dr-frmr, line 347: in this test and the collections one i'd like to create an array of different JWTs with just collection id, just user id, just app id etc and test they all work on this route

src/crud.py
```diff
@@ -902,22 +919,20 @@ async def get_metamessage(
...
         stmt = stmt.where(models.Metamessage.session_id == session_id)
-
-    # Add message filter if provided
     if message_id is not None:
         stmt = stmt.where(models.Metamessage.message_id == message_id)
 
     result = await db.execute(stmt)
-    return result.scalar_one_or_none()
+    metamessage = result.scalar_one_or_none()
+    return metamessage
```
dr-frmr: can directly return result

PR #98, opened 2025-04-10
Documentation Updates for v1.0.0

VVoruganti:


danibalcells:
Comments so far

**Overview** 
- "Natural language dialectic endpoint" - since "dialectic" in this context is new to most people, we should capitalize it (e.g. "the natural language Dialectic endpoint") or put it between quotes.

**Quickstart**
- "we can make a user" -> "we can create a user" we should stick to our own naming.
- Should we also add metamessage examples? Or show how to use the dialectic endpoint?
- We should probably briefly explain what an application is, either in Quickstart or in  Architecture

PR #107, opened 2025-05-13
Add LRU Cache + Add App_Id / User_Id to resource tables

Rajat-Ahuja1997:
## Summary by CodeRabbit

- **New Features**
  - Added explicit application and user context tracking for sessions, messages, metamessages, collections, and documents.
  - Introduced a thread-safe LRU cache to improve data retrieval efficiency.
- **Performance**
  - Simplified data queries by removing unnecessary joins, enhancing request handling speed.
- **Other**
  - Updated tests to reflect new application and user context requirements for data entities.
  - Added database migration utilities and schema updates to support new app and user associations.

### DIFF
--- a/src/crud.py
+++ b/src/crud.py
@@ -1279,12 +1279,8 @@ async def get_documents(
 ) -> Select:
     stmt = (
         select(models.Document)
-        .join(
-            models.Collection,
-            models.Collection.public_id == models.Document.collection_id,
-        )
-        .where(models.Collection.app_id == app_id)
-        .where(models.Collection.user_id == user_id)
+        .where(models.Document.app_id == app_id)
+        .where(models.Document.user_id == user_id)
         .where(models.Document.collection_id == collection_id)
     )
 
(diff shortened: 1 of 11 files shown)


src/utils/cache.py
```diff
@@ -0,0 +1,85 @@
+from collections import OrderedDict
+import threading
+
+class LRUCache:
+    """
+    A simple thread-safe LRU cache implementation.
```
VVoruganti: We don't need it to be thread safe since we're not multi-threading. What we need is for it to be asyncio safe, using an Asyncio lock/mutex

VVoruganti (changes requested):
Change to database method we discussed

src/utils/cache.py
```diff
@@ -0,0 +1,98 @@
+from collections import OrderedDict
+import asyncio
+
+class LRUCache:
```
Rajat-Ahuja1997, line 4: adding support for an LRUCache. this was originally intended to be used when we were doing crud caching for our operations, but now we have simplified many of our SQL queries, so not this optimization is not necessarily needed right now. keeping the cache here for future use.

src/crud.py
```diff
@@ -361,7 +360,6 @@ async def get_session(
 
 
 async def get_sessions(
-    db: AsyncSession,
```
Rajat-Ahuja1997, line 364: unused bc we use paginate

VVoruganti (changes requested):
Removed extra unnecessary joins from `crud.py` and `agent.py`. Otherwise logic all looks sound

VVoruganti (approved):

PR #111, opened 2025-05-19
add indexes for query documents

Rajat-Ahuja1997:
- Resolves https://linear.app/plastic-labs/issue/DEV-779
- Add HNSW index for embedding column

## Summary by CodeRabbit

- **New Features**
	- Enhanced document search and similarity queries with a new vector-based index for faster and more accurate results.
- **Chores**
	- Updated database schema to include advanced indexing on document embeddings for improved performance.

### DIFF
--- a/migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
+++ b/migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
@@ -37,34 +37,12 @@ def upgrade() -> None:
                 """
             )
         )
+        print(f"HNSW index idx_documents_embedding_hnsw created on {schema}.documents table")
     except Exception as e:
         print(f"Error creating HNSW index idx_documents_embedding_hnsw on {schema}.documents table: {e}")
         
-    print(f"HNSW index idx_documents_embedding_hnsw created on {schema}.documents table")
 
-    # Create B-tree index for filtering
-    print(f"Creating index idx_documents_query_documents_lookup on {schema}.documents table for (app_id, user_id, collection_id)")
-    try:
-        op.create_index(
-            'idx_documents_query_documents_lookup',
-            'documents',
-            ['app_id', 'user_id', 'collection_id'],
-            unique=False,
-            schema=schema
-        )
-    except Exception as e:
-        print(f"Error creating idx_documents_query_documents_lookup index on {schema}.documents table: {e}")
-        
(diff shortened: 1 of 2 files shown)


migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
```diff
@@ -0,0 +1,70 @@
...
+                """
+            )
+        )
+    except Exception as e:
+        print(f"Error creating HNSW index idx_documents_embedding_hnsw on {schema}.documents table: {e}")
+        
+    print(f"HNSW index idx_documents_embedding_hnsw created on {schema}.documents table")
+
+    # Create B-tree index for filtering
+    print(f"Creating index idx_documents_query_documents_lookup on {schema}.documents table for (app_id, user_id, collection_id)")
```
VVoruganti: Does it also need app_id and user_id. My concern is that it would be a much larger index and at the end of the day it should only ever really need to filter against collection_id. The app_id and user_id are more for validation of permissions.

migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
```diff
@@ -0,0 +1,70 @@
...
+    # Create HNSW index on the embedding column for the documents table for cosine distance
+    # Parameters:
+    # - m: max number of connections (edges) per node (default=16)
+    # - ef_construction: size of the candidate list during index construction (default=64)
+    print(f"Creating HNSW index idx_documents_embedding_hnsw on {schema}.documents table")
+    try:
+        op.execute(
+            text(
+                f"""
+                CREATE INDEX CONCURRENTLY idx_documents_embedding_hnsw ON {schema}.documents
```
VVoruganti: CONCURRENTLY may be breaking this

VVoruganti (changes requested):

migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
Rajat-Ahuja1997: removed, the migration worked after removing this

migrations/versions/66e63cf2cf77_add_indexes_to_documents_table.py
Rajat-Ahuja1997: good call, removed this since we have an existing idx for collection_id

VVoruganti (approved):

PR #112, opened 2025-05-20
Break up DB sessions in chat endpoint

Rajat-Ahuja1997:
- The current DB session used throughout the chat endpoint is scoped to the api route, meaning the sessions stay open until the API returns the chat response, even if the connection is not active. 
- Splitting up the database sessions using `tracked_db` allows us to use shorter-lived DB sessions with connections that are returned to the pool once their task is complete.

### DIFF
--- a/src/agent.py
+++ b/src/agent.py
@@ -224,8 +224,8 @@ async def chat(
     facts, tom_inference = await asyncio.gather(
         long_term_task, short_term_task
     )
-    logger.info(f"Retrieved {len(facts)} facts from long-term memory")
-    logger.info(f"TOM inference completed with {len(tom_inference)} characters")
+    logger.debug(f"Retrieved {len(facts)} facts from long-term memory")
+    logger.debug(f"TOM inference completed with {len(tom_inference)} characters")
 
     # Generate a fresh user representation
     logger.debug("Generating user representation")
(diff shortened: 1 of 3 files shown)


src/deriver/tom/long_term.py
```diff
@@ -29,7 +28,6 @@
 async def get_user_representation_long_term(
     chat_history: str,
     session_id: str,
-    embedding_store: CollectionEmbeddingStore,
```
Rajat-Ahuja1997, line 32: unused

VVoruganti (approved):

src/agent.py
```diff
@@ -178,69 +177,68 @@ async def chat(
...
+            user_id=user_id,
+            session_id=session_id,
+            chat_history=chat_history,
+            tom_inference=tom_inference,
+            facts=facts,
+            db=db_rep,
+            message_id=latest_message_id,
+            with_inference=False,
+        )
     logger.debug(
```
VVoruganti, line 222: consolidate to just use the one db session passed into the function

VVoruganti (changes requested):

src/agent.py
```diff
@@ -304,9 +283,16 @@ async def get_long_term_facts(
     async def execute_query(i: int, search_query: str) -> list[str]:
         logger.debug(f"Starting query {i + 1}/{len(search_queries)}: {search_query}")
         query_start = asyncio.get_event_loop().time()
-        facts = await embedding_store.get_relevant_facts(
-            search_query, top_k=10, max_distance=0.85
-        )
+        async with tracked_db("chat.get_long_term_facts") as query_db:
```
Rajat-Ahuja1997: makes sense to move this here

VVoruganti (approved):

PR #124, opened 2025-06-10
Allow manually setting message.created_at when creating messages; include latest deriver trace in dialectic

danibalcells:
Closes DEV-855 and DEV-830


src/crud.py
```diff
@@ -661,14 +661,20 @@ async def create_message(
...
 
-    honcho_message = models.Message(
-        session_id=session_id,
-        is_user=message.is_user,
-        content=message.content,
-        h_metadata=message.metadata,
-        user_id=user_id,
-        app_id=app_id,
-    )
+    message_kwargs = {
```
hyusap, line 664: why do this instead of just passing created_at into models.Message like this?

honcho_message = models.Message(
    session_id=session_id,
    is_user=message.is_user,
    content=message.content,
    h_metadata=message.metadata,
    user_id=user_id,
    app_id=app_id,
    created_at=message.created_at
)

danibalcells: This is a good question. I had to look into it but it turns out that if we do it as you say, if `message.created_at` is `None`, the `Message` model will map that to a `NULL` value for that column, i.e. it won't use the default `func.now()` value for that field because we're providing a value (even if the value is `None`). Therefore, if there's no `created_at` attribute, the correct thing to do is to not pass the argument, so that the model's default gets used.

PR #125, opened 2025-06-11
refactor: enhance logging with rich formatting and modular functions across main and deriver components

hyusap:
image


danibalcells (approved):
Looks good, thanks for adding this! Does the screenshot you added cover all functionality or is there anything else we should be aware of?

hyusap:
should be everything, all of the logs are colored by default now from all libraries

PR #140, opened 2025-06-24
feat: add complex arbitrary filtering on all objects

dr-frmr:
## Summary by CodeRabbit

- **New Features**
  - Introduced advanced recursive filtering for API queries with logical (AND, OR, NOT) and comparison operators, plus wildcard support.
  - Enhanced filtering capabilities on metadata fields for workspaces, peers, sessions, messages, and documents.
  - Standardized pagination across multiple API endpoints for improved consistency.
  - Removed the session active state attribute for clearer session status representation.

- **Bug Fixes**
  - Improved accuracy and consistency of filter application in API endpoints.

- **Tests**
  - Expanded test coverage for advanced filtering, including logical operators, comparison filters, wildcards, and edge cases.
  - Added tests for logical NOT filtering and ensured backward compatibility with filter formats.
  - Updated existing tests to reflect new filter structures and behaviors.


src/utils/filter.py
```diff
@@ -0,0 +1,494 @@
...
+        if condition is not None:
+            conditions.append(condition)
+
+    # Combine all conditions for this field with AND
+    if len(conditions) == 0:
+        return None
+    elif len(conditions) == 1:
+        return conditions[0]
+    else:
+        return and_(*conditions)
```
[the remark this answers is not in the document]
dr-frmr: nah

src/utils/filter.py
```diff
@@ -0,0 +1,494 @@
...
+    Returns:
+        bool: True if the column is a JSONB column, False otherwise
+    """
+    if JSONB is None:
+        # Fallback to string-based detection if JSONB type is not available
+        return (
+            hasattr(column.type, "python_type")
+            and hasattr(column.type, "impl")
+            and "JSONB" in str(column.type)
+        )
```
VVoruganti: This should never be the case. We are just importing a type directly?

src/utils/filter.py
```diff
@@ -0,0 +1,494 @@
...
+    """
+    # Map 'metadata' to 'h_metadata' for all models
+    column_name = "h_metadata" if key == "metadata" else key
+    if column_name == "id":
+        # For Message model, id maps to public_id, for others it maps to name
+        column_name = "public_id" if model_class.__name__ == "Message" else "name"
+    # Only apply _id to _name mapping for fields that are actually foreign keys
+    # and not already mapped to public_id
+    elif column_name.endswith("_id"):
+        column_name = column_name[:-3] + "_name"
```
VVoruganti: Still shaky on this would prefer if we had a more direct mapping with the contracts in code. Maybe add a TODO to optimize in the future

src/utils/filter.py
```diff
@@ -0,0 +1,494 @@
...
+        value: String value to validate as datetime
+
+    Returns:
+        Parsed datetime object if valid, None if invalid
+    """
+    # Strip whitespace and check for obviously malicious content
+    value = value.strip()
+    if not value or "'" in value or ";" in value or "--" in value:
+        logger.warning(f"Potentially malicious datetime string rejected: {value}")
+        return None
```
VVoruganti: This feels brittle, I think SQLAlchemy has some built-in sql injection prevention techniques but would double check that.

src/utils/filter.py
```diff
@@ -0,0 +1,494 @@
...
+                            else:
+                                casted_values.append(val)
+                        if casted_values:
+                            condition = column.in_(casted_values)
+                    else:
+                        condition = column.in_(op_value)
+        elif operator == "contains":
+            if column_name == "h_metadata":
+                # For JSONB columns, use JSONB contains
+                condition = column.contains(op_value)
```
VVoruganti, line 508: Claude Comment:
```
JSONB contains assumptions: The code assumes column.contains(value) will work correctly for all JSONB scenarios, but PostgreSQL's JSONB contains operator has specific semantics that might not match user expectations for nested objects.
```

src/utils/filter.py
VVoruganti: Claude:
```
Lines 395-403: The SQL injection "prevention" is still wrong
Even with JSONB context, this is bad practice. SQLAlchemy's parameterization handles this correctly, and you're blocking legitimate datetime strings.
```

src/schemas.py
```diff
@@ -156,7 +156,7 @@ class SessionCreate(SessionBase):
 
 class SessionGet(SessionBase):
     filter: dict | None = None
-    is_active: bool = False
+    is_active: bool | None = None
```
VVoruganti: Let's consolidate all Get Request Models to have any filters such as `is_active` to be a part of the filter itself. This does not include stuff like `reverse` which is more of a data transformation.

tests/test_advanced_filters.py
VVoruganti, line 1: Want to make sure we do a thorough job of testing all edge cases here. Worked with claude to analyze the code and test cases and have it list untested capabilities:
```
Missing Error Handling Tests:

Invalid datetime strings
Malformed filter dictionaries
Type mismatches in comparisons
Null/None value handling
SQL injection attempts

Missing Type Casting Edge Cases:

Boolean comparisons in JSONB (true/false vs "true"/"false")
Large number handling
Float precision issues
Mixed type comparisons (string vs number)

Missing JSONB Behavior Tests:

Nested object filtering
Array handling in JSONB fields
Complex JSONB contains scenarios
Multiple operators on same JSONB field

Missing Real DateTime Column Tests:

The tests only use date-like strings in metadata, not actual datetime columns
No validation of datetime parsing edge cases
```

src/utils/filter.py
```diff
@@ -0,0 +1,550 @@
...
+
+DISALLOWED_INTERNAL_COLUMNS = [
+    "internal_metadata",
+    ## TODO any others?
+]
+
+
+def apply_filter(
+    stmt: Select[tuple[T]], model_class: type[T], filter: dict[str, Any] | None = None
+) -> Select[tuple[T]]:
```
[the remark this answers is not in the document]
VVoruganti: Would recommend naming it `filters` instead

VVoruganti (approved):

PR #141, opened 2025-06-24
fix: remove hardcoded responses

dr-frmr:
After using Honcho 2.0 I realized that there are some hardcoded strings returned in relatively common cases. We should return either an error or an empty string so that it's easier for developers to recognize this output and route around it in their software. This PR goes with the empty string option -- LMK if an error would be better.


VVoruganti (approved):
Both of these methods are used internally so I don't think there is too much developer risk nor do I think it warrants new exception classes to be made. I think this is fine.

PR #144, opened 2025-06-25
add MessageEmbedding table

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **New Features**
  * Added support for semantic search using message embeddings, with configurable feature flags and token limits.
  * Introduced new API options to control semantic search behavior in peer and session endpoints.
  * Added batch embedding for efficient processing of multiple messages.
  * Created a dedicated table for storing message embeddings to enable fast vector similarity search.
  * Enhanced chat functionality with a new target parameter for specifying chat interaction targets.
  * Introduced a new script to generate embeddings for existing messages lacking embeddings.

* **Configuration**
  * Added new environment variables and settings for embedding token limits and feature enablement.

* **Bug Fixes**
  * Corrected parameter order in collection retrieval functions for consistency.

* **Tests**
  * Added comprehensive tests for message embedding creation, batch embedding, and semantic search feature flag handling.
  * Updated existing tests to use new structured search request format.

* **Documentation**
  * Improved environment variable documentation for embedding configuration.


src/config.py
```diff
@@ -201,6 +201,9 @@ class LLMSettings(HonchoSettings):
     # SUMMARY_SYSTEM_PROMPT_SHORT_FILE: Optional[str] = "prompts/summary_short_system.txt"
     # SUMMARY_SYSTEM_PROMPT_LONG_FILE: Optional[str] = "prompts/summary_long_system.txt"
 
+    # Embed all messages that are sent by peers
+    EMBED_MESSAGES: bool = False
```
Rajat-Ahuja1997, line 205: default false in the config -> we can set this to true for all cloud-hosted honchos via groudon?

src/crud.py
```diff
@@ -35,6 +36,92 @@ async def embed(self, query: str) -> list[float]:
         )
         return response.data[0].embedding
 
+    async def batch_embed(
```
Rajat-Ahuja1997: the idea here is to batch our embedding requests to stay within openai's limits while maintaining message order. at our max `MessageBatchCreate` payload allows for 100 messages with 50k characters each, which is ~1.25million tokens

src/crud.py
```diff
@@ -1196,16 +1300,22 @@
...
+            public_id=generate_nanoid(),
         )
-        for message in messages
-    ]
+        message_objects.append(message_obj)
+
+    id_text_dict = {message.public_id: message.content for message in message_objects}
+    embedding_dict = await embedding_client.batch_embed(id_text_dict)
+    for message_obj in message_objects:
+        message_obj.embedding = embedding_dict[message_obj.public_id]
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: done

src/crud.py
```diff
@@ -35,6 +36,92 @@
...
+        current_batch_tokens = 0
+
+        # OpenAI's official limits for embeddings API:
+        # - Max 300,000 tokens per request (total across all inputs)
+        # - Max 8,192 tokens per individual input
+        # Note: Schema limits ensure we'll never exceed the 2,048 inputs per request limit
+        MAX_TOKENS_PER_REQUEST = 300_000
+        MAX_TOKENS_PER_INPUT = 8192
+
+        for i, (text_id, text) in enumerate(zip(all_ids, all_texts, strict=False)):
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: done

src/crud.py
```diff
@@ -35,6 +36,92 @@ async def embed(self, query: str) -> list[float]:
...
+            # Get pre-computed token count
+            text_tokens = all_token_counts[i]
+
+            # Skip if single text exceeds per-input limit. This is a safety measure but realistically shouldn't really happen
+            if text_tokens > MAX_TOKENS_PER_INPUT:
+                logger.warning(
+                    f"Skipping embedding for {text_id} - {text_tokens} tokens exceeds limit"
+                )
+                continue
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: done, i am initializing these embeddings as empty

src/crud.py
```diff
@@ -35,6 +36,100 @@ async def embed(self, query: str) -> list[float]:
...
+        current_batch_ids: list[str] = []
+        current_batch_texts: list[str] = []
+        current_batch_tokens = 0
+
+        # OpenAI's official limits for embeddings API:
+        # - Max 300,000 tokens per request (total across all inputs)
+        # - Max 8,192 tokens per individual input
+        # Note: Schema limits ensure we'll never exceed the 2,048 inputs per request limit
+        MAX_TOKENS_PER_REQUEST = 300_000
+        MAX_TOKENS_PER_INPUT = 8192
```
VVoruganti: We should probably have some similar logic and validation anywhere we are adding Documents

src/crud.py
```diff
@@ -35,6 +36,92 @@
...
+        all_embeddings: dict[str, list[float]] = {}
+        for batch_ids, batch_texts in batches:
+            response = await self.client.embeddings.create(
+                model="text-embedding-3-small", input=batch_texts
+            )
+
+            # Map embeddings back to IDs
+            for i, embedding_data in enumerate(response.data):
+                original_id = batch_ids[i]
+                all_embeddings[original_id] = embedding_data.embedding
```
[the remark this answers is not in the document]
VVoruganti: In this case we'd want to throw and error and stop the migration in the case that we cannot generate embeddings

src/schemas.py
```diff
@@ -216,6 +216,14 @@ class DocumentUpdate(DocumentBase):
     metadata: dict[str, Any] | None = None
 
 
+class MessageSearchOptions(BaseModel):
+    query: str = Field(..., description="Search query")
+    use_semantic_search: bool = Field(
+        default=False,
+        description="Whether to use semantic search to filter the results",
+    )
```
VVoruganti: Open to discussing this, but it might make sense to make this a `bool | None` flag where we default to using semantic search when the flag is not included, but if semantic search is not enabled we return full-text search. 

The flag then can be used to explicitly change which one is used, but otherwise it will try to use semantic search and fallback to full text

src/crud.py
VVoruganti: These should be in settings as they could change depending on the embedding model we use

Rajat-Ahuja1997: added settings for both of these

Rajat-Ahuja1997: added the validation of 8,192 tokens to the documents embedding call with a try/catch that throws a ValidationException for anything client facing

src/agent.py
```diff
@@ -178,7 +178,7 @@ async def fetch_long_term():
         async with tracked_db("chat.get_collection") as db_embed:
             name = "global_representation" if session_name is None else ""
             collection = await crud.get_or_create_collection(
-                db_embed, workspace_name, peer_name, collection_name=name
+                db_embed, workspace_name, collection_name=name, peer_name=peer_name
             )
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this is already done

migrations/versions/917195d9b5e9_add_messageembedding_table.py
```diff
@@ -0,0 +1,116 @@
...
+
+def upgrade() -> None:
+    op.create_table(
+        "message_embeddings",
+        sa.Column("id", sa.Text(), primary_key=True),
+        sa.Column("content", sa.Text(), nullable=False),
+        sa.Column("embedding", Vector(1536), nullable=False),
+        sa.Column("message_id", sa.Text(), nullable=False),
+        sa.Column("workspace_name", sa.Text(), nullable=False),
+        sa.Column("session_name", sa.Text(), nullable=True),
```
VVoruganti, line 34: The session might be null if it is a message that is just added to the Peer

src/crud.py
```diff
@@ -16,10 +17,13 @@
 
 from . import models, schemas
 from .exceptions import (
+    DisabledException,
     ResourceNotFoundException,
+    ValidationException,
 )
 
 load_dotenv(override=True)
+EMBEDDING_COLLECTION_NAME = "embeddings"
```
VVoruganti: Unused field

src/crud.py
```diff
@@ -30,10 +34,115 @@ def __init__(self, api_key: str | None):
...
+        # - Max 300,000 tokens per request (total across all inputs)
+        # - Max 8,192 tokens per individual input
+        # Note: Schema limits ensure we'll never exceed the 2,048 inputs per request limit
+
+        # Track text IDs that are skipped due to token limits
+        skipped_text_ids: list[str] = []
+
+        for text_id, (text, text_tokens) in id_resource_dict.items():
+            # Skip if single text exceeds per-input limit
+            # TODO (Rajat): Handle chunking of message content that exceeds this limit
```
VVoruganti: Should probably implement this too, I would follow the convention of having some duplicate data between the chunks. So say it's 100 tokens, put the first 60 in the first embedding and the last 60 in the next one to have some overlap

src/crud.py
```diff
@@ -30,10 +34,115 @@ def __init__(self, api_key: str | None):
...
+                input=query, model="text-embedding-3-small"
+            )
+            return response.data[0].embedding
+        except Exception as e:
+            logger.error(
+                f"Failed to generate embedding for query: {query} - {e.__class__.__name__}: {e}"
+            )
+            raise e
+
+    async def batch_embed(
```
VVoruganti: Sorta just a general SWE comment, but I think we could clean up this logic by make it recursively create the batches

src/crud.py
```diff
@@ -1023,58 +1139,105 @@ async def search(
...
+            )
 
+    if should_use_semantic_search:
+        # Generate embedding for the search query
+        try:
+            embedding_query = await embedding_client.embed(query)
+        except ValueError as e:
+            raise ValidationException(
+                f"Query exceeds maximum token limit of {settings.LLM.MAX_EMBEDDING_TOKENS}."
+            ) from e
```
VVoruganti, line 1045: Should add this validation in pydantic on the schemas

src/crud.py
```diff
@@ -1196,16 +1359,43 @@ async def create_messages(
...
-        models.Message(
+    message_objects: list[models.Message] = []
+    for message in messages:
+        message_obj = models.Message(
             session_name=session_name,
             peer_name=message.peer_name,
             content=message.content,
             h_metadata=message.metadata or {},
             workspace_name=workspace_name,
+            public_id=generate_nanoid(),
```
VVoruganti, line 1249: Do we need to do this here, because the DB will run it automatically if not provided.

src/models.py
```diff
@@ -246,10 +227,49 @@ def __repr__(self) -> str:
...
+    content: Mapped[str] = mapped_column(TEXT)
+    embedding: MappedColumn[Any] = mapped_column(Vector(1536))
+    message_id: Mapped[str] = mapped_column(
+        ForeignKey("messages.public_id"), index=True
+    )
+    workspace_name: Mapped[str] = mapped_column(
+        ForeignKey("workspaces.name"), index=True
+    )
+    session_name: Mapped[str | None] = mapped_column(TEXT, index=True, nullable=True)
+    peer_name: Mapped[str | None] = mapped_column(TEXT, index=True, nullable=True)
```
VVoruganti: I don't think thee's a situation where the Peer ID would be null?

src/schemas.py
```diff
@@ -116,14 +116,14 @@ class MessageCreate(MessageBase):
     peer_name: str = Field(alias="peer_id")
     metadata: dict[str, Any] | None = None
 
-    _token_count: int = PrivateAttr(default=0)
+    _encoded_message: list[int] = PrivateAttr(default=[])
```
Rajat-Ahuja1997, line 119: save the encoded message instead of the token_count so we can use it for chunking later

scripts/generate_message_embeddings.py
```diff
@@ -0,0 +1,203 @@
...
+    encoding = tiktoken.get_encoding("cl100k_base")
+
+    # Prepare data for batch embedding with proper token encoding
+    id_resource_dict = {
+        message.public_id: (
+            message.content,
+            encoding.encode(message.content),  # Properly encode the content
+        )
+        for message in messages
+    }
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: the batch_embed only decodes, not encodes. for this script we need to encode before passing in to the embedding client

src/embeddings.py
```diff
@@ -0,0 +1,224 @@
...
+    step_size = max_tokens - overlap_tokens
+
+    return [
+        (
+            encoding.decode(encoded_tokens[i : i + max_tokens]),
+            min(max_tokens, len(encoded_tokens) - i),
+        )
+        for i in range(0, len(encoded_tokens), step_size)
+        if i < len(encoded_tokens)  # Ensure we don't create empty chunks
+    ]
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: my linter enforces this

tests/integration/test_message_embeddings.py
```diff
@@ -0,0 +1,394 @@
...
+        # Verify the embedding was created
+        assert embedding_record is not None
+        assert embedding_record.message_id == created_message.public_id
+        assert embedding_record.content == messages[i].content
+        assert embedding_record.workspace_name == test_workspace.name
+        assert embedding_record.session_name == test_session.name
+        assert embedding_record.peer_name == test_peer.name
+        assert embedding_record.embedding is not None
+        assert len(embedding_record.embedding) > 0
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: don't need trailing commas in the review, i don't think that's the paradigm set in this repo anyway. let me know if that's the case or not

tests/integration/test_message_embeddings.py
```diff
@@ -0,0 +1,394 @@
...
+    assert embedding_record.message_id == created_message.public_id
+    assert embedding_record.content == test_message_content
+    assert embedding_record.workspace_name == test_workspace.name
+    assert (
+        embedding_record.session_name is None
+    )  # Should be None for peer-only messages
+    assert embedding_record.peer_name == test_peer.name
+    assert embedding_record.embedding is not None
+    assert len(embedding_record.embedding) > 0
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: redundant

src/crud.py
```diff
@@ -1196,19 +1238,55 @@
...
+            token_count=len(message._encoded_message),  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: does this expose encoded_message in the schema to the client?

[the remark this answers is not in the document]
Rajat-Ahuja1997: i am fine with exposing this for now --> using the protected method still gives me type checking errors so isn't much of a value-add here

VVoruganti (approved):

PR #168, opened 2025-07-25
feat: webhooks

Rajat-Ahuja1997:


.pre-commit-config.yaml
```diff
@@ -100,20 +100,12 @@ repos:
       # Optional: Coverage check for main app only
       - id: coverage-main
         name: coverage check (main app)
-        entry: bash -c 'uv run coverage run -m pytest tests/ && uv run coverage report --fail-under=80'
+        entry: bash -c 'uv run coverage run -m pytest tests/ && uv run coverage report --fail-under=60'
```
Rajat-Ahuja1997: we are at 61 on main

docs/v2/guides/mcp.mdx
VVoruganti, line 1: Confused why there's a diff here. We already have these changes in main

src/config.py
```diff
@@ -236,6 +236,14 @@ class SummarySettings(HonchoSettings):
     THINKING_BUDGET_TOKENS: Annotated[int, Field(default=512, gt=0, le=2000)] = 512
 
 
+class WebhookSettings(HonchoSettings):
+    model_config = SettingsConfigDict(env_prefix="WEBHOOKS_", extra="ignore")  # pyright: ignore
+
+    PROXY_URL: str | None = None
```
VVoruganti: Doesn't look like this used anywhere except a log statement

src/models.py
```diff
@@ -28,6 +29,12 @@
 
 from .db import Base
 
+
+class WebhookStatus(str, Enum):
+    ENABLED = "enabled"
+    DISABLED = "disabled"
```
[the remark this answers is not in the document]
VVoruganti: +1 on this, looks unused

src/routers/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+async def list_webhook_endpoints(
+    workspace_name: str | None = None,
+    jwt_params: JWTParams = Depends(require_auth()),
+    db: AsyncSession = db,
+):
+    """
+    List all webhook endpoints, optionally filtered by workspace.
+    """
+    # Handle workspace authorization if specified
+    if workspace_name and not jwt_params.ad and jwt_params.w != workspace_name:
```
VVoruganti: Double checking the logic here. 

If not an admin & also the workspace is not specified in the JWT. The JWT could just be an admin level without specifying the workspace so this would fail in that case? Shouldn't this be an `or` statement

src/routers/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+        raise AuthenticationException("Unauthorized access to resource")
+
+    return await crud.list_webhook_endpoints(db, workspace_name)
```
VVoruganti: might want to paginate this

src/routers/webhooks.py
```diff
@@ -0,0 +1,71 @@
...
+    Get or create a webhook endpoint URL.
+    """
+    if (
+        webhook.workspace_name
+        and not jwt_params.ad
+        and jwt_params.w != webhook.workspace_name
+    ):
+        raise AuthenticationException("Unauthorized access to resource")
+
+    return await crud.get_or_create_webhook_endpoint(db, webhook)
```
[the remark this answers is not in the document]
VVoruganti: For this it makes sense to have some common sense limit. I would say let's make a config value for WEBHOOK limit by workspace maybe start at 10.

src/utils/summarizer.py
```diff
@@ -3,11 +3,11 @@
 import logging
 import time
 from enum import Enum
-from typing import TypedDict
 
 from mirascope import llm
 from sqlalchemy import update
 from sqlalchemy.ext.asyncio import AsyncSession
+from typing_extensions import TypedDict
```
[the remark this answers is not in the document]
VVoruganti: We should re-verify the minimum version of python that we're running on here. The `Dockerfile` is using 3.11 so should stick to that standard probably unless we have a strong reason to upgrade

src/webhooks/webhook_delivery.py
```diff
@@ -0,0 +1,180 @@
...
+from src.config import settings
+from src.dependencies import tracked_db
+from src.webhooks.event_emitter import webhook_emitter
+from src.webhooks.events import WebhookEvent
+
+logger = logging.getLogger(__name__)
+
+
+class WebhookDeliveryService:
+    """Service for delivering webhook events to a proxy."""
```
VVoruganti: Mentions a proxy again

src/webhooks/events.py
```diff
@@ -0,0 +1,21 @@
...
+    queue_type: str
+
+
+WebhookPayload = QueueEmptyPayload
+
+
+class WebhookEvent(BaseModel):
+    type: WebhookEventType
+    data: WebhookPayload
+    workspace_name: str
```
VVoruganti: Create WebhookEvent Base `Pydantic` schema that other event types inherit from.

src/main.py
```diff
@@ -101,8 +103,14 @@ def before_send(event: "Event", hint: "Hint") -> "Event | None":
 
 @asynccontextmanager
 async def lifespan(_: FastAPI):
-    yield
-    await engine.dispose()
+    # Start webhook delivery service
+    await webhook_delivery_service.start()
```
VVoruganti: Not sure about having the webhook service being a part of the API itself, how would events be communicated to it from the deriver or any other potential future services.

src/webhooks/event_emitter.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+            await self._event_queue.put(event)
+
+        except Exception as e:
+            logger.error(f"Failed to emit webhook event {event_type}: {e}")
+
+    async def emit_queue_empty(
+        self, workspace_name: str, payload: QueueEmptyPayload
+    ) -> None:
+        await self.emit_event(workspace_name, WebhookEventType.QUEUE_EMPTY, payload)
```
VVoruganti: looking at the codebase don't actually see anywhere that events are emitted

src/webhooks/webhook_delivery.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Args:
+            payload: JSON string of the webhook payload
+
+        Returns:
+            HMAC-SHA256 signature as hex string
+        """
+        auth_secret = os.getenv("AUTH_JWT_SECRET")
+        if not auth_secret:
+            raise ValueError("AUTH_JWT_SECRET not found - cannot sign webhook")
```
VVoruganti: Thinking about this, it would be valid to setup honcho without API auth, but with Webhooks in which we would need a new secret. Might make sense to split it up into a separate `WEBHOOK_SECRET` config value

src/config.py
```diff
@@ -266,6 +274,7 @@ class AppSettings(HonchoSettings):
     DERIVER: DeriverSettings = Field(default_factory=DeriverSettings)
     DIALECTIC: DialecticSettings = Field(default_factory=DialecticSettings)
     SUMMARY: SummarySettings = Field(default_factory=SummarySettings)
+    WEBHOOKS: WebhookSettings = Field(default_factory=WebhookSettings)
```
VVoruganti: need to add this to the section map at the top so it works properly with toml files as well

src/crud/webhook.py
```diff
@@ -0,0 +1,113 @@
...
+        await get_workspace(db, webhook.workspace_name)
+
+    # Check if endpoint already exists
+    stmt = select(models.WebhookEndpoint).where(
+        models.WebhookEndpoint.url == webhook.url,
+    )
+    if webhook.workspace_name:
+        stmt = stmt.where(
+            models.WebhookEndpoint.workspace_name == webhook.workspace_name
+        )
```
VVoruganti: I think this optional workspace scoping is just creating weird behavior. You could have duplicate urls with one being workspace scoped and another not? And then it would send duplicates everywhere. 

The more I'm looking at this the more it makes sense to just enforce workspace level webhooks

src/crud/webhook.py
```diff
@@ -0,0 +1,113 @@
...
+    await db.commit()
+    await db.refresh(webhook_endpoint)
+
+    logger.info(f"Webhook endpoint created: {webhook.url}")
+    return schemas.WebhookEndpoint.model_validate(webhook_endpoint)
+
+
+async def list_webhook_endpoints(
+    db: AsyncSession, workspace_name: str | None = None
+) -> list[schemas.WebhookEndpoint]:
```
VVoruganti: Should paginate for lists

src/webhooks/webhook_delivery.py
Rajat-Ahuja1997: added -> going to add a commit in the groudon PR to set the `WEBHOOK_SECRET` when creating the machine

src/config.py
```diff
@@ -236,6 +237,13 @@ class SummarySettings(HonchoSettings):
     THINKING_BUDGET_TOKENS: Annotated[int, Field(default=512, gt=0, le=2000)] = 512
 
 
+class WebhookSettings(HonchoSettings):
+    model_config = SettingsConfigDict(env_prefix="WEBHOOKS_", extra="ignore")  # pyright: ignore
+
+    WEBHOOK_SECRET: str | None = None  # Must be set if configuring webhooks
+    WORKSPACE_LIMIT: int = 10
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: making it WEBHOOKS.MAX_WORKSPACE_LIMIT

src/deriver/consumer.py
```diff
@@ -1,11 +1,12 @@
 import logging
 from typing import Any
 
+import httpx
 from pydantic import ValidationError
 from rich.console import Console
 
 from .deriver import Deriver
-from .queue_payload import RepresentationPayload, SummaryPayload
+from .queue_payload import RepresentationPayload, SummaryPayload, WebhookQueuePayload
```
VVoruganti: Should be consistent in naming either all be just `XYZPayload` or `XYZQueuePayload`

src/deriver/consumer.py
```diff
@@ -15,25 +16,40 @@
 deriver = Deriver()
 
 
-async def process_item(payload: dict[str, Any]) -> None:
+async def process_item(
+    client: httpx.AsyncClient, task_type: str, payload: dict[str, Any]
```
VVoruganti: a bit skeptical that we need to be passing a `httpx` client around, can't we just make it on the stop or use the `httpx` library directly. I was under the impression that you only need to do this when you want to create a custom client with pre-set flags.

src/deriver/queue_manager.py
```diff
@@ -2,15 +2,16 @@
 import signal
 from asyncio import Task
 from collections.abc import Sequence
-from dataclasses import dataclass
-from datetime import UTC, datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from logging import getLogger
+from sqlite3 import IntegrityError
```
VVoruganti: Shouldn't this be coming from sqlalchemy not sqlite3? We aren't using sqlite anywhere

src/deriver/utils.py
```diff
@@ -0,0 +1,20 @@
...
+        session_id = payload["session_name"] or "None"
+        return f"{task_type}:{session_id}:{sender_name}:{target_name}"
+
+    if task_type == "webhook":
+        workspace_name = payload["workspace_name"]
+        if not workspace_name:
+            raise ValueError("Workspace name is required for webhook tasks")
+        return f"webhook:{workspace_name}"
+
+    raise ValueError(f"Invalid task type: {task_type}")
```
VVoruganti: Have we done any testing to see how this behavior looks in a normal app. Mainly is there fear of the workers being held up with webhooks, or since there's only ever at most one worker handling webhooks for a workspace it should be ok?

src/models.py
```diff
@@ -83,6 +83,7 @@ class Workspace(Base):
     id: Mapped[str] = mapped_column(TEXT, default=generate_nanoid, primary_key=True)
     name: Mapped[str] = mapped_column(TEXT, index=True, unique=True)
     peers = relationship("Peer", back_populates="workspace")
+    webhook_endpoints = relationship("WebhookEndpoint", back_populates="workspace")
```
VVoruganti, line 86: This is the one table that is capitalized, should be consistent

VVoruganti: should that value be the class name or the table name?

src/models.py
```diff
@@ -376,26 +383,36 @@ class ActiveQueueSession(Base):
...
+    id: Mapped[str] = mapped_column(TEXT, default=generate_nanoid, primary_key=True)
+    workspace_name: Mapped[str] = mapped_column(
+        ForeignKey("workspaces.name"), index=True, nullable=False
+    )
+    url: Mapped[str] = mapped_column(TEXT, nullable=False)
+    created_at: Mapped[datetime.datetime] = mapped_column(
+        DateTime(timezone=True), default=func.now()
+    )
+
+    workspace = relationship("Workspace", back_populates="webhook_endpoints")
```
VVoruganti, line 406: Same concern about capitalization

src/routers/webhooks.py
```diff
@@ -0,0 +1,75 @@
...
+async def delete_webhook_endpoint(
+    endpoint_id: Annotated[str, Path(description="Webhook endpoint ID")],
+    jwt_params: JWTParams = Depends(require_auth()),
+    db: AsyncSession = db,
+) -> None:
+    """
+    Delete a specific webhook endpoint.
+    """
+
+    if not jwt_params.ad:
```
VVoruganti: I think a workspace level key should be able to make these requests too

src/webhooks/event_publisher.py
```diff
@@ -0,0 +1,56 @@
...
+        except Exception as e:
+            logger.exception(
+                "Failed to publish webhook event %s: %s",
+                event.type,
+                e,
+            )
+
+
+# Global event publisher instance
+webhook_publisher = WebhookQueuePublisher()
```
VVoruganti: Is there a reason to make this a class, could just be a method we call directly so we can avoid having a singleton that needs to be managed in memory.

src/webhooks/events.py
```diff
@@ -0,0 +1,24 @@
...
+    """Base class for all webhook events."""
+
+    workspace_name: str
+
+
+class QueueEmptyEvent(BaseWebhookEvent):
+    """Webhook event for when a queue becomes empty."""
+
+    type: WebhookEventType = WebhookEventType.QUEUE_EMPTY
+    queue_type: str
```
VVoruganti, line 30: Should make this Literals as well just for safety.

src/webhooks/webhook_delivery.py
```diff
@@ -0,0 +1,105 @@
...
+                    },
+                )
+                if 200 <= response.status_code < 300:
+                    logger.info(
+                        f"Successfully delivered webhook {payload.event_type} to {url}"
+                    )
+                else:
+                    logger.error(
+                        f"Failed delivery for {payload.event_type} to {url}. Status: {response.status_code}"
+                    )
```
VVoruganti: Should be able to make these in parallel or in batches

src/webhooks/webhook_delivery.py
```diff
@@ -0,0 +1,105 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from src.config import settings
+from src.deriver.queue_payload import WebhookQueuePayload
+
+logger = logging.getLogger(__name__)
+
+
+class WebhookDeliveryService:
+    """Service for delivering webhook events from a queue payload."""
```
VVoruganti: Could move the custom `httpx` client into this class if necessary to keep it

tests/routes/test_webhooks.py
VVoruganti, line 1: Add tests for creating more endpoints than the max and for creating duplicate endpoints. Also check that you can add the same endpoint on two different workspaces without error

src/webhooks/event_publisher.py
```diff
@@ -0,0 +1,56 @@
...
+from src.models import QueueItem
+from src.webhooks.events import WebhookEvent
+
+logger = logging.getLogger(__name__)
+
+
+class WebhookQueuePublisher:
+    """Places webhook events onto the database queue."""
+
+    async def publish_event(self, event: WebhookEvent) -> None:
```
VVoruganti: Don't see anything actually using publish_event, can we also implement the emitting the QueueEmpty event

migrations/versions/88b0fb10906f_add_webhooks_table.py
```diff
@@ -0,0 +1,133 @@
...
+        "queue",
+        sa.Column("task_type", sa.TEXT(), nullable=True),
+        schema=schema,
+    )
+
+    op.add_column(
+        "queue",
+        sa.Column("work_unit_key", sa.Text(), nullable=True),
+        schema=schema,
+    )
```
VVoruganti, line 77: Should add a data transformation step that'll fill up the `work_unit_key` column for old columns, otherwise this upgrade would break any existing unprocessed queue items.

VVoruganti (changes requested):
Overall structure looks good a couple of nitpicks on code style and testing. Also implement using the webhooks. 

Most important stuff
- Add more tests
- Implement Queue Empty
- Make the alembic migration more safe

src/webhooks/events.py
VVoruganti, line 1: Createa a `TESTEventType` and an endpoint that let's devs emit a test webhook.

src/deriver/queue_manager.py
Rajat-Ahuja1997: good catch

src/webhooks/event_publisher.py
Rajat-Ahuja1997: made this and the delivery service use a function

src/webhooks/webhook_delivery.py
Rajat-Ahuja1997: added basic parallelization for now

src/webhooks/webhook_delivery.py
Rajat-Ahuja1997: removed

src/deriver/consumer.py
Rajat-Ahuja1997: removed

src/deriver/utils.py
Rajat-Ahuja1997: i think it should be fine considering we only have at most one worker handling webhooks for a workspace. i'm more concerned with a single worker processing webhooks slowly if a workspace has a lot of webhooks, but we can configure a webhook+event_type worker work unit if needed in the future

Rajat-Ahuja1997: i tested this with some webhook calls (test endpoint) along with basic representation deriving and it seemed fine but didn't extensively stress test. but with our current pattern i don't think workers will be held up with webhooks unless a tenant has several workspaces that are all firing tons of webhooks. even then, webhooks should process very quickly

src/routers/webhooks.py
Rajat-Ahuja1997: fixed

src/models.py
Rajat-Ahuja1997: discussed offline, this is good

src/deriver/queue_manager.py
```diff
@@ -145,13 +116,13 @@ async def cleanup(self):
...
+    async def get_available_work_units(self, db: AsyncSession) -> Sequence[str]:
         """
         Get available work units that aren't being processed.
-        Returns a list of WorkUnit objects.
+        Returns a list of work unit keys.
         """
         # Clean up stale work units
-        five_minutes_ago = datetime.now(UTC) - timedelta(
+        five_minutes_ago = datetime.now(timezone.utc) - timedelta(
             minutes=settings.DERIVER.STALE_SESSION_TIMEOUT_MINUTES
```
Rajat-Ahuja1997, line 126: this timeout is currently 5min in testing/prod. we should conside adding an additional check to the stale cleanup query e.g. 
```
.where(
    ~exists(
        select(1)
        .select_from(models.QueueItem)
        .where(
            models.QueueItem.work_unit_key
            == models.ActiveQueueSession.work_unit_key,
            ~models.QueueItem.processed,
        )
    )
)
```
and then we can have an additional check where we delete any active queue sessions where last_update >20min (likely a crashed worker)

migrations/versions/88b0fb10906f_add_webhooks_table.py
```diff
@@ -0,0 +1,175 @@
...
+                        COALESCE(s.workspace_name, q.payload->>'workspace_name', 'None') || ':' ||
+                        COALESCE(q.payload->>'session_name', 'None') || ':' ||
+                        COALESCE(q.payload->>'sender_name', 'None') || ':' ||
+                        COALESCE(q.payload->>'target_name', 'None')
+                END
+            FROM {schema}.sessions s
+            WHERE s.id = q.session_id AND q.processed = false AND q.work_unit_key IS NULL
+            """
+        )
+    )
```
VVoruganti, line 109: Make the queue work_unit column nullable after the transformation

VVoruganti: same thing with the task_type column

src/deriver/utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ParsedWorkUnit(TypedDict):
+    task_type: str
+    workspace_name: str
+    session_name: str | None
+    sender_name: str | None
+    target_name: str | None
+
+
```
VVoruganti, line 11: Potential optimization for the future is to put this all into a pydantic object that can take in the work_unit_key and parse it automatically as well as handle the transform back and forth

Rajat-Ahuja1997: agree

VVoruganti (approved):

PR #171, opened 2025-07-30
added basic config for smithery.ai mcp 

asigdel29:
Created two config files to deploye honcho mcp on smithery.ai

smithery.json – config for the Honcho MCP 
smithery.yaml – config for the Cloudflare Worker proxy via Wrangler

### DIFF
--- a/smithery.json
+++ b/smithery.json
@@ -1,0 +1,36 @@
+{
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "startCommand": {
+    "type": "stdio",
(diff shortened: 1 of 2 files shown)


smithery.json
```diff
@@ -0,0 +1,39 @@
+{
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
```
VVoruganti, line 4: This dockerfile is the one for containerizing Honcho itself, not the MCP server, a different Dockerfile would be necessary for this to work.

PR #194, opened 2025-08-14
refactor: scope tracked_db usage within process_representation_task

Rajat-Ahuja1997:


src/utils/summarizer.py
```diff
@@ -174,7 +174,6 @@ async def create_long_summary(
 
 
 async def summarize_if_needed(
-    db: AsyncSession,
     workspace_name: str,
```
Rajat-Ahuja1997, line 177: refactored to remove this parameter because in the case where should_create_short and should_create long are both true we keep the db connection open but it's not used

VVoruganti (approved):

PR #202, opened 2025-09-03
refactor: replace mirascope with handrolled client

dr-frmr:


dr-frmr:
@coderabbit review

src/utils/clients.py
```diff
@@ -1,392 +1,757 @@
...
+                    )
+                if chunk.choices and chunk.choices[0].finish_reason:
+                    yield HonchoLLMCallStreamChunk(
+                        content="",
+                        is_done=True,
+                        finish_reasons=[chunk.choices[0].finish_reason],
+                    )
+
+
+def retry(
```
Rajat-Ahuja1997: any particular reason we're doing this ourselves instead of continuing to use tenacity?

Rajat-Ahuja1997 (approved):
looks great, just one comment about the retry logic

src/utils/clients.py
dr-frmr: good point, removed in favor of tenacity

PR #209, opened 2025-09-19
feat: get peer cards endpoint

Rajat-Ahuja1997:
## Summary by CodeRabbit

- **New Features**
  - Added an authenticated GET endpoint to retrieve a peer card in a workspace; returns the observer’s own card or a specified target’s card (null when none).

- **Schemas**
  - Introduced a PeerCardResponse model to wrap returned peer card content.

- **Tests**
  - Added tests for empty and populated peer-card retrieval, including async DB setup and verification of observer/self and target views.

### DIFF
--- a/src/routers/peers.py
+++ b/src/routers/peers.py
@@ -239,6 +239,39 @@ async def get_working_representation(
 
 
 @router.post(
+    "/{peer_id}/peer-cards",
+    response_model=schemas.PeerCardResponse,
+    dependencies=[
+        Depends(require_auth(workspace_name="workspace_id", peer_name="peer_id"))
+    ],
+)
+async def get_peer_card(
+    workspace_id: str = Path(..., description="ID of the workspace"),
+    peer_id: str = Path(..., description="ID of the observer peer"),
+    options: schemas.PeerCardGet = Body(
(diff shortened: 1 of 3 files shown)


src/routers/peers.py
```diff
@@ -238,6 +233,34 @@ async def get_working_representation(
         raise ResourceNotFoundException("Peer or session not found") from e
 
 
+@router.get(
+    "/{peer_id}/peer-cards",
```
dr-frmr: feel like this should be singular to match other endpoints like /representation, no?

dr-frmr (commented):
just one suggestion, LGTM otherwise -- why is the route `/peer-cards` and not `/peer-card`?

src/routers/peers.py
Rajat-Ahuja1997: agree, this should be singular. will fix

src/routers/peers.py
```diff
@@ -238,6 +233,34 @@ async def get_working_representation(
         raise ResourceNotFoundException("Peer or session not found") from e
 
 
+@router.get(
+    "/{peer_id}/peer-card",
```
VVoruganti: just name /card

PR #224, opened 2025-10-01
feat: Modify the Summary BaseModel to use public_id of message rather than internal ID

Rajat-Ahuja1997:
- Resolves https://linear.app/plastic-labs/issue/DEV-1093
- Moved over logic from previous PR: https://github.com/plastic-labs/honcho/pull/197

## Summary by CodeRabbit

- **New Features**
  - Summaries and summary payloads now include a public message identifier (exposed as message_id in responses) and require it for summary tasks.

- **Bug Fixes**
  - Added a fallback lookup for missing public message IDs to prevent summary processing failures.

- **Chores**
  - Threaded the public message ID through summary creation and performance tracking; SDK types updated so summary message IDs are strings.

- **Tests**
  - Updated tests to generate and validate public message IDs in summary flows.

### DIFF
--- a/src/schemas.py
+++ b/src/schemas.py
@@ -151,6 +151,7 @@ class MessageUpdate(MessageBase):
 
 
 class Message(MessageBase):
+    id: int = Field(exclude=True)
     public_id: str = Field(serialization_alias="id")
     content: str
     peer_name: str = Field(serialization_alias="peer_id")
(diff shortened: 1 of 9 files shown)


src/schemas.py
```diff
@@ -151,6 +151,7 @@ class MessageUpdate(MessageBase):
 
 
 class Message(MessageBase):
+    id: int = Field(exclude=True)
```
VVoruganti: I don't think this is necessary

VVoruganti: Running locally with this line commented out did not throw any errors for me

VVoruganti:
Can you also add the necessary changes for the SDK

src/schemas.py
Rajat-Ahuja1997: removed

PR #226, opened 2025-10-02
Integrate prometheus into Honcho

Rajat-Ahuja1997:
## Summary by CodeRabbit

- **New Features**
  - Prometheus metrics added: GET /metrics endpoint; automatic tracking for API requests, messages created, peer calls, deriver tasks/tokens, and queue processing.

- **Configuration**
  - Optional Metrics settings added to env/template and example config (ENABLED, NAMESPACE); deployment config updated with metrics listeners for API (8000) and deriver (9090).

- **Dependencies**
  - Added prometheus-client.

- **Utilities**
  - New request/route helpers and logging enhancements to support metrics labeling.

- **Tests**
  - Tests updated to include token estimation parameter.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -162,9 +162,18 @@ app.add_api_route("/metrics", prometheus.metrics, methods=["GET"])
 
 # Global exception handlers
 @app.exception_handler(HonchoException)
-async def honcho_exception_handler(_request: Request, exc: HonchoException):
+async def honcho_exception_handler(request: Request, exc: HonchoException):
     """Handle all Honcho-specific exceptions."""
     logger.error(f"{exc.__class__.__name__}: {exc.detail}", exc_info=exc)
+
+    if prometheus.METRICS_ENABLED and request.url.path != "/metrics":
+        template = get_route_template(request)
+        prometheus.API_REQUESTS.labels(
+            method=request.method,
+            endpoint=template,
+            status_code=str(exc.status_code),
+        ).inc()
+
     return JSONResponse(
         status_code=exc.status_code,
         content={"detail": exc.detail},
@@ -172,9 +181,18 @@ async def honcho_exception_handler(_request: Request, exc: HonchoException):
 
 
 @app.exception_handler(Exception)
-async def global_exception_handler(_request: Request, exc: Exception):
+async def global_exception_handler(request: Request, exc: Exception):
     """Handle all unhandled exceptions."""
     logger.error(f"Unhandled exception: {str(exc)}", exc_info=True)
+
+    if prometheus.METRICS_ENABLED and request.url.path != "/metrics":
+        template = get_route_template(request)
+        prometheus.API_REQUESTS.labels(
+            method=request.method,
+            endpoint=template,
+            status_code="500",
+        ).inc()
+
     if SENTRY_ENABLED:
         sentry_sdk.capture_exception(exc)
     return JSONResponse(
@@ -208,17 +226,5 @@ async def track_request(
             ).inc()
 
         return response
-
-    except Exception:
-        # Track Prometheus error metrics if enabled
-        if prometheus.METRICS_ENABLED and request.url.path != "/metrics":
-            template = get_route_template(request)
-            prometheus.API_REQUESTS.labels(
-                method=request.method,
-                endpoint=template,
-                status_code="500",
-            ).inc()
-        raise
-
     finally:
         request_context.reset(token)


src/config.py
```diff
@@ -271,6 +271,13 @@ class WebhookSettings(HonchoSettings):
     MAX_WORKSPACE_LIMIT: int = 10
 
 
+class MetricsSettings(HonchoSettings):
+    model_config = SettingsConfigDict(env_prefix="METRICS_", extra="ignore")  # pyright: ignore
+
+    ENABLED: bool = False
+    NAMESPACE: str = "honcho"
```
Rajat-Ahuja1997, line 293: i.e. tenant_id

src/main.py
```diff
@@ -191,6 +197,27 @@ async def track_request(
...
+        response = await call_next(request)
+
+        # Track Prometheus metrics if enabled
+        if prometheus.METRICS_ENABLED:
+            prometheus.API_REQUESTS.labels(
+                method=request.method,
+                endpoint=request.url.path,
+                status_code=str(response.status_code),
+            ).inc()
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this is valid, we should normalize the labels before aggregation

src/routers/messages.py
```diff
@@ -72,6 +72,10 @@ async def create_messages_for_session(
 
         # Enqueue all messages in one call
         background_tasks.add_task(enqueue, payloads)
+        prometheus.MESSAGES_CREATED.labels(
+            session_name=session_id,
+            workspace_name=workspace_id,
+        ).inc()
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: resolved in latest commit

src/prometheus.py
```diff
@@ -0,0 +1,56 @@
...
+    "messages_created_total",
+    "Total messages created",
+    ["namespace", "workspace_name", "session_name"],
+)
+
+DIALECTIC_CALLS = NamespacedCounter(
+    "dialectic_calls_total",
+    "Total dialectic calls",
+    ["namespace", "workspace_name", "peer_name", "target_name", "session_name"],
+)
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: will keep workspace_name as tenants should not have many workspaces, otherwise will remove other attributes

src/routers/utils.py
```diff
@@ -0,0 +1,14 @@
+from fastapi import Request
+
+
+def normalize_template_path(path: str) -> str:
```
Rajat-Ahuja1997: originally this was doing some routes mapping logic but now we're just using fastapi's route templates so this file can probably be moved up a level

src/prometheus.py
```diff
@@ -0,0 +1,66 @@
...
+METRICS_ENABLED = settings.METRICS.ENABLED
+
+
+class NamespacedCounter(Counter):
+    """Counter that automatically includes namespace label"""
+
+    def labels(self, **kwargs: str) -> "NamespacedCounter":
+        """Override labels to automatically appends namespace label"""
+        kwargs["namespace"] = settings.METRICS.NAMESPACE
+        return super().labels(**kwargs)
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this feels pretty intuitive already

src/routers/messages.py
```diff
@@ -57,6 +57,16 @@ async def create_messages_for_session(
...
 
+        input_tokens = sum(message.token_count for message in created_messages)
+
+        prometheus.MESSAGES_CREATED.labels(
+            workspace_name=workspace_id,
+        ).inc(len(created_messages))
+
+        prometheus.MESSAGE_INPUT_TOKENS.labels(
+            workspace_name=workspace_id,
+        ).inc(input_tokens)
```
VVoruganti: So don't really want to increment this value here at the beginning, but rather as they are processed by the deriver. In the case of local representation tasks the same message could be processed upwards of 10 times. 

So in queue manager let's just make a count incremented based on the length of the text in the items processed in a batch

src/main.py
```diff
@@ -191,6 +197,29 @@ async def track_request(
...
+        # Track Prometheus metrics if enabled
+        if prometheus.METRICS_ENABLED and request.url.path != "/metrics":
+            template = get_route_template(request)
+            prometheus.API_REQUESTS.labels(
+                method=request.method,
+                endpoint=template,
+                status_code=str(response.status_code),
+            ).inc()
+
+        return response
```
VVoruganti, line 228: I don't have a sense for the latency this could introduce, do you think it's worth making this a background _task

VVoruganti (changes requested):
I think just a miscommunication in the tokens processed category

src/main.py
Rajat-Ahuja1997: per claude the latency, is pretty low:

```
Prometheus Python client is highly optimized:

Counter increment: ~1-5 microseconds
Label resolution: ~0.5-2 microseconds
Total overhead: ~5-10 microseconds (0.005-0.01ms)
```
but we can make this a background task

src/deriver/deriver.py
```diff
@@ -68,6 +69,10 @@ async def critical_analysis_call(
         retry_attempts=3,
     )
 
+    prometheus.DERIVER_TOKENS_PROCESSED.labels(
+        task_type="representation",
+    ).inc(response.output_tokens + estimated_input_tokens)
```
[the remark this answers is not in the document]
VVoruganti: I don't think there's a possibility of getting a negative number from the output tokens of an llm response

src/deriver/deriver.py
```diff
@@ -247,6 +252,8 @@ async def process_representation_tasks_batch(
         observer=observer,
     )
 
+    reasoner.estimated_input_tokens = max(0, int(estimated_input_tokens))
```
VVoruganti: Should just be able to pass this into the constructor, but also the estimated_input tokens is not taking into account the tokens from the call get_session_context_formatted call. Should update this to use a sum of estimated_input_tokens + available_context tokens

src/deriver/queue_manager.py
```diff
@@ -413,6 +413,11 @@ async def process_work_unit(self, work_unit_key: str, worker_id: str) -> None:
                                     observed=work_unit.observed,
                                 )
                             )
+
+                            prometheus.DERIVER_TASKS_COMPLETED.labels(
+                                workspace_name=work_unit.workspace_name,
+                                task_type=work_unit.task_type,
+                            ).inc()
```
VVoruganti: I'm not sure I really get the point of this metric if it's only counting when we remove a work_unit_key from active queue sessions. If there a 100 tasks for a work_unit we only fire this after all 100 are finished and then it would only be incremented once.

would maybe move to mark_messages_as_processed function and take into account the length of the `messages` parameter as the number to increment by

src/main.py
VVoruganti: That sounds fine, no need to hyper optimize just yet

src/prometheus.py
```diff
@@ -0,0 +1,78 @@
...
+
+
```
VVoruganti, line 34: Can you had docstring type comments explaining what each metric is tracking and when it gets incremented

PR #228, opened 2025-10-06
Add QueueItem cleanup logic

Rajat-Ahuja1997:
## Summary by CodeRabbit

* New Features
  * Per-item error visibility for queue entries and timestamps to aid tracking.
  * Automatic background maintenance cleans up processed and old errored items.
  * Centralized error handling with clearer logs and more reliable processing.

* Configuration
  * Added QUEUE_ERROR_RETENTION_SECONDS (default: 30 days) to control retention of errored items.

* Performance/Stability
  * Graceful startup/shutdown of maintenance tasks to reduce interruptions.

* Database/Migrations
  * Added error and created_at fields to queue items with indexing and backfill migration.


src/models.py
```diff
@@ -376,6 +376,10 @@ class QueueItem(Base):
     task_type: Mapped[TaskType] = mapped_column(TEXT, nullable=False)
     payload: Mapped[dict[str, Any]] = mapped_column(JSONB, nullable=False)
     processed: Mapped[bool] = mapped_column(Boolean, default=False)
```
Rajat-Ahuja1997, line 398: we could also just have one field `status` and have the options be `pending`, `processed`, `errored`

migrations/versions/ddf7b217ca3e_add_error_field_to_queueitem.py
```diff
@@ -0,0 +1,89 @@
...
+        "queue",
+        sa.Column("error", sa.TEXT(), nullable=True),
+        schema=schema,
+    )
+
+    op.add_column(
+        "queue",
+        sa.Column("created_at", sa.DateTime(timezone=True), nullable=True),
+        schema=schema,
+    )
```
VVoruganti, line 43: Only thing is I'm unsure if this properly sets the default value to `func.now()`

src/deriver/queue_manager.py
```diff
@@ -329,16 +329,28 @@ async def process_work_unit(self, work_unit_key: str, worker_id: str) -> None:
...
+                                await self.mark_messages_as_processed(
+                                    items_to_process, work_unit_key
+                                )
+                                message_count += len(items_to_process)
+                            except Exception as e:
+                                # Mark the batch as failed
+                                error_msg = f"{e.__class__.__name__}: {str(e)}"
+                                await self.mark_messages_as_errored(
+                                    items_to_process, work_unit_key, error_msg
+                                )
```
VVoruganti: A little hesitant to throw away the entire batch, what do you think about instead just throwing away the top level item and then trying again with the next one?

src/main.py
```diff
@@ -100,10 +109,31 @@ def before_send(event: "Event", hint: "Hint") -> "Event | None":
...
+                QueueItem.processed
+                & (
+                    QueueItem.error.is_(None)
+                    | (
+                        QueueItem.error.is_not(None)
+                        & (QueueItem.created_at < thirty_days_ago)
+                    )
+                )
+            )
+        )
```
VVoruganti: Just as a function of separation of concerns could we make this a part of the queue manager instead?

src/models.py
VVoruganti: It probably doesn't make a huge difference, I'd imagine a boolean is able to index better than a string that doesn't use an enum table

src/main.py
```diff
@@ -100,10 +109,31 @@ def before_send(event: "Event", hint: "Hint") -> "Event | None":
     )
 
 
+async def cleanup_queue_items():
+    async with tracked_db("cleanup_queue_items") as db:
+        thirty_days_ago = datetime.now(timezone.utc) - timedelta(days=30)
```
VVoruganti: Could be a config value that takes milliseconds or seconds as an input instead. Max configurability depending on the load of the tenant

VVoruganti (changes requested):
Logic looks good just want to logic to be consolidated to the Queue rather than having the API manage the Queue

migrations/versions/ddf7b217ca3e_add_error_field_to_queueitem.py
Rajat-Ahuja1997: good catch, fixed

src/deriver/queue_manager.py
Rajat-Ahuja1997: changed the logic to only mark the first queue item in a batch as errored

src/deriver/queue_manager.py
```diff
@@ -288,6 +303,93 @@ async def polling_loop(self) -> None:
...
+                    & (
+                        models.QueueItem.error.is_(None)
+                        | (
+                            models.QueueItem.error.is_not(None)
+                            & (models.QueueItem.created_at < error_cutoff)
+                        )
+                    )
+                )
+            )
+            await db.commit()
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: current set-up is fine; errored queue items are just for debugging and we typically don't expect a large delta between created at and processing time

migrations/versions/ddf7b217ca3e_add_error_field_to_queueitem.py
```diff
@@ -0,0 +1,94 @@
...
+from alembic import op
+
+from migrations.utils import column_exists, index_exists
+from src.config import settings
+
+# revision identifiers, used by Alembic.
+revision: str = "ddf7b217ca3e"
+down_revision: str | None = "88b0fb10906f"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
```
VVoruganti: Update the revision to point to the latest migration in repo, currently this is creating a branchpoint.

migrations/versions/ddf7b217ca3e_add_error_field_to_queueitem.py
```diff
@@ -0,0 +1,94 @@
...
+
+schema = settings.DB.SCHEMA
+
```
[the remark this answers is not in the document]
VVoruganti: I think this is fair, we have the function anyways

src/deriver/queue_manager.py
```diff
@@ -308,6 +323,93 @@ async def polling_loop(self) -> None:
...
+                    # Timeout means it's time for next cleanup
+                    pass
+        except asyncio.CancelledError:
+            logger.debug("Maintenance loop cancelled")
+            raise
+
+    async def _handle_processing_error(
+        self,
+        error: Exception,
+        messages: list[QueueItem],
```
VVoruganti: nitpick use items or work_units here move away from referrring to things as messages

VVoruganti (commented):
Two tiny nitpicks, one for naming and the one remaining code rabbit comment seems valid. Otherwise good to go

VVoruganti (approved):

PR #231, opened 2025-10-09
Add updated get_context to SDKs

dr-frmr:
## Summary by CodeRabbit

- **New Features**
  - Context retrieval supports peer targeting and perspective, plus an optional last-user-message; peer target is required when perspective or last-user-message is provided.
  - Session context exposes peer representation and peer card; OpenAI/Anthropic context builders prepend system messages for these and summaries.
  - Session.getContext supports both options-object and positional overloads.

- **Chores**
  - Bumped core SDK dependency to 1.5.0 for Python and TypeScript.

- **Tests/Tooling**
  - Bench harness masks sensitive values by key name.
  - Tests use string message IDs.

### DIFF
--- a/sdks/typescript/src/validation.ts
+++ b/sdks/typescript/src/validation.ts
@@ -148,7 +148,12 @@ export const ContextParamsSchema = z
       .number()
       .positive('Token limit must be a positive number')
       .optional(),
-    lastUserMessage: z.union([z.string(), MessageSchema]).optional(),
+    lastUserMessage: z
+      .union([
+        z.string().min(1, 'Last user message must be a non-empty string'),
+        MessageSchema,
+      ])
+      .optional(),
     peerTarget: PeerIdSchema.optional(),
     peerPerspective: PeerIdSchema.optional(),
   })


sdks/python/src/honcho/session.py
```diff
@@ -449,7 +467,13 @@ def get_context(
...
 
         return SessionContext(
-            session_id=self.id, messages=context.messages, summary=session_summary
+            session_id=self.id,
+            messages=context.messages,
+            summary=session_summary,
+            peer_representation=str(context.peer_representation)
+            if context.peer_representation
+            else None,
+            peer_card=context.peer_card,
```
VVoruganti, line 494: Would also be None if there is no peer_representation

sdks/python/src/honcho/session.py
```diff
@@ -420,6 +432,9 @@ def get_context(
             summary: Whether to include summary information
             tokens: Maximum number of tokens to include in the context. Will default
             to Honcho server configuration if not provided.
+            last_user_message: The most recent message, used to fetch semantically relevant observations and returned as part of the context object
+            peer_target: The target of the perspective. If given without `peer_perspective`, will get the Honcho-level representation and peer card for this peer. If given with `peer_perspective`, will get the representation and card for this peer *from the perspective of that peer*.
```
VVoruganti: The wording on this was a bit confusing for me, get it now after looking at the code but I think the use of `this` and `that` when referring to peers can be confusing. Maybe be explicit and refer to them as the target peer and perspective peer?

sdks/python/src/honcho/session.py
```diff
@@ -407,6 +407,18 @@ def get_context(
         tokens: int | None = Field(
             None, gt=0, description="Maximum number of tokens to include in the context"
         ),
+        last_user_message: str | None = Field(
+            None,
+            description="The most recent message, used to fetch semantically relevant observations and returned as part of the context object",
+        ),
```
VVoruganti: Should accept a `Message` type as well and clarify that it expects a message_id.  Also clarify that this only does something if peer_target is included. 

Could also see an argument to change the order based on dependent values. peer_target -> last_user_message -> peer_perspective

sdks/python/src/honcho/session_context.py
```diff
@@ -118,12 +136,27 @@ def to_openai(
...
+            }
+            messages.insert(0, peer_representation_message)
+
+        if self.peer_card:
+            peer_card_message = {
+                "role": "system",
+                "content": f"<peer_card>{self.peer_card}</peer_card>",
+            }
+            messages.insert(0, peer_card_message)
+
```
VVoruganti, line 153: nitpick just a bit of a messy implementation, like the typescript version better where we extend two lists rather than trying to push into the beginning of a list, however idk if that's an O(n) or O(1) operation in python

sdks/typescript/src/session_context.ts
```diff
@@ -136,18 +152,36 @@ export class SessionContext {
...
+        content: `<peer_card>${this.peerCard}</peer_card>`,
+      })
+    }
+
+    if (this.summary) {
+      systemMessages.push({
+        role: 'system',
+        content: `<summary>${this.summary.content}</summary>`,
+      })
+    }
```
VVoruganti, line 182: in consistency between order of system messages between python and typescript clients

src/crud/document.py
```diff
@@ -107,6 +107,11 @@ async def create_documents(
     """
     Create multiple documents with NO duplicate detection.
 
+    If duplicate_threshold is provided, we use the embedding to check for
+    duplicates within the threshold. If a duplicate is found, instead of
+    saving a new document, we increment the times_derived field in the
+    `internal_metadata` of the duplicate document.
+
```
VVoruganti: There is no duplicate threshold parameter

src/deriver/queue_manager.py
```diff
@@ -106,6 +106,8 @@ async def initialize(self) -> None:
         """Setup signal handlers, initialize client, and start the main polling loop"""
         logger.debug(f"Initializing QueueManager with {self.workers} workers")
 
+        set_dream_scheduler(self.dream_scheduler)
```
VVoruganti: Already done in the __init__ function, I believe this is redundant unless we want  to move all of the logic here so it only starts when re run `manager.initialize()`

dr-frmr: removed

src/crud/document.py
dr-frmr: removed

sdks/python/src/honcho/session.py
dr-frmr: reworded, could still be better

sdks/python/src/honcho/session.py
dr-frmr: added and reordered

sdks/python/src/honcho/session_context.py
dr-frmr: moved to same method as ts library, consistent ordering now:
representation, peer card, summary, [regular messages]

sdks/python/src/honcho/async_client/session.py
```diff
@@ -450,11 +465,27 @@ async def get_context(
...
             session_id=self.id,
             workspace_id=self.workspace_id,
             tokens=tokens if tokens is not None else omit,
             summary=summary,
+            last_message=last_user_message_id
+            if last_user_message_id is not None
+            else omit,
+            peer_target=peer_target if peer_target is not None else omit,
+            peer_perspective=peer_perspective if peer_perspective is not None else omit,
         )
```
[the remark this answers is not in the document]
VVoruganti: We have validation logic at the beginning of the function now

sdks/python/src/honcho/session.py
```diff
@@ -430,11 +445,27 @@ def get_context(
...
             session_id=self.id,
             workspace_id=self.workspace_id,
             tokens=tokens if tokens is not None else omit,
             summary=summary,
+            last_message=last_user_message_id
+            if last_user_message_id is not None
+            else omit,
+            peer_target=peer_target if peer_target is not None else omit,
+            peer_perspective=peer_perspective if peer_perspective is not None else omit,
         )
```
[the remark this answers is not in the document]
VVoruganti: there's validation logic at the beginning of the function now

sdks/python/src/honcho/async_client/session.py
```diff
@@ -469,7 +500,13 @@ async def get_context(
...
         return SessionContext(
-            session_id=self.id, messages=context.messages, summary=session_summary
+            session_id=self.id,
+            messages=context.messages,
+            summary=session_summary,
+            peer_representation=str(context.peer_representation)
+            if context.peer_representation
+            else None,
+            peer_card=context.peer_card,
         )
```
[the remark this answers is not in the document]
VVoruganti: @dr-frmr Can you test this locally and see what the actual output looks to see if it's being formatted properly

sdks/typescript/src/validation.ts
```diff
@@ -123,16 +124,51 @@ export const ChatQuerySchema = z.object({
...
+  .object({
+    summary: z.boolean().optional(),
+    tokens: z
+      .number()
+      .positive('Token limit must be a positive number')
+      .optional(),
+    lastUserMessage: z.union([z.string(), MessageSchema]).optional(),
+    peerTarget: z.string().optional(),
+    peerPerspective: z.string().optional(),
+  })
```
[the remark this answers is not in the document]
VVoruganti: The same way we allow someone to specify to use a Message object we should probably also allow for Peer objects here. @dr-frmr

sdks/typescript/src/session.ts
```diff
@@ -465,22 +475,39 @@ export class Session {
...
-    return new SessionContext(this.id, context.messages, summary)
+    return new SessionContext(
+      this.id,
+      context.messages,
+      summary,
+      context.peer_representation
+        ? JSON.stringify(context.peer_representation)
+        : null,
+      context.peer_card ?? null
+    )
```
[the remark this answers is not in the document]
VVoruganti: @dr-frmr This seems pretty valid

sdks/typescript/src/validation.ts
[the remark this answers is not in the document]
dr-frmr: I tried but it introduces a ton of circular import errors and we can't use `@validate_call` with it if we do

sdks/typescript/src/session.ts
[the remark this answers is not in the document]
dr-frmr: fixed

PR #246, opened 2025-10-27
API & SDK improvements

dr-frmr:
Notes:
- do not take out of draft state until stainless 1.6.0 SDK is in prerelease with these changes
- a huge % of the diff here is just more generated SDK tests

## Summary by CodeRabbit

* **New Features**
  * Structured Representation model (diff/merge/format), semantic-search options, peer-card setting, observations API (list/query/delete), trigger-dream endpoint, add_messages now returns created messages, workspace/peer/session getConfig/setConfig/refresh.

* **Refactor**
  * Metadata/config caching across workspace/peer/session; session deletion performed in background; representation/context APIs accept session-scoped and semantic-search parameters; hierarchical per-session/message configuration resolution.

* **Tests**
  * Large expansion: representations, caching, sessions, observations, search, migrations, unified end-to-end scenarios.

* **Chores**
  * SDK dependency updates, docs and migration updates; example script removed.


sdks/python/pyproject.toml
VVoruganti, line 1: Note to generate production SDK

sdks/python/src/honcho/async_client/client.py
```diff
@@ -283,7 +294,13 @@ async def get_sessions(
...
+            lambda session: AsyncSession(
+                session.id,
+                self.workspace_id,
+                self._client,
+                metadata=session.metadata,
+                config=session.configuration,
+            ),
         )
 
     async def get_metadata(self) -> dict[str, object]:
```
VVoruganti, line 314: Would maybe call this `refresh` just to make it clear what it's doing.

sdks/python/src/honcho/async_client/session.py
```diff
@@ -418,6 +444,51 @@ async def set_metadata(
...
+        Args:
+            configuration: A dictionary of configuration to associate with this session.
+                          Keys must be strings, values can be any JSON-serializable type
+        """
+        await self._client.workspaces.sessions.update(
+            session_id=self.id,
+            workspace_id=self.workspace_id,
+            configuration=configuration,
+        )
+        self.configuration = configuration
```
VVoruganti: Also would just make this a combined refresh call? Open to discussing though

sdks/python/src/honcho/session.py
```diff
@@ -93,16 +104,21 @@ def __init__(
...
                 id=session_id,
                 configuration=config if config is not None else omit,
                 metadata=metadata if metadata is not None else omit,
             )
+            # Update cached values with API response
+            self.metadata = session_data.metadata
+            self.configuration = session_data.configuration
 
     def add_peers(
         self,
```
VVoruganti, line 126: Should also have this method return the Peer objects of the peers that were added or refresh the ones passed in. Think just returning them would be better in case they have updated config and metadata values. Can also rely on the api to handle default values more

sdks/typescript/src/session.ts
```diff
@@ -112,18 +117,44 @@ export class Session {
...
+   */
+  metadata?: Record<string, unknown> | null
+  /**
+   * Cached configuration for this session. May be stale if the session
+   * was not recently fetched from the API.
+   *
+   * Call getConfig() to get the latest configuration from the server,
+   * which will also update this cached value.
+   */
+  configuration?: Record<string, unknown> | null
```
VVoruganti: should set these as `public readonly` with an error that says to use the setter to change it

VVoruganti (changes requested):

sdks/typescript/src/validation.ts
```diff
@@ -138,6 +138,30 @@ const MessageSchema: z.ZodType<Message> = z.object({
...
+    .max(1.0, 'searchMaxDistance must be at most 1.0')
+    .optional(),
+  includeMostDerived: z.boolean().optional(),
+  maxObservations: z
+    .number()
+    .int()
+    .min(1, 'maxObservations must be at least 1')
+    .max(100, 'maxObservations must be at most 100')
+    .optional(),
+})
```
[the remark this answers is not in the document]
dr-frmr: No, we actually extend the schema in WorkingRepParamsSchema and PeerWorkingRepParamsSchema because there are other places where we use SearchQuerySchema alongside inputs that *do not* take a query.

migrations/versions/110bdf470272_rename_deriver_disabled_to_deriver_.py
```diff
@@ -0,0 +1,116 @@
...
+
+    while True:
+        result = bind.execute(
+            sa.text(
+                f"""
+                WITH batch AS (
+                    SELECT id
+                    FROM "{schema}".sessions
+                    WHERE configuration ? 'deriver_disabled'
+                    ORDER BY id
```
VVoruganti: remove unnecessary order by, could just slow things down

sdks/python/src/honcho/async_client/peer.py
```diff
@@ -323,6 +362,46 @@ async def set_peer_config(
...
+        return await self.set_config(config)
+
+    async def refresh(self) -> None:
+        """
+        Refresh cached metadata and configuration for this peer.
+
+        Makes async API calls to retrieve the latest metadata and configuration
+        associated with this peer and updates the cached attributes.
+        """
+        await asyncio.gather(self.get_metadata(), self.get_config())
```
VVoruganti: I get that this is good code reuse but there is an unnecessary extra api call made this way. Might be better to just implement directly

src/deriver/deriver.py
```diff
@@ -154,16 +156,25 @@ async def process_representation_tasks_batch(
...
-            speaker_peer_card: list[str] | None = await crud.get_peer_card(
+    async with tracked_db("deriver.get_peer_card") as db:
+        session_level_configuration = get_configuration(
+            await crud.get_session(
+                db, latest_message.session_name, latest_message.workspace_name
+            ),
+            await crud.get_workspace(db, workspace_name=latest_message.workspace_name),
+        )
+        if session_level_configuration.peer_cards_enabled is False:
+            speaker_peer_card = None
```
VVoruganti: So I'm wondering here if we want a difference between deriving new peer cards vs using existing peer cards. Thinking maybe we want a more deeply nested configuration object to use. 

Something like

```json
{
  "deriver": {
    "enabled": true | false,
    "representation": {
      "derive": true | false
    },
    "peer_cards": {
      "use": true | false,
      "derive": true | false
    },
    "summaries": {
      "derive": true | false,
      "use": true | false,
      "short_cadence": 20,
      "long_cadence": 60
    },
    "dreaming": {
      "enabled": true | false,
      "dedup": {
        "enabled": true | false,
        "cadence": "1/message"
      },
      "deductive": {
        "enabled": true | false,
        "cadence": "1/hr"
      },
      "etc": {
        "enabled": true | false,
        "cadence": "5 0 * 8 *"
      },
    }
  }
}
```

sdks/python/src/honcho/async_client/peer.py
dr-frmr: good point

src/deriver/deriver.py
dr-frmr: Powerful. I'd like to do this as part of a fresh PR which can include configurability at the message level and other things we have discussed.

sdks/python/src/honcho/async_client/client.py
```diff
@@ -46,6 +48,16 @@ class AsyncHoncho(BaseModel):
...
+    metadata: dict[str, object] | None = Field(
+        None,
+        frozen=True,
+        description="Cached metadata for this workspace. May be stale. Use get_metadata() for fresh data.",
+    )
+    configuration: dict[str, object] | None = Field(
+        None,
+        frozen=True,
+        description="Cached configuration for this workspace. May be stale. Use get_config() for fresh data.",
+    )
```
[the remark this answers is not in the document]
VVoruganti: The ideas is the field should be read only and can only be updated via setter functions

PR #248, opened 2025-10-28
feat: add observation endpoints, make session deletion endpoint real

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Added observations management capabilities: list observations with optional filtering, perform semantic queries, and delete observations from sessions.
  * Observations API now available in Python (sync and async) and TypeScript SDKs.

* **Changes**
  * Session deletion now permanently removes all related data instead of marking as inactive.


dr-frmr:
@coderabbit review

src/schemas.py
```diff
@@ -343,6 +343,50 @@ class DocumentCreate(DocumentBase):
     embedding: list[float] = Field()
```
Rajat-Ahuja1997, line 423: we should make a linear issue to remove the is_active column if it's not being used anymore (not sure if we still need it anywhere but looks like we don't)

src/routers/observations.py
```diff
@@ -0,0 +1,171 @@
...
+    If observer_id and observed_id are provided in filters, only observations matching
+    those criteria will be searched. Otherwise, all observations for the session are searched.
+    """
+    try:
+        # Extract observer and observed from filters if provided
+        observer = None
+        observed = None
+        if body.filters:
+            observer = body.filters.get("observer") or body.filters.get("observer_id")
+            observed = body.filters.get("observed") or body.filters.get("observed_id")
```
Rajat-Ahuja1997, line 88: not necessarily specific to this PR but do we maintain a list anywhere of the filters that can be applied to our various resources (session, peer, observations, etc)? if not, we probably should have that written down somewhere

src/crud/session.py
```diff
@@ -255,9 +264,70 @@ async def delete_session(
...
+        )
+
+        # Delete Document entries associated with this session
+        await db.execute(
+            delete(models.Document).where(
+                models.Document.session_name == session_name,
+                models.Document.workspace_name == workspace_name,
+            )
+        )
+
```
Rajat-Ahuja1997: i could see these two taking a very long time / timing out for a session with a lot of messages / lot of documents. also, we might want to batch these

src/routers/observations.py
```diff
@@ -0,0 +1,171 @@
...
+        # Now delete it using the CRUD function
+        await crud.delete_document(
+            db,
+            workspace_name=workspace_id,
+            document_id=observation_id,
+            observer=document.observer,
+            observed=document.observed,
+            session_name=session_id,
+        )
+
```
Rajat-Ahuja1997: as it stands this isn't doing an authz since we're passing in the observer/observed from the document we select and then using those two values as authz in the crud deletion. we should either remove the select and remove observer/observed from the crud deletion params or enforce some validation with JWT

sdks/python/src/honcho/async_client/session.py
```diff
@@ -393,9 +397,16 @@ async def get_messages(
...
     async def delete(self) -> None:
         """
-        Delete this session.
+        Delete this session and all associated data.
 
-        Makes an async API call to delete this session.
+        Makes an async API call to permanently delete this session and all related data including:
+        - Messages
+        - Message embeddings
+        - Documents
```
VVoruganti: Observations

sdks/python/src/honcho/async_client/session.py
```diff
@@ -393,9 +397,16 @@ async def get_messages(
 
     async def delete(self) -> None:
```
VVoruganti, line 400: I think maybe we have a hard_delete and soft_delete where the soft delete just marks it as in_active? Still see some utility marking sessions as active or inactive, but could be swayed. 

We should remove that column if it's unnecessary now

src/crud/session.py
VVoruganti: Could we handle these by adding CASCADE to db models

```
from sqlalchemy import ForeignKey
from sqlalchemy.orm import relationship

class Parent(Base):
    __tablename__ = 'parent'
    id = Column(Integer, primary_key=True)
    children = relationship("Child", back_populates="parent")

class Child(Base):
    __tablename__ = 'child'
    id = Column(Integer, primary_key=True)
    parent_id = Column(Integer, ForeignKey('parent.id', ondelete='CASCADE'))
    parent = relationship("Parent", back_populates="children")
    ```

File: src/dreamer/dreamer.py [hunk 7]
```diff
@@ -69,12 +69,14 @@ async def _process_consolidate_dream(payload: DreamPayload) -> None:
...
     async with tracked_db("dream_consolidate") as db:
         documents = await crud.get_all_documents(
-            db,
             payload.workspace_name,
             observer=payload.observer,
             observed=payload.observed,
         )
 
+        result = await db.execute(documents)
+        documents = result.scalars().all()
```
VVoruganti, line 82: Really never want to use this

src/routers/observations.py
```diff
@@ -0,0 +1,171 @@
...
+    dependencies=[
+        Depends(require_auth(workspace_name="workspace_id", session_name="session_id"))
+    ],
+)
+
+
+@router.post("/list", response_model=Page[schemas.Observation])
+async def list_observations(
+    workspace_id: str = Path(..., description="ID of the workspace"),
+    session_id: str = Path(..., description="ID of the session"),
```
VVoruganti: So thinking of a few scenarios here
- I want to get all the observations about a specific Peer
- I want to get all the observations about a Peer made in a session
- I want all of the observations Alice has made about Bob
- I want all observations made about all Peers in a session

So want to make this composable enough to let you specific representation to target and at what scope.

src/routers/observations.py
VVoruganti: There is a page on the docs that talks about filters: https://docs.honcho.dev/v2/documentation/core-concepts/features/using-filters

src/routers/observations.py
```diff
@@ -0,0 +1,171 @@
...
+
+            stmt = apply_filter(stmt, models.Document, filters)
+
+        # Order by created_at (newest first by default)
+        if reverse:
+            stmt = stmt.order_by(models.Document.created_at.asc())
+        else:
+            stmt = stmt.order_by(models.Document.created_at.desc())
+
+        return await apaginate(db, stmt)
```
VVoruganti, line 57: actual db logic should be limited to being in the crud module

src/routers/observations.py
dr-frmr: great catch, i'm gonna vote we limit `list_observations` to JWTs with session-level or higher clearance, and limit `query` and `delete` to workspace level or higher. Observations are sensitive and can influence the whole workspace...

sdks/python/src/honcho/async_client/session.py
dr-frmr: Not sure--currently in this PR there's only hard delete: should we add soft delete, or get rid of `is_active`? I don't have a strong preference either way

src/routers/observations.py
dr-frmr: Actually all need workspace clearance. `list_observations` can be used across many sessions now.

src/routers/observations.py
dr-frmr: Ok, the only change needed here was to remove session_id from the parameters and let people pass that as a filter just like observer & observed. All these scenarios are now supported via filtering.

src/dreamer/dreamer.py
dr-frmr: in the future, we can do advanced clustering, for now i just limited the dream to work on 100 recent documents

migrations/versions/121203eca967_merge_migration_heads.py
VVoruganti, line 1: Empty migration, remove gracefully

VVoruganti: Just rebase whatever new migration you made. This seems overly complicated

sdks/python/src/honcho/async_client/session.py
VVoruganti: I'm open to getting rid of is_active altogether, don't think it's really used and is more an artifact of Honcho v1. Would just worry about any cases where people have existing code that tries to use to Honcho with a delete and this causes a memory loss event. 

Wondering if we should do a token exchange or some kind of handshake method for deletion methods so that you have to confirm that you really want to delete it. Feels scary to make it so easy to just throw away a bunch of work

sdks/python/src/honcho/session.py
```diff
@@ -702,6 +713,106 @@ def search(
...
+        Args:
+            observation_id: The ID of the observation to delete
+
+        Example:
+            >>> session.delete_observation('obs_123abc')
+        """
+        self._client.workspaces.sessions.observations.delete(
+            observation_id, session_id=self.id, workspace_id=self.workspace_id
+        )
+
```
VVoruganti: From a UX standpoint I think it would be annoying to only be able to delete an observation from a session. Thinking maybe we need a new top level path for `Representations` that let's you scope by observer_observed and take in sessions and peers and objects to filter by?

src/crud/document.py
```diff
@@ -2,42 +2,101 @@
...
+
+def get_documents_with_filters(
+    workspace_name: str,
+    *,
+    filters: dict[str, Any] | None = None,
+    reverse: bool = False,
+) -> Select[tuple[models.Document]]:
+    """
+    Get all documents using custom filters.
+
```
VVoruganti, line 74: is this necessary with the previous function?

src/crud/session.py
```diff
@@ -277,11 +277,56 @@ async def update_session(
...
+        )
+        delete_stmt = delete(model).where(primary_key_column.in_(subquery))
+        delete_result = await db.execute(delete_stmt)
+        batch_deleted = delete_result.rowcount or 0
+        total_deleted += batch_deleted
+
+        if batch_deleted == 0:
+            break
+
+    return total_deleted
```
VVoruganti, line 317: Almost wondering if we need to make this a background task since this could potentially take a long time and just timeout due to normal api request limits

src/crud/session.py
```diff
@@ -296,15 +341,79 @@ async def delete_session(
...
+        # Work unit keys have format: {task_type}:{workspace_name}:{session_name}:{...}
+        await db.execute(
+            delete(models.ActiveQueueSession).where(
+                and_(
+                    func.split_part(models.ActiveQueueSession.work_unit_key, ":", 2)
+                    == workspace_name,
+                    func.split_part(models.ActiveQueueSession.work_unit_key, ":", 3)
+                    == session_name,
+                )
+            )
```
VVoruganti, line 361: Do we need to create any logic in the deriver to stop working on these if it has claimed a workunit? Also is there any lock contention to worry about?

src/utils/search.py
```diff
@@ -184,6 +187,35 @@ async def search(
     """
     # Base query conditions
     stmt = select(models.Message)
+
+    # Handle special peer_perspective filter
+    if filters and "peer_perspective" in filters:
+        peer_name = filters["peer_perspective"]
```
VVoruganti, line 193: Update documentation to talk about this new custom filter

tests/alembic/revisions/test_110bdf470272_rename_deriver_disabled_to_deriver_.py
VVoruganti, line 1: I imagine this file will get changed in an upstream branch

VVoruganti (changes requested):
In general just worried about the fragility of these deletion endpoints. 
1. Worried about any backwards compatability with is_active
2. With the performance of deleting a very large sesison 

Also thinking maybe we need to update the data mode to explicitly call our Representations and Observations

```mermaid
 graph TD
      W[Workspaces] -->|have| P[Peers]
      W -->|have| S[Sessions]

      S -->|have| SM[Messages]

      P <-.->|many-to-many| S
      
      P -->|have| R[Representations]

       R -->|have| O[Observations]


      style W fill:#FF5A7E,stroke:#333,stroke-width:2px,color:#fff
      style P fill:#e1f5fe,stroke:#0277bd,color:#000
      style S fill:#f3e5f5,stroke:#7b1fa2,color:#000
      style SM fill:#e8f5e9,stroke:#2e7d32,color:#000
```

PR #251, opened 2025-10-29
feat: add messages-per-summary config options to workspace and sessions

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Added configurable summary thresholds at workspace and session levels
  * Summary settings now support hierarchical configuration (session overrides workspace, workspace overrides global defaults)
  * Configuration merging preserves existing settings when updating

* **Tests**
  * Added comprehensive integration and unit tests for summary configuration across workspaces and sessions

### DIFF
--- a/migrations/versions/110bdf470272_rename_deriver_disabled_to_deriver_.py
+++ b/migrations/versions/110bdf470272_rename_deriver_disabled_to_deriver_.py
@@ -30,41 +30,42 @@ def upgrade() -> None:
     - deriver_disabled: false -> deriver_enabled: true
     - Remove deriver_disabled key from configuration
     """
-    # Update sessions table
-    # Set deriver_enabled to false where deriver_disabled is true
-    op.execute(
-        sa.text(
-            f"""
-            UPDATE "{schema}".sessions
-            SET configuration = configuration - 'deriver_disabled' || jsonb_build_object('deriver_enabled', false)
-            WHERE configuration ? 'deriver_disabled'
-              AND (configuration->>'deriver_disabled')::boolean = true
-            """
-        )
-    )
+    # Update sessions table in batches
+    # Combine all cases in one pass: set deriver_enabled based on deriver_disabled value,
+    # or just remove deriver_disabled if it's null or other value
+    bind = op.get_bind()
+    batch_size = 5000
 
-    # Set deriver_enabled to true where deriver_disabled is false
-    op.execute(
-        sa.text(
-            f"""
-            UPDATE "{schema}".sessions
-            SET configuration = configuration - 'deriver_disabled' || jsonb_build_object('deriver_enabled', true)
-            WHERE configuration ? 'deriver_disabled'
-              AND (configuration->>'deriver_disabled')::boolean = false
-            """
+    while True:
+        result = bind.execute(
+            sa.text(
+                f"""
+                WITH batch AS (
+                    SELECT id
+                    FROM "{schema}".sessions
+                    WHERE configuration ? 'deriver_disabled'
+                    ORDER BY id
+                    LIMIT :batch_size
+                )
+                UPDATE "{schema}".sessions s
+                SET configuration = CASE
+                    WHEN s.configuration->>'deriver_disabled' = 'true' THEN
+                        s.configuration - 'deriver_disabled' || jsonb_build_object('deriver_enabled', false)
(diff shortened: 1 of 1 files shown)


src/schemas.py
```diff
@@ -241,6 +334,38 @@ class SessionGet(SessionBase):
...
+        # Merge summary config into configuration dict if provided
+        if short is not None or long is not None:
+            if self.configuration is None:
+                self.configuration = {}
+            if short is not None:
+                self.configuration["messages_per_short_summary"] = short
+            if long is not None:
+                self.configuration["messages_per_long_summary"] = long
+
+        return self
```
VVoruganti: I think I'd rather take the approach of having a SessionConfiguration pydantic model that is more tightly defined and leave it all in there. There's a weirdness here of having both a configuration field and separate fields for the request. Can also then just add to that one configuration model for subsequent updates

src/schemas.py
```diff
@@ -230,9 +291,41 @@ class SessionCreate(SessionBase):
...
```
VVoruganti: Same comment as Session One. We could probably also make the Configuration models inherit from the same base? Have a ConfigurationBase and then the Session and Workspace ones just inherit

src/schemas.py
```diff
@@ -230,9 +291,41 @@ class SessionCreate(SessionBase):
...
+    )
+    messages_per_long_summary: int | None = Field(
+        default=None,
+        ge=20,
+        description="Number of messages per long summary. Must be positive, greater than or equal to 20, and greater than messages_per_short_summary.",
+    )
 
     model_config = ConfigDict(populate_by_name=True)  # pyright: ignore
 
+    @model_validator(mode="after")
```
VVoruganti: Actually not sure when we would need validation here, because this model is only used in responses and not in requests so probably doesn't need validation like this? doesn't hurt I suppose

src/schemas.py
```diff
@@ -30,9 +30,38 @@ class WorkspaceCreate(WorkspaceBase):
...
+                "messages_per_short_summary must be less than messages_per_long_summary"
+            )
+
+        # Merge summary config into configuration dict if provided
+        if short is not None:
+            self.configuration["messages_per_short_summary"] = short
+        if long is not None:
+            self.configuration["messages_per_long_summary"] = long
+
+        return self
```
VVoruganti, line 83: Code reuse a lot of this code with a unified model

src/utils/config_helpers.py
```diff
@@ -0,0 +1,64 @@
...
+
+import logging
+
+from src import models
+from src.config import settings
+
+logger = logging.getLogger(__name__)
+
+
+def get_summary_config(
```
VVoruganti: minor nit, can probably do some short circuiting by checking order of precedence and returning whenever both options are fulfilled.

src/deriver/enqueue.py
```diff
@@ -277,8 +291,8 @@ async def generate_queue_records(
     records: list[dict[str, Any]] = []
 
     if settings.SUMMARY.ENABLED and (
```
VVoruganti: Just a callout probably want to also make this a configurable value that'll fall back to the .env value

migrations/versions/110bdf470272_rename_deriver_disabled_to_deriver_.py
```diff
@@ -0,0 +1,112 @@
...
+    op.execute(
+        sa.text(
+            f"""
+            UPDATE "{schema}".sessions
+            SET configuration = configuration - 'deriver_disabled'
+            WHERE configuration ? 'deriver_disabled'
+            """
+        )
+    )
+
```
Rajat-Ahuja1997: can probably combine these in one pass. also we should batch

migrations/versions/110bdf470272_rename_deriver_disabled_to_deriver_.py
```diff
@@ -0,0 +1,112 @@
...
+    op.execute(
+        sa.text(
+            f"""
+            UPDATE "{schema}".sessions
+            SET configuration = configuration - 'deriver_enabled' || jsonb_build_object('deriver_disabled', false)
+            WHERE configuration ? 'deriver_enabled'
+              AND (configuration->>'deriver_enabled')::boolean = true
+            """
+        )
+    )
```
Rajat-Ahuja1997: same as above

Rajat-Ahuja1997 (approved):

PR #252, opened 2025-10-29
Fix Server Default Migrations and add in appropriate Server Defaults

VVoruganti:


src/models.py
```diff
@@ -429,10 +453,12 @@ class QueueItem(Base):
...
     task_type: Mapped[TaskType] = mapped_column(TEXT, nullable=False)
     payload: Mapped[dict[str, Any]] = mapped_column(JSONB, nullable=False)
-    processed: Mapped[bool] = mapped_column(Boolean, default=False)
+    processed: Mapped[bool] = mapped_column(
+        Boolean, default=False, server_default=text("false")
+    )
     error: Mapped[str | None] = mapped_column(TEXT, nullable=True)
     created_at: Mapped[datetime.datetime] = mapped_column(
-        DateTime(timezone=True), default=func.now()
+        DateTime(timezone=True), server_default=func.now()
```
Rajat-Ahuja1997, line 461: do we want both default and server_default here?

Rajat-Ahuja1997: this question applies to all of the non jsonb columns that you have amended from `default` to `server_default`

migrations/env.py
```diff
@@ -74,6 +75,51 @@ def run_migrations_offline() -> None:
...
+    # Get current port, default to 5432 if not specified
+    current_port = parsed.port or 5432
+
+    # If using transaction pooler port (6543), switch to session pooler (5432)
+    if current_port == 6543:
+        # Replace the port in the netloc
+        if parsed.port:
+            # If port is explicitly in the URL, replace it
+            new_netloc = parsed.netloc.replace(f":{current_port}", ":5432")
+        else:
```
Rajat-Ahuja1997, line 112: nit: isn't the implication here that if we don't have a parsed.port then current port will be 5432? so this block is unreachable

src/models.py
VVoruganti: So supposedly using default=func.now() would still just make it a server side then and just add into the insert SQL a `NOW()` value so don't think it's helpful for the timestamp ones. 

Don't think there's a real benefit to having both. Only place where we want it on the client side is for the nanoids where postgres can't natively generate those.

PR #254, opened 2025-10-29
feat: codify queue columns

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **Refactor**
  * Queue items now record workspace name and message ID in dedicated columns, with new indexes and foreign key for faster, reliable lookups.

* **Behavioral**
  * Processing, delivery, scheduling, and work‑unit key construction accept explicit workspace context and message IDs; serialized queue payloads omit those migrated fields.
  * Documents now expose reasoning level and derivation count as top‑level fields with sensible defaults.

* **Tests**
  * Added comprehensive migration tests and updated unit/integration tests to validate the migration and new queue semantics.

* **Chores**
  * Test tooling updated to run migration tests selectively.

### DIFF
--- a/migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
+++ b/migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
@@ -127,6 +127,17 @@ def upgrade() -> None:
         referent_schema=schema,
     )
 
+    # Step 7b: Add foreign key constraint on message_id -> messages.id
+    op.create_foreign_key(
+        "fk_queue_message_id",
+        "queue",
+        "messages",
+        ["message_id"],
+        ["id"],
+        source_schema=schema,
+        referent_schema=schema,
+    )
+
     # Step 8: Add index on workspace_name (for FK performance and filtering)
     op.create_index(
         op.f("ix_queue_workspace_name"),
@@ -191,7 +202,10 @@ def downgrade() -> None:
             op.f("ix_queue_workspace_name"), table_name="queue", schema=schema
         )
 
-    # Drop foreign key constraint
+    # Drop foreign key constraints
+    if fk_exists("queue", "fk_queue_message_id", inspector):
+        op.drop_constraint("fk_queue_message_id", "queue", schema=schema)
+
     if fk_exists("queue", "fk_queue_workspace_name", inspector):
         op.drop_constraint("fk_queue_workspace_name", "queue", schema=schema)
 
--- a/src/models.py
+++ b/src/models.py
@@ -469,7 +469,9 @@ class QueueItem(Base):
     workspace_name: Mapped[str] = mapped_column(
         ForeignKey("workspaces.name"), nullable=False
     )
-    message_id: Mapped[int | None] = mapped_column(BigInteger, nullable=True)
+    message_id: Mapped[int | None] = mapped_column(
+        BigInteger, ForeignKey("messages.id"), nullable=True
+    )
 
     __table_args__ = (
         Index("ix_queue_created_at", "created_at"),
--- a/tests/alembic/revisions/test_ec8f94139b02_codify_workspace_name_and_message_id_in_.py
+++ b/tests/alembic/revisions/test_ec8f94139b02_codify_workspace_name_and_message_id_in_.py
@@ -36,6 +36,14 @@ def prepare_codify_workspace_name_and_message_id_in(
     verifier.assert_column_exists("queue", "workspace_name", exists=False)
     verifier.assert_column_exists("queue", "message_id", exists=False)
 
+    # Verify foreign key constraints don't exist yet
+    verifier.assert_constraint_exists(
+        "queue", "fk_queue_workspace_name", "foreign_key", exists=False
+    )
+    verifier.assert_constraint_exists(
+        "queue", "fk_queue_message_id", "foreign_key", exists=False
+    )
+
     # Verify indexes don't exist yet
     verifier.assert_indexes_not_exist(_INDEXES)
 
@@ -217,8 +225,9 @@ def verify_codify_workspace_name_and_message_id_in(verifier: MigrationVerifier) 
     # Verify all indexes were created
     verifier.assert_indexes_exist(_INDEXES)
 
-    # Verify foreign key constraint exists
+    # Verify foreign key constraints exist
     verifier.assert_constraint_exists("queue", "fk_queue_workspace_name", "foreign_key")
+    verifier.assert_constraint_exists("queue", "fk_queue_message_id", "foreign_key")
 
     conn = verifier.conn
     schema = verifier.schema
--- a/tests/deriver/test_queue_processing.py
+++ b/tests/deriver/test_queue_processing.py
@@ -711,7 +711,6 @@ class TestQueueProcessing:
         token_counts = [500, 600]
         messages = [
             models.Message(
-                id=999,
                 session_name=session.name,
                 workspace_name=session.workspace_name,
                 peer_name=peer.name,
@@ -720,7 +719,6 @@ class TestQueueProcessing:
                 seq_in_session=1,
             ),
             models.Message(
-                id=1000,
                 session_name=session.name,
                 workspace_name=session.workspace_name,
                 peer_name=peer.name,
@@ -729,6 +727,15 @@ class TestQueueProcessing:
                 seq_in_session=2,
             ),
         ]
+
+        # Save messages to database first
+        for message in messages:
+            db_session.add(message)
+        await db_session.commit()
+
+        # Refresh to get the actual IDs
+        for message in messages:
+            await db_session.refresh(message)
 
         # Create payloads and queue items
         queue_items: list[models.QueueItem] = []


migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,289 @@
...
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+from migrations.utils import column_exists, fk_exists, get_schema, index_exists
+
+# revision identifiers, used by Alembic.
+revision: str = "ec8f94139b02"
```
Rajat-Ahuja1997, line 17: will likely need to update this depending on which migration(s) get into main

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,289 @@
...
+    # Step 1: Drop rows where workspace_name is NULL in payload
+    # This removes invalid/corrupted queue items
+    op.execute(
+        sa.text(
+            f"""
+            DELETE FROM "{schema}".queue
+            WHERE payload->>'workspace_name' IS NULL
+            """
+        )
+    )
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: i expect this not to actually do anything because workspace name has been enforced in all queue payload schemas for quite some time but adding this defensively just in case.

table scan might take quite a while though so we might want to batch this..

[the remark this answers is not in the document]
Rajat-Ahuja1997: batched in latest commit. also as a learning, op.execute doesn't support parameterized queries like this. you have to use conn.execute() where conn = op.get_bind()

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,298 @@
...
+                UPDATE "{schema}".queue q
+                SET payload = q.payload - 'message_id'
+                FROM batch
+                WHERE q.id = batch.id
+                """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
+            break
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: unneeded, we run our migrations as all or nothing

scripts/run_alembic_tests.py
```diff
@@ -0,0 +1,141 @@
...
+
+This script determines which specific alembic tests to run based on:
+1. If a test file changed, run the test for that revision
+2. If a migration file changed, run the corresponding test
+
+The alembic test system uses a parameterized test in test_pipeline.py that runs
+for each revision. We filter these tests using pytest's -k flag with the revision IDs.
+
+Usage: python scripts/run_alembic_tests_selective.py <file1> <file2> ...
+"""
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed  in latest commit

.pre-commit-config.yaml
```diff
@@ -74,11 +74,11 @@ repos:
...
         name: pytest (alembic migrations)
-        entry: uv run pytest tests/alembic/
+        entry: uv run python scripts/run_alembic_tests.py
         language: system
-        files: ^(migrations/.*\.py|tests/alembic/.*\.py)$
+        files: ^(migrations/versions/.*\.py|tests/alembic/revisions/.*\.py)$
         stages: [pre-push]
-        pass_filenames: false
+        pass_filenames: true
         require_serial: true
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed  in latest commit

src/deriver/enqueue.py
```diff
@@ -159,19 +159,29 @@ def create_representation_record(
...
     )
     return {
-        "work_unit_key": get_work_unit_key(processed_payload),
+        "work_unit_key": construct_work_unit_key(workspace_name, processed_payload),
         "payload": processed_payload,
         "session_id": session_id,
         "task_type": "representation",
+        "workspace_name": workspace_name,
+        "message_id": message_id,
     }
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed  in latest commit

src/deriver/consumer.py
```diff
@@ -56,25 +59,29 @@ async def process_item(task_type: str, queue_payload: dict[str, Any]) -> None:
             )
             raise ValueError(f"Invalid payload structure: {str(e)}") from e
 
+        # Use workspace_name and message_id from QueueItem columns
+        message_id = queue_item.message_id
+
+        if message_id is None:
+            raise ValueError("Summary tasks require a message_id")
+
         message_public_id = validated.message_public_id
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed  in latest commit

src/models.py
VVoruganti, line 1: Alembic Check Results
```
alembic check
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.autogenerate.compare] Detected removed index 'ix_queue_created_at' on 'queue'
INFO  [alembic.autogenerate.compare] Detected removed index 'ix_queue_session_id' on 'queue'
ERROR [alembic.util.messaging] New upgrade operations detected: [('remove_index', Index('ix_queue_created_at', Column('created_at', TIMESTAMP(timezone=True), table=<queue>, nullable=False, server_default=DefaultClause(<sqlalchemy.sql.elements.TextClause object at 0x1065eec90>, for_update=False)))), ('remove_index', Index('ix_queue_session_id', Column('session_id', TEXT(), table=<queue>)))]
  FAILED: New upgrade operations detected: [('remove_index', Index('ix_queue_created_at', Column('created_at',
  TIMESTAMP(timezone=True), table=<queue>, nullable=False, server_default=DefaultClause(<sqlalchemy.sql.elements.TextClause object
  at 0x1065eec90>, for_update=False)))), ('remove_index', Index('ix_queue_session_id', Column('session_id', TEXT(),
  table=<queue>)))]
  ```

File: src/models.py [hunk 8]
```diff
@@ -325,6 +325,9 @@ class MessageEmbedding(Base):
     )
 
 
+DocumentLevel = Literal["explicit", "deductive"]
```
VVoruganti: Move this to the `utils/types.py` file

src/deriver/consumer.py
```diff
@@ -98,7 +103,7 @@ async def process_item(task_type: str, queue_payload: dict[str, Any]) -> None:
                     queue_payload,
                 )
                 raise ValueError(f"Invalid payload structure: {str(e)}") from e
-            await process_dream(validated)
+            await process_dream(validated, workspace_name)
```
VVoruganti, line 106: Not sure I see the logic for why we can't just use the DreamPayload here, don't see any additional validation to check for the workspace_name and it's constructed from the same source data

migrations/versions/b8183c5ffb48_codify_document_level_and_times_derived.py
```diff
@@ -0,0 +1,207 @@
...
+                    1
+                )
+                FROM batch
+                WHERE d.id = batch.id
+            """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
+            break
```
VVoruganti: Could combine this into one batching operation:

```python
 Step 3: Populate level and times_derived from internal_metadata in batches
batch_size = 5000
while True:
    result = connection.execute(
        text(
            f"""
            WITH batch AS (
                SELECT id
                FROM {schema}.documents
                WHERE level IS NULL OR times_derived IS NULL
                LIMIT :batch_size
            )
            UPDATE {schema}.documents d
            SET 
                level = COALESCE(
                    d.internal_metadata->>'level',
                    'explicit'
                ),
                times_derived = COALESCE(
                    (d.internal_metadata->>'times_derived')::integer,
                    1
                )
            FROM batch
            WHERE d.id = batch.id
        """
        ),
        {"batch_size": batch_size},
    )
    if result.rowcount == 0:
        break
```

migrations/versions/b8183c5ffb48_codify_document_level_and_times_derived.py
```diff
@@ -0,0 +1,207 @@
...
+                    to_jsonb(d.times_derived)
+                )
+                FROM batch
+                WHERE d.id = batch.id
+            """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
+            break
```
VVoruganti: Could probably consolidate these too

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,298 @@
...
+    batch_size = 5000
+    while True:
+        result = conn.execute(
+            sa.text(
+                f"""
+                WITH batch AS (
+                    SELECT id
+                    FROM "{schema}".queue
+                    WHERE workspace_name IS NULL
+                    ORDER BY id
```
VVoruganti: remove the order to improve perf

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,298 @@
...
+                    LIMIT :batch_size
+                )
+                UPDATE "{schema}".queue q
+                SET workspace_name = q.payload->>'workspace_name'
+                FROM batch
+                WHERE q.id = batch.id
+                AND q.workspace_name IS NULL
```
VVoruganti: unnecessary repeated check that'll do an entire scan

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
```diff
@@ -0,0 +1,298 @@
...
+                UPDATE "{schema}".queue q
+                SET payload = q.payload - 'workspace_name'
+                FROM batch
+                WHERE q.id = batch.id
+                """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
+            break
```
VVoruganti: See where these could be combined with previous batches

migrations/versions/ec8f94139b02_codify_workspace_name_and_message_id_in_.py
VVoruganti, line 114: Same comment as above, see if you can consolidate

src/utils/queue_payload.py
```diff
@@ -30,59 +35,73 @@ class RepresentationPayloads(BasePayload):
...
+    """Payload for webhook delivery tasks.
+
+    Note: workspace_name is optional for backward compatibility during migration.
+    New queue items should not include this field as it is now stored in a
+    dedicated column on the queue table.
+    """
 
     task_type: Literal["webhook"] = "webhook"
-    workspace_name: str
+    workspace_name: str | None = None  # Migrated to column
```
VVoruganti: So we've changed the code everywhere to just use the workspace_name and message_id columns from the QueueItem rather than the payload and pydantic by default will just discard extra values when you try to validate a dictionary or coerce it into a pydantic model.

I don't see how keeping these here actually protect us especially if the migration ensures all of the QueueItems are fixed. I don't see any fallback logic in the consumer that would then go back and use these fields

PR #255, opened 2025-10-29
feat: add optional backup providers that kick in on retry

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Per-component BACKUP_PROVIDER/BACKUP_MODEL support and automatic failover to backups on final retry; LLM calls now use unified per-component LLM settings.
  * Retry tracking added to coordinate failover attempts and logging.

* **Validation**
  * Config validation requires backup provider and model be set together or both omitted.

* **Behavior**
  * When failing over to backups, provider-specific unsupported parameters are filtered with warnings.

* **Documentation**
  * Example env/config updated with backup fields and revised model defaults.


src/config.py
```diff
@@ -198,6 +207,8 @@ class DeriverSettings(HonchoSettings):
 
     PROVIDER: SupportedProviders = "google"
     MODEL: str = "gemini-2.5-flash-lite"
+    BACKUP_PROVIDER: SupportedProviders | None = None
```
Rajat-Ahuja1997: i would just make sure to add all of the new config to config.toml.example and .env.template

src/utils/clients.py
```diff
@@ -183,6 +183,73 @@ async def honcho_llm_call(
...
+        and llm_settings.BACKUP_MODEL is not None
+        and llm_settings.BACKUP_PROVIDER in CLIENTS
+    )
+
+    if (
+        use_backup
+        and llm_settings.BACKUP_PROVIDER is not None
+        and llm_settings.BACKUP_MODEL is not None
+    ):
+        provider = llm_settings.BACKUP_PROVIDER
```
Rajat-Ahuja1997: these two are redundant because they're inside the use_backup check

src/utils/clients.py
```diff
@@ -193,13 +260,28 @@ async def honcho_llm_call(
     if track_name:
         decorated = ai_track(track_name)(decorated)
 
-    # apply retry logic
+    # apply retry logic - retries on ANY exception
     if enable_retry:
+
+        def before_retry_callback(retry_state: Any) -> None:
```
Rajat-Ahuja1997, line 306: I think when retry happens, it calls honcho_llm_call_inner again with the same provider and model arguments that were initially determined in the function. The backup provider selection logic never runs again because that logic is in honcho_llm_call, not in honcho_llm_call_inner.

Rajat-Ahuja1997 (commented):
setup looks good but I don't think the retry logic is working as intended, should be a simple fix though

src/utils/clients.py
Rajat-Ahuja1997: we also might want to retry with the same provider once (for certain exceptions) and then go to the backup provider if needed

dr-frmr: agreed, backup should come into play on 3rd and final retry

src/utils/clients.py
dr-frmr: oops

src/utils/clients.py
```diff
@@ -208,53 +208,92 @@ async def honcho_llm_call(
...
-        # Filter out incompatible parameters when using backup
-        if provider != "anthropic" and thinking_budget_tokens:
-            logger.warning(
-                f"thinking_budget_tokens not supported by {provider}, ignoring"
+        """
+        Inner function that selects provider/model based on current attempt.
+        This function is retried, so provider selection happens on each attempt.
+        """
+        attempt = _current_attempt.get()
+
```
Rajat-Ahuja1997, line 223: could be interesting to increment a prometheus metric whenever we switch backups. then we could have a dashboard in grafana tracking this to clue us in on model/provider issues. not necessary for this PR though

PR #262, opened 2025-11-03
feat: add jinja templating for prompts

Rajat-Ahuja1997:
- Resolves https://linear.app/plastic-labs/issue/DEV-1194

## Summary by CodeRabbit

* **New Features**
  * Prompt system moved to template-driven rendering with several new customizable templates and a template rendering utility.
  * New dream consolidation settings and related template path added; additional deriver and dialectic template paths added.

* **Documentation**
  * Example config and env templates updated with new template path entries and commented placeholders.

* **Chores**
  * Added Jinja2 dependency.

* **Tests**
  * Added tests for template rendering and prompt generation.

### DIFF
--- a/tests/utils/test_templates.py
+++ b/tests/utils/test_templates.py
@@ -1,5 +1,6 @@
 """Tests for the Jinja2 template rendering system."""
 
+import json
 from collections.abc import Callable
 from datetime import datetime, timezone
 from typing import Any, cast
@@ -74,10 +75,11 @@ class TestDeriverPrompts:
     def test_critical_analysis_prompt_basic(self):
         """Test critical_analysis_prompt renders with minimal inputs."""
         representation = Representation()
+        created_at = datetime(2025, 1, 1, tzinfo=timezone.utc)
         result = critical_analysis_prompt(
             peer_id="test_user",
             peer_card=None,
-            message_created_at=datetime.now(timezone.utc),
+            message_created_at=created_at,
             working_representation=representation,
             history="Previous context here",
             new_turns=["User: Hello", "Assistant: Hi there"],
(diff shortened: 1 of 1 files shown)


tests/utils/test_templates.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+class TestTemplateManager:
+    """Tests for the TemplateManager class."""
+
+    def test_template_manager_singleton(self):
+        """Verify TemplateManager is cached as a singleton."""
+        manager1 = get_template_manager()
+        manager2 = get_template_manager()
+        assert manager1 is manager2
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: the caching you're recommending is implemented in this PR

src/config.py
```diff
@@ -220,6 +220,10 @@ class DeriverSettings(HonchoSettings):
 
     MAX_INPUT_TOKENS: Annotated[int, Field(default=23000, gt=0, le=23000)] = 23000
 
+    # Template paths for prompt templates
+    CRITICAL_ANALYSIS_TEMPLATE: str = "deriver/critical_analysis.jinja"
+    PEER_CARD_TEMPLATE: str = "deriver/peer_card.jinja"
```
VVoruganti, line 225: Should just put these in the corresponding sections as a TEMPLATE or PROMPT variable so 

DERIVER_TEMPLATE or PEER_CARD_TEMPLATE.

Also think we can move away from the wording of critical analysis call. That name is fairly arbitrary. What we could do is potentially add support parsing lists based on a `,` in the string so that we could have more than 1 value in the case we want pipelines of prompts (`settings.DERIVER.TEMPLATE[0]`)

src/config.py
```diff
@@ -263,6 +267,10 @@ class DialecticSettings(HonchoSettings):
         int, Field(default=100_000, gt=10_000, le=200_000)
     ] = 100_000
 
+    # Template paths for prompt templates
+    DIALECTIC_TEMPLATE: str = "dialectic/dialectic.jinja"
+    QUERY_GENERATION_TEMPLATE: str = "dialectic/query_generation.jinja"
+
 
 class SummarySettings(HonchoSettings):
```
VVoruganti, line 275: We should be consistent and have all prompts be templates

tests/utils/test_templates.py
VVoruganti, line 1: These tests are good but I would like if we could have more comprehensive testing that looked the jinja templates and prompts.py files and validated the contracts match. The prompts.py files are providing all required variables and are not providing anything extra

PR #266, opened 2025-11-10
rm: level_valid constraint

Rajat-Ahuja1997:
- this feature is still developing and we shouldn't need a migration every time we add a new level e.g., `implicit`

## Summary by CodeRabbit

* **Tests**
  * Added new migration verification tests to assert constraint presence before upgrade and absence after upgrade.

* **Chores**
  * Added a migration that removes the database check enforcing allowed values for the document level field and provides a downgrade path that normalizes existing rows.
  * Adjusted schema/validation declarations to use a shared level type instead of a literal value set.

### DIFF
--- a/migrations/versions/29ade7350c19_remove_document_level_valid_constraint.py
+++ b/migrations/versions/29ade7350c19_remove_document_level_valid_constraint.py
@@ -42,6 +42,30 @@ def downgrade() -> None:
     connection = op.get_bind()
     inspector = sa.inspect(connection)
 
+    # Normalize any rows with values outside the legacy domain before recreating constraint
+    # This ensures downgrade succeeds even if new values like 'implicit' were added after upgrade
+    batch_size = 5000
+    while True:
+        result = connection.execute(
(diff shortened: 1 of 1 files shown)


VVoruganti (changes requested):
Do this as a new migration rather than modifying old ones

Rajat-Ahuja1997:
> Do this as a new migration rather than modifying old ones

done

migrations/versions/29ade7350c19_remove_document_level_valid_constraint.py
```diff
@@ -0,0 +1,76 @@
...
+                )
+                UPDATE {schema}.documents d
+                SET level = 'explicit'
+                FROM batch
+                WHERE d.id = batch.id
+                """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
```
Rajat-Ahuja1997, line 66: could also just drop these rows but this preserves data

PR #274, opened 2025-11-18
metrics for deriver / dialectic input + output tokens

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **New Features**
  * Streaming responses now emit final-chunk output-token counts.

* **Improvements**
  * More granular token metrics: separate input vs. output tracking, per-component labels, and new dialectic/deriver metrics for better observability.
  * Summary and prompt flows now include explicit input-token accounting and prompt-token estimators for more accurate usage reporting.

* **Other**
  * Token encoding updated, which may change reported token counts.

### DIFF
--- a/src/utils/summarizer.py
+++ b/src/utils/summarizer.py
@@ -360,15 +360,10 @@ async def _create_and_save_summary(
     )
 
     messages_tokens = sum([message.token_count for message in messages])
-    print("latest_summary", latest_summary)
-    print(
-        "latest_summary token count",
-        latest_summary["token_count"] if latest_summary else 0,
-    )
     previous_summary_tokens = latest_summary["token_count"] if latest_summary else 0
     input_tokens = messages_tokens + previous_summary_tokens
 
-    new_summary = await _create_summary(
+    new_summary, is_fallback = await _create_summary(
         messages=messages,
         previous_summary_text=previous_summary_text,
         summary_type=summary_type,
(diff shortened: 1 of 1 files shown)


src/deriver/deriver.py
```diff
@@ -341,8 +341,11 @@ async def reason(
                 working_representation=working_representation,
                 history=history,
                 new_turns=new_turns,
-                estimated_input_tokens=self.estimated_input_tokens,
             )
+            prometheus.DERIVER_TOKENS_PROCESSED.labels(
+                task_type="critical_analysis",
+                token_type="input",  # nosec B106
+            ).inc(self.estimated_input_tokens)
```
VVoruganti: So I think we'd want to break this out even more granularly to understand just the diffrent dimensions of the input. 
- Representation Batch
- Peer Card
- Working Representation 

etc. 

Is there a way we could send a log with payload that breaks it down or add additional labels?

src/utils/clients.py
VVoruganti, line 1: Looks good as long as all of these were tested

VVoruganti (changes requested):

src/deriver/deriver.py
Rajat-Ahuja1997: this should say `task_type="representation"` not task_type="critical_analysis", will fix.

but i think in prometheus to get the understand to different parts of the input we would have to do something like this:

```
prometheus.DERIVER_TOKENS_PROCESSED.labels(
    task_type="representation",
    token_type="input",
    resource="working_representation"  # or "peer_card" etc.
).inc(self.estimated_working_rep_tokens)
```

and then at query time we could aggregate on task_type="representation", "token_type="input"

src/schemas.py
```diff
@@ -136,7 +136,7 @@ def encoded_message(self) -> list[int]:
 
     @model_validator(mode="after")
     def validate_and_set_token_count(self) -> Self:
-        encoding = tiktoken.get_encoding("cl100k_base")
+        encoding = tiktoken.get_encoding("o200k_base")
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: our embedding model aligns with cl100k_base. do these two need to align?

PR #287, opened 2025-12-04
Turbopuffer and LanceDB Integration

Rajat-Ahuja1997:
- Resolves https://linear.app/plastic-labs/issue/DEV-1244

## Summary by CodeRabbit

* **New Features**
  * Configurable, pluggable external vector stores (LanceDB, Turbopuffer) with namespace scoping and resilient upsert retries.
  * Background reconciler scheduler to run vector sync, re-embedding and periodic queue cleanup.

* **Behavior Changes / Bug Fixes**
  * Embeddings may be stored externally; documents and embeddings gain sync state, retry metadata, and soft-delete semantics with vector cleanup.
  * Semantic search uses external vector store when enabled.

* **Tests**
  * In-memory mock vector store and extensive reconciliation tests added for CI.

### DIFF
--- a/sdks/typescript/bun.lock
+++ b/sdks/typescript/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@honcho-ai/sdk",


src/crud/document.py
```diff
@@ -446,8 +549,16 @@ async def is_rejected_duplicate(
...
         )
+        vector_store = get_vector_store()
+        namespace = vector_store.get_document_namespace(
+            workspace_name, observer, observed
+        )
+        await vector_store.delete_many(namespace, [existing_doc.id])
+
+        # Delete from database after vector store succeeds
         await db.delete(existing_doc)
         await db.flush()  # Flush to make deletion visible in this transaction
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: added soft-deletion handling + retries here

src/vector_store/turbopuffer.py
```diff
@@ -0,0 +1,299 @@
...
+        ns = self._get_namespace(namespace)
+
+        rows: list[dict[str, Any]] = [
+            {
+                "id": v.id,
+                "vector": v.embedding,
+                **v.metadata,
+            }
+            for v in vectors
+        ]
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed in latest commit

src/crud/document.py
```diff
@@ -194,27 +241,70 @@ async def create_documents(
...
+                async for attempt in AsyncRetrying(
+                    stop=stop_after_attempt(3),
+                    wait=wait_exponential(multiplier=0.5, min=0.5, max=2.0),
+                    reraise=True,
+                ):
+                    with attempt:
+                        await vector_store.upsert_many(namespace, vector_records)
+            except Exception as e:
+                # Final attempt failed - log but don't raise
+                # Documents exist in DB, vectors can be added manually later
```
VVoruganti: How can we add a reconciliation layer to let honcho self-heal? Should this be a dreaming task. What could be the reasons for a query to fail that aren't systemic? Just have a lot of questions about this part tbh

VVoruganti: By splitting up the documents between our own table and the vector store we definitely need some kind of way to ensure they stay in sync

src/vector_store/__init__.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Args:
+            namespace: The namespace to delete
+        """
+        ...
+
+
+# Singleton instance
+_vector_store_instance: VectorStore | None = None
+
```
VVoruganti: This may not be asyncio and or thread safe

src/vector_store/lancedb.py
```diff
@@ -0,0 +1,345 @@
...
+from src.config import settings
+
+from . import QueryResult, VectorRecord, VectorStore
+
+logger = logging.getLogger(__name__)
+
+# Schema for LanceDB tables
+# id: string, vector: fixed_size_list of float32 (1536 dimensions for OpenAI embeddings)
+# Additional metadata columns are added dynamically
+VECTOR_DIMENSION = 1536
```
VVoruganti: Should maybe make this a top level config value

src/vector_store/__init__.py
```diff
@@ -0,0 +1,225 @@
...
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class QueryResult:
+    """A single result from a vector query."""
+
+    id: str
+    score: float  # Distance/similarity score (lower = more similar for cosine distance)
+    metadata: dict[str, Any] = field(default_factory=dict)
```
VVoruganti: Not super into these being top level dataclasses. Also don't think we need to have a `metadata` field since the actual entries and flattened and more tightly defined in both of the storage backends. think we could be more explicit here of having types for MessageEmbeddings and Documents with top level fields for validation. Could probably make these pydantic objects too

src/config.py
```diff
@@ -356,6 +357,36 @@ class DreamSettings(BackupLLMSettingsMixin, HonchoSettings):
...
+
+    # Vector store type: "turbopuffer" or "lancedb"
+    TYPE: Literal["turbopuffer", "lancedb"] = "lancedb"
+
+    # Global namespace prefix for all vector namespaces
+    # Namespaces follow the pattern:
+    # - Documents: {NAMESPACE}-{workspace}-{observer}-{observed}
+    # - Messages: {NAMESPACE}-{workspace}-messages
+    NAMESPACE: str = "honcho"
+
```
VVoruganti, line 550: Should add dimensions here as well

src/models.py
```diff
@@ -371,7 +379,6 @@ class Document(Base):
     times_derived: Mapped[int] = mapped_column(
         Integer, nullable=False, server_default=text("1")
     )
-    embedding: MappedColumn[Any] = mapped_column(Vector(1536))
```
VVoruganti, line 377: What is our data migration strategy here?

src/models.py
```diff
@@ -273,13 +271,21 @@ def __repr__(self) -> str:
 
 @final
```
VVoruganti, line 273: Does it make sense to even still have this table if we are moving things externally?

VVoruganti:
## Code Review: Consistency & Migration Concerns

I've reviewed this PR and the overall architecture is solid - clean abstractions, good use of retries, and the soft-delete pattern for documents is well thought out. However, I have some concerns about data consistency and migration paths that should be addressed before merging.

---

### 1. Missing Data Migration Path (High Priority)

The migration (`f1a2b3c4d5e6`) only makes schema changes (nullable columns, adds `chunk_index`, adds `deleted_at`) but **does not migrate existing embeddings** from PostgreSQL to the external vector store.

**Impact for existing deployments:**
- Existing embeddings in PostgreSQL become orphaned (column is nullable but data stays)
- The new code doesn't read from PostgreSQL vectors at all
- All semantic search queries will return empty results until data is re-created

**Suggested fix:** Add a data migration script that:
1. Reads existing embeddings from PostgreSQL
2. Writes them to the configured vector store
3. Optionally: drops the old HNSW indexes after migration

---

### 2. CREATE Flow: No Recovery for Failed Vector Upserts (Medium Priority)

When creating documents/messages, if the vector store upsert fails after retries, the code logs but doesn't raise:

```python
except Exception as e:
    # Final attempt failed - log but don't raise
    # Documents exist in DB, vectors can be added manually later
    logger.error(f"Failed to upsert vectors after retries: {e}")
```

**Problem:** Records exist in PostgreSQL but are invisible to semantic search, with no automated recovery mechanism.

**Suggested fix:** Track sync state and retry in background:
```python
# Add to models
vector_synced: Mapped[bool] = mapped_column(default=False)
vector_sync_attempts: Mapped[int] = mapped_column(default=0)

# Background job retries failed syncs
async def sync_unsynced_vectors():
    docs = await db.execute(
        select(Document)
        .where(Document.vector_synced == False)
        .where(Document.vector_sync_attempts < MAX_RETRIES)
    )
    # ... retry logic
```

---

### 3. Asymmetric Handling: Documents vs Messages (Medium Priority)

Documents have a soft-delete fallback pattern, but **messages do not**:

| Operation | Documents | Messages |
|-----------|-----------|----------|
| Delete fails | Soft delete → cleanup job | Silent failure (log only) |
| Create fails | Log only | Log only |

The message delete path in `delete_session` is best-effort:
```python
try:
    await vector_store.delete_many(namespace, vector_ids)
except Exception as e:
    # Log warning but continue
    logger.warning(f"Failed to delete message vectors...")
```

**Suggested fix:** Apply the same soft-delete pattern to `MessageEmbedding`, or document why the asymmetry is intentional.

---

### 4. No Consistency Audit Mechanism (Low Priority)

There's no way to detect if PostgreSQL and the vector store have drifted apart over time. Consider adding:

1. A CLI command or admin endpoint to audit consistency
2. Metrics for failed sync operations (not just logs)
3. A repair/reconciliation job that can be triggered manually

---

### Summary

The core implementation is good, but the lack of a migration path for existing data is a blocker for any production deployment with existing embeddings. The consistency concerns are lower priority but worth addressing for production hardening.

src/crud/session.py
```diff
@@ -420,6 +421,36 @@ async def delete_session(
...
+                await vector_store.delete_many(namespace, vector_ids)
+                logger.debug(
+                    f"Deleted {len(vector_ids)} message vectors for session {session_name}"
+                )
+            except Exception as e:
+                # Log warning but continue - workspace deletion will clean up eventually
+                logger.warning(
+                    f"Failed to delete message vectors for session {session_name}: {e}"
+                )
+
```
VVoruganti, line 468: Generally just have concerns about the synchronization state now that we don't have ACID guarantees from postgres. I think we need to be a little more hardened in our approach to ensure synchronization. Here are some comments from claude about potential for data to enter a bad state if either postgres or a vector store fails an operation but not both:

  Gaps & Risks

  | Scenario                    | Documents                  | Messages                    |
  |-----------------------------|----------------------------|-----------------------------|
  | Create: vector upsert fails | Orphaned in DB (no search) | Orphaned in DB (no search)  |
  | Delete: vector delete fails | Soft delete → cleanup job  | No fallback, silent failure |
  | Vector store down for hours | Accumulates orphans        | Accumulates orphans         |
  | Backfill orphaned records   | Not implemented            | Not implemented             |
  | Detect drift/inconsistency  | Not implemented            | Not implemented             |

  What's Missing

  1. No soft-delete for messages - Delete failures are silently logged:
  # Messages just log and continue
  except Exception:
      logger.exception("Failed to upsert message vectors after retries")
  2. No backfill/repair mechanism - If creates fail, there's no way to retry later:
  # After 3 retries, it just gives up
  except Exception as e:
      # Final attempt failed - log but don't raise
      # Documents exist in DB, vectors can be added manually later  ← "manually"?
      logger.error(f"Failed to upsert vectors after retries: {e}")
  3. No consistency audit - No way to detect if PostgreSQL and vector store have drifted
  4. Cleanup only handles deletes - The background job only processes soft-deleted documents, not failed creates

  Suggested Improvements

  # 1. Add a "vector_synced" flag to track sync state
  class Document(Base):
      vector_synced: Mapped[bool] = mapped_column(default=False)
      vector_sync_attempts: Mapped[int] = mapped_column(default=0)

  # 2. Background job to retry failed syncs
  async def sync_unsynced_vectors():
      docs = await db.execute(
          select(Document)
          .where(Document.vector_synced == False)
          .where(Document.vector_sync_attempts < MAX_RETRIES)
      )
      for doc in docs:
          try:
              await vector_store.upsert(...)
              doc.vector_synced = True
          except Exception:
              doc.vector_sync_attempts += 1

  # 3. Apply same pattern to messages (currently no fallback at all)

  The current implementation is optimistic - it assumes the vector store is generally available and failures are rare. For production hardening, you'd want the ability to detect and repair inconsistencies.

VVoruganti (changes requested):
I think we need to spend a little more time hardening this to ensure consistency and thinking through the migration strategy

src/models.py
Rajat-Ahuja1997: no we should eventually get rid of it

src/deriver/queue_manager.py
```diff
@@ -343,30 +348,88 @@ async def cleanup_queue_items(self) -> None:
...
+                    try:
+                        logger.info("Running vector reconciliation cycle")
+                        await self._run_reconciliation()
+                    except Exception:
+                        logger.exception("Error during vector reconciliation")
+                        if settings.SENTRY.ENABLED:
+                            sentry_sdk.capture_exception()
+                    next_vector_reconciliation = now + timedelta(
+                        seconds=RECONCILIATION_INTERVAL_SECONDS
+                    )
```
VVoruganti: should we add any claim logic so that two independent deriver machines don't try to run these jobs separately? Can we create specialized AQS records for maintenance tasks

src/vector_store/__init__.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        if namespace_type == "document":
+            if observer is None or observed is None:
+                raise ValueError(
+                    "observer and observed are required for document namespaces"
+                )
+            return f"{self.namespace_prefix}.{workspace_name}.{observer}.{observed}"
+        if namespace_type == "message":
+            return f"{self.namespace_prefix}.{workspace_name}.messages"
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: your proposed fix would be unreachable code

src/crud/workspace.py
```diff
@@ -312,6 +321,51 @@ async def delete_workspace(db: AsyncSession, workspace_name: str) -> schemas.Wor
...
+                        doc_namespace,
+                        collection.observer,
+                        collection.observed,
+                    )
+                except Exception as e:
+                    logger.warning(
+                        "Failed to delete document namespace %s: %s",
+                        doc_namespace,
+                        e,
+                    )
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: will track this in a follow up

src/vector_store/pgvector.py
```diff
@@ -0,0 +1,315 @@
...
+            observer = parts[2]
+            observed = parts[3]
+            return "documents", {
+                "workspace_name": workspace,
+                "observer": observer,
+                "observed": observed,
+            }
+        else:
+            raise ValueError(f"Invalid namespace format: {namespace}")
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: will take this as a note

src/vector_store/pgvector.py
```diff
@@ -0,0 +1,315 @@
...
+                    "workspace_name": row.workspace_name,
+                    "observer": row.observer,
+                    "observed": row.observed,
+                    "session_name": row.session_name,
+                    "level": row.level,
+                },
+            )
+            for row in rows
+        ]
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: we removed PgVectorStore

tests/utils/test_agent_tools.py
```diff
@@ -311,9 +312,14 @@ class TestSearchMemory:
...
 
     async def test_returns_matching_observations(
-        self, make_tool_context: Callable[..., ToolContext]
+        self,
+        make_tool_context: Callable[..., ToolContext],
+        monkeypatch: pytest.MonkeyPatch,
     ):
         """Returns observations matching semantic query."""
+        # Force pgvector queries since test documents are created directly in postgres
+        monkeypatch.setattr("src.config.settings.VECTOR_STORE.MIGRATED", False)
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed

src/crud/message.py
```diff
@@ -136,25 +137,114 @@ async def create_messages(
...
+                            await db.execute(
+                                update(models.MessageEmbedding)
+                                .where(models.MessageEmbedding.id.in_(embedding_ids))
+                                .values(
+                                    sync_attempts=models.MessageEmbedding.sync_attempts
+                                    + 1,
+                                    last_sync_at=func.now(),
+                                )
+                            )
+                            await db.commit()
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed this to always create MessageEmbedding in latest commits

src/reconciler/sync_vectors.py
```diff
@@ -0,0 +1,554 @@
...
+    for doc in documents:
+        if cast(list[float] | None, doc.embedding) is None:
+            missing_docs.append(doc)
+    reembedded_by_id: dict[str, list[float]] = {}
+
+    if missing_docs:
+        try:
+            # Re-embed all missing documents in one batch
+            contents = [doc.content for doc in missing_docs]
+            embeddings = await embedding_client.simple_batch_embed(contents)
```
VVoruganti: in what situation could this actually occur?

src/reconciler/sync_vectors.py
```diff
@@ -0,0 +1,554 @@
...
+
+        except Exception as e:
+            logger.warning(
+                f"Failed to sync documents to external vector store {namespace}: {e}"
+            )
+            # Increment attempts and mark as failed if we've hit max attempts
+            await _bump_document_sync_attempts(db, docs_with_vectors)
+            failed_count += len(docs_with_vectors)
+
+    return synced_count, failed_count
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: we don't need to have retries in a loop that is already constantly retrying over a long period of time

src/vector_store/lancedb.py
```diff
@@ -0,0 +1,351 @@
...
+                self._db = await lancedb.connect_async(self._db_path)
+        return self._db
+
+    async def _get_table(self, namespace: str) -> AsyncTable | None:
+        """Get a table if it exists, otherwise return None."""
+        db = await self._get_db()
+        table_names = await db.table_names()
+        if namespace in table_names:
+            return await db.open_table(namespace)
+        return None
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this paradigm is working consistently across turbopuffer, lancedb, and pgvector

src/reconciler/scheduler.py
```diff
@@ -0,0 +1,283 @@
...
+            db.add(queue_item)
+            try:
+                await db.commit()
+            except IntegrityError:
+                # Another instance already enqueued this task (unique constraint)
+                await db.rollback()
+                logger.debug(
+                    "Task %s already enqueued by another instance, skipping", task.name
+                )
+                return False
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: we have a partial unique index in the support external embeddings migration. we don't want a general unique index on this constraint because the queue can have multiple items with the same work_unit_key

src/reconciler/sync_vectors.py
```diff
@@ -0,0 +1,565 @@
...
+                external_vector_store,
+                batch_size=RECONCILIATION_BATCH_SIZE,
+            )
+            if cleaned:
+                metrics.documents_cleaned += cleaned
+                did_work = True
+
+            if not did_work:
+                logger.debug("No work done, breaking reconciliation loop")
+                break
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed

src/reconciler/scheduler.py
```diff
@@ -0,0 +1,283 @@
...
+
+            logger.info("Enqueued reconciler task: %s", task.name)
+            return True
+
+    async def _ensure_system_workspace(self, db: AsyncSession) -> None:
```
Rajat-Ahuja1997: @VVoruganti the reason we're creating a __system__ workspace is that `workspace_name` is a required field on QueueItems. what we could do to avoid this is queueing reconciliation tasks per workspace. lmk what you think. I don't think making `workspace_name` nullable on the QueueItem model is necessarily the answer here

src/reconciler/scheduler.py
```diff
@@ -0,0 +1,283 @@
...
+        """Ensure the system workspace exists for reconciler tasks."""
+        # Use upsert to create workspace if it doesn't exist
+        stmt = pg_insert(models.Workspace).values(name=SYSTEM_WORKSPACE_NAME)
+        stmt = stmt.on_conflict_do_nothing(index_elements=["name"])
+        await db.execute(stmt)
```
VVoruganti: Not super into this, I think maybe we can lift the restriction on workspace_name being required and add that to the db migration.

VVoruganti (changes requested):

src/reconciler/scheduler.py
VVoruganti: I think it's fine to allow the Queue to turn into a more general purpose thing. Would lean towards that than making custom internal resources

src/deriver/queue_manager.py
Rajat-Ahuja1997: resolved, we're using a reconciliation scheduler now

src/utils/work_unit.py
```diff
@@ -136,4 +142,17 @@ def parse_work_unit_key(work_unit_key: str) -> ParsedWorkUnit:
...
+                f"Invalid work_unit_key format for task_type {task_type}: {work_unit_key}"
+            )
+        return ParsedWorkUnit(
+            task_type=task_type,
+            workspace_name=None,
+            session_name=None,
+            observer=None,
+            observed=None,
+        )
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: we only really need the task_type though

migrations/versions/119a52b73c60_support_external_embeddings.py
```diff
@@ -0,0 +1,350 @@
...
+                DELETE FROM "{schema}".queue
+                WHERE workspace_name IS NULL
+                LIMIT :batch_size
+                """
+            ),
+            {"batch_size": batch_size},
+        )
+        if result.rowcount == 0:
+            break
+
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: fixed migration in latest commit.

VVoruganti (dismissed):

PR #312, opened 2026-01-05
fix: summary creation if start_message_id > end_message_id

Rajat-Ahuja1997:
## Race condition explanation
- When summary queue items are processed out of order (e.g., a queue item for message 3000 processes after the summary has message_id 5000), the code for `get_messages_id_range` will fetch an empty list of messages. we then make an LLM call to create a summary and save the message_id for the summary as `messages[-1].id if messages else 0` which gets stamped onto the session as 0. all future summarizations will then attempt to fetch the entire message history up until a given message_id.

## Summary by CodeRabbit

* **New Features**
  * Added sequence-aware message retrieval and integrated sequence-based summarization so summaries use message sequence ranges and configuration settings.

* **Bug Fixes**
  * Prevents redundant summary work by skipping when newer summaries exist.
  * Ensures configuration and message-sequence info propagate through the summary flow; logs and exits gracefully when no messages are available.

<h3>Greptile Summary</h3>

Added validation check to prevent summary creation when `start_message_id >= end_message_id`, which can occur when queue items are processed out of order and the summary is already ahead of the requested message ID. This prevents unnecessary database queries and LLM calls.

<h3>Confidence Score: 5/5</h3>

- This PR is safe to merge with minimal risk
- The fix correctly handles an edge case where queue items are processed out of order, preventing unnecessary database queries and LLM calls. The logic is sound (early return when start_id >= message_id), includes helpful debug logging, and follows the existing code patterns. No breaking changes or risky modifications.
- No files require special attention

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| src/utils/summarizer.py | Added early return to prevent redundant summary creation when messages are already summarized |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant BG as Background Queue
    participant SUM as _create_and_save_summary
    participant DB as Database
    participant CRUD as crud.get_messages_id_range
    
    Note over BG,CRUD: Out-of-order processing scenario
    
    BG->>SUM: Process message_id=100
    SUM->>DB: get_summary(workspace, session, type)
    DB-->>SUM: latest_summary (message_id=100)
    
    Note over SUM: start_id = 100<br/>requested message_id = 100
    
    alt start_id >= message_id (NEW CHECK)
        SUM->>SUM: Early return with debug log
        Note over SUM: Skip redundant summary creation
    else start_id < message_id
        SUM->>CRUD: get_messages_id_range(start_id, end_id)
        CRUD-->>SUM: messages list
        SUM->>SUM: Generate new summary
        SUM->>DB: Save summary to session metadata
    end
```

### DIFF
--- a/src/utils/summarizer.py
+++ b/src/utils/summarizer.py
@@ -376,6 +376,11 @@ async def _create_and_save_summary(
     summary_start = time.perf_counter()
 
     latest_summary = await get_summary(db, workspace_name, session_name, summary_type)
+    if latest_summary:
+        latest_summary_message_id = latest_summary["message_id"]
+        # Skip if latest summary already covers message.
+        if latest_summary_message_id >= message_id:
+            return
 
     previous_summary_text = latest_summary["content"] if latest_summary else None
 


dr-frmr (changes requested):
Not 100% certain but I think the logic here is flawed -- `message_id` in `_create_and_save_summary` is a global ID and not one that increments one-by-one within a session. Need to pass in `message_seq_in_session` or something, no?

Rajat-Ahuja1997:
> Not 100% certain but I think the logic here is flawed -- `message_id` in `_create_and_save_summary` is a global ID and not one that increments one-by-one within a session. Need to pass in `message_seq_in_session` or something, no?

you're correct, we can't just rely on subtracting by the config. i'll fix this

PR #323, opened 2026-01-13
Oolong Benchmark

3un01a:
In this PR we introduce the Oolong Benchmark which includes tests for both synthetic datasets + real datasets.
Linear: https://linear.app/plastic-labs/issue/AI-125/oolong

## Summary by CodeRabbit

* **New Features**
  * Added a full OOLONG benchmarking suite and CLI: dataset loaders, context/answer parsing, scoring, batching, multi-session/multi-observer execution, pooling for parallel runs, timing/metrics, and JSON summary output.

* **Chores**
  * Added pyarrow dependency to support parquet data handling.
  * Updated bench ignore list to exclude benchmark data artifacts.

* **Tests**
  * Test runner invocation changed to enforce single-process execution (-n0).

### DIFF
--- a/tests/bench/oolong_common.py
+++ b/tests/bench/oolong_common.py
@@ -13,7 +13,7 @@ import re
 from collections.abc import Callable, Sequence
 from datetime import date, datetime
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 
 import dateutil.parser
 import pyarrow.parquet as pq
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "redis>=7.0.0,<8.0.0",
     "cashews[redis]==7.4.4",
     "scikit-learn>=1.6.0",
+    "datasets>=4.4.2",
```
VVoruganti: add this to the dev dependency group shouldn't be necessary for production usage

tests/bench/oolong_common.py
```diff
@@ -0,0 +1,540 @@
...
+def load_oolong_real_dataset(split: str = "test") -> Any:
+    """Load the OOLONG-real dataset from HuggingFace.
+
+    Args:
+        split: Dataset split to load (default: "test")
+
+    Returns:
+        HuggingFace dataset object
+    """
+    return load_dataset("oolongbench/oolong-real", "dnd", split=split)
```
VVoruganti: I'd prefer if we stayed consistent in our data loading strategy, either we load datasets from huggingface or we have them locally, but now we'd have a hybrid setup

3un01a: Will stick to loading it directly from disk to maintain consistency with other benchmarks.

pyproject.toml
3un01a: Removing dependency since we are now loading from disk now.

VVoruganti (approved):

PR #327, opened 2026-01-14
feat: message ingestion forced-batching

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Token-based batching for representation tasks to improve processing efficiency and fairness.
  * Stronger final-response behavior so the model converges to a text answer after tool-iteration limits are reached.

* **Documentation**
  * Clarified queue-status docs to explain asynchronous reasoning, batching expectations, and minor wording/format tweaks.

* **Configuration**
  * Reduced default token batch-size threshold and adjusted low-level tool-choice default.

* **Tests**
  * Added tests for forced-batching behavior and a test fixture to ensure queue table cleanup.

* **Bug Fixes**
  * Enforced single pending dream task constraint to prevent duplicate pending work.

### DIFF
--- a/tests/deriver/conftest.py
+++ b/tests/deriver/conftest.py
@@ -1,17 +1,34 @@
 import asyncio
-from collections.abc import Awaitable, Callable, Sequence
+from collections.abc import AsyncGenerator, Awaitable, Callable, Sequence
 from datetime import datetime, timezone
 from typing import Any, Literal, TypeAlias, cast
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from nanoid import generate as generate_nanoid
-from sqlalchemy import select
+from sqlalchemy import delete, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from src import crud, models, schemas
 from src.utils.queue_payload import create_payload
 from src.utils.work_unit import construct_work_unit_key
+
+
+@pytest.fixture(autouse=True)
+async def clean_queue_tables(db_session: AsyncSession) -> AsyncGenerator[None, None]:
+    """Clean up queue-related tables before each test to ensure isolation.
+
+    This prevents webhook queue items and active queue sessions from previous tests
+    from polluting subsequent tests, which can cause issues when tests have a limit
+    on how many work units can be claimed (e.g., WORKERS=1).
+    """
+    # Clean up before the test
+    await db_session.execute(delete(models.ActiveQueueSession))
+    await db_session.execute(delete(models.QueueItem))
+    await db_session.commit()
+
+    yield
+
 
 QueuePayload: TypeAlias = dict[str, Any]
 QueuePayloadEntry: TypeAlias = QueuePayload | tuple[QueuePayload, int | None]


src/deriver/queue_manager.py
```diff
@@ -204,14 +204,40 @@ async def cleanup_stale_work_units(self) -> None:
...
+                        "max_single_token"
+                    ),
+                )
+                .join(
+                    models.Message,
+                    models.QueueItem.message_id == models.Message.id,
+                )
+                .where(~models.QueueItem.processed)
+                .where(models.QueueItem.work_unit_key.isnot(None))
+                .where(models.QueueItem.message_id.isnot(None))
```
VVoruganti: would these be covered automatically with a different type of join, like a specifically inner join or do we not want that?

src/deriver/queue_manager.py
```diff
@@ -220,9 +246,24 @@ async def get_and_claim_work_units(self) -> dict[str, str]:
...
                 .where(models.ActiveQueueSession.work_unit_key.is_(None))
+                # Forced batching: only claim representation work units if ready
+                .where(
+                    or_(
+                        # Non-representation tasks: process immediately
+                        ~models.QueueItem.work_unit_key.startswith("representation:"),
+                        # Accumulated tokens >= threshold
+                        token_stats_subq.c.total_tokens >= batch_max_tokens,
+                        # Single message >= threshold (prevents infinite wait)
+                        token_stats_subq.c.max_single_token >= batch_max_tokens,
```
VVoruganti: wouldn't this case already be covered by the total token stats? I don't see how the single token could be greater than the batch_max, but the total_token isn't. The actual message claiming logic happens later so it's not like we're determining which messages get handled in a batch

src/deriver/queue_manager.py
VVoruganti: Actually looking at SQLAlchemy the default join is already an inner join, so the message_id check is a bit redundant. The query planner would probably remove it though

src/deriver/queue_manager.py
```diff
@@ -204,14 +204,40 @@ async def cleanup_stale_work_units(self) -> None:
...
+            # Also track max single message token count (for single-large-message case)
+            token_stats_subq = (
+                select(
+                    models.QueueItem.work_unit_key,
+                    func.coalesce(func.sum(models.Message.token_count), 0).label(
+                        "total_tokens"
+                    ),
+                    func.coalesce(func.max(models.Message.token_count), 0).label(
+                        "max_single_token"
+                    ),
```
VVoruganti: Claude comments:

Similarly, the func.coalesce(..., 0) wrappers might be overly defensive since:
  - With an inner join, you'll always have matching Message rows
  - The GROUP BY means you'll always have at least one row contributing to the aggregate

  The only case where coalesce would matter is if token_count itself could be NULL on a Message, which might be a valid concern depending on your schema.

  Cleaner version:
```
  token_stats_subq = (
      select(
          models.QueueItem.work_unit_key,
          func.sum(models.Message.token_count).label("total_tokens"),
          func.max(models.Message.token_count).label("max_single_token"),
      )
      .join(
          models.Message,
          models.QueueItem.message_id == models.Message.id,
      )
      .where(~models.QueueItem.processed)
      .where(models.QueueItem.work_unit_key.isnot(None))
      .group_by(models.QueueItem.work_unit_key)
      .subquery()
  )
  ```

VVoruganti (changes requested):

VVoruganti (approved):

PR #333, opened 2026-01-19
Telemetry Overhaul

VVoruganti:
## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry replaces Prometheus with CloudEvents telemetry: batched emitter, retries, lifecycle hooks, and rich telemetry events.
  * Per-level reasoning and Dream specialist controls (deduction/induction, surprisal sampling) and expanded token/tool budgets.
  * Workspace/session deletion now returns cascade deletion counts.

* **Infrastructure**
  * Added Mimir and Grafana services; updated container mounts and volumes.

* **Configuration**
  * New OTEL/telemetry settings and environment variables; updated defaults for vector store, cache, and reasoning traces.

### DIFF
--- a/src/utils/summarizer.py
+++ b/src/utils/summarizer.py
@@ -88,7 +88,7 @@ class SummaryType(Enum):
 
 
 def short_summary_prompt(
-    messages: list[models.Message],
+    formatted_messages: str,
     output_words: int,
     previous_summary_text: str,
 ) -> str:
@@ -112,7 +112,7 @@ Return only the summary without any explanation or meta-commentary.
 </previous_summary>
 
 <conversation>
-{_format_messages(messages)}
+{formatted_messages}
 </conversation>
 
 Produce as thorough a summary as possible in {output_words} words or less.
@@ -120,7 +120,7 @@ Produce as thorough a summary as possible in {output_words} words or less.
 
 
 def long_summary_prompt(
-    messages: list[models.Message],
+    formatted_messages: str,
     output_words: int,
     previous_summary_text: str,
 ) -> str:
@@ -146,7 +146,7 @@ Return only the summary without any explanation or meta-commentary.
 </previous_summary>
 
 <conversation>
-{_format_messages(messages)}
+{formatted_messages}
 </conversation>
 
 Produce as thorough a summary as possible in {output_words} words or less.
@@ -159,7 +159,7 @@ def estimate_short_summary_prompt_tokens() -> int:
     try:
         return estimate_tokens(
             short_summary_prompt(
-                messages=[],
+                formatted_messages="",
                 output_words=0,
                 previous_summary_text="",
             )
@@ -175,7 +175,7 @@ def estimate_long_summary_prompt_tokens() -> int:
     try:
         return estimate_tokens(
             long_summary_prompt(
-                messages=[],
+                formatted_messages="",
                 output_words=0,
                 previous_summary_text="",
             )
@@ -187,7 +187,7 @@ def estimate_long_summary_prompt_tokens() -> int:
 
 @conditional_observe(name="Create Short Summary")
 async def create_short_summary(
-    messages: list[models.Message],
+    formatted_messages: str,
     input_tokens: int,
     previous_summary: str | None = None,
 ) -> HonchoLLMCallResponse[str]:
@@ -203,7 +203,9 @@ async def create_short_summary(
     else:
         previous_summary_text = "There is no previous summary -- the messages are the beginning of the conversation."
 
-    prompt = short_summary_prompt(messages, output_words, previous_summary_text)
+    prompt = short_summary_prompt(
+        formatted_messages, output_words, previous_summary_text
+    )
 
     return await honcho_llm_call(
         llm_settings=settings.SUMMARY,
@@ -214,7 +216,7 @@ async def create_short_summary(
 
 @conditional_observe(name="Create Long Summary")
 async def create_long_summary(
-    messages: list[models.Message],
+    formatted_messages: str,
     previous_summary: str | None = None,
 ) -> HonchoLLMCallResponse[str]:
     # the word/token ratio is roughly 4:3 so we multiply by 0.75.
@@ -226,7 +228,9 @@ async def create_long_summary(
     else:
         previous_summary_text = "There is no previous summary -- the messages are the beginning of the conversation."
 
-    prompt = long_summary_prompt(messages, output_words, previous_summary_text)
+    prompt = long_summary_prompt(
+        formatted_messages, output_words, previous_summary_text
+    )
 
     return await honcho_llm_call(
         llm_settings=settings.SUMMARY,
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -180,22 +191,20 @@ async def lifespan(_: FastAPI):
...
-        template = get_route_template(request)
-        prometheus.API_REQUESTS.labels(
+    template = get_route_template(request)
+    if settings.OTEL.ENABLED:
+        otel_metrics.record_api_request(
             method=request.method,
             endpoint=template,
             status_code=str(exc.status_code),
-        ).inc()
+        )
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this might be valid

src/config.py
```diff
@@ -439,10 +441,69 @@ class WebhookSettings(HonchoSettings):
...
+
+    # Master toggle for OTel metrics
     ENABLED: bool = False
+
+    # OTLP HTTP endpoint for metrics (e.g., "https://mimir.example.com/otlp/v1/metrics")
+    # For Mimir, the endpoint is typically: <mimir-url>/otlp/v1/metrics
+    # For Grafana Cloud: https://otlp-gateway-<region>.grafana.net/otlp/v1/metrics
+    ENDPOINT: str | None = None
+
+    # Optional headers for authentication (e.g., {"X-Scope-OrgID": "tenant"})
```
Rajat-Ahuja1997: nit: do we want to explicitly mention the concept of `tenant` here? we typically haven't done that in honcho

src/deriver/deriver.py
```diff
@@ -209,3 +217,24 @@ async def process_representation_tasks_batch(
...
 
     log_performance_metrics("minimal_deriver", f"{latest_message.id}_{observer}")
+
+    # Emit telemetry event
+    emit(
+        RepresentationCompletedEvent(
+            workspace_id=latest_message.workspace_name,
+            workspace_name=latest_message.workspace_name,
+            session_id=latest_message.session_name,
+            session_name=latest_message.session_name,
```
Rajat-Ahuja1997: do we need both workspace_id/workspace_name and session_id/session_name if the pairs are being set to the same value?

src/deriver/consumer.py
```diff
@@ -276,9 +304,29 @@ async def process_reconciler(payload: ReconcilerPayload) -> None:
...
+        # Emit telemetry event
+        emit(
+            SyncVectorsCompletedEvent(
+                documents_synced=metrics.documents_synced,
+                documents_failed=metrics.documents_failed,
+                message_embeddings_synced=metrics.message_embeddings_synced,
+                message_embeddings_failed=metrics.message_embeddings_failed,
+                total_duration_ms=duration_ms,
+            )
+        )
```
Rajat-Ahuja1997: do we want to emit when nothing has synced or failed? it could be useful as a heartbeat but otherwise it feels like telemetry noise.

src/utils/summarizer.py
```diff
@@ -469,14 +479,37 @@ async def _create_and_save_summary(
...
+    # Emit telemetry event (only for non-fallback summaries)
+    # Note: Using AgentToolSummaryCreatedEvent with dummy run_id/iteration since
+    # this is called from the deriver, not from an agentic loop
+    if not is_fallback:
+        emit(
+            AgentToolSummaryCreatedEvent(
+                run_id="deriver",  # Placeholder - not from an agentic run
+                iteration=0,  # Placeholder - not from an agentic loop
+                parent_category="deriver",
+                agent_type="summarizer",
```
Rajat-Ahuja1997, line 499: double checking on if agent_type should be set bc this isn't called from an agentic loop.

src/dialectic/core.py
```diff
@@ -308,19 +325,42 @@ def _log_response_metrics(
...
+        # Emit telemetry event
+        emit(
+            DialecticCompletedEvent(
+                run_id=self._run_id,
+                workspace_id=self.workspace_name,
+                workspace_name=self.workspace_name,
+                peer_id=self.observed,
+                peer_name=self.observed,
+                session_id=self.session_name,
+                session_name=self.session_name,
```
Rajat-Ahuja1997, line 349: same comment regarding session_id/session_name and workspace_id/workspace_name

src/routers/workspaces.py
```diff
@@ -114,7 +115,22 @@ async def delete_workspace(
...
     This action cannot be undone.
     """
-    await crud.delete_workspace(db, workspace_name=workspace_id)
+    result = await crud.delete_workspace(db, workspace_name=workspace_id)
+
+    # Emit telemetry event with cascade counts
+    emit(
+        DeletionCompletedEvent(
+            workspace_id=workspace_id,
+            workspace_name=workspace_id,
```
Rajat-Ahuja1997, line 123: we probably want to emit an event for session deletions as well

src/dreamer/orchestrator.py
```diff
@@ -199,6 +249,51 @@ async def run_dream(
...
+    total_output_tokens = (
+        deduction_result.output_tokens if deduction_result else 0
+    ) + (induction_result.output_tokens if induction_result else 0)
+
+    # Emit DreamRunEvent with aggregated metrics
+    emit(
+        DreamRunEvent(
+            run_id=run_id,
+            workspace_id=workspace_name,
+            workspace_name=workspace_name,
```
Rajat-Ahuja1997, line 267: we are emitting `DreamRunEvent in both `run_dream` and process_dream`, is that intentional?

src/deriver/consumer.py
```diff
@@ -276,9 +304,29 @@ async def process_reconciler(payload: ReconcilerPayload) -> None:
...
+
     elif reconciler_type == ReconcilerType.CLEANUP_QUEUE:
         logger.debug("Processing cleanup_queue task")
         await cleanup_queue_items()
 
+        duration_ms = (time.perf_counter() - start_time) * 1000
+
+        # Emit telemetry event for cleanup stale items
+        emit(
+            CleanupStaleItemsCompletedEvent(
```
Rajat-Ahuja1997, line 336: should we populate the queue_items_cleaned field here? we should be able to return the rowcount from the db call

src/deriver/consumer.py
```diff
@@ -276,9 +304,29 @@ async def process_reconciler(payload: ReconcilerPayload) -> None:
...
             )
 
+        # Emit telemetry event
+        emit(
+            SyncVectorsCompletedEvent(
+                documents_synced=metrics.documents_synced,
+                documents_failed=metrics.documents_failed,
+                message_embeddings_synced=metrics.message_embeddings_synced,
+                message_embeddings_failed=metrics.message_embeddings_failed,
+                total_duration_ms=duration_ms,
```
Rajat-Ahuja1997: should add `documents_cleaned` (cleaning up soft-deleted records)

src/telemetry/logging.py
```diff
@@ -229,8 +229,7 @@ def log_performance_metrics(
         return
     if metrics is None:
         metrics = []
-    metrics = accumulated_metrics.get(task_name, []) + metrics
-    accumulated_metrics[task_name].clear()
+    metrics = accumulated_metrics.pop(task_name, []) + metrics
```
Rajat-Ahuja1997, line 258: this helps but we still have the smaller memory leak issue of metrics staying in the dict if an exception is thrown before `log_performance_metrics` is called. could fix or add a TODO for later

src/dreamer/orchestrator.py
VVoruganti: fixed in latest commit

src/deriver/consumer.py
VVoruganti: Yea don't want to extra noise, especially since these happen continuously moving into the if statement.

src/utils/summarizer.py
VVoruganti: this is a temporary workaround, eventually we want to move the summarizer to be a part of the dreamer

PR #347, opened 2026-01-26
feat: make session_name nullable for documents and update related SDKs

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Observations and conclusions can be created without a session, enabling global/sessionless items
  * Dreams can be scheduled without specifying a session

* **Database**
  * Schema migration enables nullable session associations for documents and related models

* **SDKs**
  * Python and TypeScript SDKs updated to accept optional session identifiers

* **Tests**
  * New tests covering sessionless creation, mixed batches, and migration behavior

### DIFF
--- a/migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
+++ b/migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
@@ -4,14 +4,16 @@ Allow dreams to run without a session_id by making the session_name
 column nullable on the documents table.
 
 Revision ID: e4eba9cfaa6f
-Revises: 7c0d9a4e3b1f
+Revises: 119a52b73c60
 Create Date: 2026-01-26
 
 """
 
 from collections.abc import Sequence
 
+import sqlalchemy as sa
 from alembic import op
+from nanoid import generate as generate_nanoid
 
 from migrations.utils import get_schema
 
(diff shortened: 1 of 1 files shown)


migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
```diff
@@ -0,0 +1,37 @@
...
+down_revision: str | None = "119a52b73c60"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+schema = get_schema()
+
+
+def upgrade() -> None:
+    """Make session_name nullable on documents table."""
+    op.alter_column("documents", "session_name", nullable=True, schema=schema)
```
VVoruganti, line 75: do we want a constraint that explicit conclusions do require a session name? I think that would be a good in-between compromise here

migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
```diff
@@ -0,0 +1,37 @@
+"""make document session_name nullable for sessionless dreams
+
+Allow dreams to run without a session_id by making the session_name
+column nullable on the documents table.
+
+Revision ID: e4eba9cfaa6f
+Revises: 7c0d9a4e3b1f
```
Rajat-Ahuja1997: think this should be `Revises: 119a52b73c60`

src/dreamer/orchestrator.py
```diff
@@ -114,9 +114,13 @@ async def run_dream(
         f"[{run_id}] Starting dream cycle for {workspace_name}/{observer}/{observed}"
     )
 
-    session = await crud.get_session(
-        db, workspace_name=workspace_name, session_name=session_name
-    )
+    if session_name is not None:
+        session = await crud.get_session(
+            db, workspace_name=workspace_name, session_name=session_name
+        )
```
VVoruganti, line 120: should throw a error if the session is not valid

VVoruganti: unless it is validated earlier

dr-frmr: that get_session function will throw, it's not get_or_creating

migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
```diff
@@ -0,0 +1,37 @@
...
+    """Make session_name nullable on documents table."""
+    op.alter_column("documents", "session_name", nullable=True, schema=schema)
+
+
+def downgrade() -> None:
+    """Make session_name non-nullable again, backfilling NULLs first."""
+    op.execute(
+        f"UPDATE \"{schema}\".documents SET session_name = '__orphaned__' WHERE session_name IS NULL"
+    )
+    op.alter_column("documents", "session_name", nullable=False, schema=schema)
```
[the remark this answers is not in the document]
Rajat-Ahuja1997: this is valid, we would need to create a `__orphaned__` session in each workspace

migrations/versions/e4eba9cfaa6f_make_document_session_name_nullable.py
dr-frmr: true

PR #359, opened 2026-02-02
fix: N+1 query in search_messages

Rajat-Ahuja1997:
## Release Notes

* **Refactor**
  * Optimized message context retrieval to improve performance when processing multiple message ranges.

### DIFF
--- a/src/crud/message.py
+++ b/src/crud/message.py
@@ -3,7 +3,7 @@ from logging import getLogger
 from typing import Any
 
 from nanoid import generate as generate_nanoid
-from sqlalchemy import ColumnElement, Select, and_, func, select, text, update
+from sqlalchemy import ColumnElement, Select, and_, func, or_, select, text, update
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from src import models, schemas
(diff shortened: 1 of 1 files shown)


src/crud/message.py
```diff
@@ -109,18 +109,31 @@ async def _build_merged_snippets(
...
+        context_result = await db.execute(context_stmt)
+        all_context_messages = list(context_result.scalars().all())
 
+        # Partition results back into their respective ranges
+        for start_seq, end_seq, range_matches in merged_ranges:
+            context_messages = [
+                msg
+                for msg in all_context_messages
+                if start_seq <= msg.seq_in_session <= end_seq
+            ]
```
Rajat-Ahuja1997, line 136: Note: we trade multiple DB round-trips increased CPU usage by scanning all_context_messages per range in merged_ranges, but I think this should be pretty negligible with our caps on # of ranges (<20)

dr-frmr: good catch, we should enforce that limit somewhere, usually it's closer to 5 so this should be totally fine

dr-frmr (approved):

PR #361, opened 2026-02-02
fix: make FLUSH_ENABLED a config value

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **New Features**
  * Added a configurable "deriver flush" flag (config + env) to allow immediate processing when enabled.

* **Refactor**
  * Removed Redis-based signaling for forcing deriver flush; queue flush now follows local configuration rather than Redis.

* **Chores**
  * Simplified test and benchmark runners and CLI by removing Redis flush-related arguments and calls.


src/cache/client.py
```diff
@@ -128,16 +125,8 @@ async def is_deriver_flush_enabled() -> bool:
...
-        parsed = urlparse(settings.CACHE.URL)
-        clean_url = urlunparse(parsed._replace(query=""))
-
-        redis_client = aioredis.from_url(clean_url)  # pyright: ignore[reportUnknownMemberType]
-        try:
-            result = await redis_client.get(DERIVER_FLUSH_KEY)
-            return result == b"1"
-        finally:
-            await redis_client.aclose()
+        result = await cache.get(DERIVER_FLUSH_KEY)
```
Rajat-Ahuja1997: right now this is a global key for all consumers of the cache—is that the expected behavior? if we want to enable flush mode for individual namespaces we'll need to prepend the key with the cache namespace

tests/bench/harness.py
```diff
@@ -489,6 +489,8 @@ def start_deriver(self) -> subprocess.Popen[str]:
         # Create environment with instance-specific database connection
         env = os.environ.copy()
         env.update(self.get_database_env_vars())
+        # Enable flush mode for tests - process messages immediately without waiting for batch threshold
+        env["DERIVER_FLUSH_ENABLED"] = "true"
```
Rajat-Ahuja1997, line 493: this is currently fine since all unified tests use flush: true right now, but if we ever need tests that run with flush disabled, I think we'd need to use a redis-based approach (otherwise I don't know a way to alter the config in our subprocesses during runtime)

VVoruganti: I don't think we really have a case of needing to change the config on a test by test basis, but if there is we can engineer a solution around it

PR #371, opened 2026-02-05
sdks: add set peer card function

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Added getCard/get_card and setCard/set_card APIs with deprecated wrappers to retrieve and update peer cards, including target-specific updates.

* **Behavior Changes**
  * Peer-card operations now ensure observer peers exist and validate peers before processing to avoid not-found issues.

* **Documentation**
  * Compatibility guide recommends installing latest patches; SDK docs include peer-card usage and migration notes.

* **Tests**
  * New unit and integration tests covering peer-card set/get and endpoint behavior.

* **Chores**
  * Bumped SDK versions to 2.0.1; pre-commit test hook now stops on first failure.

### DIFF
--- a/src/crud/peer_card.py
+++ b/src/crud/peer_card.py
@@ -6,7 +6,7 @@ from typing import cast
 from sqlalchemy import update
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from src import models, schemas
+from src import exceptions, models, schemas
 from src.cache.client import cache
 from src.crud.peer import get_or_create_peers, get_peer, peer_cache_key
 
@@ -84,7 +84,11 @@ async def set_peer_card(
             )
         )
     )
-    await db.execute(stmt)
+    result = await db.execute(stmt)
+    if result.rowcount == 0:
+        raise exceptions.ResourceNotFoundException(
+            f"Peer {observer} not found in workspace {workspace_name}"
+        )
     await db.commit()
 
     # Invalidate cache - read-through pattern
--- a/src/routers/peers.py
+++ b/src/routers/peers.py
@@ -338,11 +338,7 @@ async def set_peer_card(
         observed=observed,
     )
 
-    # Return the updated peer card
-    peer_card = await crud.get_peer_card(
-        db, workspace_id, observer=peer_id, observed=observed
-    )
-    return schemas.PeerCardResponse(peer_card=peer_card)
+    return schemas.PeerCardResponse(peer_card=peer_card_data.peer_card)
 
 
 @router.get(


docs/v3/documentation/reference/sdk.mdx
```diff
@@ -396,6 +408,46 @@ const searchedContext = await alice.context({
...
+const updated = await alice.setCard(["Likes TypeScript", "Lives in NYC"]);
+
+// Set card about another peer
+const updatedBob = await alice.setCard(["Works at Acme", "Enjoys hiking"], "bob");
+```
+</CodeGroup>
+
+<Info>
+Peer cards are automatically maintained by the deriver agent during message processing. Use `set_card()` / `setCard()` when you need to manually override or seed the card — the peer will be created automatically if it doesn't already exist.
+</Info>
```
VVoruganti: should say dreaming agent

src/crud/peer_card.py
```diff
@@ -34,16 +34,16 @@ async def get_peer_card(
...
+    result = await get_or_create_peers(
+        db, workspace_name, [schemas.PeerCreate(name=observer)]
+    )
+    peer = result.resource[0]
+    return cast(
+        list[str] | None,
+        peer.internal_metadata.get(
+            construct_peer_card_label(observer=observer, observed=observed)
+        ),
+    )
```
VVoruganti: So I think our standards are that a get function should return null if the underlying resource doesn't exist and shouldn't call a get_or_create. Our other setter methods like set_config and set_metadata call a get_or_create but I don't think the getters do. Let's keep the old behavior

VVoruganti (changes requested):

src/crud/peer_card.py
dr-frmr: done!

src/crud/peer_card.py
```diff
@@ -34,16 +34,19 @@ async def get_peer_card(
...
-            ),
-        )
-    except exceptions.ResourceNotFoundException:
+    peer = await db.scalar(
+        select(models.Peer)
+        .where(models.Peer.workspace_name == workspace_name)
+        .where(models.Peer.name == observer)
+    )
+    if peer is None:
         return None
```
VVoruganti: use the get_peer function to make use of the redis cache

src/crud/peer_card.py
```diff
@@ -34,16 +34,19 @@ async def get_peer_card(
...
     """
-    try:
-        peer = await get_peer(db, workspace_name, schemas.PeerCreate(name=observer))
-        return cast(
-            list[str] | None,
-            peer.internal_metadata.get(
-                construct_peer_card_label(observer=observer, observed=observed)
-            ),
-        )
-    except exceptions.ResourceNotFoundException:
```
VVoruganti, line 45: I actually think it is accurate to send a resource not found exception in this case rather than failing silently.

tests/crud/test_peer_card.py
```diff
@@ -50,27 +49,11 @@ async def test_peer_card_get_set_roundtrip(
...
-            observed="missing-peer",
-        )
-
-
 @pytest.mark.asyncio
 async def test_get_peer_card_missing_peer_returns_none(
     db_session: AsyncSession, sample_data: tuple[models.Workspace, models.Peer]
 ):
-    """Getting a peer card for a non-existent peer should return None."""
+    """Getting a peer card for a non-existent peer should return None while creating the peer."""
```
VVoruganti: Docstring innaccurate should not create the peer

PR #372, opened 2026-02-06
perf: add redis caching and reduce embedding API calls

Rajat-Ahuja1997:
- Redis caching for collection + ~~session peer configs~~
  - Session soft-delete cache invalidation fix
  - Embedding deduplication (prefetch, search_memory fallback)
  - Batch embedding in create_observations and extract_preferences

## Summary by CodeRabbit

* **New Features**
  * Exposed collection metadata update API at package level.
  * Observation creation supports batch embeddings with per-item fallback and richer result reporting.
  * Search/memory APIs accept optional pre-computed embeddings.
  * Added ObservationInput schema with validation.

* **Performance**
  * Collection and session caching with immediate post-commit caching and targeted invalidation.
  * Embedding batching and reuse to reduce duplicate calls.

* **Reliability**
  * Transient-safe cache helpers with retry/backoff and broader cache invalidation on deletes.

* **Tests**
  * Added tests for batch-embedding fallback and partial-success handling.

### DIFF
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -12,7 +12,6 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from src import crud, models, schemas
 from src.config import settings
 from src.embedding_client import embedding_client
-from src.exceptions import ValidationException
 from src.models import Document
 from src.schemas import ResolvedConfiguration
 from src.telemetry.events import (
@@ -628,8 +627,8 @@ async def create_observations(
         )
     except Exception as e:
         logger.warning(
-            "Batch embedding failed for create_observations; "
-            + f"falling back to per-observation embedding: {e}"
+            "Batch embedding failed for create_observations; falling back to per-observation embedding: %s",
+            e,
         )
 
     # Build document objects with pre-computed embeddings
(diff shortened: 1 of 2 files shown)


src/utils/agent_tools.py
```diff
@@ -619,8 +621,42 @@ async def create_observations(
         else:
             level = "explicit"
 
-        # Generate embedding for the observation
-        embedding = await embedding_client.embed(content)
+        valid_observations.append((obs, level))
+        contents.append(content)
```
VVoruganti: nitpick but I think validation should be done with pydantic if possible

src/utils/agent_tools.py
```diff
@@ -862,6 +901,20 @@ async def extract_preferences(
...
+    try:
+        query_embeddings = await embedding_client.simple_batch_embed(semantic_queries)
+        query_embeddings_by_query = dict(
+            zip(semantic_queries, query_embeddings, strict=True)
+        )
+    except Exception as e:
+        logger.warning(
+            "Batch embedding failed for extract_preferences; "
+            + f"falling back to per-query embedding in search_messages: {e}"
+        )
```
VVoruganti, line 920: should add a return type to this of ObservationsCreatedResult or something like that that reports if all of the observations were created or not so that we report any silent failures if any of the list of requested observations was malformed in some way

PR #378, opened 2026-02-10
feat: implement async workspace deletion with active session checks

dr-frmr:
- Updated the DELETE /workspaces/:id endpoint to return 202 Accepted, indicating that the deletion request is processed in the background.
- Added a check for active sessions before allowing workspace deletion, raising a ConflictException if any exist.
- Updated related tests to ensure proper handling of active sessions during workspace deletion.

## Summary by CodeRabbit

* **New Features**
  * Workspace deletion now runs asynchronously in the background and returns 202 Accepted with a confirmation message.

* **Bug Fixes**
  * Deletion is blocked while active sessions exist; you must delete or deactivate sessions first.
  * Cascading cleanup of related workspace data occurs once session preconditions are met.

### DIFF
--- a/src/crud/workspace.py
+++ b/src/crud/workspace.py
@@ -3,7 +3,7 @@ from logging import getLogger
 from typing import Any
 
 from cashews import NOT_NONE
-from sqlalchemy import Select, delete, func, select
+from sqlalchemy import Select, delete, exists, func, select
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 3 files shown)


src/crud/workspace.py
```diff
@@ -227,6 +227,32 @@ async def update_workspace(
...
+                models.Session.workspace_name == workspace_name,
+                models.Session.is_active == True,  # noqa: E712
+            )
+        )
+        or 0
+    )
+    if active_sessions_count > 0:
+        raise ConflictException(
+            f"Cannot delete workspace '{workspace_name}': {active_sessions_count} active session(s) remain. Delete all sessions first."
+        )
```
Rajat-Ahuja1997: nit: you can do `result = db.scalar(exists().where...` here and then just check if result is true. more efficient

src/crud/workspace.py
```diff
@@ -250,6 +276,9 @@ async def delete_workspace(
         logger.warning("Workspace %s not found", workspace_name)
         raise ResourceNotFoundException()
 
+    # Check for active sessions — workspace cannot be deleted while active sessions exist
+    await check_no_active_sessions(db, workspace_name)
+
```
Rajat-Ahuja1997: nit: i think we should remove `check_no_active_sessions` from `crud.delete_workspace` and only keep it in the router. the router enforces that the user needs to delete their sessions before deleting the workspace. this check in the crud function only catches the window between the user triggering deletion and the deletion being processed by the queue, and if we throw a Conflict Exception there, it just marks the queue item as errored and keeps the workspace around, with no signaling to the user.

VVoruganti (approved):

PR #383, opened 2026-02-11
Parallelize Async Calls

VVoruganti:
## Release Notes

* **Performance Improvements**
  * Optimized session context retrieval with concurrent data fetching for improved responsiveness
  * Refined token budget management for context summaries and messages
  * Enhanced efficiency of context message retrieval operations

* **Tests**
  * Strengthened test fixture reliability and data isolation for more robust testing

### DIFF
--- a/src/routers/sessions.py
+++ b/src/routers/sessions.py
@@ -1,5 +1,5 @@
-import asyncio
 import logging
+from contextlib import suppress
 
 from fastapi import APIRouter, Body, Depends, Path, Query, Response
 from fastapi_pagination import Page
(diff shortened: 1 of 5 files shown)


src/routers/sessions.py
```diff
@@ -592,38 +672,35 @@ async def get_session_context(
...
-        search_top_k=search_top_k,
-        search_max_distance=search_max_distance,
-        include_most_derived=include_most_frequent,
-        max_observations=max_conclusions,
-        embedding=embedding,
-    )
-    card = await _get_peer_card_task(
-        db, workspace_id, observer=observer, observed=observed
+    # Phase 1: Fetch representation, card, and summaries concurrently — all independent
+    representation, card, (short_summary, long_summary) = await asyncio.gather(
```
Rajat-Ahuja1997: this change makes sense but is incompatible with my changes in https://github.com/plastic-labs/honcho/pull/380 which use a single AsyncSession to process the tasks sequentially. arguments for this PR is that we parallelize the work but are opening 3 db connections. in mine we're processing work sequentially but with a single db connection. seems like we should either keep this PR or 380 but not both but will let you decide what's best

PR #386, opened 2026-02-13
fix: Various Codex Audits

VVoruganti:
## Summary by CodeRabbit

* **Chores**
  * Switched startup and runtime messages to the standard logging system; removed printing of the admin JWT at startup.
  * Improved metrics summary formatting and logging; added safety around metrics file cleanup.

* **Tests**
  * Added extensive unit tests (dependencies, webhook delivery, queue cleanup, reconciliation, client/session helpers).
  * Introduced runtime-mock gating and pytest warning filters to reduce noisy warnings.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -364,18 +364,18 @@ async def sample_data(
     # Create test app
     test_workspace = models.Workspace(name=str(generate_nanoid()))
     db_session.add(test_workspace)
-    await db_session.flush()
 
     # Create test user
     test_peer = models.Peer(
         name=str(generate_nanoid()), workspace_name=test_workspace.name
     )
     db_session.add(test_peer)
-    await db_session.flush()
+
+    # Commit so data is visible to independent tracked_db sessions.
+    # _truncate_all_tables handles cleanup between tests.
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -67,6 +67,29 @@ def emit(self, record: logging.LogRecord):
 )
 CONNECTION_URI = make_url(DB_URI)
 
+_RUNTIME_MOCK_TEST_PREFIXES = (
```
Rajat-Ahuja1997: imo given that most tests do want these mocks i feel we should use a blocklist instead of an allowlist

Rajat-Ahuja1997 (commented):
lgtm although i left a comment on our runtime mocks allowlist. 

codex also had this one note: [P2] Print metrics summary instead of info-level logs — /Users/rajatahuja/plastic/honcho/src/telemetry/metrics_collector.py:236-239
    In print_summary, the output was changed from print(...) to logger.info(...), but the benchmark runners configure logging at
    WARNING by default (tests/bench/runner_common.py), so this method now emits nothing in the normal benchmark path.

PR #425, opened 2026-03-14
fix: optimizes prefix-based caching

adavyas:
## Summary

Improves prompt prefix caching by separating stable system instructions from higher-variance request content.

This keeps the cacheable prefix stable across repeated calls and adds supporting metrics/tests for cache behavior.

## What Changed

- split deriver prompt construction into stable system prompt + variable user prompt
- improve summary prompt construction to use stable system instructions plus request-specific user content
- add cache-control normalization for cacheable system blocks in client handling
- add Gemini cached-token extraction to telemetry response handling
- add cache comparison tooling for prefix-cache evaluation
- add/update tests around:
  - client cache-control behavior
  - summarizer prompt structure
  - cache token extraction

## Why

Previously, stable instructions and changing context were more tightly coupled. That reduces prefix-cache reuse when only part of the prompt changes.

This PR isolates stable prompt sections so provider-side prompt caching can reuse more of the unchanged prefix.

## Notes

- no functional feature change is intended beyond prompt structuring and cache-related telemetry
- branch-specific CI/type-check fixes were included where needed

## Testing

Verified locally with:
- `py_compile` on changed files
- prompt/caching structure checks

## Summary by CodeRabbit

* **New Features**
  * Add a comparison tool to probe prompt-prefix cache behavior across code variants, producing per-call metrics, human-readable results, delta summaries, and optional JSON export.
  * Surface improved cache read/create metrics for additional providers (including Gemini).

* **Chores**
  * Move many prompt flows to explicit system/user message pairs for derivation, summarization, and response construction.
  * Stop mutating base system content; append session history as a separate message.

* **Tests**
  * Add tests covering cache token extraction, system-message cache metadata, and message-based summary creation.


src/deriver/prompts.py
```diff
@@ -41,10 +30,17 @@ def minimal_deriver_prompt(
 - Contextualize each observation sufficiently (e.g. "Ann is nervous about the job interview at the pharmacy" not just "Ann is nervous")
 
 EXAMPLES:
-- EXPLICIT: "I just had my 25th birthday last Saturday" → "{peer_id} is 25 years old", "{peer_id}'s birthday is June 21st"
-- EXPLICIT: "I took my dog for a walk in NYC" → "{peer_id} has a dog", "{peer_id} lives in NYC"
-- EXPLICIT: "{peer_id} attended college" + general knowledge → "{peer_id} completed high school or equivalent"
+- EXPLICIT: "I just had my 25th birthday last Saturday" → "{peer_id} is 25 years old", "{peer_id} celebrated their birthday last Saturday"
+- EXPLICIT: "I took my dog for a walk in NYC" → "{peer_id} has a dog", "{peer_id} said they took their dog for a walk in NYC"
+- EXPLICIT: "{peer_id} attended college" → "{peer_id} said they attended college"
```
Rajat-Ahuja1997: i would not make these example changes because the goal of ther deriver is to extract meaningful explicit facts + not to repeat back the raw message

src/deriver/prompts.py
```diff
@@ -11,27 +11,16 @@
...
-    Generate minimal prompt for fast observation extraction.
-
-    Args:
-        peer_id: The ID of the user being analyzed.
-        messages: All messages in the range (interleaving messages and new turns combined).
-
-    Returns:
-        Formatted prompt string for observation extraction.
-    """
+def minimal_deriver_system_prompt(peer_id: str) -> str:
```
Rajat-Ahuja1997: keeping the `peer_id` in the system prompt results in the system prompt only being cached for deriver calls to the same peer. we could move the peer context to minimal_deriver_user_prompt but not sure if the deriver would perform worse in this case. what do you think @VVoruganti

src/utils/clients.py
```diff
@@ -92,6 +92,128 @@ def count_message_tokens(messages: list[dict[str, Any]]) -> int:
...
+                    "cache_control", {"type": "ephemeral"}
+                )
+            normalized_blocks.append(normalized_block)
+        return normalized_blocks
+
+    return []
+
+
+def _build_gemini_contents_from_messages(
+    messages: list[dict[str, Any]],
```
Rajat-Ahuja1997, line 128: should we use this in the streaming gemini path too?

Rajat-Ahuja1997 (changes requested):

src/utils/clients.py
Rajat-Ahuja1997: in the dialectic, session_history is now its own messages block. that info would be lost here as we only takes `messages[0].content. this was a pre-existing issue but accidentally worked because we bundled session history into `messages[0] but we should fix this now

Rajat-Ahuja1997:
@adavyas can you resolve the merge conflicts?

PR #428, opened 2026-03-16
chore: improving structure

ajspig:


docs/v3/guides/integrations/mcp.mdx
VVoruganti, line 1: Should we rename the file as well?

PR #430, opened 2026-03-17
fix: implement deriver custom instructions

adavyas:
## Summary

Implements `reasoning.custom_instructions` for the deriver on top of the prefix-cache prompt split from #425.

## What Changed

- keeps `custom_instructions` in resolved reasoning config
- adds `custom_instructions` to the deriver user prompt
- updates deriver prompt token estimation to include optional custom instructions
- adds tests for config precedence, prompt rendering, and deriver wiring

## Scope

This only implements `custom_instructions` for the deriver.

It does not add `custom_instructions` support to:
- dialectic
- dreamer

## Why

The schema already exposed `reasoning.custom_instructions`, but it was not used at runtime. This wires it into the deriver while preserving the cache-friendly split introduced in #425 by keeping the instructions in the user message rather than the stable system prompt.

## Testing

Verified locally with:
- `python -m py_compile` on changed files
- `basedpyright` on changed files
- targeted pytest on:
  - `tests/deriver/test_deriver_processing.py`
  - `tests/deriver/test_prompts.py`
  - `tests/utils/test_config_helpers.py`

Targeted pytest result:
- `15 passed`

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ Then modify the values as needed. The TOML file is organized into sections:
 - `[auth]` - Authentication configuration
 - `[cache]` - Redis cache configuration
 - `[llm]` - LLM provider API keys and general settings
-- `[deriver]` - Background worker settings and representation configuration
+- `[deriver]` - Background worker settings and representation configuration, including reasoning input budgets
 - `[peer_card]` - Peer card generation settings
 - `[dialectic]` - Dialectic API configuration with per-level reasoning settings
 - `[summary]` - Session summarization settings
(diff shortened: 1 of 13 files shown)


VVoruganti (changes requested):
This implementation looks good, but we need to make some changes to the top level Reasoning Configuration Schema to add common sense limits or calculated limits based on the token budget. 

The src/config.py sets max input tokens for the deriver, there should probably be another configurable field that influences max token count for custom instructions and discussion in the docs about how they interact.

PR #437, opened 2026-03-20
Add Voyage AI embedding provider

Marshall2HD:
Adds Voyage AI as an embedding provider.
VECTOR_STORE_DIMENSIONS was being silently ignored before because dimensions were hardcoded. That actually works now.

Changes:

- Voyage as a new embedding provider with dimension validation (256/512/1024/2048)
- All providers now pass dimensions through to their API calls instead of hardcoding them
- Migration to resize vector columns when dimensions change
- 18 unit tests + verified against live Voyage API

Set LLM_EMBEDDING_PROVIDER=voyage and LLM_VOYAGE_API_KEY=... to use it.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -65,6 +65,9 @@ LLM_OPENAI_API_KEY=your-openai-api-key-here
 # Anthropic API key for dialectic and deriver functionality
 LLM_ANTHROPIC_API_KEY=your-anthropic-api-key-here
 
+# Voyage AI API key (if using Voyage for embeddings)
+# LLM_VOYAGE_API_KEY=your-voyage-api-key-here
+
 # Google API key for summarization (if using Gemini)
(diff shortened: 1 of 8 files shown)


ajspig:
Hey @Marshall2HD — thanks for the Voyage AI integration! The dimensions wiring piece (making `VECTOR_STORE_DIMENSIONS` actually work) is addressed by #459. 
For Voyage specifically — the refactored architecture in #459 moves away from bespoke embedding providers entirely. Embedding transports are now just `openai` or `gemini`, with configurable `base_url` and `model` overrides. Since Voyage has an OpenAI-compatible API, it works by pointing the `openai` transport at Voyage's endpoint with the desired model name. 

We got a wave of community PRs all targeting embedding flexibility, and it endedup needing a sig refactor. But thank you for taking the time to make this contribution! Closing as duplicate of #459.

PR #465, opened 2026-03-30
fix: use VECTOR_STORE.DIMENSIONS for SQLAlchemy vector column size

paralizeer:
## Problem

`Document` and `MessageEmbedding` models hardcode `Vector(1536)`:

```python
```

This assumes OpenAI `text-embedding-3-small` dimensions. Using any other embedding model produces a runtime error:

```
ValueError: expected 1536 dimensions, not 768
```

## Fix

Replace `Vector(1536)` with `Vector(settings.VECTOR_STORE.DIMENSIONS)`.

`VECTOR_STORE.DIMENSIONS` already exists (default: 1536) and is used for LanceDB/TurboPuffer — this extends it consistently to the pgvector SQLAlchemy models.

## Impact

No breaking change for default deployments (DIMENSIONS defaults to 1536). Users with non-OpenAI embedding models can now set:

```toml
[vector_store]
DIMENSIONS = 768  # e.g. nomic-embed-text, all-MiniLM-L6-v2
```

## Summary by CodeRabbit

* **Chores**
  * Embedding dimensionality is now configurable via settings instead of a fixed value.
  * Gemini-based embedding requests now use the configured embedding dimension.
  * On application startup the system validates database vector column dimensions, logs mismatches or confirmations, and raises an error if stored dimensions differ from the configured value.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -133,11 +133,10 @@ async def lifespan(_: FastAPI):
             "Error initializing cache in api process; proceeding without cache: %s", e
         )
 
-    # Fail fast if Alembic migration dims diverge from VECTOR_STORE.DIMENSIONS (pgvector only)
-    async with SessionLocal() as session:
-        await check_vector_dimensions(session)
-
     try:
+        # Fail fast if Alembic migration dims diverge from VECTOR_STORE.DIMENSIONS (pgvector only)
+        async with SessionLocal() as session:
+            await check_vector_dimensions(session)
         yield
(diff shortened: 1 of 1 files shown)


VVoruganti (changes requested):
Also modify the alembic migrations to use a configuration variable instead, otherwise table creation will not work properly. Another approach is to introduce an alembic migration that checks the vector size and applies the change programmatically

PR #468, opened 2026-03-31
fix(deriver): disable metrics server to prevent port 9090 conflicts

roadtogideon222-byte:
## Problem
The deriver container becomes unhealthy because it tries to bind port 9090 for its Prometheus metrics server (METRICS_ENABLED=true), but port 9090 is already used by the Prometheus sidecar container defined in the same docker-compose file.

## Solution
- Set METRICS_ENABLED=false for the deriver service (the API service handles metrics)
- Added a process-based healthcheck (kill -0 1) since the deriver is a queue processor with no HTTP endpoint

## Files changed
- Added docker-compose.yml (production-ready config with example file as base)

## Testing
Verified on VPS: deriver container stays healthy after restart with no port conflicts.

## Summary by CodeRabbit

* **Chores**
  * Added Docker Compose configuration for the complete application stack, including database, cache, API services, and monitoring tools with health checks and persistent storage.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,116 @@
+services:
+  api:
+    image: honcho:latest
+    build:
+      context: .
+      dockerfile: Dockerfile
+    entrypoint: ["sh", "docker/entrypoint.sh"]
+    depends_on:
+      database:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+    ports:
+      - 8000:8000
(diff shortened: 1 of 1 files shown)


Rajat-Ahuja1997:
hey @roadtogideon222-byte thanks for the contribution. prometheus and the deriver run in their own network namespaces, so there is no port conflict here

PR #483, opened 2026-04-03
fix: stabilize self-hosted Docker restarts for api/deriver

DevNewbie1826:
## Summary
- remove the development bind mounts (`.:/app` and `venv:/app/.venv`) from the Docker compose example for `api` and `deriver`
- replace the API entrypoint with an image-local command that runs DB migrations and then starts FastAPI
- add a deriver-specific healthcheck instead of probing `/openapi.json`

## Why
The previous compose example could fail after `docker compose restart api deriver` with errors like:

- `ModuleNotFoundError: No module named 'sqlalchemy'`
- deriver reporting unhealthy because it inherited an API-style web healthcheck

The root cause was the dev-oriented runtime mount structure overlaying `/app` and `/app/.venv`, which made dependency resolution unstable across restarts.

## Verification
I verified all of the following locally:
- `docker compose up -d --build api deriver`
- `docker compose restart api deriver`
- `api` becomes healthy and `http://localhost:8000/openapi.json` responds
- `deriver` becomes healthy with the new healthcheck
- `hermes honcho status` connects successfully after restart

## Summary by CodeRabbit

* **Chores**
  * Updated Docker Compose configuration to automatically provision the database on API startup.
  * Optimized Docker setup by streamlining volume and mount configurations.
  * Added health monitoring to the deriver service.

### DIFF
--- a/docker-compose.yml.example
+++ b/docker-compose.yml.example
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    entrypoint: ["sh", "docker/entrypoint.sh"]
+    entrypoint: ["sh", "-lc", "/app/.venv/bin/python scripts/provision_db.py && exec /app/.venv/bin/fastapi run --host 0.0.0.0 src/main.py"]
     depends_on:
       database:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


ajspig:
Thanks @DevNewbie1826! 
You correctly identified the bind mount issue — those dev-oriented `.:/app` and `venv:/app/.venv` mounts were definitely causing restart instability. PR #510 (self-hosting docs overhaul) addresses both of your main fixes: it comments out the bind mounts as optional dev-only configuration and binds all ports to `127.0.0.1`. 

The deriver healthcheck is a nice touch — if you'd like to contribute that as a small focused PR after #510 merges, we'd welcome it. Closing as covered.

PR #485, opened 2026-04-03
feat: add observation categories for scoped memory retrieval

### LINKED ISSUE #484: feat: Add observation categories for scoped memory retrieval
## Problem

Currently all observations are stored in a single flat namespace per peer. When a peer has many observations spanning different domains (e.g., training data, business context, automation preferences), semantic search must scan all observations regardless of domain. This reduces retrieval precision for queries that target a specific knowledge area.

## Proposed Solution

Add a `category` column to the Document model (observations) that enables filtering search results by semantic category. Categories are free-form strings (not an enum), allowing developers to define domain-specific categories while the Deriver auto-assigns categories from a recommended set.

### Key Design Decisions

- **Dedicated indexed column** on the `documents` table rather than burying in JSONB metadata — enables efficient SQL filtering and external vector store filtering
- **Free-form text, not an enum** — developers can define their own categories; the Deriver suggests from a recommended set (preferences, personal, professional, goals, relationships, health, technical, hobbies, beliefs, habits)
- **Nullable & backward-compatible** — existing observations get NULL category; queries without a category filter match everything (no breaking changes)

### Changes Required

1. **Model + Migration**: Add nullable `category` TEXT column with index to `documents` table
2. **Schemas**: Add `category` to `ObservationInput`, `DocumentCreate`, `DocumentMetadata`, `ConclusionCreate`, `Conclusion` response
3. **Filter layer**: Add `category` to document filter mapping and external vector store filter passthrough
4. **Agent tools**: Extend `search_memory` tool with optional `category` filter param; extend `create_observations` tool to accept `category`
5. **Prompts**: Update Deriver prompt to auto-categorize; update Dialectic prompt to document category filtering

### API Usage

**Creating categorized observations:**
```json
POST /workspaces/{id}/conclusions
{
  "conclusions": [
    {
      "content": "User prefers dark mode",
      "observer_id": "agent-1",
      "observed_id": "user-1",
      "category": "preferences"
    }
  ]
}
```

**Querying by category:**
```json
POST /workspaces/{id}/conclusions/query
{
  "query": "color preferences",
  "filters": {
    "observer_id": "agent-1",
    "observed_id": "user-1",
    "category": "preferences"
  },
  "top_k": 10
}
```

**Agent tool usage (search_memory):**
```json
{
  "query": "marathon training data",
  "category": "health",
  "top_k": 20
}
```

## Alternatives Considered

1. **JSONB metadata field** — already exists but not indexed for efficient filtering; would require nested JSONB queries
2. **Predefined enum** — too restrictive; different use cases need different category taxonomies
3. **Tags (multi-value)** — more flexible but adds complexity to filtering and the LLM tool interface; categories can be extended to tags later if needed

UnbrandedTech:
## Summary

- Adds a nullable `category` column to the Document (observation) model with an indexed column and Alembic migration
- Extends `search_memory` and `create_observations` agent tool schemas to support category filtering/assignment
- Updates Deriver prompt to auto-categorize observations; updates Dialectic prompt to document category filtering
- Threads category through all layers: schemas, CRUD, filter mapping, external vector store filters, and Conclusions API

Closes #484

## Motivation

All observations are currently in a single flat namespace per peer. When a peer has many observations spanning different domains (e.g., "training data" vs "automation preferences"), semantic search must scan everything. Adding a `category` filter enables more precise retrieval by scoping search to a specific knowledge domain.

## Design Decisions

- **Dedicated indexed column** (not JSONB metadata) for efficient SQL and vector store filtering
- **Free-form text** (not enum) so developers can define domain-specific categories
- **Nullable & backward-compatible** — existing observations get NULL; queries without category match everything

## Files Changed

| File | Change |
|------|--------|
| `src/models.py` | Add `category` column to Document |
| `migrations/versions/a2b3c4d5e6f7_*.py` | Alembic migration |
| `src/schemas/internal.py` | Add `category` to DocumentMetadata, DocumentCreate, ObservationInput |
| `src/schemas/api.py` | Add `category` to ConclusionCreate, Conclusion, filter support |
| `src/utils/filter.py` | Add `category` to document filter mapping |
| `src/crud/document.py` | Thread `category` through create_documents, create_observations, external vector store filters |
| `src/utils/agent_tools.py` | Add `category` to search_memory + create_observations tool schemas and handlers |
| `src/deriver/prompts.py` | Instruct Deriver to auto-categorize observations |
| `src/dialectic/prompts.py` | Document category filter on search_memory tool |

## Test plan

- [ ] Run `uv run pytest tests/` to verify no regressions
- [ ] Verify migration applies cleanly: `uv run alembic upgrade head`
- [ ] Verify migration rolls back cleanly: `uv run alembic downgrade -1`
- [ ] Create conclusions via API with and without `category` field
- [ ] Query conclusions with `category` filter and verify scoped results
- [ ] Verify Deriver assigns categories to new observations
- [ ] Verify Dialectic agent can use `category` param in search_memory

## Summary by CodeRabbit

* **New Features**
  * Documents and observations can be tagged with optional semantic categories (e.g., preferences, business_context, personal, goals, technical).
  * Search/memory tools support optional category filtering to narrow results and scoped retrieval.
  * Category field available in create/update flows for documents and observations.

* **Documentation**
  * Tool prompts and descriptions updated to show category usage and examples.

### DIFF
--- a/migrations/versions/a2b3c4d5e6f7_add_category_column_to_documents.py
+++ b/migrations/versions/a2b3c4d5e6f7_add_category_column_to_documents.py
@@ -53,5 +53,10 @@ def upgrade() -> None:
 
 def downgrade() -> None:
     """Remove category column and index."""
-    op.drop_index("ix_documents_category", table_name="documents", schema=schema)
-    op.drop_column("documents", "category", schema=schema)
+    connection = op.get_bind()
+    inspector = sa.inspect(connection)
+
+    if index_exists("documents", "ix_documents_category", inspector):
+        op.drop_index("ix_documents_category", table_name="documents", schema=schema)
+    if column_exists("documents", "category", inspector):
+        op.drop_column("documents", "category", schema=schema)
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -557,6 +557,7 @@ async def create_documents(
                                 "observed": observed,
                                 "session_name": doc.session_name,
                                 "level": doc.level,
+                                "category": doc.category,
                             },
                         )
                     )
@@ -845,6 +846,7 @@ async def create_observations(
                                 "observed": observed,
                                 "session_name": doc.session_name,
                                 "level": doc.level,
+                                "category": doc.category,
                             },
                         )
                     )
--- a/src/reconciler/sync_vectors.py
+++ b/src/reconciler/sync_vectors.py
@@ -243,6 +243,7 @@ async def _sync_documents(
                         "observed": doc.observed,
                         "session_name": doc.session_name,
                         "level": doc.level,
+                        "category": doc.category,
                     },
                 )
             )
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -1288,7 +1288,9 @@ async def _handle_search_memory(ctx: ToolContext, tool_input: dict[str, Any]) ->
         # this stage, and be efficient with tool calls, and make sure the model
         # doesn't short-circuit and think there's nothing here, we
         # automatically search the message history for relevant information.
-        if ctx.agent_type == "dialectic":
+        # Skip fallback when a category filter is applied — zero results may
+        # simply mean no observations in that category, not an empty memory.
+        if ctx.agent_type == "dialectic" and not category:
             limit = min(_safe_int(tool_input.get("top_k"), 20), 20)
             message_output = None
             async with tracked_db("tool.search_memory.fallback") as db:


UnbrandedTech:
Thanks for the thorough review! All findings were verified against the code and addressed in 45afa53. Here's the breakdown:

### Outside diff comments — all fixed

**`src/crud/document.py` — VectorRecord metadata missing `category` (two sites)**
Good catch. Both the `create_documents` upsert loop (~line 548) and the `create_observations` vector store block (~line 838) were building VectorRecord metadata without `"category"`, which would silently prevent category-based filtering in external vector stores (Turbopuffer/LanceDB). Added `"category": doc.category` to both.

I also found a **third site** with the same gap: `src/reconciler/sync_vectors.py:236` — the reconciler that re-syncs pending documents to the external store was also missing `"category"` in its VectorRecord metadata. Fixed that too, since without it any documents re-synced after a failed upsert would lose their category in the external store.

**`src/utils/agent_tools.py` — fallback fires incorrectly with category filter**
Confirmed. The fallback message-history search (`if total_count == 0`) was designed for the "memory is truly empty" case early in a workspace's lifecycle. With category filtering, zero results just means no observations in that category — not that memory is empty. Triggering the fallback here would confusingly return raw message snippets instead of an accurate "no observations found in this category" response. Fixed by adding `and not category` to the guard.

### Nitpick — fixed

**Migration `downgrade()` — unguarded drops**
Agreed. The `upgrade()` already used `column_exists`/`index_exists` guards for idempotency, but `downgrade()` didn't. Added the same guards so partial or manual schema changes don't cause the downgrade to crash.

PR #491, opened 2026-04-04
fix: use configured VECTOR_STORE.DIMENSIONS instead of hardcoded 1536

voider1:
## Summary

- Replace hardcoded `Vector(1536)` in SQLAlchemy models (`Document`, `MessageEmbedding`) with `Vector(settings.VECTOR_STORE.DIMENSIONS)` so the dimension reads from config
- Replace hardcoded `"output_dimensionality": 1536` in Gemini embedding calls with the same config value
- This allows using embedding models with non-1536 dimensions (e.g. 768-dim nomic-embed-text) without patching source

## Context

When self-hosting Honcho with a local embedding model (e.g. nomic-embed-text-v1.5 via llama.cpp, which outputs 768 dimensions), the `VECTOR_STORE.DIMENSIONS` config setting is ignored because `src/models.py` hardcodes `Vector(1536)`. This causes pgvector to reject embeddings at query time with `ValueError: expected 1536 dimensions, not 768`.

The config setting already exists and defaults to 1536, so this is a no-op for existing deployments.

## Test plan

- [ ] Verify existing deployments with 1536-dim embeddings are unaffected (default config)
- [ ] Verify self-hosted setup with `VECTOR_STORE.DIMENSIONS = 768` works without source patching
- [ ] Verify no circular import issues (`src.config` imported in `src.models`)

## Release Notes

* **Refactor**
  * Embedding vector dimensions are now configurable through application settings instead of hardcoded values, providing greater flexibility for vector store configurations and system customization.
  * Updated database models and embedding services to consistently use configurable dimensions for improved system adaptability and configuration management.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -82,7 +82,7 @@ class _EmbeddingClient:
             response = await self.client.aio.models.embed_content(
                 model=self.model,
                 contents=query,
-                config={"output_dimensionality": 1536},
+                config={"output_dimensionality": settings.VECTOR_STORE.DIMENSIONS},
             )
             if not response.embeddings or not response.embeddings[0].values:
                 raise ValueError("No embedding returned from Gemini API")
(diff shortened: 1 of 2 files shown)


ajspig:
Thanks @voider1! Great catch on the hardcoded 1536 — you and several others spotted this independently, which really validates the need for the fix. PR #459 tackles this as part of a broader transport-based refactor of clients.py: dimensions are now fully configurable via `settings.EMBEDDING.VECTOR_DIMENSIONS` within a nested config structure that replaces the old flat env vars. We're closing this as a duplicate. Your instinct to make this configurable was spot on. 
We'd love for you to take a look at #459 and weigh in if anything stands out. See: https://github.com/plastic-labs/honcho/pull/459

PR #492, opened 2026-04-05
fix: make observer/observed optional in conclusions query endpoint

TreyDong:
## Summary

Previously, the conclusions query endpoint required both observer and observed to be specified, raising ValidationException if either was missing.

## Changes

- query_documents: observer/observed now optional (default None)  
- _query_documents_pgvector: conditionally applies WHERE clause only when param is not None
- query_conclusions router: removed ValidationException guard

## Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| Without filters | ValidationException | Returns results |
| SDK path | Works | Works (unchanged) |

## Summary by CodeRabbit

* **New Features**
  * Document and conclusion semantic search now supports optional peer filtering—users can omit `observer` and/or `observed` parameters to query across all peer relationships in the workspace.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -278,20 +278,23 @@ async def fetch_documents_by_ids(
 async def _query_documents_pgvector(
     db: AsyncSession,
     workspace_name: str,
-    observer: str,
-    observed: str,
+    observer: str | None,
+    observed: str | None,
     embedding: list[float],
     filters: dict[str, Any] | None,
     max_distance: float | None,
     top_k: int,
 ) -> list[models.Document]:
     """pgvector similarity search — pure DB operation."""
+    stmt = select(models.Document).where(
+        models.Document.workspace_name == workspace_name
+    )
+    if observer is not None:
+        stmt = stmt.where(models.Document.observer == observer)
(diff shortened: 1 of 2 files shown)


ajspig:
Hey @TreyDong — thanks for this PR! 

Currently, the `/conclusions/query` endpoint requires both `observer` and `observed`to be specified. Omitting them raises a ValidationException. 

I opened an issue to track this as a feature (#520) since the full solution touches the external vector store path (Turbopuffer/LanceDB) in a non-trivial way — namespace hashing requires both observer and observed, so omitting them crashes on those backends. There's also a type mismatch in fetch_documents_by_ids that would
need addressing.

Feel free to weigh in there and/or try and tackle the improvements needed!

VVoruganti:
Generally we want to stick to a strict boundary for conclusions queries to a singular representation. For external vector dbs such as turbopuffer or lancedb we enforce a strict `collection` boundary. Instead we would have to do a fanout query across multiple representations simultaneously and then rerank

VVoruganti:
Closing for inactivity, please reopen if comments are able to be addressed

PR #523, opened 2026-04-08
fix(docker)+feat(k8s): resolve deriver healthcheck bug and add Kubernetes deployment

### LINKED ISSUE #521: Move HTTP healthcheck from shared image to API service only
## Problem description

Honcho currently defines an HTTP-based `HEALTHCHECK` in the shared Docker image (`Dockerfile`) that probes `http://localhost:8000/openapi.json`.

That works for the `api` service, but the same image is also used for the `deriver` service. `deriver` is a background worker (`python -m src.deriver`), not an HTTP server, so inheriting the image-level HTTP healthcheck makes Docker report the worker as unhealthy even when the process is running normally.

In local verification:

- `deriver` started successfully and logged:
  - `Starting deriver queue processor`
  - `Running main loop`
  - `ReconcilerScheduler started ...`
- but container health failed with:
  - `urllib.error.URLError: <urlopen error [Errno 111] Connection refused>`
  - because the healthcheck still probes `localhost:8000/openapi.json`

This makes it look like `deriver` is broken, when the real issue is that the healthcheck assumes every consumer of the image exposes the API server.

## Desired solution

Move the HTTP healthcheck out of the shared `Dockerfile` and define it only on the `api` service in `docker-compose.yml` (or equivalent service-specific config).

That would let:

- `api` keep its HTTP/OpenAPI healthcheck
- `deriver` run without inheriting an irrelevant API probe
- future worker-style services reuse the image without false unhealthy states

## Alternatives considered

- **Disable healthcheck only for `deriver` in compose**: works as a local workaround, but still leaves the image with an API-specific assumption.
- **Create a worker-specific process healthcheck for `deriver`**: possible, but still does not solve the more general design issue that the base image encodes API behavior.

## Why this matters

This is easy to misdiagnose as an LLM / local-model / embedding-server failure, even though the root cause is only healthcheck placement. Moving the healthcheck to the `api` service makes the container role boundaries clearer and avoids false alarms during deployment and debugging.

b-mackenzie-alexander:
## Summary

- Removes the API-specific `HEALTHCHECK` from the shared `Dockerfile` that permanently marked the deriver worker container as unhealthy — the deriver is a background queue worker with no HTTP server, so `GET localhost:8000/openapi.json` could never succeed.
- Adds a `/k8s` directory with Kustomize-based manifests for deploying all Honcho services on Kubernetes (local and production), with a comprehensive README and a self-contained pytest test suite.

## Motivation

**Healthcheck bug (Closes #521)**

The shared image's `HEALTHCHECK` directive probes `http://localhost:8000/openapi.json`. This is correct for `honcho-api` but causes `honcho-deriver` to be permanently marked `unhealthy` in every container runtime that respects Docker health status — including Kubernetes when the image-level check is surfaced. The deriver is a queue processor; it has no HTTP server. The fix removes the `HEALTHCHECK` from the shared image. In the Kubernetes manifests, the API gets a proper `httpGet` readiness probe and a `tcpSocket` liveness probe; the deriver gets no HTTP probe and relies on `restartPolicy: Always`.

**Kubernetes deployment (References #514)**

Teams running existing Kubernetes infrastructure had no supported deployment path. The new `/k8s` directory provides:

- **postgres** — StatefulSet (`pgvector/pgvector:pg15`, 10 Gi PVC) with headless Service and differentiated readiness/liveness probes
- **redis** — StatefulSet (`redis:8.2`, 2 Gi PVC) with headless Service and differentiated probes
- **honcho-api** — Deployment (HPA-managed replicas, no static `replicas` field), ClusterIP Service, HPA (1–5 replicas at 70% CPU), PDB (`minAvailable: 1`); httpGet readiness + tcpSocket liveness to avoid the identical-probe anti-pattern
- **honcho-deriver** — Deployment with no HTTP liveness probe; health managed by the Deployment restart policy
- **NetworkPolicies** — default-deny ingress + explicit allow rules; postgres and redis reachable only from API and deriver pods
- **Security hardening** — `runAsNonRoot`, `readOnlyRootFilesystem`, `allowPrivilegeEscalation: false`, `capabilities.drop: ALL`, `seccompProfile: RuntimeDefault`, `automountServiceAccountToken: false` on all application containers; init containers run as UID/GID 65534 (nobody/nogroup)
- **Secrets/ConfigMap split** — sensitive values in a gitignored `Secret`; non-sensitive tunables in a `ConfigMap`
- **README** — covers prerequisites, per-distro image loading (kind, k3s, Docker Desktop), secrets setup, deploy, verify, useful commands, autoscaling, networking, and production notes (managed DB/cache, external secrets operators, Ingress, TLS, observability)

## Test plan

- [ ] **Manifest unit tests** (no cluster required):
  ```bash
  pytest tests/test_k8s_manifests.py --noconftest -p no:xdist --override-ini="addopts=" -v
  ```
  30 tests covering namespace isolation, all 4 NetworkPolicies, headless Services, HPA/PDB wiring, probe types, security contexts, init container hardening, ConfigMap/Secret split, and resource limits.

- [ ] **Schema validation** (optional, requires `kubeconform`):
  ```bash
  kubectl kustomize k8s/ | kubeconform -strict -summary
  ```

- [ ] **Security scoring** (optional, requires `kube-score`):
  ```bash
  kubectl kustomize k8s/ | kube-score score -
  ```

- [ ] **Misconfiguration scan** (optional, requires `trivy`):
  ```bash
  trivy config k8s/
  ```

- [ ] **End-to-end** (requires a local k8s cluster — kind, k3s, or Docker Desktop):
  1. `docker build -t honcho:latest .`
  2. Load image into cluster per README instructions
  3. `cp k8s/secrets.yaml.example k8s/secrets.yaml` and fill in values
  4. `kubectl apply -f k8s/secrets.yaml && kubectl apply -k k8s/`
  5. `kubectl get pods -n honcho --watch` — all pods reach `Running`
  6. `kubectl port-forward svc/honcho-api 8000:80 -n honcho` + `curl http://localhost:8000/openapi.json`
  7. `kubectl logs deploy/honcho-deriver -n honcho` — no healthcheck errors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a complete Kubernetes deployment stack for API, background worker, PostgreSQL, and Redis with autoscaling, pod disruption budget, and hardened pod/security settings

* **Documentation**
  * Added comprehensive Kubernetes deployment guide with quick-start, validation, and operational guidance

* **Tests**
  * Added manifest validation suite to verify rendered manifests, resource alignment, network policies, and security posture

* **Chores**
  * Ignore local secrets file
  * Removed container healthcheck from image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,6 +185,9 @@ timing_logs.csv
 config.toml
 .aider*
 
+# Kubernetes secrets — filled-in copies of k8s/secrets.yaml.example
+k8s/secrets.yaml
+
 CRUSH.md
 .crush/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,4 @@ USER app
 
 EXPOSE 8000
 
-HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD python -c "import urllib.request; urllib.request.urlopen('http://localhost:8000/openapi.json')" || exit 1
-
 CMD ["fastapi", "run", "--host", "0.0.0.0", "src/main.py"]
--- a/k8s/README.md
+++ b/k8s/README.md
@@ -1,0 +1,338 @@
+# Running Honcho on Kubernetes
+
+This directory contains Kubernetes manifests for deploying all Honcho services. The setup is managed with [Kustomize](https://kustomize.io/), which is built into `kubectl` v1.14+.
+
+## What gets deployed
+
+| Resource | Kind | Notes |
+|----------|------|-------|
+| `postgres` | StatefulSet + ClusterIP Service | pgvector/pgvector:pg15, 10 Gi PVC |
+| `redis` | StatefulSet + ClusterIP Service | redis:8.2, 2 Gi PVC |
+| `honcho-api` | Deployment + ClusterIP Service | FastAPI server; runs migrations on start |
+| `honcho-deriver` | Deployment (no Service) | Background queue worker |
+| `honcho-api` | HorizontalPodAutoscaler | 1–5 replicas at 70% CPU |
+| `honcho-api` | PodDisruptionBudget | minAvailable: 1 |
+| NetworkPolicies | default-deny + allow rules | Postgres/Redis reachable only from API/Deriver |
+
+## Testing the manifests
+
+A pytest test suite ships with the manifests at `tests/test_k8s_manifests.py`. It renders the Kustomize output and asserts structural correctness, security posture, and architectural invariants without requiring a running cluster.
+
+```bash
+# Requires: kubectl on PATH, pytest, pyyaml
+pytest tests/test_k8s_manifests.py --noconftest -p no:xdist --override-ini="addopts=" -v
+```
+
+To also validate schemas against the official Kubernetes API spec and run security scoring:
+
+```bash
+# Schema validation (install: brew install kubeconform)
+kubectl kustomize k8s/ | kubeconform -strict -summary
+
+# Security and best-practices scoring (install: brew install kube-score)
+kubectl kustomize k8s/ | kube-score score -
+
+# Misconfiguration scanning (install: brew install trivy)
+trivy config k8s/
+```
+
+---
+
+## Prerequisites
+
+- `kubectl` v1.14+ (Kustomize is built in)
+- Docker (to build the Honcho image)
+- One of the following local Kubernetes distributions:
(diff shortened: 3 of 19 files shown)


b-mackenzie-alexander:
Closing this in favour of two focused PRs per reviewer feedback — splitting the Docker healthcheck fix and the Kubernetes manifests so each can be evaluated independently.

PR #534, opened 2026-04-09
Support dedicated embedding endpoint config

counterposition:
## Summary

- add `LLM_EMBEDDING_API_KEY` and `LLM_EMBEDDING_BASE_URL` for the `openrouter` embedding path
- pass the resolved embedding base URL explicitly through the `EmbeddingClient` wrapper
- improve the missing-key error message to mention both supported embedding key settings
- add unit coverage for dedicated endpoint preference, fallback behavior, and missing-key errors
- make the `tracked_db` test fixture lazy so these pure unit tests do not force DB setup

## Motivation

A self-hosted Honcho deployment can route chat/reasoning traffic through one OpenAI-compatible provider while needing embeddings from another endpoint. Without separate embedding configuration, writes that depend on embeddings (for example, conclusion creation) can fail even though the primary LLM path is configured correctly.

## Test plan

- [x] Run the focused unit tests from the repo root:
  ```bash
  uv run pytest tests/test_embedding_client.py -q
  ```
- [x] Result: `4 passed`

## Notes

- dedicated embedding endpoint support is intentionally scoped to the `openrouter` embedding path
- the fixture change is included to keep the new tests unit-level rather than requiring database initialization

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -208,6 +208,10 @@ class LLMSettings(HonchoSettings):
     GEMINI_API_KEY: str | None = None
     GROQ_API_KEY: str | None = None
     OPENAI_COMPATIBLE_BASE_URL: str | None = None
+    # Optional dedicated embedding endpoint/key. Useful when chat traffic goes
+    # through one OpenAI-compatible provider but embeddings must use another.
+    EMBEDDING_API_KEY: str | None = None
+    EMBEDDING_BASE_URL: str | None = None
 
     # Separate vLLM endpoint (for local models)
     VLLM_API_KEY: str | None = None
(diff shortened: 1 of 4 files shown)


ajspig:
Thanks @counterposition! Dedicated embedding endpoint configuration is exactly the right idea. The existing PR #459 adds this via EMBEDDING_MODEL_CONFIG__OVERRIDES__BASE_URL and EMBEDDING_MODEL_CONFIG__OVERRIDES__API_KEY_ENV, letting you point embeddings at any endpoint independently of the LLM providers.

Closing as covered by #459. Thanks for the contribution!

PR #535, opened 2026-04-09
feat: add Vertex AI support for Anthropic and Google providers

lucharo:
Vertex AI support for self-hosted Honcho -- use GCP service accounts instead of direct API keys.

## What changed

**`src/utils/clients.py`** -- when `ANTHROPIC_VERTEX_PROJECT_ID` and `ANTHROPIC_VERTEX_REGION` are both set, initializes an `AsyncAnthropicVertex` client. When `GOOGLE_GENAI_USE_VERTEXAI=true` is set, initializes `genai.Client(vertexai=True)`. Both only kick in when no direct API key is configured.

**`src/embedding_client.py`** -- same Vertex AI fallback for the Gemini embedding client. Also makes the embedding model name configurable via `LLM_EMBEDDING_MODEL`.

**`src/config.py`** -- adds `EMBEDDING_MODEL` field (optional, defaults to `gemini-embedding-001`).

## Why

If you're self-hosting on GCP and already have Vertex AI access through a service account, you shouldn't need to go get separate API keys from Anthropic and Google. This wires up the existing Vertex AI auth path.

## Tested with

- Claude haiku-4-5, sonnet-4-6, opus-4-6 via `AnthropicVertex`
- Gemini 2.5 Flash generation via Vertex AI
- Gemini embeddings (`gemini-embedding-2-preview`) via Vertex AI

Existing behavior is unaffected -- the Vertex paths only activate when env vars are set and no API key exists.

## Summary by CodeRabbit

* **New Features**
  * New configurable embedding model setting to explicitly override the default embedding model.
  * Optional Google Vertex AI backend for Gemini-style embeddings with automatic selection when configured.
  * Optional Anthropic Vertex AI backend for Anthropic-style embeddings with automatic selection when configured.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -214,6 +214,7 @@ class LLMSettings(HonchoSettings):
     VLLM_BASE_URL: str | None = None
 
     EMBEDDING_PROVIDER: Literal["openai", "gemini", "openrouter"] = "openai"
+    EMBEDDING_MODEL: str | None = None
 
     # General LLM settings
     DEFAULT_MAX_TOKENS: Annotated[int, Field(default=1000, gt=0, le=100_000)] = 2500
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import os
 import threading
 from collections import defaultdict
 from typing import NamedTuple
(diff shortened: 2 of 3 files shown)


Rajat-Ahuja1997:
@lucharo thanks for the contribution! the vertex support is worth keeping but please rebase onto main / merge main and re-apply your changes, as we've heavily refactored the src/config, embedding client, and clients.py. happy to review once that's done

VVoruganti:
Thanks for this @lucharo — Vertex AI support is something we want. This predates a heavy refactor of src/config, the embedding client, and clients.py, so it no longer applies cleanly and has gone quiet. Closing for now; please reopen with a rebase onto main if you'd like to carry it forward, otherwise we'll pick the feature up on our side.

PR #541, opened 2026-04-10
fix(api): allow optional observer and observed parameters for cross-peer search

### LINKED ISSUE #520: Cross-peer semantic search for conclusions
Currently, the `/conclusions/query` endpoint requires both `observer` and `observed` to be specified. Omitting them raises a ValidationException. This is by design — collections are scoped by (workspace, observer, observed) — but it creates a poor UX for ad-hoc REST API queries where a user wants to search across all peer relationships in a workspace.

## Context

PR #492 by @TreyDong proposed making these optional. The pgvector path is straightforward (conditional `.where()` clauses), but the external vector store path (Turbopuffer/LanceDB) is the hard part:

- Namespaces are computed by hashing (workspace, observer, observed). Passing None crashes or produces a garbage namespace.
- `fetch_documents_by_ids` expects non-optional observer/observed — None values leak through as IS NULL filters instead of being skipped.
- A proper solution for external stores would require enumerating all relevant namespaces in the workspace and fanning out queries across them.

## Design questions

1. Should cross-peer search be pgvector-only for now, with a clear error for external store deployments?
2. Or should we support it across all backends? If so:
   - How do we enumerate namespaces? Scan the documents table for distinct (observer, observed) pairs in the workspace?
   - Should top_k apply per-namespace or globally after merging results?
   - What are the perf bounds? A workspace with many peer relationships could fan out to dozens of namespace queries.
3. Should the same relaxation apply to the /list endpoint or just /query?

## Minimum viable approach

Guard with `_uses_pgvector()` — allow optional observer/observed on pgvector, raise an informative error on external stores. Revisit external store support based on demand.

Related: #492

Pamf1973:
Fixes #520 

What changed?
We adopted the minimum viable approach recommended by the maintainers by updating the /conclusions/query endpoint and the underlying DB query logic. 

Specifically:

We modified src/crud/document.py to accept optional observer and observed parameters, ensuring the SQL constraints are dynamically bypassed when they are omitted.
We updated src/routers/conclusions.py to route the ValidationException inside a _uses_pgvector() guard logic.
We added explicitly targeted guardrails across these paths to raise a clear validation error ensuring the users of external vector stores (like LanceDB/Turbopuffer) receive a helpful message instead of a silent failure, as namespaces currently require specific scoping there.

Why was this prioritized?

I prioritized this contribution because the current constraint was creating unnecessary friction for developers wanting to query global, workspace-level insights. Fixing this improves the overall developer experience and REST API utility, without blocking future optimizations for external storage solutions. It's a quick win for the community's usability.

What problem does it solve for Honcho?

Previously, the endpoint demanded precise (workspace, observer, observed) scopes for fetching semantic conclusions. If a user simply wanted to search across all peer conversations/facts within a workspace, it would fail. By making the parameters optional when operating on standard DB infrastructure (pgvector), Honcho allows for much broader semantic searches and ad-hoc analytics without breaking security or backward compatibility.

## Summary by CodeRabbit

* **New Features**
  * Made observer and observed parameters optional for document searches, enabling more flexible query capabilities.

* **Bug Fixes**
  * Fixed semantic search validation to properly handle external vector store configurations, ensuring appropriate error handling when required parameters are missing in specific search scenarios.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -278,8 +278,8 @@ async def fetch_documents_by_ids(
 async def _query_documents_pgvector(
     db: AsyncSession,
     workspace_name: str,
-    observer: str,
-    observed: str,
+    observer: str | None,
+    observed: str | None,
     embedding: list[float],
     filters: dict[str, Any] | None,
     max_distance: float | None,
@@ -289,11 +289,14 @@ async def _query_documents_pgvector(
     stmt = (
         select(models.Document)
         .where(models.Document.workspace_name == workspace_name)
-        .where(models.Document.observer == observer)
-        .where(models.Document.observed == observed)
         .where(models.Document.embedding.isnot(None))
         .where(models.Document.deleted_at.is_(None))
     )
+
+    if observer:
+        stmt = stmt.where(models.Document.observer == observer)
+    if observed:
+        stmt = stmt.where(models.Document.observed == observed)
 
     if max_distance is not None:
         stmt = stmt.where(
@@ -314,8 +317,8 @@ async def query_documents(
     workspace_name: str,
     query: str,
     *,
-    observer: str,
-    observed: str,
+    observer: str | None = None,
+    observed: str | None = None,
     filters: dict[str, Any] | None = None,
     max_distance: float | None = None,
     top_k: int = 5,
@@ -378,6 +381,11 @@ async def query_documents(
             for doc in docs:
                 managed_db.expunge(doc)
             return docs
+
+    if not observer or not observed:
+        raise ValidationException(
+            "observer and observed must be specified for semantic search on external vector stores"
+        )
 
     # External vector store — network call first, DB only for the ID fetch
     document_ids = await query_external_vector_document_ids(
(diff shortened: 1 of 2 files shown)


Pamf1973:
@ajspig can you please review my PR? Thank you!

ajspig:
Closing this PR as a duplicate of #492.
The core issue here isn't the pgvector path — that's straightforward. The blocker is external vector store support (Turbopuffer/LanceDB):

Namespaces are computed by hashing (workspace, observer, observed). Passing None crashes or produces garbage.
fetch_documents_by_ids expects non-optional observer/observed — None values leak through as IS NULL filters instead of being skipped.
A proper solution requires enumerating all relevant namespaces in the workspace and fanning out queries across them.

We've captured this in #520 which tracks the design work. The minimum viable approach is to guard with _uses_pgvector() — allow optional observer/observed on pgvector only, and raise an informative error on external stores.

Thanks for the contribution — please follow #520 for updates on when this can move forward properly.

PR #551, opened 2026-04-11
Fix/506 session bootstrap fails

Pamf1973:
Resolves #506

The Problem: Currently, session bootstrap fails with a too_long error when a session context attempts to retrieve more than ~100 messages (triggering internal schema list limits inside the underlying structure). This causes clients like OpenClaw to crash when resuming long-running sessions, effectively locking developers out of recovering historical context.

The Solution: This PR establishes a secure "safety valve" by introducing a declarative max_messages parameter to session.context(). This allows clients to cap the exact number of recent messages fetched during context retrieval. The backend intelligently truncates the conversation history at the database level before anything can crash validation, ensuring sessions of any size remain fast, stable, and easily retrievable.

Technical Changes
Backend Persistence: Updated crud.get_messages_id_range() in src/crud/message.py to accept the max_messages parameter. This securely wraps the query in a descending limit subquery mapped ascending, ensuring exactly the most recent requested window is returned chronologically. Plumbed max_messages cleanly backward through the _get_session_context_task and fastAPI routing configurations.
TypeScript SDK: Updated ContextParamsSchema (sdks/typescript/src/validation.ts) to validate numeric options (maxMessages) correctly. Integrated into SessionContextParams and seamlessly bound it into the TypeScript Session.context() mapping logic.
Python SDK: Mirrored the max_messages query param integration accurately onto sdks/python/src/honcho/session.py and the async bindings identically in sdks/python/src/honcho/aio.py.

Testing
✅ Compiled Python SDK and Backend cleanly.
✅ Compiled TypeScript SDK npm run build cleanly.
✅ The database properly limits output based natively on max_messages while perfectly respecting tokens rules.

## Summary by CodeRabbit

* **New Features**
  * Added optional `max_messages` parameter to session context retrieval, enabling users to cap the number of recent messages included in context (minimum: 1 message). Available across Python SDK, TypeScript SDK, and REST API.

* **Tests**
  * Added test coverage for message limiting functionality in session context queries.


ajspig:
Hey @Pamf1973 — thanks for looking into this! A few things need attention before this can move forward:
1. Out-of-scope changes: The document.py and conclusions.py changes (making observer/observed optional for pgvector queries) are unrelated to the max_messages feature and #506. That's a separate concern being tracked in #520, and @VVoruganti has flagged architectural considerations around it. Please remove those changes and keep this PR focused on max_messages only.
2. package-lock.json: The +245 line diff shouldn't be in here — that's likely from running npm install locally.
3. Testing: The PR description says it compiles cleanly but doesn't include actual test cases. Could you add a test that exercises max_messages on a session with >100 messages to validate the fix against #506?

Pamf1973:
Here is exactly what I did:

Reverted document.py and conclusions.py: I checked out both files from main to remove the out-of-scope observer/observed changes, ensuring this PR stays strictly focused on the max_messages fix.
Removed package-lock.json: I deleted the sdks/typescript/package-lock.json file which introduced the +245 line diff originally spawned by a local npm install.
Added Test Case: I added a new test test_get_session_context_with_max_messages in tests/routes/test_sessions.py which:
Sets up a session.
Pushes 110 messages sequentially.
Validates that max_messages=10 query successfully constrains the returned result down to 10 context messages while keeping chronological ordering for the most recent items.

ajspig:
Thanks for the contribution @Pamf1973! We looked into this and don't think this change is needed.
The too_long error from #506 doesn't originate in Honcho. There's no such error anywhere in the codebase. It was almost certainly an LLM provider error (context window exceeded) on the consumer side. We tested against a session with 1739 messages using session.context() with --tokens 2000 and it returned correctly without issue.

Session.context() already accepts a tokens parameter that caps the response at the database level. That's the intended mechanism for this. The issue pointed to LimitSchema in the TS SDK as the culprit, but that's only used for .search() result pagination and is a completely separate code path from context retrieval. Adding max_messages as a secondary cap overlaps with what token budgets already handle. Going to close this along with #506. Appreciate you digging in!

PR #570, opened 2026-04-17
feat: allow thinking_budget_tokens=0 for non-Anthropic providers (MiniMax/OpenAI-compatible support)

Lunox-1337:
## Summary

Changes the lower-bound validation for `THINKING_BUDGET_TOKENS` from `gt=0` (must be > 0) to `ge=0` (must be >= 0) in three settings classes. This enables using Honcho with providers that do not support extended thinking/reasoning budgets — specifically MiniMax-M2.7 and other OpenAI-compatible APIs.

## Changes

- `config.py:259` — `DeriverSettings.THINKING_BUDGET_TOKENS`: `gt=0` → `ge=0`
- `config.py:441` — `SummarySettings.THINKING_BUDGET_TOKENS`: `gt=0` → `ge=0`
- `config.py:545` — `DreamSettings.THINKING_BUDGET_TOKENS`: `gt=0` → `ge=0`

## Why

The existing code at `clients.py:1385` already guards the thinking budget parameter — it is only passed to the API when `provider == "anthropic"` and `thinking_budget_tokens > 0`. However, the Pydantic validation was blocking `thinking_budget_tokens=0` at config load time, preventing Honcho from starting with providers that don't support the parameter.

With this change, users can configure:

```toml
[deriver]
provider = "custom"
model = "MiniMax-M2.7"
thinking_budget_tokens = 0

[summary]
provider = "custom"
model = "MiniMax-M2.7"
thinking_budget_tokens = 0
```

via the OpenAI-compatible API:

```toml
OPENAI_COMPATIBLE_BASE_URL = "https://api.minimax.io/v1"
OPENAI_COMPATIBLE_API_KEY = "${MINIMAX_API_KEY}"
```

## No Behaviour Change for Existing Users

- Dialectic level settings already used `ge=0` (unchanged)
- Anthropic users with `thinking_budget_tokens > 0` are unaffected — validation still requires values >= 0
- Non-Anthropic providers were already not sending thinking budget to the API — this just removes the spurious config-level error

## Summary by CodeRabbit

* **Refactor**
  * Thinking-budget constraints updated to allow a value of zero for relevant configuration settings, preserving previous defaults and upper bounds.

* **Changes**
  * Per-reasoning-level defaults updated: provider set to a custom provider, model switched to "MiniMax-M2.7" across levels; thinking-budget set to 0 for medium/high/max; minimal level's max output increased from 250 to 1024.


VVoruganti:
Closed due to #459

PR #573, opened 2026-04-17
fix(dreamer): threshold and time-guard semantics

lilyplasticlabs:
## Summary

Two design-level fixes that correct what the dreamer's scheduling guards actually measure. First of four PRs in the dreamer systems rollout.

- **Finding 2 — threshold filter:** the document-count threshold used to count all levels (including the dreamer's own `deductive` / `inductive` / `contradiction` output). That inflated the trigger for the next dream and created a feedback loop (effective threshold of ~15 instead of 50 on the meeting-intel tenant). Fix: filter the count to `level == "explicit"` only — raw material from the deriver is what should count.
- **Finding 3 (code-level) — `last_dream_at` relocation:** the 8-hour time guard's timestamp was written at *enqueue* time, so duplicate enqueues reset the clock and failed runs still advanced the guard. Fix: move the write to `process_dream` inside `if result is not None:`, using a short-lived `tracked_db()` session.

Together these two fixes directly answer the original recursion question from the March Discord thread — not by direct recursion (verified that doesn't happen), but indirectly: the threshold was lying about how much unconsolidated material had accumulated.

## `last_dream_at` success criteria — defaulting to lenient

`DreamResult` has separate `deduction_success` / `induction_success` flags. Three options for when to write `last_dream_at`:

- **Strict** — both specialists succeed
- **Middle** — at least one specialist succeeds
- **Lenient (chosen)** — any non-null `DreamResult`

Rationale: the time guard's job is rate-limiting LLM resource use, not quality assurance. Lenient matches that intent — failures still count as "the guard ran recently, don't fire another one right away." If you want strict/middle instead, the change is a one-line condition in `orchestrator.py` and I'll update the failure-path test accordingly.

## No data migration

Existing `last_dream_at` values were written at enqueue and stay as-is. Worst case: one cycle of slight imprecision on existing collections. Not worth the migration complexity.

## Test plan

- [x] Unit — `TestThresholdFilter` (3 tests): mixed levels below threshold → no trigger; explicit-only at threshold → triggers; contradiction excluded (covers B1–B3 edge of positive filter)
- [x] Unit — `TestEnqueueDreamMetadataShape`: AsyncMock verifies `update_data` lacks `last_dream_at` at enqueue
- [x] Integration (real Postgres) — `TestLastDreamAtCompletionWrite` (3 tests): happy path writes, failure path doesn't, exception path doesn't (and `process_dream` still swallows the exception, queue-processed semantics preserved)
- [x] `uv run pytest tests/dreamer/ tests/deriver/test_enqueue_dream.py -v` — 14/14 pass
- [x] Pre-push full suite — 946/946 pass (after local `bun install` in `sdks/typescript/` to pick up `zod`)
- [x] `uv run ruff check` — clean
- [x] `uv run basedpyright` — 0 errors, 0 warnings, 0 notes on touched files

## Bug IDs

- Finding 2 — threshold counts all document levels, not just `explicit`
- Finding 3 (code-level) — `last_dream_at` written at enqueue, never at completion

Both from Finding 2 and 3 in the 2026-04-16 findings doc. Finding 3's architectural remedy (workspace concurrency cap) is deferred per your feedback that the queue + deriver worker pool already throttles execution.

## Summary by CodeRabbit

* **Refactor**
  * Scheduling now counts only explicit observations; derived/other levels no longer affect eligibility.
  * Enqueue records only the document-count; completion records the completion timestamp. Metadata updates use row-locked read-modify-write semantics and preserve existing sibling fields.

* **Tests**
  * Added unit and integration tests covering explicit-only scheduling, enqueue/completion metadata shape, and preservation of existing dream metadata.


src/deriver/enqueue.py
```diff
@@ -516,19 +515,26 @@ async def enqueue_dream(
...
                 workspace_name,
                 observer,
                 observed,
-                update_data={
-                    "dream": {
-                        "last_dream_document_count": document_count,
-                        "last_dream_at": now_iso,
-                    }
-                },
+                update_data={"dream": dream_meta},
```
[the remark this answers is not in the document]
lilyplasticlabs: Loop 2 addressed this with Option A (row lock via `with_for_update=True`) at both call sites: `enqueue.py:536` and `orchestrator.py:341`. Both reads run inside a `tracked_db` transaction, so `SELECT ... FOR UPDATE` serializes concurrent enqueue + completion writes.

Didn't take Option B (`jsonb_set` on nested paths) in this PR even though it's the more durable structural fix. Option B changes the crud API surface — `update_collection_internal_metadata` has other callers that would need auditing — and deserves its own review lens rather than riding alongside a guard-semantics fix. Tracked as a follow-up for a dedicated PR.

Resolving.

src/deriver/enqueue.py
```diff
@@ -516,19 +521,28 @@ async def enqueue_dream(
             stmt = insert(QueueItem).returning(QueueItem)
             await db_session.execute(stmt, [dream_record])
 
-            # Update collection metadata (CRUD handles cache invalidation)
-            now_iso = datetime.now(timezone.utc).isoformat()
+            # Update collection metadata (CRUD handles cache invalidation).
+            # last_dream_at is written at completion in process_dream, not here.
```
VVoruganti, line 525: Moving last_dream_at from enqueue time to completion time seems to reopen duplicate scheduling inside the 8-hour guard. check_and_schedule_dream() only consults last_dream_at, and execute_dream() does not re-check that guard before enqueueing, so a second timer can still be armed while the first dream is running and later enqueue inside the guard window.

src/dreamer/orchestrator.py
```diff
@@ -323,6 +324,32 @@ async def process_dream(
                         + f"duration={result.total_duration_ms:.0f}ms"
                     )
 
+                    # Write last_dream_at at completion (not enqueue) so duplicate
```
VVoruganti, line 327: Writing last_dream_at only on completion changes the guard semantics: while a dream is queued/running, last_dream_at is still stale, so additional explicit documents can schedule another dream that will later enqueue inside MIN_HOURS_BETWEEN_DREAMS.

src/deriver/enqueue.py
```diff
@@ -516,19 +521,28 @@ async def enqueue_dream(
...
+            # Row-lock the collection during read-modify-write to prevent concurrent enqueue/completion from clobbering each other's dream metadata writes.
+            collection = await crud.get_collection(
+                db_session,
+                workspace_name,
+                observer=observer,
+                observed=observed,
+                with_for_update=True,
+            )
+            dream_meta = dict(collection.internal_metadata.get("dream", {}))
+            dream_meta["last_dream_document_count"] = document_count
```
VVoruganti, line 539: last_dream_document_count is still advanced at enqueue time. If the dream later fails, the current explicit corpus has already been consumed as the new baseline, so the same corpus will not auto-trigger another dream; only future explicit observations can move the delta back over threshold.

VVoruganti (changes requested):

lilyplasticlabs:
Changes ready for review — commits `d24958d` + `8cf76fa` (comment trim).

**Root cause:** your three comments are the same bug seen in three places. The two dream-guard fields were being written at different steps — `last_dream_at` at completion, `last_dream_document_count` at enqueue. During the in-flight window (a dream queued but not yet processed), the two fields disagreed, which broke `check_and_schedule_dream`'s guards.

**Fix — two-part:**

1. **Atomic guard-pair write at completion** (`orchestrator.py:process_dream`). Both `last_dream_at` and `last_dream_document_count` now write together inside the same `with_for_update=True` transaction, only on `result is not None`. Count is recomputed inside the locked block rather than passed via `DreamPayload`. Closes comments 2 and 3.
2. **In-flight check via the queue** (`dream_scheduler.py:check_and_schedule_dream`). Before arming a new timer, query `QueueItem.work_unit_key.in_(pending_keys)` with `processed=False`. Mirrors `uq_queue_dream_pending_work_unit_key`. Closes comment 1.

The rule is pinned to `check_and_schedule_dream`'s docstring. Two new tests under `TestGuardPairCoherence` in `tests/dreamer/test_dreamer_integration.py` cover both halves. 951/951 tests pass, ruff + basedpyright clean.

**Observability note for telemetry:** a raising `run_dream` currently lands the QueueItem as `processed=True` without populating the `error` field (`process_dream`'s except branch swallows). Worth a backlog item.

src/deriver/enqueue.py
lilyplasticlabs: Addressed in `d24958d`. `enqueue_dream` no longer touches `internal_metadata["dream"]` at all — the block on duplicate scheduling is now explicit in `check_and_schedule_dream` via a `QueueItem` query with `processed=False` (mirrors `uq_queue_dream_pending_work_unit_key`). Coverage: `TestGuardPairCoherence::test_pending_queue_item_blocks_second_schedule`.

src/dreamer/orchestrator.py
lilyplasticlabs: Addressed in `d24958d`. Both `last_dream_at` and `last_dream_document_count` now write atomically inside the same `with_for_update=True` transaction in `process_dream`, only on `result is not None`. The window where the two guards disagree is closed by the `QueueItem` query in `check_and_schedule_dream` (`dream_scheduler.py:~295-324`).

src/deriver/enqueue.py
lilyplasticlabs: Addressed in `d24958d`. `last_dream_document_count` write moved from enqueue to `process_dream` on success, in the same transaction as `last_dream_at`. Failed dreams (`run_dream` returns `None` or raises) leave both fields untouched, so the same corpus retries on the next scheduler tick. Coverage: `TestGuardPairCoherence::test_silent_failure_allows_retry_on_same_corpus`.

VVoruganti (approved):

PR #576, opened 2026-04-18
fix(json): extract structured payloads from noisy model output

Niko96-dotcom:
## Summary
- strip reasoning wrappers before repair without corrupting already-valid JSON
- prefer tagged JSON fences over generic fenced blocks
- collect and rank balanced JSON candidates instead of trusting the first blob that looks vaguely object-shaped
- add regression tests for fenced JSON preference, schema-vs-payload selection, and literal `<think>` text inside valid JSON

## Test Plan
- `set -a && source /Users/nikolaymohr/honcho/.env && set +a && uv run pytest tests/utils/test_json_parser.py -q`

## Why
OpenAI-compatible providers keep wrapping structured responses in junk. The old path was too trusting and regularly grabbed the wrong payload.

## Summary by CodeRabbit

* **New Features**
  * Improved JSON payload extraction with enhanced handling of noisy input, fenced code blocks, and embedded JSON structures for increased parsing robustness.

* **Tests**
  * Added comprehensive test coverage for JSON extraction and validation scenarios, including edge cases and various formatting styles.

### DIFF
--- a/src/utils/json_parser.py
+++ b/src/utils/json_parser.py
@@ -8,6 +8,135 @@ from json_repair import repair_json
 logger = logging.getLogger(__name__)
 # logging.getLogger("sqlalchemy.engine.Engine").disabled = True
 
+_JSON_FENCE_RE = re.compile(
+    r"```(?P<lang>[a-zA-Z0-9_+-]*)\s*(?P<body>.*?)```", re.IGNORECASE | re.DOTALL
+)
+_THINK_TAG_RE = re.compile(r"<think>.*?</think>", re.IGNORECASE | re.DOTALL)
+
+
+def _strip_reasoning_wrappers(text: str) -> str:
+    """Remove common reasoning/prose wrappers before JSON extraction."""
+    return _THINK_TAG_RE.sub("", text.strip()).strip()
+
+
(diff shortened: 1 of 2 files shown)


akattelu:
Hey @Niko96-dotcom, thanks for opening this.

We ended up consolidating this whole cluster onto #820, which merged in June. A self-hoster sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object`, which switches to `response_format={"type":"json_object"}`, injects the schema with a fence-forbidding instruction, and routes failed parses through `json_repair`. I already closed the companion #577 for the same reason. Because of that I am going to close this too.

Two extra reasons I would not fold this parser change in as-is:
- `validate_and_repair_json` is meant to stay generic. Ranking candidates by deriver keys (`explicit`, `deductive`, `inductive`, ...) bakes one agent's schema into a shared utility. Schema-aware repair already lives downstream in `repair_response_model_json`.
- The scanner still has a few holes (unclosed prefix, a schema-shaped `explicit` winning on length, `<think>` inside string values when the whole blob is not already valid JSON). Those are fixable, but they are also a sign this is more machinery than we want at this layer.

PR #577, opened 2026-04-18
fix(clients): harden structured-output fallback paths

Niko96-dotcom:
## Summary
- tighten the deriver prompt to require exact JSON output and forbid unstated general-knowledge leakage
- add structured-output fallback after parse/schema drift for OpenAI-compatible providers
- scope the MiniMax custom-provider fallback to `PromptRepresentation` instead of every response model
- avoid retrying generic provider/runtime failures through the structured fallback path
- add tests covering validation fallback, non-fallback runtime errors, and the MiniMax prompt-representation branch

## Test Plan
- `set -a && source /Users/nikolaymohr/honcho/.env && set +a && uv run pytest tests/utils/test_clients.py -q`

## Why
The current structured-output path is too brittle for messy OpenAI-compatible providers. This makes fallback behavior explicit instead of magical and reduces the odds of deriver batches dying on parse drift.

## Release Notes

* **Bug Fixes**
  * Added fallback retry mechanism to gracefully handle structured output parsing failures, improving system resilience.
  * Stricter fact extraction now recognizes only explicitly stated information, avoiding inferred facts.

* **Improvements**
  * Enhanced output formatting constraints for consistency and reliability.

### DIFF
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -33,17 +33,25 @@ Analyze messages from {peer_id} to extract **explicit atomic facts** about them.
    - Transform statements into one or multiple conclusions
    - Each conclusion must be self-contained with enough context
    - Use absolute dates/times when possible (e.g. "June 26, 2025" not "yesterday")
+   - Do NOT add facts from general knowledge or unstated implications
 
 RULES:
 - Properly attribute observations to the correct subject: if it is about {peer_id}, say so. If {peer_id} is referencing someone or something else, make that clear.
 - Observations should make sense on their own. Each observation will be used in the future to better understand {peer_id}.
 - Extract ALL observations from {peer_id} messages, using others as context.
 - Contextualize each observation sufficiently (e.g. "Ann is nervous about the job interview at the pharmacy" not just "Ann is nervous")
+- Return ONLY valid JSON.
+- The top-level response MUST be a single JSON object with exactly one key: "explicit".
+- The value of "explicit" MUST be an array of objects shaped like {{"content": "..."}}.
+- Do NOT include markdown fences.
+- Do NOT include explanatory prose.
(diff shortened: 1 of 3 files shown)


akattelu:
Hey @Niko96-dotcom, thanks for opening this. The underlying bug is real: OpenAI-compatible providers that reject or ignore json_schema structured output were making the deriver silently produce zero observations, and that needed a fix.

We ended up addressing this cluster at the config layer in #820, which is now merged, so I'm going to consolidate onto that. It adds a per-model-config `STRUCTURED_OUTPUT_MODE` option, and since the deriver is the only feature using structured output, a self-hoster sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object`. That sends `response_format={"type":"json_object"}`, injects the schema into the prompt with a no-fences instruction that deliberately includes the literal word "JSON" (which satisfies the Alibaba/Qwen "messages must contain 'json'" contract), and routes failed parses through `json_repair`. It also targets the post-refactor LLM layer, whereas this PR is against the old `src/utils/clients.py` path that no longer exists, so it wouldn't apply cleanly anymore.

One thing here that I think is genuinely worth keeping: your reactive try-json_schema-then-fall-back-on-exception angle is different from #820's static opt-in, and it could auto-recover providers that reject json_schema without the self-hoster knowing to set the flag. If we go after zero-config provider compatibility later, that pattern (minus the over-broad TypeError/AttributeError catch CodeRabbit flagged) is a good starting point. Separately, the prompt change to extract only explicitly-stated facts is an unrelated behavioral shift to deriver semantics that I'd want to evaluate on its own rather than fold in here.

Closing in favor of #820 as the single tracked fix. Really appreciate you digging into this, and please give that env var a try on your provider.

PR #581, opened 2026-04-19
fix(surprisal): use correct filter format for level observations

### LINKED ISSUE #559: Bug: Surprisal filter format error - should use {"in": [...]} syntax
## Description

The Surprisal module in `src/dreamer/surprisal.py` line 264 uses an incorrect filter format when calling `get_all_documents()`.

## Problem

```python
# Current code (incorrect)
filters = {"level": levels}
```

This format is not recognized by the `apply_filter` function in `src/utils/filter.py`, which expects:

```python
# Expected format
filters = {"level": {"in": levels}}
```

## Impact

When Surprisal tries to filter observations by level, it returns 0 results, causing:
- Surprisal computation to fail silently
- Dream cycle to skip the Surprisal phase
- No high-surprisal observations are identified

## Fix

Change line 264 in `src/dreamer/surprisal.py`:

```python
# Before
filters = {"level": levels}

# After
filters = {"level": {"in": levels}}
```

## Environment

- Honcho version: latest (commit from April 2026)
- Python: 3.11+

## Verification

After fix:
```bash
docker compose up -d --build deriver
docker exec honcho-deriver-1 python -c "
from src.dreamer.surprisal import get_surprisal_observations
import asyncio
result = asyncio.run(get_surprisal_observations(workspace='test', peer_id='test'))
print(f'Found {len(result)} high-surprisal observations')
"
```

Should return non-zero count of observations.

## Related

This bug was discovered during practical testing with SURPRISAL_ENABLED=true. The fix was verified with a complete Dream cycle that successfully identified 15 high-surprisal observations from 152 total.

Sanjays2402:
## Problem

The Surprisal module in `src/dreamer/surprisal.py` passes a bare list to the `filters` parameter of `get_all_documents()`:

```python
filters={"level": levels}
```

However, `apply_filter()` in `src/utils/filter.py` expects operator-based syntax with an explicit `"in"` key for list membership checks:

```python
filters={"level": {"in": levels}}
```

Without the `"in"` operator wrapper, the filter doesn't match any recognized operator pattern and is silently ignored. This causes `_fetch_level_observations()` to return 0 results, making the entire Surprisal phase of the Dream cycle a no-op — no high-surprisal observations are ever identified.

## Fix

One-line change on line 264 of `src/dreamer/surprisal.py`:

```diff
- filters={"level": levels} if levels else None,
+ filters={"level": {"in": levels}} if levels else None,
```

## Verification

After this fix, `_fetch_level_observations()` correctly filters by observation level, allowing the Surprisal computation to identify and surface high-surprisal observations during the Dream cycle.

Fixes #559

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved level filtering accuracy in the recent sampling strategy to properly handle multiple level values.

### DIFF
--- a/src/dreamer/surprisal.py
+++ b/src/dreamer/surprisal.py
@@ -261,7 +261,7 @@ async def _fetch_recent_observations(
         workspace_name=workspace_name,
         observer=observer,
         observed=observed,
-        filters={"level": levels} if levels else None,
+        filters={"level": {"in": levels}} if levels else None,
         limit=limit,
     )
 


Rajat-Ahuja1997 (approved):
good catch! thanks for the contribution

PR #594, opened 2026-04-21
fix(llm): robust JSON handling for OpenAI-compatible proxies

laychiman:
## Problem

When using OpenAI-compatible proxies (Ollama, vLLM, Together, OpenRouter), the structured output pipeline fails silently — the deriver generates zero observations on every call.

Three failure modes:

1. **Proxies ignore `response_format=json_schema`** — `_create_structured_response` falls back to `json_schema` format, which most proxies don't support. They reject it or ignore it, returning empty/malformed content.

2. **Proxies return markdown-wrapped JSON** — Even with `response_format=json_object`, models like `gemma3:4b` on Ollama return `` ```json {...} ``` `` instead of bare JSON. `json.loads()` fails with "Expecting value: line 1 column 1".

3. **Proxies don't follow the exact Pydantic schema** — They return string arrays instead of object arrays (e.g. `"explicit": ["observation1", "observation2"]` instead of `"explicit": [{"content": "observation1"}]`), or use alternate field names like `"observation"` instead of `"content"`.

## Changes

- **`src/utils/json_parser.py`** — New `extract_json_from_text()` function that strips markdown code fences and extracts bare JSON from prose before validation. Called at the top of `validate_and_repair_json()`.

- **`src/llm/backends/openai.py`** — `_create_structured_response()` now uses `{"type": "json_object"}` instead of `json_schema` format. The repair pipeline handles any schema drift downstream.

- **`src/llm/structured_output.py`** — `repair_response_model_json()` now normalizes schema variants:
  - `explicit`: `string` → `{"content": s}`, `{"observation": x}` → `{"content": x}`
  - `deductive`: `string` → `{"conclusion": s, "premises": []}`

## Testing

Tested with `gemma3:4b-cloud` on Ollama (self-hosted). Before: deriver generates 0 observations on every call. After: deriver generates 4+ observations per message batch, stored successfully in pgvector.

```
╭──────────────── ⚡ PERFORMANCE ─────────────────╮
│   Observation Count    4    count              │
╰────────────────────────────────────────────────╯
```

## Impact

- Zero impact on OpenAI native users (their `chat.completions.parse()` already works)
- Enables self-hosted users on Ollama/vLLM/etc. to use the deriver
- Backward compatible — all normalization is additive to the repair pipeline

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON extraction from text containing surrounding content or markdown formatting for more reliable parsing.
  * Enhanced normalization and repair of structured response data to ensure consistent output handling.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -378,13 +378,13 @@ class OpenAIBackend:
         response_format: type[BaseModel],
     ) -> Any:
         structured_params = dict(params)
-        structured_params["response_format"] = {
-            "type": "json_schema",
-            "json_schema": {
-                "name": response_format.__name__,
-                "schema": response_format.model_json_schema(),
-            },
-        }
+        # Use json_object instead of json_schema — OpenAI-compatible proxies
+        # (Ollama, vLLM, etc.) don't support json_schema and will reject it
+        # or ignore it, returning empty/malformed content. json_object is
+        # widely supported and the repair pipeline (extract_json_from_text
+        # + validate_and_repair_json + repair_response_model_json) handles
+        # markdown-wrapped output correctly.
+        structured_params["response_format"] = {"type": "json_object"}
         return await self._client.chat.completions.create(**structured_params)
 
     @staticmethod
--- a/src/llm/structured_output.py
+++ b/src/llm/structured_output.py
@@ -33,24 +33,48 @@ def repair_response_model_json(
         final = validate_and_repair_json(raw_content)
         repaired_data = json.loads(final)
 
-        if (
-            response_model is PromptRepresentation
-            and "deductive" in repaired_data
-            and isinstance(repaired_data["deductive"], list)
+        if response_model is PromptRepresentation and isinstance(
+            repaired_data, dict
         ):
-            for item in repaired_data["deductive"]:
-                if isinstance(item, dict):
-                    if "conclusion" not in item and "premises" in item:
-                        if item["premises"]:
-                            item["conclusion"] = (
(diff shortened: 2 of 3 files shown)


akattelu:
Hey @laychiman, thanks for opening this. You've diagnosed a real bug: OpenAI-compatible proxies that ignore `response_format=json_schema` make the deriver silently produce zero observations, and your gemma3:4b repro (0 to 4+ observations) nails it.

We ended up consolidating this whole cluster onto #820, which just merged. It adds a per-model-config option, `STRUCTURED_OUTPUT_MODE`, so a self-hoster opts in with `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object`. That sends `{"type":"json_object"}`, injects the schema into the prompt (the instruction deliberately contains the literal word "JSON", which also satisfies the Alibaba/Qwen "messages must contain json" contract), and routes failed parses through `json_repair`. We went opt-in and fail-fast on anthropic/gemini rather than switching the openai backend unconditionally, so native OpenAI users keep `json_schema`.

The one piece of yours I think is genuinely worth harvesting is the schema-drift normalization in `repair_response_model_json()`: coercing weakly-formatted proxy output like `"explicit": ["str1"]` into `[{"content": "str1"}]` and the deductive string case. `json_repair` alone won't fix alternate field names or string arrays from low-capability models, so that layer could harden #820's parsing. If you'd be up for opening a focused PR against #820's `json_object` path for just that normalization (with CodeRabbit's note about coercing a scalar `premises` into a list folded in), I'd happily review it.

Closing this in favor of #820 as the single tracked fix. Thanks again for the clear root-cause analysis.

PR #602, opened 2026-04-22
feat(embedding): forward dimensions to OpenAI embeddings.create

### LINKED ISSUE #601: embedding_client.py: OpenAI path doesn't send dimensions, silently ignoring EMBEDDING_VECTOR_DIMENSIONS
### Summary

In `src/embedding_client.py`, the Gemini transport forwards `output_dimensionality=self.vector_dimensions` to the provider, but the OpenAI transport never passes the corresponding `dimensions` parameter on `embeddings.create(...)`. As a result, `EMBEDDING_VECTOR_DIMENSIONS` is *validated* against the response via `_validate_embedding_dimensions`, but never actually *requested* from the provider. Any OpenAI-compatible server whose native embedding size differs from the configured value fails validation, even when the underlying model fully supports MRL (Matryoshka) truncation.

### Where (on `main`)

Three OpenAI call sites in `src/embedding_client.py`:

1. `_EmbeddingClient.embed` (single query)
2. `_EmbeddingClient.simple_batch_embed`
3. `_EmbeddingClient._process_batch`

Each invokes `self.client.embeddings.create(model=..., input=...)` with no `dimensions=` argument. The Gemini branch alongside each does pass `config={"output_dimensionality": self.vector_dimensions}`.

### Related work

- #463 (closed) made the Gemini `output_dimensionality` read from configuration instead of hardcoded `1536`. That covered the Gemini branch and the `pgvector` column — not the OpenAI branch.
- #585 (open) is a user hitting the same symptom against LM Studio (`VECTOR_STORE_DIMENSIONS=768` but honcho still expects 1536), but it conflates model-name hardcoding, deployment concerns, and the dimensions issue; not a clean report.

### Motivating example

Running [Qwen3-Embedding-4B](https://huggingface.co/Qwen/Qwen3-Embedding-4B) (2560-dim native, MRL-trained) locally via [oMLX](https://github.com/jundot/omlx) with honcho's pgvector mode at `EMBEDDING_VECTOR_DIMENSIONS=1536`:

1. Honcho calls `self.client.embeddings.create(model="qwen3-embedding-4b", input=[query])` — no `dimensions`.
2. oMLX's OpenAI-compatible `/v1/embeddings` endpoint returns 2560 floats (the model's native size).
3. `_validate_embedding_dimensions` raises: `Embedding dimension mismatch ... Expected 1536, got 2560`.

oMLX's endpoint already honors the OpenAI `dimensions` request parameter end-to-end (slices + L2 renormalizes server-side — correct for MRL-trained models), so if honcho passed `dimensions=1536`, it would receive a valid, correctly-renormalized 1536-dim vector and pgvector ingestion would work. The same applies to any MRL-capable model served by an OpenAI-compatible backend (Nomic v1.5, mxbai-embed-large-v1, OpenAI's own `text-embedding-3-small`/`-large`, etc.).

Related server-side work on oMLX: jundot/omlx#901 (adding a per-model default so even clients that don't send `dimensions` get the right size). That is a complementary fix on the *server* side; honcho is the right place to fix it on the *client* side regardless of which backend it talks to.

### Proposal

Pass `dimensions=self.vector_dimensions` on the three OpenAI `embeddings.create` calls, mirroring the existing Gemini behavior.

### Backward-compatibility note

OpenAI rejects the `dimensions` parameter for pre-`text-embedding-3` models (notably `text-embedding-ada-002` — the API returns a 400 if `dimensions` is supplied). To avoid regressing ada-002 users, the cleanest option is a config toggle:

```
EMBEDDING.SEND_DIMENSIONS: bool = True  # opt-out for ada-002 or providers that reject the field
```

An alternative would be to condition on model-name heuristics (e.g., skip when the model starts with `text-embedding-ada`), but that gets fragile against OpenAI-compatible servers using custom model names. A config flag is clearer.

### Scope

Happy to open a PR from a fork with the three-call-site change plus the `SEND_DIMENSIONS` flag and a docs update (`.env.template`, `docs/v3/contributing/configuration.mdx`) if the direction above works for you. Let me know if you'd prefer a different approach (e.g., always-send vs. flag-gated, or model-name heuristic) before I put the PR together.

0xAlcibiades:
Fixes #601.

## Summary

- Forward `dimensions=self.vector_dimensions` on the three OpenAI `embeddings.create` call sites (`embed`, `simple_batch_embed`, `_process_batch`), mirroring the existing Gemini `output_dimensionality` passthrough.
- Add `EMBEDDING_SEND_DIMENSIONS: bool` (default `true`) as an opt-out for providers that reject the field — notably `text-embedding-ada-002`.
- Thread the new setting through the `EmbeddingClient` singleton (init + settings signature) so hot-reloads rebuild cleanly.
- Update `.env.template` and `docs/v3/contributing/configuration.mdx` with the new knob and the ada-002 caveat.

## Why

Today `EMBEDDING_VECTOR_DIMENSIONS` is validated against the response in `_validate_embedding_dimensions` but never requested from the provider on the OpenAI path. That means MRL-capable models (Qwen3-Embedding, Nomic v1.5, mxbai-embed-large-v1, OpenAI's own `text-embedding-3-small`/`-large`) can't negotiate a smaller output size, and any OpenAI-compatible server whose native size ≠ configured value throws on every embed. The Gemini branch already does this correctly via `config={"output_dimensionality": self.vector_dimensions}`; this PR brings the OpenAI branch in line.

Full motivation, backward-compat analysis, and reproduction via Qwen3-Embedding-4B + [oMLX](https://github.com/jundot/omlx) in #601.

## Backward compatibility

`EMBEDDING_SEND_DIMENSIONS=true` is the default. For modern OpenAI-compatible servers this is transparent (they either respect the hint or ignore it). For `text-embedding-ada-002` — which rejects the parameter — set `EMBEDDING_SEND_DIMENSIONS=false`. Documented in both `.env.template` and the v3 configuration doc.

## Test plan

- [x] `uv run pytest tests/llm/test_embedding_client.py` — 4 passed.
  - Updated `test_openai_embedding_client_uses_configured_model_and_dimensions` to assert `dimensions` is now present on the call.
  - New `test_openai_embedding_client_omits_dimensions_when_disabled` covers the opt-out path.
  - Existing Gemini and dimension-mismatch tests unchanged.
- [x] `uv run ruff check` + `uv run ruff format` clean on all edited files.
- [x] `uv run basedpyright` — 0 errors, 0 warnings, 0 notes.
- [ ] End-to-end against a self-hosted MRL-capable backend (oMLX + Qwen3-Embedding-4B → pgvector) — validating on my side; will update this checkbox once confirmed.

## Summary by CodeRabbit

* **New Features**
  * Added configurable embedding-dimension control via EMBEDDING_SEND_DIMENSIONS (default: enabled) to include or omit the dimensions field in embedding requests for provider compatibility.

* **Documentation**
  * Updated configuration docs and examples to describe the new setting and its default behavior.

* **Tests**
  * Added/updated tests to verify behavior with dimensions enabled and disabled.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -18,6 +18,10 @@ LOG_LEVEL=INFO
 # EMBEDDING_VECTOR_DIMENSIONS=1536
 # EMBEDDING_MAX_INPUT_TOKENS=8192
 # EMBEDDING_MAX_TOKENS_PER_REQUEST=300000
+# Forward `dimensions` on OpenAI embeddings.create requests so MRL-capable
+# models return EMBEDDING_VECTOR_DIMENSIONS-sized vectors. Set to false for
+# providers that reject the field (e.g. text-embedding-ada-002).
+# EMBEDDING_SEND_DIMENSIONS=true
 # EMBEDDING_MODEL_CONFIG__TRANSPORT=openai
 # EMBEDDING_MODEL_CONFIG__MODEL=text-embedding-3-small
 # EMBEDDING_MODEL_CONFIG__OVERRIDES__BASE_URL=
--- a/docs/v3/contributing/configuration.mdx
+++ b/docs/v3/contributing/configuration.mdx
@@ -208,6 +208,7 @@ Embeddings use their own nested model config, separate from the main text-genera
 EMBEDDING_VECTOR_DIMENSIONS=1536
 EMBEDDING_MAX_INPUT_TOKENS=8192
 EMBEDDING_MAX_TOKENS_PER_REQUEST=300000
+EMBEDDING_SEND_DIMENSIONS=true  # Forward `dimensions` on OpenAI requests
 
 # Embedding transport/model selection
 EMBEDDING_MODEL_CONFIG__TRANSPORT=openai  # openai, gemini
@@ -218,8 +219,9 @@ EMBEDDING_MODEL_CONFIG__OVERRIDES__BASE_URL=http://localhost:8000/v1
 EMBEDDING_MODEL_CONFIG__OVERRIDES__API_KEY_ENV=EMBEDDING_CUSTOM_API_KEY
 ```
 
-Current constraint:
+Current constraints:
 - `EMBEDDING_VECTOR_DIMENSIONS` can be changed for fully migrated external vector stores, but pgvector and dual-write mode still require `1536` until the schema migration lands.
+- `EMBEDDING_SEND_DIMENSIONS` controls whether the OpenAI transport forwards the `dimensions` parameter on `embeddings.create` (default `true`). Enables Matryoshka (MRL) truncation for MRL-capable models served via OpenAI or OpenAI-compatible backends. Set to `false` for providers that reject the field, notably `text-embedding-ada-002`.
 
 ### Feature-Specific Model Configuration
 
--- a/src/config.py
+++ b/src/config.py
@@ -683,6 +683,10 @@ class EmbeddingSettings(HonchoSettings):
     VECTOR_DIMENSIONS: Annotated[int, Field(default=1536, gt=0)] = 1536
     MAX_INPUT_TOKENS: Annotated[int, Field(default=8192, gt=0)] = 8192
     MAX_TOKENS_PER_REQUEST: Annotated[int, Field(default=300_000, gt=0)] = 300_000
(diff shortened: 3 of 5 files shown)


VVoruganti:
superseded by #678

PR #615, opened 2026-04-25
fix(deriver): ignore blank observations before embedding

flyespresso:
## Summary
- filter blank/whitespace-only observations before batch embedding
- skip the save path entirely when every observation is blank
- add focused unit coverage that avoids database/runtime fixtures

## Motivation
Embedding providers reject empty string inputs. If the deriver produces an empty explicit or deductive observation, saving the representation can fail before any valid observations are persisted.

## Tests
- `.venv/bin/python -m pytest tests/llm/test_representation_manager.py -q`
- `.venv/bin/python -m ruff check src/crud/representation.py tests/llm/test_representation_manager.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Observation content is trimmed of leading/trailing whitespace before use.
  * Empty or whitespace-only observations are excluded from embedding and persistence, with processing skipped when none remain.
  * Embeddings and documents now use the normalized (trimmed) observation text.

* **Tests**
  * Added tests verifying trimming, filtering, embedding behavior, and the early-return/no-op case when all observations are blank.


flyespresso:
Addressed the follow-up review feedback in `aebd54d`.

Changes made:
- extracted shared observation text normalization logic to remove the duplicated explicit/deductive branching
- normalized observation text before both embedding and persistence so stored content matches the embedded payload
- added `DeductiveObservation` coverage for the blank/non-blank filtering path
- updated the existing explicit-observation test to assert normalized persisted content

Re-ran the focused checks from the PR:
- `.venv/bin/python -m pytest tests/llm/test_representation_manager.py -q`
- `.venv/bin/python -m ruff check src/crud/representation.py tests/llm/test_representation_manager.py`

flyespresso:
Addressed the follow-up nit on the test mock access in `9b121e2`.

Changes:
- added a small `_saved_observations()` helper in `tests/llm/test_representation_manager.py`
- both tests now read `await_args` via kwargs-first fallback instead of assuming positional arg layout

Re-ran:
- `./.venv/bin/python -m pytest tests/llm/test_representation_manager.py -q`
- `./.venv/bin/python -m ruff check src/crud/representation.py tests/llm/test_representation_manager.py`

tests/llm/test_representation_manager.py
VVoruganti, line 1: move these tests to the existing `tests/crud/test_representation_manager.py`

src/crud/representation.py
VVoruganti, line 1: The same bug exists in `src/utils/agent_tools` for the agentic use cases. The `create_observations` tool should use the same system. Please unify the fix across both surface areas

VVoruganti (changes requested):

flyespresso:
Addressed the latest review in `7947964`.

Changes made:
- moved the `save_representation` blank-observation tests into the existing `tests/crud/test_representation_manager.py`
- removed the now-redundant `tests/llm/test_representation_manager.py`
- applied the same blank/whitespace filtering + normalization behavior to the agentic tool path in `src/utils/agent_tools.py#create_observations`
- added tool-path coverage in `tests/utils/test_agent_tools.py` for mixed blank/non-blank input and all-blank short-circuiting

Verification:
- `ruff check src/utils/agent_tools.py tests/utils/test_agent_tools.py tests/crud/test_representation_manager.py`
- local `pytest` on the touched test files still requires the repo's Postgres test DB on `localhost:5432`, which is not available in this workspace, so I could not complete those assertions locally here

flyespresso:
Addressed the latest follow-up in `a63ca65`:

- moved `test_query_documents_most_derived_excludes_soft_deleted` back under `TestRepresentationManagerSoftDelete`, where `_setup()` is defined
- re-ran `ruff check tests/crud/test_representation_manager.py`
- verified the class/method layout directly so the soft-delete tests are back on the right fixture class

VVoruganti (approved):

PR #618, opened 2026-04-26
fix(deriver): filter empty-string observations before embedding

24601:
**Problem**

The deriver's `save_representation` builds `observation_texts` from `Representation.explicit` and `.deductive` without guarding against empty or whitespace-only content. The OpenAI embeddings API rejects any batch containing an empty string with HTTP 400:

```
Error code: 400 - Invalid 'input[2]': input cannot be an empty string.
```

This causes `save_representation` to raise, which is caught in the deriver loop and logged as an error — silently dropping all observations for that message batch.

**Fix**

Filter observation/text pairs before the `simple_batch_embed` call, dropping any entry whose content is empty or whitespace-only. If all observations are filtered out, return early with 0 saved — consistent with the existing early return for a fully empty `Representation`.

```diff
-        observation_texts = [
-            obs.conclusion if isinstance(obs, DeductiveObservation) else obs.content
-            for obs in all_observations
-        ]
+        # Filter empty strings — OpenAI embeddings API rejects them with 400
+        obs_text_pairs = [
+            (obs, obs.conclusion if isinstance(obs, DeductiveObservation) else obs.content)
+            for obs in all_observations
+        ]
+        obs_text_pairs = [(obs, text) for obs, text in obs_text_pairs if text and text.strip()]
+        if not obs_text_pairs:
+            logger.debug("All observations have empty content; skipping embedding")
+            return new_documents
+        all_observations, observation_texts = map(list, zip(*obs_text_pairs))
```

**Testing**

Two new tests in `tests/crud/test_representation_manager.py`:

1. `test_empty_content_observations_are_filtered_before_embedding` — mix of valid and empty/whitespace observations; asserts embedding client receives only non-empty texts and the correct document count is returned.
2. `test_all_empty_content_observations_skips_embedding` — all observations empty; asserts embedding client is never called and 0 is returned.

Observed in production logs with observers `moltar`, `basit`, `claude-basits-macbook-air-1`, and `webhook-github` on both `input[2]` and `input[7]`, confirming the index varies with message batch size.

## Bug Fixes
* Fixed API errors that occurred when processing observations with empty or whitespace-only content. The system now gracefully filters out invalid entries before processing.

### DIFF
--- a/src/crud/representation.py
+++ b/src/crud/representation.py
@@ -72,10 +72,17 @@ class RepresentationManager:
         # Batch embed all observations
         batch_embed_start = time.perf_counter()
 
-        observation_texts = [
-            obs.conclusion if isinstance(obs, DeductiveObservation) else obs.content
+        # Filter empty strings — OpenAI embeddings API rejects them with 400
+        obs_text_pairs = [
+            (obs, obs.conclusion if isinstance(obs, DeductiveObservation) else obs.content)
             for obs in all_observations
         ]
+        obs_text_pairs = [(obs, text) for obs, text in obs_text_pairs if text and text.strip()]
+        if not obs_text_pairs:
+            logger.debug("All observations have empty content; skipping embedding")
+            return new_documents
+        all_observations, observation_texts = map(list, zip(*obs_text_pairs))
+
         try:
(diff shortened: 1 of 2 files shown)


VVoruganti:
closing as addressed in #615

PR #620, opened 2026-04-27
fix(deriver): use json_object + prompt schema injection for openai-compatible backends

kaishi00:
## Problem

The OpenAI backend sends `response_format: {"type": "json_schema", ...}` when making structured output calls. In practice, most OpenAI-compatible endpoints — proxies, aggregators (OpenRouter, ZAI, opencode, etc.), and local vLLM servers without guided decoding enabled — silently ignore this field and return either empty content or free-form text.

This causes a cascade of failures:
1. `validate_and_repair_json` raises `ValueError` on empty/non-JSON content
2. `ValueError` is not always in the except clause, so it propagates up
3. Tenacity retries 3 times, then raises `RetryError`
4. The deriver hard-crashes on every task, queue never drains

## Fix

Two changes to `src/llm/backends/openai.py`:

**1. Switch from `json_schema` to `json_object`**
`json_object` is supported by virtually every OpenAI-compatible endpoint. Schema correctness is already handled by the manual parse + fallback logic that follows (`repair_response_model_json` → `PromptRepresentation` fallback).

**2. Inject schema into the prompt**
Append the JSON schema to the last user message so the model knows what structure to produce. This is identical to what the Anthropic path already does. Without this, models receiving `json_object` with no schema guidance return arbitrary JSON keys.

**3. Add `ValueError` to the except clause**
`validate_and_repair_json` raises `ValueError` (not `json.JSONDecodeError`) when repair fails. Adding it to the except ensures the existing fallback path (`PromptRepresentation(explicit=[])`) is reached instead of crashing.

## Testing

Verified against ZAI (GLM models) and opencode.go proxy — both now correctly extract and store observations. The existing fallback logic (`validate_and_repair_json` → schema-aware repair → `PromptRepresentation` fallback) is unchanged.

## Summary by CodeRabbit

* **Chores**
  * Updated internal handling of structured response formatting for OpenAI-compatible API calls. Schema guidance is now provided through prompt instructions rather than API parameters, enhancing compatibility with the API's JSON response format requirements.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -241,14 +241,23 @@ class OpenAIBackend:
         params["stream_options"] = {"include_usage": True}
         if isinstance(response_format, type):
             # parse() supports BaseModel types but streaming create() does not —
-            # convert to a json_schema dict so the streaming path works.
-            params["response_format"] = {
-                "type": "json_schema",
-                "json_schema": {
-                    "name": response_format.__name__,
-                    "schema": response_format.model_json_schema(),
-                },
-            }
+            # convert to a json_object dict so the streaming path works.
+            # json_schema is ignored by many OpenAI-compatible proxies; json_object
+            # is near-universally supported. Schema validation is handled manually
+            # by the repair logic downstream.
+            params["response_format"] = {"type": "json_object"}
+            # Inject schema into the last user message so the model knows the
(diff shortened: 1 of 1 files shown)


kaishi00:
Superseded by #629 — same fix rebased against refactored openai.py backend (code moved from utils/clients.py)

PR #627, opened 2026-04-28
feat(models): make vector embedding dimensions configurable

kaishi00:
## Summary

Hardcoded \`Vector(1536)\` in both \`Document\` and \`MessageEmbedding\` SQLAlchemy models locks users into OpenAI embedding dimensionality. Any alternative embedding model (bge-m3=1024, nomic-embed=768, Cohere v3=1024, etc.) produces a \`ValueError: expected 1536 dimensions, not N\` on every insert.

## Changes

- **\`src/models.py\`**: Read vector dimensions from existing \`VECTOR_STORE.DIMENSIONS\` config setting (default 1536) instead of hardcoding
- Graceful fallback to 1536 if config is unavailable at import time
- Zero behavior change for default OpenAI/Gemini users (still 1536 dims)

## Motivation

The config system already supports \`VECTOR_STORE.DIMENSIONS\` for LanceDB/Turbopuffer vector stores, but the pgvector ORM models ignored this setting. This disconnect meant:
- Users **could** configure a custom embedding provider via \`EMBEDDING_MODEL_CONFIG\`
- But **could not** actually use it because pgvector rejected non-1536 vectors

This unblocks local/Ollama embeddings (bge-m3), open-source models (nomic, e5), and any model with different output dimensions.

## Test Plan

1. Set \`VECTOR_STORE_DIMENSIONS=1024\` in \`.env\`
2. Configure embedding via \`EMBEDDING_MODEL_CONFIG\` for Ollama bge-m3
3. Run deriver — observations save without dimension errors  
4. Default config (no changes) — existing 1536-dim behavior unchanged

## Summary by CodeRabbit

* **Refactor**
  * Embedding configuration now supports dynamic dimension settings with automatic fallback handling.

### DIFF
--- a/src/models.py
+++ b/src/models.py
@@ -5,6 +5,14 @@ from typing import Any, final
 from dotenv import load_dotenv
 from nanoid import generate as generate_nanoid
 from pgvector.sqlalchemy import Vector
+
+# Resolve embedding dimensions from config; fallback to 1536 for backward compat.
+# Must be set before any model class that references Vector(...) is defined.
+try:
+    from .config import settings as _settings
+    _EMBEDDING_DIMS: int = _settings.VECTOR_STORE.DIMENSIONS  # type: ignore[union-attr]
+except Exception:
+    _EMBEDDING_DIMS = 1536
 from sqlalchemy import (
     BigInteger,
     Boolean,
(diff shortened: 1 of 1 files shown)


kaishi00:
Audit complete — this PR contains regressions that would hurt Honcho if merged:

- Removes `delete_documents()` batch function (performance regression)
- Reverts telemetry schema v2→v1 (breaking change for consumers)
- Removes reconciler backoff mechanism (removes resilience feature)
- Deletes tests for the above features

The one legitimate change (configurable embedding dims) needs proper implementation with DB migration. The OpenCode docs update can be resubmitted as a standalone PR.

Closing to re-submit clean fixes separately.

PR #628, opened 2026-04-28
fix(messages): content-hash dedup to prevent gateway restart duplicates

kaishi00:
## Summary

Gateways re-send previously synced messages on restart (new public_id each time). The existing \`public_id\`-based dedup in query paths does **not** catch these at insert time, causing:
- Duplicate messages in the database
- Redundant deriver queue entries (wasted LLM calls)
- Observation bloat from re-processing identical conversations

## Changes

- **\`src/crud/message.py\`**: Add MD5 content-hash deduplication in \`create_messages()\` before insert
- Checks against messages from the last 1 hour in the same session/workspace
- Logs skipped duplicates for observability
- Preserves existing public_id dedup as secondary safety net

## Root Cause

Gateway message tracking is session-scoped. On restart, the gateway loses its
position and re-sends from the last acknowledged offset. Honcho assigns a
**new** \`nanoid\` public_id on each insert, so the same conversation content
enters the DB multiple times with different IDs.

## Test Plan

1. Send messages through a gateway, verify they insert normally
2. Restart gateway, allow it to re-send the same messages
3. Verify second batch is silently dropped (check logs for "Skipping duplicate")
4. Verify no duplicate rows in messages table
5. Verify different content in same session still inserts normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Incoming messages are deduplicated by sender+content within a recent lookback window; duplicates are skipped with informational logging while overall insertion behavior remains unchanged.

* **Configuration**
  * Validation now requires an explicit migrated flag before enforcing vector-dimension constraints.
  * Embedding vector dimensions follow runtime configuration with a safe fallback.

* **Integration**
  * LLM structured responses now use a JSON-object response format with the schema injected into the prompt; injection is applied safely for both streaming and non-streaming paths to avoid schema accumulation.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -1271,11 +1271,12 @@ class AppSettings(HonchoSettings):
             self.METRICS.NAMESPACE = self.NAMESPACE
 
         if self.EMBEDDING.VECTOR_DIMENSIONS != 1536 and (
-            self.VECTOR_STORE.TYPE == "pgvector" or not self.VECTOR_STORE.MIGRATED
+            self.VECTOR_STORE.TYPE == "pgvector" and not self.VECTOR_STORE.MIGRATED
         ):
             raise ValueError(
                 "EMBEDDING.VECTOR_DIMENSIONS must remain 1536 while pgvector is "
-                + "active or vector-store migration is incomplete"
+                + "active and vector-store migration is incomplete "
+                + "(set VECTOR_STORE_MIGRATED=true after migrating)"
             )
 
         return self
--- a/src/crud/message.py
+++ b/src/crud/message.py
@@ -259,10 +259,11 @@ async def create_messages(
                 models.Message.created_at >= _lookback,
             )
         )
-        for _row in _existing_rows.scalars().all():
-            # Composite key: peer_name + content prevents false positives
-            # when different peers say the same thing (e.g. "ok", "thanks")
-            _composite = f"{_row[0]}|{_row[1]}" if isinstance(_row, tuple) else str(_row)
+        # Use .all() NOT .scalars() — scalars() only returns the first
+        # selected column, which would silently drop `content` from our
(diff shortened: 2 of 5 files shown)


VVoruganti:
This PR seems to be related to a different project or a specific integration. It should be possible to send the same exact message content to honcho if that is actually what happened. De-duplication of messages should happen at the integration level

kaishi00:
> This PR seems to be related to a different project or a specific integration. It should be possible to send the same exact message content to honcho if that is actually what happened. De-duplication of messages should happen at the integration level

The problem was if hermes gateway restarts before honcho finishes it's queue, it'll add all the queue items again and do both. You're right it might be hermes specific (maybe), but not exactly a hermes problem either.

VVoruganti:
> > This PR seems to be related to a different project or a specific integration. It should be possible to send the same exact message content to honcho if that is actually what happened. De-duplication of messages should happen at the integration level
> 
> The problem was if hermes gateway restarts before honcho finishes it's queue, it'll add all the queue items again and do both. You're right it might be hermes specific (maybe), but not exactly a hermes problem either.

should be address in the hermes agent repo where the integration is made with some kind of local WAL or cache

kaishi00:
Audit complete — two issues with this PR:

1. **#629 is byte-for-byte identical to #628** (duplicate PR, closing #629)
2. **#628 is a Frankenstein bundle** containing:
   - ✅ Genuine Honcho fixes (json_object fallback, deep-copy, None sentinel in openai.py) — will be resubmitted as clean PR
   - ⚠️ Hermes-specific changes (content-hash dedup, configurable dims) — moving to plugin layer instead
   - ❌ Regressions from #627 (batch delete removal, telemetry revert, backoff removal) — reverting

Closing both. A clean PR with only the upstream-valid openai.py fixes will follow.

PR #629, opened 2026-04-28
fix(openai): use json_object for structured output on compatible endpoints

kaishi00:
## Summary

OpenAI-compatible endpoints (vLLM, Ollama, ZAI, LiteLLM proxies) do NOT reliably support the json_schema response_format. They silently ignore it and return empty string content, causing structured output parsing to fail.

## Changes

- src/llm/backends/openai.py (streaming + non-streaming):
  - Replace json_schema with json_object (universally supported)
  - Inject JSON schema into last user message so model knows expected shape
  - Mirrors existing Anthropic backend approach for non-native structured output

## Root Cause

OpenAI added json_schema structured output in late 2024 requiring server-side validation. OpenAI-compatible endpoints ignore it:
- vLLM: ignores json_schema, returns empty content
- Ollama: partial support, model-dependent
- ZAI/LiteLLM proxies: strip unknown parameters

json_object is universally supported since 2023. The existing repair_response_model_json() fallback handles malformed output.

## Test Plan

1. Configure Honcho with vLLM or ZAI as LLM provider
2. Trigger deriver run — should produce observations (not empty)
3. Test with native OpenAI — should still work
4. Verify schema injection appears in request payloads

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent duplicate messages by deduplicating incoming content within a recent time window.
  * Improve structured and multimodal responses so JSON outputs are more consistent; schema prompts are embedded into messages and retries no longer produce repeated edits.

* **Chores**
  * Config validation tightened to only error for pgvector dimension mismatches when migration is incomplete; clarifies setting VECTOR_STORE_MIGRATED=true.
  * Embedding vector dimensionality is now configurable via settings.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -1271,11 +1271,12 @@ class AppSettings(HonchoSettings):
             self.METRICS.NAMESPACE = self.NAMESPACE
 
         if self.EMBEDDING.VECTOR_DIMENSIONS != 1536 and (
-            self.VECTOR_STORE.TYPE == "pgvector" or not self.VECTOR_STORE.MIGRATED
+            self.VECTOR_STORE.TYPE == "pgvector" and not self.VECTOR_STORE.MIGRATED
         ):
             raise ValueError(
                 "EMBEDDING.VECTOR_DIMENSIONS must remain 1536 while pgvector is "
-                + "active or vector-store migration is incomplete"
+                + "active and vector-store migration is incomplete "
+                + "(set VECTOR_STORE_MIGRATED=true after migrating)"
             )
 
         return self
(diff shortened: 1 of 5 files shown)


kaishi00:
Audit complete — two issues with this PR:

1. **#629 is byte-for-byte identical to #628** (duplicate PR, closing #629)
2. **#628 is a Frankenstein bundle** containing:
   - ✅ Genuine Honcho fixes (json_object fallback, deep-copy, None sentinel in openai.py) — will be resubmitted as clean PR
   - ⚠️ Hermes-specific changes (content-hash dedup, configurable dims) — moving to plugin layer instead
   - ❌ Regressions from #627 (batch delete removal, telemetry revert, backoff removal) — reverting

Closing both. A clean PR with only the upstream-valid openai.py fixes will follow.

PR #632, opened 2026-04-28
fix(llm): add json_object fallback for non-OpenAI structured output

kaishi00:
## Summary

When using non-OpenAI LLM backends (vLLM, Ollama, etc.), the `json_schema` response format is not supported. This PR adds a `json_object` fallback mode that works across all OpenAI-compatible endpoints while maintaining `json_schema` as the preferred format for native OpenAI users.

### Changes
- **openai.py**: Detect when transport does not support json_schema; inject json_object format instructions as fallback
- **openai.py**: Deep-copy message list before mutation to prevent state corruption
- **openai.py**: Add None sentinel for `_last_user_idx` edge case
- **openai.py**: Skip multimodal content parts with empty text
- **config.py**: Fix vector dimension validator logic (pgvector migration guard)

### Behavior
- **OpenAI-native**: unchanged (uses `json_schema` with better enforcement)
- **vLLM/Ollama/other**: uses `json_object` with prompt-injected schema (portable, works everywhere)

Fixes #629 (clean re-submission after audit).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation logic for vector store configurations to accurately detect incomplete migrations with non-standard embedding dimensions
  * Improved OpenAI structured output handling to prevent schema duplication across retries and ensure consistent JSON formatting

* **Improvements**
  * Enhanced error messaging for vector store migration status

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -1271,11 +1271,12 @@ class AppSettings(HonchoSettings):
             self.METRICS.NAMESPACE = self.NAMESPACE
 
         if self.EMBEDDING.VECTOR_DIMENSIONS != 1536 and (
-            self.VECTOR_STORE.TYPE == "pgvector" or not self.VECTOR_STORE.MIGRATED
+            self.VECTOR_STORE.TYPE == "pgvector" and not self.VECTOR_STORE.MIGRATED
         ):
             raise ValueError(
                 "EMBEDDING.VECTOR_DIMENSIONS must remain 1536 while pgvector is "
-                + "active or vector-store migration is incomplete"
+                + "active and vector-store migration is incomplete "
+                + "(set VECTOR_STORE_MIGRATED=true after migrating)"
             )
 
         return self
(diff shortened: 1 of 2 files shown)


akattelu:
Hey @kaishi00, thanks for opening this. The underlying bug is real: on OpenAI-compatible providers that reject or ignore `json_schema` response format (vLLM, Ollama, Z.AI GLM, and similar), the deriver was silently producing zero observations, and your `json_object` plus prompt-schema-injection approach targets exactly the right root cause.

We've since consolidated this onto #820, which landed the same strategy as an explicit per-model-config opt-in. Self-hosters can now turn it on with `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` (the deriver is the only feature using structured output, though other features can set their own). The injected instruction deliberately includes the literal word "JSON," which also satisfies the Alibaba/Qwen "messages must contain 'json'" contract, and parsing is routed through `json_repair` so fenced or prefixed output is largely tolerated. Since #820 is merged and covers the same ground, I'm going to close this one to keep a single tracked fix.

One piece here is genuinely orthogonal and worth keeping: the `src/config.py` change narrowing the embedding-dimension validator to only raise when the store is pgvector and `VECTOR_STORE_MIGRATED==false`, with the clearer message pointing at `VECTOR_STORE_MIGRATED=true`. That's separate from the structured-output work and #820 doesn't touch it, so I think it'd be worth a small standalone PR if you're up for splitting it out. Thanks again for the contribution.

PR #633, opened 2026-04-30
fix(llm/openai): normalize tool_choice "any" → "required"

thrialectics:
OpenAI rejects tool_choice="any" with a 400 (only accepts none/auto/required), but Anthropic and Gemini backends both treat "any" and "required" as equivalent (anthropic.py:321, gemini.py:568), and tool_loop.py:382 assumes the same. Callers passing "any" — including via env/config overrides not covered by the default-only patch in #630 — would still hit the error.

Map "any" → "required" inside the OpenAI backend so callers can use either spelling without provider-specific awareness.

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenAI backend now properly converts tool choice parameters to ensure full compatibility with the OpenAI API specification. Tools will function correctly with proper parameter formatting.

* **Tests**
  * Added comprehensive test coverage to verify tool choice parameter normalization is functioning as expected in the OpenAI backend.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -309,6 +309,10 @@ class OpenAIBackend:
         if tools:
             params["tools"] = self._convert_tools(tools)
             if tool_choice is not None:
+                # OpenAI accepts only "none" | "auto" | "required" (or a function spec).
+                # Other backends use "any" with the same semantics as "required".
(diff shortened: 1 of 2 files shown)


VVoruganti:
Closing as still and addressed

PR #636, opened 2026-05-01
feat(llm): add azure_openai transport

jlec:
## Summary

Adds a new `"azure_openai"` transport so every `ModelConfig` site (deriver, dialectic levels, summary, dream specialists, embedding, fallback) can target a native **Azure OpenAI** resource — or any Azure-compatible gateway that speaks the Azure URL shape and `api-key` header.

This is the last missing piece needed to run Honcho against enterprise Azure OpenAI deployments, where the classic `AsyncOpenAI` client cannot reach the upstream model at all (it produces the wrong URL shape and wrong auth header).

## Motivation

Before this change, `DERIVER_MODEL_CONFIG__OVERRIDES__BASE_URL` can be pointed at any OpenAI-compatible endpoint, but that specifically does **not** include Azure OpenAI. Azure expects:

1. URL: `{endpoint}/openai/deployments/{deployment}/chat/completions?api-version={v}` — not `{base_url}/chat/completions`
2. Auth header: `api-key: <key>` — not `Authorization: Bearer <key>`
3. Reasoning models require `max_completion_tokens` (already handled by the existing `_uses_max_completion_tokens` heuristic)

I confirmed this from inside a running container against a real Azure-compatible gateway: classic `AsyncOpenAI` → `HTTP 500 "UnhandledError"` on every call; `AsyncAzureOpenAI` with the same key → `HTTP 200` with a real completion. The key and deployments were fine — just the client shape was wrong.

## Design

**One insight collapsed most of the surface area:** `AsyncOpenAI` and `AsyncAzureOpenAI` from the `openai` SDK expose the identical `chat.completions.create(...)` / `embeddings.create(...)` surface. Only the constructor differs (`azure_endpoint=` + `api_version=` instead of `base_url=`). Therefore:

- **No new backend class.** `OpenAIBackend` works unchanged — it receives either client type.
- **No new history adapter.** `history_adapter_for_provider` already routes non-Anthropic / non-Gemini through `OpenAIHistoryAdapter`.
- The change is purely: new transport literal, new client factory, and threading a single new `api_version` field through the existing config resolvers.

### New transport + config field

- `ModelTransport` and `EmbeddingTransport` literals extend with `"azure_openai"`.
- `api_version: str | None = None` added to `ModelOverrideSettings`, `ModelConfig`, `ResolvedFallbackConfig`, and `EmbeddingModelConfig`.
- `AZURE_OPENAI_API_KEY` added to `LLMSettings` (picks up `LLM_AZURE_OPENAI_API_KEY` via the existing `env_prefix="LLM_"` pattern).
- `api_version` propagates end-to-end: `resolve_model_config`, `_resolve_fallback_config`, `resolve_embedding_model_config`, and fallback swap-in in `select_model_config_for_attempt`. Legacy `"provider/model"` shorthand sets recognize `"azure_openai"`.

### Client factories + dispatch

- `src/llm/registry.py` gains `get_azure_openai_client` (singleton) and `get_azure_openai_override_client(azure_endpoint, api_key, api_version)` (LRU-cached). The override factory validates both `azure_endpoint` and `api_version` are present — they're structurally required by Azure, so failing fast here prevents confusing runtime errors later.
- `client_for_model_config` and `backend_for_provider` each get an `"azure_openai"` branch; the backend branch returns `OpenAIBackend` since the surface matches.

### Credentials + embedding

- `default_transport_api_key("azure_openai")` → `settings.LLM.AZURE_OPENAI_API_KEY`.
- `_EmbeddingClient.__init__` gets an `azure_openai` branch constructing `AsyncAzureOpenAI(azure_endpoint=..., api_version=...)`. Downstream dispatch (`isinstance(client, genai.Client)`) is unchanged, so the existing OpenAI code path handles embedding calls verbatim.

### Drive-by fix

While developing and running against a partially-configured dialectic (`DIALECTIC_LEVELS__minimal__*` only), the API crashed at startup with:

```
ValidationError: Missing configuration for reasoning levels: {'low', 'medium', 'high', 'max'}
```

Root cause: `DialecticSettings._merge_level_defaults` iterated only over user-supplied level overrides, so partially overriding one level silently dropped the other four defaults, and `_validate_all_levels_present` rejected the result. Rewritten to iterate over every built-in default and layer the user's override on top when present (existing `MODEL_CONFIG` merge semantics + transport-change thinking-param scrubbing preserved). Regression test included.

### Docs

- `.env.template`, `config.toml.example`: commented Azure examples for each module (deriver / dialectic levels / summary / dream / embedding), mirroring the existing OpenRouter-style examples.
- `CHANGELOG.md`: entry under `## [Unreleased] / ### Added`.

### Tests

- `tests/llm/test_azure_openai_routing.py` (new): asserts `client_for_model_config("azure_openai", cfg)` constructs `AsyncAzureOpenAI(azure_endpoint=..., api_version=...)`, and `backend_for_provider("azure_openai", client)` returns `OpenAIBackend`.
- `tests/llm/test_credentials.py`: adds `default_transport_api_key("azure_openai")`.
- `tests/llm/test_model_config.py`: Azure literal acceptance on `Configured*Settings` / `Fallback*Settings`, `api_version` propagation through all three resolvers, end-to-end env-var binding (`DERIVER_MODEL_CONFIG__OVERRIDES__API_VERSION=...`), and the dialectic-level merge regression.
- `tests/llm/test_embedding_client.py`: Azure branch wires up correctly and raises when `api_version` is missing.

## Usage

```env
LLM_AZURE_OPENAI_API_KEY=<key>

DERIVER_MODEL_CONFIG__TRANSPORT=azure_openai
DERIVER_MODEL_CONFIG__MODEL=gpt-4o-mini                # Azure deployment name
DERIVER_MODEL_CONFIG__OVERRIDES__BASE_URL=https://your-resource.openai.azure.com
DERIVER_MODEL_CONFIG__OVERRIDES__API_VERSION=2024-10-21
DERIVER_MODEL_CONFIG__OVERRIDES__API_KEY_ENV=LLM_AZURE_OPENAI_API_KEY
```

Same shape works for `EMBEDDING_*`, `DIALECTIC_LEVELS__*`, `SUMMARY_*`, and `DREAM_*` — fallbacks can also target Azure.

## Verification

- `uv run ruff check src/ tests/llm/` — clean.
- `uv run basedpyright src/config.py src/llm/ src/embedding_client.py` — 0 errors / 0 warnings.
- `uv run pytest tests/llm/` — **90 passed** (89 previous + 1 regression + 4 new Azure-specific in existing files + 2 new in `test_azure_openai_routing.py` → netting out to +1 over the prior total because some new tests replaced scaffolding assertions).
- End-to-end: swapped the gateway endpoint into `.env`, rebuilt the compose stack, and confirmed the container receives real completions through the Azure URL shape with no more `UnhandledError` from the gateway.

## Out of scope (noted for follow-ups)

- Entra-ID / `DefaultAzureCredential` / `azure_ad_token_provider=` auth — would need a separate `api_key_provider` field. Only `api-key` auth is in scope here.
- Separate `deployment_name` field — the SDK's convention of reusing `model` as the Azure deployment name is kept.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added Azure OpenAI as a supported transport for both LLM completions and embeddings, including Azure endpoint + `api_version` routing.
  * Added optional Entra ID (Azure AD) authentication support (`use_entra_id`) for Azure OpenAI requests.

* **Documentation**
  * Updated env/config examples and changelog with Azure OpenAI setup guidance, placeholders, and per-module configuration examples.

* **Tests**
  * Added coverage for Azure routing, credentials, embedding behavior, and configuration propagation.

* **Bug Fixes**
  * Improved validation and error messages for missing Azure-specific settings (e.g., `api_version`) and Entra ID prerequisites.


jlec:
Addressed the three inline review comments in c7f23a7:

- **`src/embedding_client.py`** — added `runtime_config.api_version` to `_get_settings_signature()` so the cached client is invalidated when only the api_version changes. Good catch; would have been a nasty silent-staleness bug.
- **`tests/llm/test_azure_openai_routing.py`** — wrapped the test body in `try`/`finally` and re-clear the `get_azure_openai_override_client` LRU cache on exit. Prevents the monkeypatched `FakeAsyncAzureOpenAI` from leaking into later tests via the `(endpoint, key, version)` cache key.
- **`tests/llm/test_model_config.py::test_env_vars_bind_azure_openai_api_version`** — now calls `_clear_deriver_env(monkeypatch)` at the top, matching the three sibling direct-construction tests in this file.

Also took the `get_azure_openai_client` nitpick and **deleted** the function rather than "fix" it. `AsyncAzureOpenAI(api_key=...)` without `azure_endpoint`/`api_version` will fail at construction, and the singleton had zero callers — I originally added it for symmetry with the other default-client helpers, but that symmetry was never going to be usable for Azure (which requires per-config endpoint + version). The override factory `get_azure_openai_override_client` is now the only Azure entry point.

---

On the last nitpick — swapping `ValueError` → `ValidationException` in the `_EmbeddingClient` Azure branch — I'm going to **leave this as-is**. The neighboring gemini and openai branches (lines 44 and 74) already raise plain `ValueError` for identical missing-credential shape. My new Azure branch matches that local pattern. Converting only the Azure branch would introduce inconsistency, and converting all three is outside the scope of this PR — happy to do it in a follow-up if the maintainers want the swap.

Verified: `uv run ruff check src/ tests/llm/` clean, `uv run basedpyright` clean on all touched modules, `uv run pytest tests/llm/` 90/90 passing.

tgoodyear:
Great work on this PR! We have the same need at our org and confirmed the exact same problem: `AsyncOpenAI` produces the wrong URL shape and auth header for Azure OpenAI endpoints.

I noticed you explicitly scoped out Entra ID / `DefaultAzureCredential` auth. We need exactly that because our Azure OpenAI resource has `disableLocalAuth=true` (Entra ID tokens only, no API keys).

I built a small follow-up (59 insertions, 6 files) on top of your branch that adds a `use_entra_id: bool` config field. When enabled, it swaps `api_key=` for `azure_ad_token_provider=` using `DefaultAzureCredential` from `azure-identity`. Lazy import so the dependency is only required when used.

- **Issue**: #651
- **Branch**: [tgoodyear/honcho@feat/azure-ad-token-auth](https://github.com/tgoodyear/honcho/tree/feat/azure-ad-token-auth)

Happy to open a follow-up PR once this lands, or squash it into this PR as an additional commit if you'd prefer. Let me know!

jlec:
Let's get it into this pr and drop a single working integration

jlec:
Squashed @tgoodyear's Entra ID branch into this PR as commit e1daf51 (kept him as author). Closes #651.

Added two small dep-hygiene tweaks on top of his diff before the squash:

1. **Moved `azure-identity` from top-level `dependencies` to an optional extra.** Now:

   ```toml
   [project.optional-dependencies]
   azure = ["azure-identity>=1.15.0"]
   ```

   Rationale: `azure-identity` drags in `msal`, `cryptography`, `azure-core` (~5 MB + native deps). API-key users shouldn't pay that cost. Entra ID users opt in with `pip install 'honcho[azure]'`.

2. **Wrapped the lazy imports in `try/except ImportError`** so a misconfigured user (sets `use_entra_id=true` but hasn't installed the extra) gets a clear actionable error instead of a bare `ModuleNotFoundError`:

   ```
   azure_openai use_entra_id requires the azure extra.
   Install with: pip install 'honcho[azure]'
   ```

   Raised via `ValidationException` in `registry.py` and `ValueError` in `embedding_client.py` to match each file's local idiom (what the neighboring gemini / openai branches use). Added regression tests in both `test_azure_openai_routing.py` and `test_embedding_client.py` that stub `builtins.__import__` to force the ImportError and assert the install hint is in the raised message.

### Verification

- `uv run ruff check src/ tests/llm/` — clean
- `uv run basedpyright src/config.py src/llm/ src/embedding_client.py` — 0 errors (12 warnings are all cascading `reportUnknownVariableType` from the suppressed lazy imports, expected)
- `uv run pytest tests/llm/` — **100 passed** (89 before Entra ID + 7 Trevor added + 2 install-hint regression + ... netting to 100)
- Tested in-container end-to-end is unchanged vs the API-key path (the Entra ID path needs a real Azure tenant to verify, which @tgoodyear confirmed works on their deployment)

Thanks @tgoodyear — the patch was a near-drop-in; only the dep packaging needed adjustment. The combined PR now lands Azure OpenAI transport with both auth modes in a single coherent changeset.

zboyles:
I previously added a comment in #651 noting success applying the referenced branch commits but I've now performed a second test, still using **Hermes-Agent** integration, this time running in a container together with **Honcho** in the same docker compose project. I'm happy to again report the success of locally merging this PR into into the repo used to build the `honcho` image, allowing all of the **Honcho** tooling to function correctly while using Azure API key auth. Cheers.

jerbob92:
Anything that can be done to push this forward? Would really like to try out Honcho using the Azure hosted endpoints.

PR #638, opened 2026-05-02
fix: add dimensions parameter to OpenAI-compatible embeddings.create() calls (#625)

### LINKED ISSUE #625: OpenAI transport embeddings.create() missing dimensions parameter
Problem

  Root Cause

  OpenRouter, etc.).

  it entirely.

  Affected Code

  # Current
  )
  # Should be
  )

  # Current
      input=batch,
      model=self.model,
  )
  # Should be
      input=batch,
      model=self.model,
  )

  # Current
  )
  # Should be
      model=self.model,
  )

  Compatibility Concern

  Environment

shellybotmoyer:
### Problem
When using an OpenAI-compatible embedding provider (e.g., DashScope/Bailian text-embedding-v4) that supports a `dimensions` parameter but defaults to a different output dimensionality than what Honcho expects, the embedding client fails with a dimension mismatch error:

```
```

### Root Cause
The OpenAI `embeddings.create()` API supports a `dimensions` parameter that lets the caller specify the desired output dimensionality. Many OpenAI-compatible providers also support this parameter. However, none of the 3 `embeddings.create()` call sites in `embedding_client.py` passed `dimensions=self.vector_dimensions`, so providers with different default dimensions would always produce wrong-sized vectors.

### Fix
Added `dimensions=self.vector_dimensions` to all 3 OpenAI transport `embeddings.create()` call sites:
1. Single embed (line 101) — `model=self.model, input=[query], dimensions=self.vector_dimensions`
2. Simple batch embed (line 138) — `input=batch, model=self.model, dimensions=self.vector_dimensions`
3. Batch embed with text IDs (line 292) — `model=self.model, input=[item.text for item in batch], dimensions=self.vector_dimensions`

The Gemini transport path already correctly passes `config={"output_dimensionality": self.vector_dimensions}`.

### Testing
- Verified that OpenAI-compatible providers with `dimensions` support will now produce correctly-sized embeddings
- Providers that don't support `dimensions` will ignore the parameter (per OpenAI API spec, unknown parameters are silently ignored by most providers)

## Release Notes

* **Documentation**
  * Added upgrade notes documenting new LLM client configuration requirements and embedding resilience improvements.

* **Bug Fixes**
  * System now gracefully handles missing embeddings instead of failing, enabling continued operation and later backfill.
  * Improved quota exhaustion detection and error handling to fail fast on rate limits.

* **Refactor**
  * Updated duplicate detection logic to only run when embeddings are available.

### DIFF
--- a/memory/2026-05-02-upgrade-notes.md
+++ b/memory/2026-05-02-upgrade-notes.md
@@ -1,0 +1,51 @@
+# Honcho Upgrade Notes (2026-05-02)
+
+## Critical: LLM Module Restructuring
+
+The upstream `plastic-labs/honcho` repo completely restructured the LLM client code:
+- `src/utils/clients.py` → **DELETED** (was the "custom" provider path for Ollama)
+- New: `src/llm/` module with registry, backends (openai, anthropic, gemini), structured output
+- `ModelTransport = Literal["anthropic", "openai", "gemini"]` — no "custom" transport
+- OpenAI-compatible endpoints (Ollama, OpenRouter, etc.) use `transport=openai` with `base_url` override
+
+## New Config Format
+
+Old (DEPRECATED, silently ignored):
+```
+DERIVER_PROVIDER=custom
+DERIVER_MODEL=qwen3-coder:480b
+LLM_OPENAI_COMPATIBLE_BASE_URL=https://ollama.com/v1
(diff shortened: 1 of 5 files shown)


shellybotmoyer:
Superseded by #640 which is branched from clean origin/main

PR #640, opened 2026-05-02
fix: add dimensions parameter to OpenAI-compatible embeddings.create() calls

### LINKED ISSUE #625: OpenAI transport embeddings.create() missing dimensions parameter

shellybotmoyer:
## Summary

Fixes #625 — OpenAI-compatible embedding providers (DashScope, Bailian, etc.) that support a `dimensions` parameter now receive the correct vector dimensionality, preventing `Embedding dimension mismatch` errors (e.g., "Expected 1536, got 1024").

## Problem

`embedding_client.py` has four `embeddings.create()` call sites for the OpenAI-compatible path. None of them pass the `dimensions` parameter, so providers that support it (like DashScope `text-embedding-v4`) return their default dimensionality instead of the configured `EMBEDDING_VECTOR_DIMENSIONS` value. This causes a mismatch between the expected and actual embedding dimensions.

## Fix

Add `dimensions=self.vector_dimensions` to all OpenAI-compatible `embeddings.create()` call sites in `src/embedding_client.py`. The Gemini path already correctly passes `output_dimensionality` — this brings the OpenAI path to parity.

## Testing

- Verified that the Gemini path already uses `output_dimensionality` correctly
- Confirmed all 4 OpenAI-compatible call sites now pass `dimensions=self.vector_dimensions`
- Existing `_validate_embedding_dimensions()` method catches dimension mismatches if any provider ignores the parameter

## Bug Fixes
* Fixed OpenAI embedding requests to ensure configured dimensionality settings are explicitly applied across all embedding operations, improving consistency and reliability.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -99,7 +99,7 @@ class _EmbeddingClient:
             return self._validate_embedding_dimensions(response.embeddings[0].values)
         else:  # openai
             response = await self.client.embeddings.create(
-                model=self.model, input=[query]
+                model=self.model, input=[query], dimensions=self.vector_dimensions
             )
             return self._validate_embedding_dimensions(response.data[0].embedding)
 
@@ -138,6 +138,7 @@ class _EmbeddingClient:
                     response = await self.client.embeddings.create(
                         input=batch,
                         model=self.model,
+                        dimensions=self.vector_dimensions,
                     )
                     embeddings.extend(
                         [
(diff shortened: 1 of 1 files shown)


VVoruganti:
superseded by #678

PR #642, opened 2026-05-03
fix: pass provider params to OpenAI-compatible embeddings

LittleLory:
Allow embedding provider-specific parameters to be passed through to OpenAI-compatible embedding APIs. This enables providers that support configurable output dimensions (e.g., via a dimensions parameter) to return embeddings matching Honcho's configured VECTOR_DIMENSIONS.

Changes:
- Add provider_params field to EmbeddingModelConfig
- Preserve embedding overrides.provider_params during config resolution
- Pass provider params to all three OpenAI embedding call sites
- Add tests for single and batch embedding calls with provider params
- Add backward compatibility test proving empty provider_params works

The implementation reuses the existing provider_params pattern from ModelOverrideSettings, keeping the change backward-compatible. Existing configs without provider_params behave identically.

## Summary by CodeRabbit

* **New Features**
  * Embedding models now accept and forward provider-specific parameters (e.g., dimensions) to the embedding provider for single and batch requests.

* **Bug Fixes**
  * Client re-initialization now respects changes to provider-specific parameters so updates take effect.

* **Tests**
  * Added tests verifying provider-parameter preservation, forwarding, and omission when not configured.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -330,6 +330,7 @@ class EmbeddingModelConfig(BaseModel):
     transport: EmbeddingTransport = "openai"
     api_key: str | None = None
     base_url: str | None = None
+    provider_params: dict[str, Any] = Field(default_factory=dict)
 
     @model_validator(mode="before")
     @classmethod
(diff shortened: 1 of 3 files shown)


VVoruganti:
superseded by #678

PR #644, opened 2026-05-03
fix: add JSON-only output instructions to deriver prompt

LittleLory:
## Summary

- Adds explicit JSON-only output format instructions to the minimal deriver prompt

## Rationale

Some OpenAI-compatible providers accept `response_format=json_schema` but do not reliably enforce it, returning natural language instead of JSON. Adding explicit natural-language instructions in the prompt gives these providers a fallback signal to return valid JSON even when they ignore the schema constraint.

## Test plan

- `uv run pytest tests/llm/test_deriver_prompts.py -q`
- `uv run ruff check src/deriver/prompts.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the deriver prompt to enforce strict JSON-only output format, eliminating potential markdown interference with response parsing.

* **Tests**
  * Added test coverage validating JSON-only output requirements and response format specifications.

### DIFF
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -45,6 +45,13 @@ EXAMPLES:
 - EXPLICIT: "I took my dog for a walk in NYC" → "{peer_id} has a dog", "{peer_id} lives in NYC"
 - EXPLICIT: "{peer_id} attended college" + general knowledge → "{peer_id} completed high school or equivalent"
 
+OUTPUT FORMAT:
+Return ONLY valid JSON. No markdown. No explanation. No prose before or after the JSON.
+The JSON must exactly match this shape:
+{{"explicit":[{{"content":"self-contained fact about {peer_id}"}}]}}
+If there are no explicit facts about {peer_id}, return exactly:
+{{"explicit":[]}}
+
 Messages to analyze:
 <messages>
 {messages}
--- a/tests/llm/test_deriver_prompts.py
+++ b/tests/llm/test_deriver_prompts.py
@@ -1,0 +1,13 @@
+from src.deriver.prompts import minimal_deriver_prompt
+
+
+def test_minimal_deriver_prompt_requires_json_only_output() -> None:
+    prompt = minimal_deriver_prompt(
+        peer_id="alice",
+        messages="[2026-05-03] alice: I enjoy hiking on weekends.",
+    )
+
+    assert "Return ONLY valid JSON" in prompt
+    assert "No markdown" in prompt
+    assert '"explicit"' in prompt
+    assert '{"explicit":[]}' in prompt


Mertolgahan:
## Root Cause Analysis: Why Prompt-Only Fix May Not Be Sufficient

We've been debugging the same issue in our self-hosted Honcho deployment and found that **the prompt changes alone may not resolve the root cause** for all OpenAI-compatible providers.

### What We Discovered

After extensive debugging, we traced the issue to how `json_mode=True` interacts with `response_model` in `src/llm/api.py`: the `response_format` is set to `{"type": "json_object"}` for ALL requests when `json_mode=True`, regardless of whether `response_model` is also specified. Some providers (like Ollama with certain model families) **accept `json_schema` but silently return empty `content`** — the JSON ends up in `reasoning` instead.

### Evidence

From our deriver logs:
```
2026-05-04 10:41:21 - ERROR - ❌ Repair failed: Expecting value: line 1 column 1 (char 0)
```

This happens because:
1. Model receives prompt + `response_format={type: "json_object"}` + JSON schema
2. Model returns empty `content` (structured data goes into `reasoning` or `refusal`)
3. `response.content` is `None` or empty string
4. JSON parser → `Expecting value: line 1 column 1`

### Suggested Addition

In addition to the prompt changes (which are great!), we recommend also reviewing the `json_mode` → `response_format` mapping in `src/llm/api.py`. When `response_model` is specified alongside `json_mode=True`, some providers behave differently than OpenAI.

We're working on a complementary fix at our fork — happy to collaborate if useful!

cc: @plastic-labs team

akattelu:
Hey @LittleLory, thanks for opening this. The bug is real: on OpenAI-compatible providers that accept `response_format=json_schema` but don't enforce it, the deriver was silently returning zero observations, so the instinct to give the model an explicit JSON directive is the right one.

We ended up consolidating this whole cluster onto #820 (now merged), which adds a per-model-config `STRUCTURED_OUTPUT_MODE` option. A self-hoster sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` and we then send `response_format={"type":"json_object"}`, inject the schema into the prompt with a no-markdown instruction, and deliberately include the literal word "JSON" (which also satisfies the Alibaba/Qwen "messages must contain 'json'" contract). The main reason I'd lean on that approach over this one is gating: #820 only changes behavior when you opt in, so Anthropic, Gemini, and well-behaved `json_schema` providers are untouched, whereas appending the directive unconditionally in `src/deriver/prompts.py` affects every provider. Parsing is also routed through `json_repair` to tolerate fenced or prefixed output.

One thing I want to flag from @Mertolgahan's comment: the empty-content case they traced (output landing in the reasoning/refusal field, producing `Expecting value: line 1 column 1`) is a distinct failure mode. `json_repair` is noise-tolerant but there's no dedicated `<think>`-block stripper, so it's largely covered but not guaranteed. If that specific empty-content case persists on Ollama after switching to `json_object`, a failing sample would help us a lot.

Going to close this in favor of #820 as the single tracked fix. Thanks again for the report and the test, and please reopen or ping if `json_object` doesn't resolve it on your provider.

PR #654, opened 2026-05-06
fix: remove default_query stream param causing streaming format response

xchliu:
## Problem

The \`default_query={"stream": "false"}\` in \`src/utils/clients.py\` uses a **string** \`"false"\` instead of boolean \`False\`. This causes OpenAI-compatible APIs (like Finna/custom providers) to return **streaming format responses** (plain strings) instead of proper \`ChatCompletion\` objects.

This breaks the Deriver pipeline:
- \`response.usage\` fails with \`AttributeError: str object has no attribute usage\`
- 3 retries all fail, no observations/conclusions are produced  
- Deriver becomes completely non-functional

## Root Cause

HTTP query parameters are always strings. When \`default_query={"stream": "false"}\` is set, the URL gets \`?stream=false\` appended. Some APIs interpret this as enabling streaming (since its not the proper boolean \`False\` in the request body), while others default to streaming mode when they encounter an unrecognized query parameter.

The \`stream=False\` parameter should be passed in the **request body** (as a proper boolean), not in the URL query string. The OpenAI SDK already defaults \`stream=False\` when no stream parameter is provided, so removing \`default_query\` entirely fixes the issue.

## Fix

Remove \`default_query={"stream": "false"}\` from the custom OpenAI client initialization in \`src/utils/clients.py\`. The SDK default \`stream=False\` behavior is sufficient.

## Testing

- Verified that Deriver now produces observations correctly (7 observations generated from test messages)
- No more \`AttributeError\` on \`response.usage\`
- LLM calls return proper ChatCompletion objects  
- Conclusion generation pipeline restored

## Summary by CodeRabbit

* **Style**
  * Minor formatting adjustment with no impact on user experience.
* **Chore**
  * Internal handling and parsing logic refined to make non-streaming behavior more explicit and robust; no changes to public interfaces or visible behavior.

### DIFF
--- a/src/utils/clients.py
+++ b/src/utils/clients.py
@@ -267,6 +267,7 @@ if settings.LLM.OPENAI_COMPATIBLE_API_KEY and settings.LLM.OPENAI_COMPATIBLE_BAS
     CLIENTS["custom"] = AsyncOpenAI(
         api_key=settings.LLM.OPENAI_COMPATIBLE_API_KEY,
         base_url=settings.LLM.OPENAI_COMPATIBLE_BASE_URL,
+        
     )
 
 # vLLM uses separate settings for local model serving
--- a/src/utils/clients.py
+++ b/src/utils/clients.py
@@ -1876,6 +1876,7 @@ async def honcho_llm_call_inner(
             openai_params: dict[str, Any] = {
                 "model": params["model"],
                 "messages": processed_messages,
+                "stream": False,
             }
 
             if temperature is not None and "gpt-5" not in model:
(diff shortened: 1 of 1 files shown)


shellybotmoyer:
## Diagnosis: PR targets a file deleted in upstream restructure

**Root cause of DIRTY merge state:** This PR modifies `src/utils/clients.py`, but that file was **deleted** in upstream's LLM module restructure (commit range around `f37338b`). The `clients.py` module was replaced by `src/llm/registry.py`, `src/llm/backends/openai.py`, and other files in `src/llm/`.

**Verification:**
- `src/utils/clients.py` no longer exists on `origin/main`
- The new `src/llm/registry.py` and `src/llm/backends/openai.py` do **not** contain `default_query` at all
- Therefore, the bug this PR fixes (`default_query={"stream": "false"}` string/boolean mismatch) does not exist in current `main`

**Recommendation:**
This PR is safe to close as superseded by upstream's module restructure. If a similar `default_query` bug reappears in `src/llm/`, a new targeted PR should be opened against the new files.

Refs: `plastic-labs/honcho#654` (this PR), LLM restructure notes in `src/llm/`.

VVoruganti:
closing for staleness as this predates the refactor of `clients.py` into the `llm/` module

PR #655, opened 2026-05-06
fix(deriver): scope search_messages by observed peer to prevent cross-peer bio leak

thrialectics:
## Summary

Fix a cross-peer information leak in the dreamer's deduction specialist. The `search_messages` tool was returning all session messages regardless of author, so when the deduction specialist ran with `observed=<peer_A>` and asked the LLM to "extract bio facts about {observed}", it could pick up text from `<peer_B>`'s messages and write it as facts about `<peer_A>`. End result: user bio info ending up in the assistant's peer card (and vice versa).

## Mechanism

`_handle_search_messages` in `src/utils/agent_tools.py` was calling `crud.search_messages` with `observer` and `session_name` filters but no author filter. The same applied at every layer below — neither the pgvector backend nor the external-vector-store path filtered matched messages by author.

The dreamer specialist's `ToolContext` already carries the `observed` peer (from `create_tool_executor(... observed=observed ...)` in `src/dreamer/specialists.py:202`), so the fix is plumbing it from the executor context through to the SQL filter.

## Changes

**Production:**
- `src/utils/agent_tools.py` — `_handle_search_messages` now passes `peer_name=ctx.observed` to `crud.search_messages`
- `src/crud/message.py` — added `peer_name: str | None = None` parameter to:
  - `search_messages` (public API)
  - `_semantic_search_messages` (orchestrator)
  - `_search_messages_pgvector` (pgvector backend; SQL filter `models.Message.peer_name == peer_name`)
  - `_fetch_messages_by_ids` (external vector-store path; same SQL filter)

**Behavior:**
- When `peer_name` is set, **matched messages** are restricted to those authored by that peer.
- **Context messages** (the surrounding conversation, ±2 turns) are unaffected so conversational flow remains readable.
- Default is `None` for backward compatibility — only the agent-tools handler opts in.

## Tests

**TDD:** `tests/utils/test_agent_tools.py::TestSearchMessages::test_filters_by_observed_peer_author` was written first as a RED test, asserting the handler passes `peer_name=ctx.observed` to `crud.search_messages`. Confirmed RED with `AssertionError: search_messages was called without a peer_name filter`. Then implemented the fix and confirmed GREEN.

**Regression:** Two `fake_fetch_messages_by_ids` fixtures in `tests/integration/test_message_embeddings.py` needed updating to accept the new `peer_name` kwarg in their signatures. Caught by running the broader test suite after the production change.

Full test run after fix: **56 passed**.

## Test plan

- [x] New unit test asserts handler passes `peer_name=ctx.observed`
- [x] All existing `tests/utils/test_agent_tools.py` tests still pass
- [x] All existing `tests/integration/test_message_embeddings.py` tests still pass after fixture updates
- [ ] Verify in production: re-run a dream cycle on a multi-peer session and confirm user bio facts no longer appear in the assistant's representation/peer card
- [ ] Verify the same bug doesn't recur via the search_memory empty-fallback path (also calls into message search — could be a follow-up if it matters)

## Summary by CodeRabbit

**New Features**
- Added peer-scoping capability to message search, enabling search queries to be contextually filtered and isolated by the specific peer being observed, improving result relevance and ensuring proper data isolation.

**Tests**
- Added unit tests to verify peer-scoped message search functionality is properly implemented and integrates correctly with the search system.

### DIFF
--- a/src/crud/message.py
+++ b/src/crud/message.py
@@ -722,6 +722,7 @@ async def _fetch_messages_by_ids(
     *,
     after_date: datetime | None = None,
     before_date: datetime | None = None,
+    peer_name: str | None = None,
 ) -> list[models.Message]:
     """Fetch messages by ID, preserving the supplied ordering."""
     fetch_stmt = (
@@ -733,6 +734,8 @@ async def _fetch_messages_by_ids(
         fetch_stmt = fetch_stmt.where(models.Message.created_at >= after_date)
     if before_date:
         fetch_stmt = fetch_stmt.where(models.Message.created_at <= before_date)
+    if peer_name:
+        fetch_stmt = fetch_stmt.where(models.Message.peer_name == peer_name)
 
     result = await db.execute(fetch_stmt)
     messages_by_id = {msg.public_id: msg for msg in result.scalars().all()}
@@ -751,6 +754,7 @@ async def _search_messages_pgvector(
     before_date: datetime | None = None,
     limit: int = 10,
     context_window: int = 2,
+    peer_name: str | None = None,
 ) -> list[tuple[list[models.Message], list[models.Message]]]:
     """Run semantic message search against pgvector-backed embeddings."""
     # pgvector path: cosine distance in SQL
@@ -781,6 +785,8 @@ async def _search_messages_pgvector(
         match_stmt = match_stmt.where(models.Message.created_at >= after_date)
     if before_date:
         match_stmt = match_stmt.where(models.Message.created_at <= before_date)
+    if peer_name:
+        match_stmt = match_stmt.where(models.Message.peer_name == peer_name)
 
     result = await db.execute(match_stmt)
     matched_messages = _deduplicate_messages(result.scalars().all(), limit)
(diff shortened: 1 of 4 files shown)


Rajat-Ahuja1997 (changes requested):
my concern is that the `search_messages` tool is shared by the dreamer and the dialectic, and this change impacts dialectic query semantics. a valid dialectic query can ask about something that another peer said in a session — `search_messages` in this case wouldn't be able to fetch messages created by other peers. one way you can fix this is to gate the `peer_name` author scoping by `ctx.agent_type` so dialectic can keep the unscoped message-match behavior

thrialectics:
Closing in favor of a corrected fix at the right layer.

@Rajat-Ahuja1997 and Vineeth's reviews together pointed at the layer mismatch:

1. Rajat noted that `search_messages` is shared between the dreamer and dialectic, and unconditional peer-author scoping breaks legitimate dialectic queries about other peers.
2. Vineeth noted that the broad retrieval is intentional — the deduction specialist needs to see other peers' messages as context to interpret the observed peer's statements. So the bug isn't in retrieval; it's in how observations get attributed at peer-card-write time.

The actual bug surface is the deduction/induction specialist prompts (`src/dreamer/specialists.py`), not `search_messages` plumbing. New PR coming with:
- Prompt updates to `DeductionSpecialist.build_system_prompt` and `InductionSpecialist.build_system_prompt` PEER CARD sections, making attribution discipline explicit ("entries must be facts about {observed}, not about another peer who appears as context").
- New test in `tests/unified/test_cases/` that asserts cross-peer bio facts don't end up in the wrong peer's card.

Thanks both for catching this.

PR #664, opened 2026-05-08
fix(llm): ensure json keyword in messages for Alibaba/Qwen compatibility

### LINKED ISSUE #663: [Bug] `json_mode=True` fails on Alibaba/Qwen providers — "messages must contain the word 'json'"
The OpenAI API spec requires that when `response_format` of type `json_object` is used, the word "json" must appear somewhere in the messages. Most OpenAI-compatible providers (OpenAI itself, NVIDIA, etc.) are forgiving about enforcing this — they'll return valid JSON regardless. However, Alibaba-backed Qwen models strictly enforce it and reject the call:

```
[400]: Error from provider (Alibaba): <400> InternalError.Algo.InvalidParameter: 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.
```

This affects the deriver, which calls `honcho_llm_call(json_mode=True, response_model=PromptRepresentation)`. The OpenAI SDK's `parse()` method sends `response_format` as `json_schema`, but Alibaba internally converts this to `json_object` and enforces the keyword check — the word "json" must appear somewhere in the messages.

**Root cause:**

Alibaba/Qwen strictly enforces the OpenAI spec requirement that "json" (case-insensitive) must appear in the messages whenever `response_format` is set, regardless of whether it's `json_object` or `json_schema`. Other providers are lenient about this check. The OpenAI backend never ensures the keyword is present, so any `json_mode=True` call routed to a Qwen model fails immediately.

**Proposed fix:**

Add an `_ensure_json_keyword()` helper to the OpenAI backend that checks whether "json" appears in the messages when `json_mode=True` is set. If not, appends a brief instruction ("Respond in valid JSON format.") to the system message. This is gated on `json_mode` specifically — not on every `response_format` call — so it only activates when the caller explicitly requests JSON output, avoiding noise for structured output calls to providers that don't need it.

**Reproduction:**

```python
# Via any OpenAI-compatible proxy pointing to Alibaba Qwen
response = await client.chat.completions.create(
    model="qwen3.6-plus",
    messages=[{"role": "user", "content": "Hello"}],
    response_format={"type": "json_object"},
)
# → 400: 'messages' must contain the word 'json'
```

kevin-ho:
Closes #663

When `json_mode=True` is set, Alibaba/Qwen providers strictly enforce the OpenAI spec requirement that the word "json" must appear in messages. Most providers are lenient about this check but Qwen rejects the call with a 400.

Adds `_ensure_json_keyword()` helper that appends a brief instruction to the system message when `json_mode` is active and no message contains the keyword. Gated on `json_mode` specifically to avoid noise for structured output calls to providers that don't need it.

**Changes:**
- New module-level constant `_JSON_KEYWORD_HINT` and helper `_ensure_json_keyword()` in `src/llm/backends/openai.py`
- Called in both `complete()` and `stream()` when `extra_params["json_mode"]` is set
- Early return if any message already contains "json" (case-insensitive) — no unnecessary injection
- If a system message exists, the hint is appended to it; otherwise a new system message is prepended

**Testing:**
Verified against `opencode-go/qwen3.6-plus` via OmniRoute gateway — previously 400'd immediately, now succeeds on first attempt.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for structured JSON output across OpenAI-compatible providers by ensuring chat input reliably triggers JSON-formatted responses when structured-output mode is enabled.
  * Enhances reliability of returned JSON without changing public APIs or altering existing response formatting behavior.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -18,6 +18,45 @@ from src.llm.structured_output import (
 logger = logging.getLogger(__name__)
 
 
+_JSON_KEYWORD_HINT = "Respond in valid JSON format."
+
+
+def _ensure_json_keyword(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Ensure the word "json" appears in messages when response_format is set.
+
+    Many OpenAI-compatible providers (notably Alibaba/Qwen) require the word
+    "json" to appear in the messages when ``response_format`` is set, even if
+    the format uses ``json_schema`` rather than ``json_object``.  Alibaba
+    internally converts ``json_schema`` to ``json_object`` and then enforces
+    the keyword check.
+
+    If no message contains "json" (case-insensitive), a brief instruction is
+    appended to the system message (or a new system message is prepended).
+    """
+    if any("json" in str(m.get("content", "")).lower() for m in messages):
+        return messages
+
+    result = list(messages)
+    # Find an existing system message and append the hint
+    for i, msg in enumerate(result):
+        if msg.get("role") == "system":
+            updated = dict(msg)
+            content = updated.get("content", "")
+            if isinstance(content, str):
+                updated["content"] = content.rstrip() + f"\n\n{_JSON_KEYWORD_HINT}"
+            elif isinstance(content, list):
+                updated["content"] = list(content) + [
+                    {"type": "text", "text": _JSON_KEYWORD_HINT}
+                ]
+            else:
+                continue  # unrecognised content shape — fall through to prepend
+            result[i] = updated
+            return result
+    # No system message — prepend one
+    result.insert(0, {"role": "system", "content": _JSON_KEYWORD_HINT})
+    return result
+
+
 def _uses_max_completion_tokens(model: str) -> bool:
     """OpenAI reasoning models (gpt-5 family + o-series) require
     ``max_completion_tokens`` instead of the classic ``max_tokens`` parameter.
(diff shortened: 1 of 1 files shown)


kevin-ho:
Thanks for the review @coderabbitai[bot] — I looked into the `AttributeError` concern on the `.rstrip()` call.

This is a false positive. The `isinstance` branches are mutually exclusive:

- **Line 49** (`if isinstance(content, str)`): `.rstrip()` is only called here, inside the string branch.
- **Line 52** (`elif isinstance(content, list)`): Appends a new `{"type": "text", "text": ...}` block — no `.rstrip()` anywhere.

The detection path (line 36) uses `str()` as a catch-all for the keyword check, but the mutation path already gates on type before touching string methods. A list-typed `content` will never reach `.rstrip()`.

No changes needed.

akattelu:
Hey @kevin-ho, thanks for opening this. The diagnosis in #663 is exactly right: Qwen/Alibaba strictly enforce the OpenAI requirement that the literal word "json" appear in messages, and without it the call 400s, so structured output silently breaks on those providers.

We ended up consolidating this whole class of OpenAI-compatible provider issues onto #820, which just merged. It adds a per-model-config `STRUCTURED_OUTPUT_MODE` option, and setting `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` switches the deriver to `response_format={"type":"json_object"}` plus a prompt-injected schema. That injected instruction deliberately contains the literal word "JSON", which satisfies the same Qwen/Alibaba contract your `_ensure_json_keyword()` helper targets. Parsing is also routed through `json_repair` to tolerate fenced or prefixed output from these providers.

One thing from your PR I think is worth keeping in mind: the idempotent "only inject if the keyword is absent" guard is more surgical than #820's always-inject approach. If we ever add a path where a caller supplies their own system prompt for `json_object` mode, that early-return is a nice way to avoid redundant hint text.

Going to close this in favor of #820 as the single tracked fix, but the #663 diagnosis was a real help in getting there. Thanks again.

PR #665, opened 2026-05-09
fix(config,llm): add default base_url support for OpenAI, Anthropic, Gemini clients

### LINKED ISSUE #641: Self-hosted Honcho with OpenAI-compatible providers (OpenRouter / vLLM): default `AsyncOpenAI` client lacks `base_url` → 401 against `api.openai.com`
## Summary

When deploying self-hosted Honcho with an OpenAI-**compatible** provider (OpenRouter, vLLM, Together, Anyscale, etc.) — i.e. setting `LLM_OPENAI_API_KEY` to a non-OpenAI key — every dialectic / deriver / summary call fails with:

```
openai.AuthenticationError: Error code: 401 - {'error': {'message': 'Incorrect API key provided: sk-…',
  'type': 'invalid_request_error', 'code': 'invalid_api_key'}, 'status': 401}
```

The SDK error mentions `platform.openai.com` because the client is hitting `https://api.openai.com/v1` instead of the configured provider.

There is no warning that `config.toml` placed `PROVIDER` / `MODEL` at the wrong nesting level either — the operator silently runs against the hardcoded default model.

## Repro

`.env`
```
LLM_OPENAI_API_KEY=sk-or-v1-…              # OpenRouter key
LLM_VLLM_API_KEY=sk-or-v1-…
LLM_VLLM_BASE_URL=https://openrouter.ai/api/v1
LLM_EMBEDDING_API_KEY=sk-or-v1-…
LLM_EMBEDDING_BASE_URL=https://openrouter.ai/api/v1
```

`config.toml` excerpt (legacy flat shape, common in the wild):
```toml
[dialectic.levels.minimal]
PROVIDER = "vllm"
MODEL = "z-ai/glm-4.7-flash"
```

Then call `peer.chat()` / `dialectic.chat()`. Result: `tenacity.RetryError` wrapping `AuthenticationError`. After `retry_attempts=3`, the 401 surfaces to clients.

## Root cause #1 — default OpenAI client lacks `base_url`

`src/llm/registry.py:98`
```python
if settings.LLM.OPENAI_API_KEY:
    CLIENTS["openai"] = AsyncOpenAI(
        api_key=settings.LLM.OPENAI_API_KEY,
        # no base_url → SDK defaults to https://api.openai.com/v1
    )
```

`LLMSettings` (`src/config.py:643`) has `OPENAI_API_KEY` but **no** `OPENAI_BASE_URL` field, so any operator value is silently ignored.

## Root cause #2 — flat `PROVIDER`/`MODEL` in `config.toml` silently fall back to default model

`DialecticLevelSettings` expects nested:
```toml
[dialectic.levels.minimal.model_config]
transport = "openai"
model = "z-ai/glm-4.7-flash"
```

But `config.toml.example` and most copy-pasted configs in the wild use the flat shape. When the parsed `model_config` is missing, Honcho falls back to `_default_dialectic_levels()` which hardcodes `model="gpt-5.4-mini"` — a model that doesn't exist on any provider, so OpenRouter responds `401 User not found.`. **No warning is logged.**

## Workaround

1. Add `OPENAI_BASE_URL=https://openrouter.ai/api/v1` to `.env` (the OpenAI Python SDK reads this env var on `AsyncOpenAI()` instantiation).
2. Reformat `config.toml`: replace each `PROVIDER`/`MODEL` pair with a nested `[<section>.model_config]` block (`transport=`, `model=`).
3. Note: `config.toml` is **baked into the Docker image** — `docker compose restart` won't pick up host-side edits. Use `docker cp` or mount it as a volume in `docker-compose.yml`.

## Proposed fix

1. **Add `OPENAI_BASE_URL` to `LLMSettings`** and pass it to `AsyncOpenAI(base_url=…)` in `registry.py`. Same for `EmbeddingSettings`.
2. **Validate `config.toml` shape on startup**: if a `[dialectic.levels.*]` (or `[deriver]`, `[summary]`, `[dream]`) section contains `PROVIDER`/`MODEL` at top level (instead of nested `model_config`), log a `WARNING` and either auto-migrate or refuse to start.
3. **Mount `config.toml` as a volume** in `docker-compose.yml.example` so operator edits survive restarts.

## Environment

- Honcho: latest `main` as of May 2026
- Deploy: docker compose self-hosted
- Provider: OpenRouter (also reproduces with vLLM)
- SDK: `openai` Python

shellybotmoyer:
## What

Self-hosted Honcho users who set `LLM_OPENAI_BASE_URL` (or `LLM_ANTHROPIC_BASE_URL`, `LLM_GEMINI_BASE_URL`) for OpenRouter / vLLM / local endpoints get 401 errors because the **default** client constructors ignore base_url entirely, always routing to the hardcoded provider endpoint (e.g. `api.openai.com`).

The per-service **override** path (`ModelOverrideSettings.base_url`) already works for model configs, but the top-level default path does not. This means self-hosted Honcho silently falls back to the wrong endpoint unless the user also sets a per-service `MODEL_CONFIG__OVERRIDES__BASE_URL`, which is non-obvious.

## Root Cause

- `LLMSettings` does not define `OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`, or `GEMINI_BASE_URL` — because of `extra="ignore"`, the env vars are silently dropped.
- `src/llm/registry.py` default client factories (`get_openai_client`, `get_anthropic_client`, `get_gemini_client`) and the module-level `CLIENTS` dict never read a base_url.

## Changes

- Added `ANTHROPIC_BASE_URL`, `OPENAI_BASE_URL`, `GEMINI_BASE_URL` to `LLMSettings` (env prefix `LLM_`, so `LLM_OPENAI_BASE_URL` etc. work).
- Wired each base_url into:
  - `get_openai_client()` / `get_anthropic_client()` / `get_gemini_client()`
  - The module-level `CLIENTS` dict entries
- Used `genai_types.HttpOptions` for Gemini base_url wiring.

## Test Plan

- Run `hermes doctor` or start Honcho with `LLM_OPENAI_BASE_URL="https://api.openrouter.ai/api/v1"` and verify no `404/401` on warmup checks.

## Related

Closes #641

## Release Notes

* **New Features**
  * Added configurable base URL settings for LLM providers (Anthropic, OpenAI, Gemini).

* **Bug Fixes**
  * Improved error handling for embedding service rate limiting and quota exhaustion.
  * Enhanced duplicate detection logic for documents with missing embeddings.
  * System can now save documents without embeddings when generation fails.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -648,6 +648,11 @@ class LLMSettings(HonchoSettings):
     OPENAI_API_KEY: str | None = None
     GEMINI_API_KEY: str | None = None
 
+    # Base URLs for LLM providers (optional — override the SDK default endpoint)
+    ANTHROPIC_BASE_URL: str | None = None
+    OPENAI_BASE_URL: str | None = None
+    GEMINI_BASE_URL: str | None = None
+
     # General LLM settings
     DEFAULT_MAX_TOKENS: Annotated[int, Field(default=1000, gt=0, le=100_000)] = 2500
 
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -453,7 +453,8 @@ async def create_documents(
             # for each document, if deduplicate is True, perform a process
             # that checks against existing documents and either rejects this document
             # as a duplicate OR deletes an existing document that is a duplicate.
-            if deduplicate:
+            # Skip dedup when embedding is None (backfill by reconciler later)
+            if deduplicate and doc.embedding is not None:
                 is_duplicate = await is_rejected_duplicate(
                     db, doc, workspace_name, observer=observer, observed=observed
                 )
--- a/src/crud/representation.py
+++ b/src/crud/representation.py
@@ -102,6 +102,15 @@ class RepresentationManager:
                 "Observation content exceeds maximum token limit of "
                 + f"{settings.EMBEDDING.MAX_INPUT_TOKENS}."
             ) from e
+        except Exception as e:
+            # Embedding API failure (e.g. Gemini daily quota exhausted) —
+            # save observations without embeddings; the reconciler will backfill
+            # after quota resets at midnight Pacific.
+            logger.warning(
+                f"Embedding API rate limited, saving {len(all_observations)} "
+                f"observations without embeddings (reconciler will backfill): {e}"
+            )
+            embeddings = None
 
         batch_embed_duration = (time.perf_counter() - batch_embed_start) * 1000
         accumulate_metric(
@@ -138,7 +147,7 @@ class RepresentationManager:
         self,
         db: AsyncSession,
         all_observations: list[ExplicitObservation | DeductiveObservation],
-        embeddings: list[list[float]],
+        embeddings: list[list[float]] | None,
         message_ids: list[int],
         session_name: str,
         message_created_at: datetime.datetime,
@@ -154,7 +163,7 @@ class RepresentationManager:
 
         # Prepare all documents for bulk creation
         documents_to_create: list[schemas.DocumentCreate] = []
-        for obs, embedding in zip(all_observations, embeddings, strict=True):
+        for i, obs in enumerate(all_observations):
             # NOTE: will add additional levels of reasoning in the future
             if isinstance(obs, DeductiveObservation):
                 obs_level = "deductive"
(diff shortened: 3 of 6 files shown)


shellybotmoyer:
Superseded by #643. This PR (#665) was opened from <code>main</code> and contains the same commits as #643 (<code>fix/llm-default-client-base-url</code>). Closing in favor of the properly-branched PR.

PR #668, opened 2026-05-10
fix(llm): inject system prompt with JSON schema for structured output

leugenea:
## Summary
Some OpenAI-compatible models ignore the `response_format` parameter and return 
markdown-formatted JSON or include explanations, causing downstream parsing failures.

## Changes
- In `OpenAIBackend.generate_structured`, inject a system prompt containing the 
  target JSON schema when no existing system message is present.
- The injected prompt instructs the model to respond with raw JSON only, without 
  markdown or explanations.

## Motivation
Increases the robustness of structured output generation across models that do not 
strictly adhere to the `response_format` contract.

## Testing
- Existing structured output tests continue to pass.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced structured output handling in the AI backend with automatic system message injection when needed, improving JSON schema compliance and response reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/668)

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -357,9 +357,11 @@ class OpenAIBackend:
 
         cache_creation, cache_read = extract_openai_cache_tokens(usage)
         return CompletionResult(
-            content=content_override
-            if content_override is not None
-            else (message.content or ""),
+            content=(
+                content_override
+                if content_override is not None
+                else (message.content or "")
+            ),
             input_tokens=usage.prompt_tokens if usage else 0,
             output_tokens=usage.completion_tokens if usage else 0,
             cache_creation_input_tokens=cache_creation,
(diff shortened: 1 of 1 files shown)


akattelu:
Hey @leugenea, thanks for opening this. You're right that this is a real bug: on OpenAI-compatible providers that ignore or reject the `json_schema` contract, the deriver silently produces zero observations, and your instinct to inject the schema into the prompt is the right shape of fix.

We've since consolidated this whole cluster onto #820, which is now merged, so I'm going to close this in favor of that. The main reason is that #820 makes the behavior opt-in via a new per-model-config option (`STRUCTURED_OUTPUT_MODE`) rather than changing every openai-transport structured-output call, and it actually switches the request to `response_format={"type":"json_object"}` instead of keeping `json_schema`. That last part matters for providers like Alibaba/Qwen that reject the `json_schema` request outright, and its prompt instruction deliberately contains the literal word "JSON" to satisfy their "messages must contain 'json'" contract. It also doesn't have the `if not has_system` gap, so it applies even when a system prompt is already present, and it hardens the parse side through `json_repair`.

The deriver is the only feature that uses structured output, so a self-hoster turns it on with:

```
DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object
```

Docs are in `docs/v3/contributing/configuration.mdx` and the "Deriver produces no observations" section of `docs/v3/contributing/troubleshooting.mdx`. Really appreciate you taking the time to dig into this. If you flip on that setting and still hit a parsing issue on your provider, ping me with the failing output sample and I'll take a look.

PR #670, opened 2026-05-11
feat(openai): extra_body passthrough + embeddings dimensions + json_object default

ZVin-Chen:
## Summary

Three small, additive improvements to the OpenAI backend (`src/llm/backends/openai.py` + `src/embedding_client.py`) that broaden compatibility with OpenAI-compatible providers (DeepSeek, Alibaba Bailian, vLLM-hosted models, etc.) without changing behaviour on OpenAI itself.

### 1. `provider_params` → OpenAI SDK `extra_body` passthrough

`ConfiguredModelSettings.overrides.provider_params` is documented as a free-form passthrough for backend-specific knobs, but the OpenAI backend's `_build_params` was cherry-picking a small whitelist (`top_p`, `frequency_penalty`, `presence_penalty`, `seed`, `verbosity`) and silently dropping the rest. Any unrecognised key now flows through to the OpenAI SDK's `extra_body` parameter, which is the canonical way to pass provider-specific fields.

This unblocks e.g.:
```toml
[dialectic.levels.minimal.model_config.overrides]
provider_params = { thinking = { type = "disabled" } }
```
…on DeepSeek's v4 family, which defaults `thinking.type=enabled` (returns `reasoning_content` + rejects `tool_choice=required`).

### 2. Forward `dimensions` on embedding calls

`embedding_client.py` already validates that returned embedding length matches `vector_dimensions`, but it doesn't pass `dimensions=` on the wire. With Alibaba Bailian's `text-embedding-v4`, default output is 1024-d but the deployment's pgvector schema typically uses 1536. Calling with `dimensions=self.vector_dimensions` lets the operator-configured `EMBEDDING_VECTOR_DIMENSIONS` actually take effect — OpenAI's own `text-embedding-3-*` accepts the same parameter, so this is portable.

Applied at all three OpenAI embedding call sites (single embed, batch embed, batch embed with text ids).

### 3. Default structured-output to `json_object` + schema-in-prompt

When a Pydantic class is passed as `response_format`, the OpenAI backend was first trying `chat.completions.parse()` (Structured Outputs, `json_schema` mode), then falling back to a `json_schema` payload via `_create_structured_response`. **Both paths use `json_schema`**, which DeepSeek-v4 family and several Bailian / vLLM models reject (`This response_format type is unavailable now` / `Failed to deserialize the JSON body`).

Default now goes straight to `{"type": "json_object"}` (universally supported) with the target Pydantic schema injected as a system-message instruction. The existing `repair_response_model_json` machinery already handles minor JSON shape drift, so the looser enforcement is acceptable in exchange for portability. Streaming path (`stream`) gets the same treatment.

Trade-off: on OpenAI itself, this gives up strict-mode `json_schema` in favour of `json_object` + prompted schema. In practice the parsed output is virtually identical because `gpt-*` models comply with schema descriptions reliably; the repair logic catches any drift.

## Why these three together

They surfaced together while wiring honcho into a stack that uses DeepSeek as the dialectic / deriver LLM and Alibaba Bailian (`text-embedding-v4`) for embeddings. Each change is independent and small but they share the same motivation: make the OpenAI backend gracefully cover the long tail of OpenAI-compatible providers.

## End-to-end verification

Tested against:
- LLM: DeepSeek v4-flash via `https://api.deepseek.com/v1` with `thinking={type:disabled}` injected through provider_params
- Embeddings: Alibaba Bailian `text-embedding-v4` via `https://dashscope.aliyuncs.com/compatible-mode/v1`, `dimensions=1536`

Full dialectic chat across all 5 reasoning levels (`minimal` / `low` / `medium` / `high` / `max`) returns synthesized answers; deriver creates observations end-to-end; dreamer specialists (deduction + induction) complete a full `update_peer_card` cycle.

## Notes

- No new dependencies.
- No behaviour change on OpenAI's own endpoints other than (a) embeddings now explicitly request `dimensions` (matches OpenAI's documented param) and (b) structured outputs route through `json_object` instead of `json_schema` (still produces valid Pydantic-parseable JSON via repair logic).
- The submodule is wired into a downstream consumer ([ZVin-Chen/emotional_agent#17](https://github.com/ZVin-Chen/emotional_agent/pull/17)) that exercises the stack end-to-end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed embedding request dimension consistency across all operation types to ensure proper vector generation.
  * Improved structured JSON output handling with better validation, error detection, and automatic correction mechanisms.

* **Improvements**
  * Enhanced provider parameter compatibility by expanding support for additional configuration options.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/670)

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -99,7 +99,9 @@ class _EmbeddingClient:
             return self._validate_embedding_dimensions(response.embeddings[0].values)
         else:  # openai
             response = await self.client.embeddings.create(
-                model=self.model, input=[query]
+                model=self.model,
+                input=[query],
+                dimensions=self.vector_dimensions,
             )
             return self._validate_embedding_dimensions(response.data[0].embedding)
 
@@ -138,6 +140,7 @@ class _EmbeddingClient:
                     response = await self.client.embeddings.create(
                         input=batch,
                         model=self.model,
+                        dimensions=self.vector_dimensions,
                     )
                     embeddings.extend(
                         [
@@ -283,7 +286,9 @@ class _EmbeddingClient:
                                 )
                 else:  # openai
                     response = await self.client.embeddings.create(
-                        model=self.model, input=[item.text for item in batch]
+                        model=self.model,
+                        input=[item.text for item in batch],
+                        dimensions=self.vector_dimensions,
                     )
                     for item, embedding_data in zip(batch, response.data, strict=True):
                         result[item.text_id][item.chunk_index] = (
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -311,14 +311,27 @@ class OpenAIBackend:
             if tool_choice is not None:
                 params["tool_choice"] = tool_choice
         if extra_params:
-            for key in (
+            # Recognised top-level params we already consume above.
+            _recognised: set[str] = {
                 "top_p",
                 "frequency_penalty",
                 "presence_penalty",
                 "seed",
-            ):
+                "verbosity",
+                "json_mode",
+            }
+            for key in ("top_p", "frequency_penalty", "presence_penalty", "seed"):
                 if key in extra_params:
                     params[key] = extra_params[key]
+            # Pass anything else through as OpenAI SDK extra_body so
+            # provider-specific fields (e.g. DeepSeek's ``thinking``,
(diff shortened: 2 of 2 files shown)


ZVin-Chen:
Sorry, opened against the wrong base. Re-targeting to my fork's main.

PR #671, opened 2026-05-11
fix(openai): ensure "json" keyword in messages for strict providers (fixes #663)

### LINKED ISSUE #663: [Bug] `json_mode=True` fails on Alibaba/Qwen providers — "messages must contain the word 'json'"

shellybotmoyer:
## Summary

Alibaba/Qwen strictly enforce the OpenAI spec requirement that the word "json" must appear somewhere in the messages whenever `response_format` is set (whether `json_object` or `json_schema`). Most providers are lenient, so this has been a hidden compatibility issue that only surfaces when routing `json_mode=True` calls through Qwen models.

## Proposed Fix

- Add `_ensure_json_keyword()` static method to `OpenAIBackend`
- When `json_mode=True` in `complete()` and `stream()`, call `_ensure_json_keyword(messages)` before the API call
- If any message already contains "json" (case-insensitive), return messages unchanged (fast path)
- Otherwise, append a brief instruction to the last system message, or add a new system message if none exists

## Changes

- `src/llm/backends/openai.py`: added `_ensure_json_keyword()` and wired it into both `complete()` and `stream()` when `json_mode=True`

## Verification

- Fast path preserves existing messages when "json" is already present
- System-message injection avoids mutating the messages list in-place
- No effect on non-JSON calls (gated on `extra_params.get("json_mode")`)

Closes #663

## Release Notes

* **New Features**
  * Configure alternate API endpoints for Anthropic, OpenAI, and Gemini providers without changing model settings.
  * Improved document processing to gracefully handle embedding generation failures while preserving data.

* **Bug Fixes**
  * Fixed OpenAI embedding dimension alignment with configured vector settings.
  * Enhanced rate-limit detection for embedding requests.
  * Corrected JSON mode formatting for OpenAI responses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/671)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -648,6 +648,11 @@ class LLMSettings(HonchoSettings):
     OPENAI_API_KEY: str | None = None
     GEMINI_API_KEY: str | None = None
 
+    # Base URLs for LLM providers (optional — override the SDK default endpoint)
+    ANTHROPIC_BASE_URL: str | None = None
+    OPENAI_BASE_URL: str | None = None
+    GEMINI_BASE_URL: str | None = None
+
     # General LLM settings
     DEFAULT_MAX_TOKENS: Annotated[int, Field(default=1000, gt=0, le=100_000)] = 2500
 
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -453,7 +453,8 @@ async def create_documents(
             # for each document, if deduplicate is True, perform a process
             # that checks against existing documents and either rejects this document
             # as a duplicate OR deletes an existing document that is a duplicate.
-            if deduplicate:
+            # Skip dedup when embedding is None (backfill by reconciler later)
+            if deduplicate and doc.embedding is not None:
                 is_duplicate = await is_rejected_duplicate(
                     db, doc, workspace_name, observer=observer, observed=observed
                 )
--- a/src/crud/representation.py
+++ b/src/crud/representation.py
@@ -102,6 +102,15 @@ class RepresentationManager:
                 "Observation content exceeds maximum token limit of "
                 + f"{settings.EMBEDDING.MAX_INPUT_TOKENS}."
             ) from e
+        except Exception as e:
+            # Embedding API failure (e.g. Gemini daily quota exhausted) —
+            # save observations without embeddings; the reconciler will backfill
+            # after quota resets at midnight Pacific.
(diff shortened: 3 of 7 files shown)


shellybotmoyer:
Superseded by #664 (kevin-ho), which is a cleaner single-commit fix for the same issue. Closing in favor of the focused PR.

shellybotmoyer:
Superseded by #664.

PR #678, opened 2026-05-13
Make embeddings configurable

### LINKED ISSUE #554: Feature Request: Configurable embedding model
## Problem

I'm currently using a local embedding service with a custom base URL via \`LLM_OPENAI_COMPATIBLE_BASE_URL\`. However, the embedding model ID is currently hardcoded in \`src/embedding_client.py\`:

- \`openai\`: \`text-embedding-3-small\`
- \`openrouter\`: \`openai/text-embedding-3-small\`  
- \`gemini\`: \`gemini-embedding-001\`

This means even when I point to my own embedding endpoint, I cannot specify which model to use. I had to manually modify the source code to change the model (e.g., to \`openai/nomic-embed-text\`), which isn't ideal for maintenance.

## Suggested Solution

Would it be possible to add a new configuration parameter, perhaps something like:

\`\`\`bash
LLM_EMBEDDING_MODEL=text-embedding-3-large
\`\`\`

This would allow users to override the default model without modifying the codebase.

## Benefits

- Greater flexibility for self-hosted deployments
- Easier to switch between different embedding models
- No code changes required when upgrading/changing models

Thanks for considering this! Happy to provide more context if needed.

### LINKED ISSUE #564: Bug: VECTOR_STORE_DIMENSIONS not respected in models + Feature: separate custom embedding provider
## Bug: `VECTOR_STORE_DIMENSIONS` setting has no effect

The config option `VECTOR_STORE_DIMENSIONS` (default: 1536) exists in `src/config.py` but is never used in `src/models.py`. Both `MessageEmbedding.embedding` and `Document.embedding` are hardcoded to `Vector(1536)`, which means changing the setting has no effect on the actual database column or SQLAlchemy model.

**Affected files:**
- `src/models.py` line 281: `mapped_column(Vector(1536), nullable=True)`
- `src/models.py` line 389: `mapped_column(Vector(1536), nullable=True)`

**Suggested fix:**

Add `from src.config import settings` to `src/models.py` and replace the hardcoded values:

```python
# Before

# After
embedding: MappedColumn[Any] = mapped_column(Vector(settings.VECTOR_STORE.DIMENSIONS), nullable=True)
```

---

## Feature Request: Separate custom embedding provider endpoint

Currently, `LLM_EMBEDDING_PROVIDER=openrouter` reuses `LLM_OPENAI_COMPATIBLE_BASE_URL` for both LLM inference and embeddings. This makes it impossible to use a self-hosted embedding service independently from the LLM inference endpoint.

**Use case:** Running a local embedding model (e.g. llama.cpp, Ollama) on a separate server while using a different endpoint for LLM inference.

**Suggested addition to `LLMSettings` in `src/config.py`:**

```python
EMBEDDING_PROVIDER: Literal["openai", "gemini", "openrouter", "custom"] = "openai"
EMBEDDING_COMPATIBLE_BASE_URL: str | None = None
EMBEDDING_COMPATIBLE_API_KEY: str | None = None
EMBEDDING_MODEL: str | None = None
```

**And a new branch in `src/embedding_client.py`:**

```python
elif self.provider == "custom":
    if api_key is None:
        api_key = settings.LLM.EMBEDDING_COMPATIBLE_API_KEY
    if not api_key:
        raise ValueError("LLM_EMBEDDING_COMPATIBLE_API_KEY is required")
    base_url = settings.LLM.EMBEDDING_COMPATIBLE_BASE_URL
    if not base_url:
        raise ValueError("LLM_EMBEDDING_COMPATIBLE_BASE_URL is required")
    self.client = AsyncOpenAI(api_key=api_key, base_url=base_url)
    self.model = settings.LLM.EMBEDDING_MODEL or "text-embedding-3-small"
    self.max_embedding_tokens = settings.MAX_EMBEDDING_TOKENS
    self.max_batch_size = 2048
```

**Example `.env` usage:**

```env
LLM_EMBEDDING_PROVIDER=custom
LLM_EMBEDDING_COMPATIBLE_BASE_URL=http://my-embedding-server:8080/v1
LLM_EMBEDDING_COMPATIBLE_API_KEY=my-key
LLM_EMBEDDING_MODEL=my-embed-model
VECTOR_STORE_DIMENSIONS=1024
```

This allows full decoupling of embedding and inference providers, which is especially useful for self-hosted deployments.

### LINKED ISSUE #578: Allow configurable embedding model name + custom base URL for self-hosted embeddings
## Summary

Honcho's embedding client (`src/embedding_client.py`) hardcodes both the **model name** and, for two of three providers, the **base URL**. That makes it impossible to use a self-hosted embedding endpoint (e.g. Ollama with `bge-m3`, llama.cpp, TEI, Infinity) without forking the repo or deploying a translation proxy.

## Current state

Three providers supported:

| Provider | Model (hardcoded) | Base URL |
|---|---|---|
| `openai` | `text-embedding-3-small` | `api.openai.com` (no override) |
| `gemini` | `gemini-embedding-001` | Google (no override) |
| `openrouter` | `openai/text-embedding-3-small` | honors `LLM_OPENAI_COMPATIBLE_BASE_URL` |

The `openrouter` path is the closest to self-hosted-friendly because it accepts a custom base URL, but it still sends a fixed model string (`openai/text-embedding-3-small`) that a local embedder typically doesn't recognize.

## Why this matters

1. **Sovereignty / cost** — operators running Honcho alongside their own LLM stack (e.g. Ollama on a Jetson) want to keep embedding workloads local for both cost and data-residency reasons.
2. **Model flexibility** — some domains need specific embedders (multilingual, code, long-context). `text-embedding-3-small` and `gemini-embedding-001` are reasonable defaults but not universally optimal.
3. **Consistency with chat provider config** — Honcho already lets users set `DERIVER_PROVIDER`, `DERIVER_MODEL`, `DIALECTIC_PROVIDER`, `DIALECTIC_MODEL`, `DREAM_PROVIDER`, `DREAM_MODEL`, including a `custom` provider with `LLM_OPENAI_COMPATIBLE_BASE_URL` for chat. Embeddings are the odd one out.

## Proposed

Add these env vars (matching the chat-side naming convention):

- `LLM_EMBEDDING_MODEL` — free-form model string; default to current hardcoded value per provider
- `LLM_EMBEDDING_BASE_URL` — custom base URL override; applies regardless of provider when set
- Accept `custom` as a new `EMBEDDING_PROVIDER` value (mirrors the chat-side `custom` provider) that uses `LLM_EMBEDDING_BASE_URL` + `LLM_EMBEDDING_MODEL` + reuses `LLM_OPENAI_COMPATIBLE_API_KEY` for auth.

Minimal code change in `_EmbeddingClient.__init__` — thread `api_key`, `base_url`, `model` through the existing provider branches rather than hardcoding them.

## Workaround today

Operators are left writing a proxy service that accepts the OpenAI embeddings shape, rewrites the hardcoded model name, and forwards to their embedder. Happy to contribute a PR if the design above is acceptable.

## Environment

Honcho **3.0.6** (image `ghcr.io/plastic-labs/honcho:latest`) — self-hosted in K3s behind an existing `LLM_OPENAI_COMPATIBLE_BASE_URL=https://api.x.ai/v1` for chat providers.

### LINKED ISSUE #585: [Bug] How to use a local embedding model ?
Hello, 

I try to use honcho in local homelab, with hermes and openclaw, everything is local, 
no paid models services or subscriptions,
minimum use of internet, all data is local.

The goal is to survive zombi apocalypse, 
and have minimal use of internet (only if no other choice).
Data and services must stay local.
For example i use self hosted synapse/element/matrix chat instead of Telegram, 
local Searxng (search engine) that search into local Zimi and Kiwix (local selfhosted Wikipedia and other ZIM files). Instead of Tailscale i use self hosted Headscale, local Rusdesk for RDP, etc ...

All my models are running in LM Studio on a gaming PC + RTX 4090 (+ other LM Studio instances distributed in lan).
My homelab is an old gaming PC with RTX 3070 and proxmox.
Proxmox runs Truenas VM, Home assistant VM, and linux Mint VM with GPU passthrough.
The Mint VM has docker with 100 containers, including AI containers, some of them use GPU (from Mint).
Hermes and Openclaw are installed directly inside Mint VM (i tryed docker versions and had many problems, so installed them inside linux Mint via bash one liner (no docker), and they work better).

Now i try to configure Honcho with both Openclaw and Hermes (the idea is they will share same memory through Honcho, and run in parallel using different LM Studio instances).
Honcho will be docker container, will use existing redis/postgres(pgvector18) docker containers, 
and also LM Studio witch serve models from lan gaming PC (http://${PROART_IP}:1234/v1).

Everything is configured and running but there is two problems : 
1/
Looks like there is no way to configure an LM Studio embedding model compatible with Honcho, in relation with this : 
( Embedding dimensions (default: 1536 for OpenAI text-embedding-3-small)
VECTOR_STORE_DIMENSIONS=768 )
Honcho always sends "text-embedding-3-small" to LM Studio and forces 1536 dimensions instead of 768 ?
And i did not find any LM Studio embedding model with 1536 dimensions compatible with Honcho.
So Hermes and Openclaw crash due to embedding problem with Honcho.
2/
LM Studio lazy load models when recieve request, but this does not work with Honcho embedding, i need to load an embedding model manually (perhaps related to 1/).
This is not optimal, because you need to do manual action in LM Studio, and i have multiple instances.

Second env file (i have main env file with variables for all containers) :  
```
# models : Lm Studio : honcho
MD_LMS_DIALECTIC=hermes-3-llama-3.1-8b      
MD_LMS_DERIVER=llama-3.2-3b-instruct       
MD_LMS_EMBEDDING=gte-qwen2-1.5b-instruct   
MD_LMS_SUMMARY=llama-3.2-3b-instruct        
...

```

Honcho compose : 
```
# https://github.com/plastic-labs/honcho/blob/main/docker-compose.yml.example
# https://github.com/elkimek/honcho-self-hosted/tree/main

services:

  srv_honcho_api:
    extends:
      file: ../_common/all.yml
      service: all

    image: ${C_HONA}:latest
    build:
      context: ../../git/honcho
      dockerfile: Dockerfile

    env_file:
      - ./.env

    entrypoint: ["sh", "docker/entrypoint.sh"]

    container_name: ${C_HONA}
    ports:
      - ${C_HONA_8000}:8000 # 8800
    environment:
      # postgres + redis
      - DB_CONNECTION_URI=postgresql+psycopg://${C_HON_PG_USER}:${C_HON_PG_PWD}@${C_PG18v_HOST}:5432/${C_HON_PG_DB}
      - CACHE_URL=${C_RED_URL}?suppress=true
      # --- LM Studio ---
      - LLM_OPENAI_COMPATIBLE_BASE_URL=http://${PROART_IP}:1234/v1
      # --- Dialectic Levels (custom provider + model) ---
      - DIALECTIC_LEVELS__minimal__MODEL=${MD_LMS_DERIVER} 
      - DIALECTIC_LEVELS__low__MODEL=${MD_LMS_DERIVER}
      - DIALECTIC_LEVELS__medium__MODEL=${MD_LMS_DIALECTIC}
      - DIALECTIC_LEVELS__high__MODEL=${MD_LMS_DIALECTIC}
      - DIALECTIC_LEVELS__max__MODEL=${MD_LMS_DIALECTIC}
      # --- Embeddings ---
      - LLM_EMBEDDING_BASE_URL=http://${PROART_IP}:1234/v1
      - LLM_EMBEDDING_MODEL=${MD_LMS_EMBEDDING} 
      # --- Summary ---
      - SUMMARY_MODEL=${MD_LMS_SUMMARY}
      # --- Deriver ---
      - DERIVER_MODEL=${MD_LMS_DERIVER}

    hostname: ${C_HONA_HOST}  # hona
    networks:
      proxy:
        ipv4_address: ${NET_TFK}.${C_HONA_NUM} # 168
        mac_address: ${MAC_TFK}:${C_HONA_MAC}  # a8
    #  proxmox:
    #    ipv4_address: ${NET_PX}.${C_HONA_NUM}
    #    mac_address: ${MAC_PX}:${C_HONA_MAC} 
    #dns:
    #  - ${NET_PX_DNS}
      
    depends_on:
          - srv_honcho_deriver


  srv_honcho_deriver:
    extends:
      file: ../_common/all.yml
      service: all

    image: ${C_HONA}:latest
    build:
      context: ../../git/honcho
      dockerfile: Dockerfile

    env_file:
      - ./.env

    #command: python -m honcho.deriver
    entrypoint: ["/app/.venv/bin/python", "-m", "src.deriver"]

    container_name: ${C_HOND}
    environment:
      # postgres + redis
      - DB_CONNECTION_URI=postgresql+psycopg://${C_HON_PG_USER}:${C_HON_PG_PWD}@${C_PG18v_HOST}:5432/${C_HON_PG_DB}
      - CACHE_URL=${C_RED_URL}?suppress=true
      # --- LM Studio ---
      - LLM_OPENAI_COMPATIBLE_BASE_URL=http://${PROART_IP}:1234/v1
      # --- Dialectic Levels (custom provider + model) ---
      - DIALECTIC_LEVELS__minimal__MODEL=${MD_LMS_DERIVER} 
      - DIALECTIC_LEVELS__low__MODEL=${MD_LMS_DERIVER}
      - DIALECTIC_LEVELS__medium__MODEL=${MD_LMS_DIALECTIC}
      - DIALECTIC_LEVELS__high__MODEL=${MD_LMS_DIALECTIC}
      - DIALECTIC_LEVELS__max__MODEL=${MD_LMS_DIALECTIC}
      # --- Embeddings ---
      - LLM_EMBEDDING_BASE_URL=http://${PROART_IP}:1234/v1
      - LLM_EMBEDDING_MODEL=${MD_LMS_EMBEDDING}
      # --- Summary ---
      - SUMMARY_MODEL=${MD_LMS_SUMMARY}
      # --- Deriver ---
      - DERIVER_MODEL=${MD_LMS_DERIVER}

    hostname: ${C_HOND_HOST} # hond
    networks:
      proxy:
        ipv4_address: ${NET_TFK}.${C_HOND_NUM} # 169
        mac_address: ${MAC_TFK}:${C_HOND_MAC}  # a9
    #  proxmox:
    #    ipv4_address: ${NET_PX}.${C_HOND_NUM}
    #    mac_address: ${MAC_PX}:${C_HOND_MAC} 
    #dns:
    #  - ${NET_PX_DNS}

networks: 
  proxy: 
    external: true
  proxmox:
    external: true

# docker exec -it ai-honcho-api python -m honcho.manage upgrade
# curl http://192.168.50.54:1234/v1/models
# docker compose run --rm ai-honcho-api alembic upgrade head
# cd /Volumes/mb-docker/git
# git clone https://github.com/plastic-labs/honcho.git

```

Honcho .env ( env_file:  - ./.env) :

```

# Honcho Environment Variables Template
# Copy this file to .env and fill in the appropriate values
#
# Required variables are marked with (REQUIRED)
# Optional variables have default values and can be left commented out

# =============================================================================
# Application Settings
# =============================================================================
LOG_LEVEL=INFO
# SESSION_OBSERVERS_LIMIT=10
# GET_CONTEXT_MAX_TOKENS=100000
# MAX_FILE_SIZE=5242880 # Bytes
# MAX_MESSAGE_SIZE=25000 # Characters

# Embedding settings
# EMBED_MESSAGES=true
# MAX_EMBEDDING_TOKENS=8192
# MAX_EMBEDDING_TOKENS_PER_REQUEST=300000

# LANGFUSE_HOST=
# LANGFUSE_PUBLIC_KEY=

COLLECT_METRICS_LOCAL=false
# LOCAL_METRICS_FILE=metrics.jsonl
# REASONING_TRACES_FILE=traces.jsonl  # Path to JSONL file for reasoning traces

# NAMESPACE="honcho"

# =============================================================================
# Database Settings (REQUIRED)
# =============================================================================
# Connection URI for PostgreSQL database with pgvector support
# Must use postgresql+psycopg prefix for SQLAlchemy compatibility
DB_CONNECTION_URI=postgresql+psycopg://postgres:postgres@localhost:5432/postgres

# Optional database settings
# DB_SCHEMA=public
# DB_POOL_CLASS=default
# DB_POOL_SIZE=10
# DB_MAX_OVERFLOW=20
# DB_POOL_TIMEOUT=30
# DB_POOL_RECYCLE=300
# DB_POOL_PRE_PING=true
# DB_POOL_USE_LIFO=true
# DB_SQL_DEBUG=false
# DB_TRACING=false

# =============================================================================
# Authentication Settings
# =============================================================================
# Whether to enable authentication (set to true for production)
AUTH_USE_AUTH=false

# JWT secret key (REQUIRED if AUTH_USE_AUTH=true)
# Generate with: python scripts/generate_jwt_secret.py
# AUTH_JWT_SECRET=your-secret-key-here

# =============================================================================
# LLM Provider (REQUIRED)
# =============================================================================
# Honcho uses LLMs for memory extraction, summarization, dialectic chat, and
# dream consolidation. The server will fail to start without a provider configured.
#
# Quick start: uncomment the two lines below, set your endpoint and API key,
# then uncomment the provider/model lines in each feature section below.
# Any OpenAI-compatible endpoint works (OpenRouter, Together, Fireworks, etc.).
# Models must support tool calling (function calling).
#
#LLM_OPENAI_COMPATIBLE_BASE_URL=https://openrouter.ai/api/v1
LLM_OPENAI_COMPATIBLE_API_KEY=none
#
# Provider options for each feature: custom, vllm, google, anthropic, openai, groq
# "custom" routes through the OpenAI-compatible endpoint above.
# Model name format depends on your provider (e.g., OpenRouter: vendor/model-name).
#
# ---- Alternative: vLLM self-hosted ------------------------------------------
# LLM_VLLM_BASE_URL=http://localhost:8000/v1
# LLM_VLLM_API_KEY=not-needed
#
# ---- Alternative: direct vendor keys (no endpoint needed) -------------------
# LLM_GEMINI_API_KEY=
# LLM_ANTHROPIC_API_KEY=
# LLM_OPENAI_API_KEY=
# LLM_GROQ_API_KEY=
#
# ---- General LLM settings ---------------------------------------------------
# Embedding provider — defaults to openai (requires LLM_OPENAI_API_KEY).
# Set to openrouter to route embeddings through your custom endpoint instead.
LLM_EMBEDDING_PROVIDER=openrouter
LLM_EMBEDDING_API_KEY=none
# LLM_DEFAULT_MAX_TOKENS=2500
# LLM_MAX_TOOL_OUTPUT_CHARS=10000
# LLM_MAX_MESSAGE_CONTENT_CHARS=2000

# =============================================================================
# Deriver (Background Worker)
# =============================================================================
DERIVER_ENABLED=true
DERIVER_PROVIDER=custom
#DERIVER_MODEL=your-model-here  # e.g. google/gemini-2.5-flash
# DERIVER_THINKING_BUDGET_TOKENS=1024  # gt=0 required; omit for non-thinking models
# DERIVER_WORKERS=1
# DERIVER_POLLING_SLEEP_INTERVAL_SECONDS=1.0
# DERIVER_STALE_SESSION_TIMEOUT_MINUTES=5
# DERIVER_QUEUE_ERROR_RETENTION_SECONDS=2592000
# DERIVER_TEMPERATURE=
# DERIVER_DEDUPLICATE=true
# DERIVER_MAX_OUTPUT_TOKENS=4096
# DERIVER_LOG_OBSERVATIONS=false
# DERIVER_MAX_INPUT_TOKENS=23000
# DERIVER_WORKING_REPRESENTATION_MAX_OBSERVATIONS=100
# DERIVER_REPRESENTATION_BATCH_MAX_TOKENS=1024
# DERIVER_FLUSH_ENABLED=false

# =============================================================================
# Peer Card
# =============================================================================
# PEER_CARD_ENABLED=true

# =============================================================================
# Dialectic
# =============================================================================
# DIALECTIC_MAX_OUTPUT_TOKENS=8192
# DIALECTIC_MAX_INPUT_TOKENS=100000
# DIALECTIC_HISTORY_TOKEN_LIMIT=8192
# DIALECTIC_SESSION_HISTORY_MAX_TOKENS=4096
#
# Per-level provider, model, and tuning:
DIALECTIC_LEVELS__minimal__PROVIDER=custom
#DIALECTIC_LEVELS__minimal__MODEL=your-model-here  # e.g. google/gemini-2.5-flash
DIALECTIC_LEVELS__minimal__THINKING_BUDGET_TOKENS=1
DIALECTIC_LEVELS__minimal__MAX_TOOL_ITERATIONS=1
DIALECTIC_LEVELS__minimal__MAX_OUTPUT_TOKENS=250
DIALECTIC_LEVELS__low__PROVIDER=custom
#DIALECTIC_LEVELS__low__MODEL=your-model-here
DIALECTIC_LEVELS__low__THINKING_BUDGET_TOKENS=1
DIALECTIC_LEVELS__low__MAX_TOOL_ITERATIONS=5
DIALECTIC_LEVELS__medium__PROVIDER=custom
#DIALECTIC_LEVELS__medium__MODEL=your-model-here
DIALECTIC_LEVELS__medium__THINKING_BUDGET_TOKENS=1
DIALECTIC_LEVELS__medium__MAX_TOOL_ITERATIONS=2
DIALECTIC_LEVELS__high__PROVIDER=custom
#DIALECTIC_LEVELS__high__MODEL=your-model-here
DIALECTIC_LEVELS__high__THINKING_BUDGET_TOKENS=1
DIALECTIC_LEVELS__high__MAX_TOOL_ITERATIONS=4
DIALECTIC_LEVELS__max__PROVIDER=custom
#DIALECTIC_LEVELS__max__MODEL=your-model-here
DIALECTIC_LEVELS__max__THINKING_BUDGET_TOKENS=1
DIALECTIC_LEVELS__max__MAX_TOOL_ITERATIONS=10

# =============================================================================
# Summary
# =============================================================================
# SUMMARY_ENABLED=true
SUMMARY_PROVIDER=custom
#SUMMARY_MODEL=your-model-here  # e.g. google/gemini-2.5-flash
SUMMARY_THINKING_BUDGET_TOKENS=512  # gt=0 required; omit for non-thinking models
SUMMARY_MESSAGES_PER_SHORT_SUMMARY=20
SUMMARY_MESSAGES_PER_LONG_SUMMARY=60
SUMMARY_MAX_TOKENS_SHORT=1000
SUMMARY_MAX_TOKENS_LONG=4000

# =============================================================================
# Dream
# =============================================================================
# DREAM_ENABLED=true
DREAM_PROVIDER=custom
#DREAM_MODEL=your-model-here  # e.g. google/gemini-2.5-flash
#DREAM_DEDUCTION_MODEL=your-model-here
#DREAM_INDUCTION_MODEL=your-model-here
DREAM_THINKING_BUDGET_TOKENS=8192  # gt=0 required; omit for non-thinking models
# DREAM_DOCUMENT_THRESHOLD=50
# DREAM_IDLE_TIMEOUT_MINUTES=60
# DREAM_MIN_HOURS_BETWEEN_DREAMS=8
# DREAM_ENABLED_TYPES=["omni"]
# DREAM_MAX_OUTPUT_TOKENS=16384
# DREAM_MAX_TOOL_ITERATIONS=20
# DREAM_HISTORY_TOKEN_LIMIT=16384
#
# Surprisal sampling (advanced):
# DREAM_SURPRISAL__ENABLED=false
# DREAM_SURPRISAL__TREE_TYPE=kdtree
# DREAM_SURPRISAL__TREE_K=5
# DREAM_SURPRISAL__SAMPLING_STRATEGY=recent
# DREAM_SURPRISAL__SAMPLE_SIZE=200
# DREAM_SURPRISAL__TOP_PERCENT_SURPRISAL=0.10
# DREAM_SURPRISAL__MIN_HIGH_SURPRISAL_FOR_REPLACE=10
# DREAM_SURPRISAL__INCLUDE_LEVELS=["explicit","deductive"]

# =============================================================================
# Webhook Settings
# =============================================================================
# WEBHOOK_SECRET=
# WEBHOOK_MAX_WORKSPACE_LIMIT=10

# =============================================================================
# Monitoring and Observability (Optional)
# =============================================================================
# Sentry error tracking
# SENTRY_ENABLED=false
# SENTRY_DSN=your-sentry-dsn-here
# SENTRY_RELEASE=your-release-semver
# SENTRY_ENVIRONMENT=development
# SENTRY_TRACES_SAMPLE_RATE=0.1
# SENTRY_PROFILES_SAMPLE_RATE=0.1

# =============================================================================
# Prometheus Metrics Settings (Pull-based metrics)
# =============================================================================
METRICS_ENABLED=true
# METRICS_NAMESPACE=honcho  # Inherits from NAMESPACE if not set

# =============================================================================
# CloudEvents Telemetry Settings (Analytics events)
# =============================================================================
# TELEMETRY_ENABLED=false
# TELEMETRY_ENDPOINT=https://telemetry.honcho.dev/v1/events
# TELEMETRY_HEADERS={"Authorization": "Bearer your-token"}  # JSON string for auth headers
# TELEMETRY_BATCH_SIZE=100
# TELEMETRY_FLUSH_INTERVAL_SECONDS=1.0
# TELEMETRY_FLUSH_THRESHOLD=50
# TELEMETRY_MAX_RETRIES=3
# TELEMETRY_MAX_BUFFER_SIZE=10000
# TELEMETRY_NAMESPACE=honcho  # Inherits from NAMESPACE if not set

# =============================================================================
# Cache
# =============================================================================
CACHE_ENABLED=true
# CACHE_URL="redis://localhost:6379/0?suppress=true"
# CACHE_NAMESPACE="honcho"  # Inherits from NAMESPACE if not set
# CACHE_DEFAULT_TTL_SECONDS=300
# CACHE_DEFAULT_LOCK_TTL_SECONDS=5

# =============================================================================
# Vector Store Settings
# =============================================================================
VECTOR_STORE_TYPE=pgvector  # pgvector, turbopuffer, lancedb
VECTOR_STORE_MIGRATED=true # true= migrated

# Global namespace prefix for all vector namespaces
# Namespaces follow the pattern: {NAMESPACE}.{type}.{hash}
# where hash is a base64url-encoded SHA-256 of the workspace/peer names
# - Documents: {NAMESPACE}.doc.{hash(workspace, observer, observed)}
# - Messages: {NAMESPACE}.msg.{hash(workspace)}
# VECTOR_STORE_NAMESPACE=honcho  # Inherits from NAMESPACE if not set

# Embedding dimensions (default: 1536 for OpenAI text-embedding-3-small)
VECTOR_STORE_DIMENSIONS=768

# Turbopuffer-specific settings (required if TYPE is "turbopuffer")
# VECTOR_STORE_TURBOPUFFER_API_KEY=your-turbopuffer-api-key
# VECTOR_STORE_TURBOPUFFER_REGION=gcp-us-east4

# LanceDB-specific settings (local embedded mode)
# VECTOR_STORE_LANCEDB_PATH=./lancedb_data

# Reconciliation interval for background sync (default: 5 minutes)
# VECTOR_STORE_RECONCILIATION_INTERVAL_SECONDS=300
```

LM Studio log shows requests/responses with Honcho : related to openai/text-embedding-3-small

```
2026-04-20 10:31:36  [INFO]
 Received request to embed multiple:  [
  "embedding test",
  "Sure thing! What would you like to embed?"
]
2026-04-20 10:31:36  [INFO]
 Returning embeddings (not shown in logs)
2026-04-20 10:31:37 [DEBUG]
 Received request: POST to /v1/embeddings with body  {
  "input": [
    "embedding test",
    "Sure thing! What would you like to embed?"
  ],
  "model": "openai/text-embedding-3-small",
  "encoding_format": "base64"
}
2026-04-20 10:31:37  [INFO]
 Received request to embed multiple:  [
  "embedding test",
  "Sure thing! What would you like to embed?"
]
2026-04-20 10:31:37  [INFO]
 Returning embeddings (not shown in logs)
2026-04-20 10:31:38 [DEBUG]
 Received request: POST to /v1/embeddings with body  {
  "input": [
    "embedding test",
    "Sure thing! What would you like to embed?"
  ],
  "model": "openai/text-embedding-3-small",
  "encoding_format": "base64"
}
2026-04-20 10:31:38  [INFO]
 Received request to embed multiple:  [
  "embedding test",
  "Sure thing! What would you like to embed?"
]
2026-04-20 10:31:38  [INFO]
 Returning embeddings (not shown in logs)
```

Honcho API log shows errors related to : expected 1536 dimensions, not 768

```
2026-04-20 10:16:58,451 - __main__ - INFO - Starting deriver queue processor
2026-04-20 10:16:58,452 - __main__ - INFO - Prometheus metrics server started on port 9090
2026-04-20 10:16:58,452 - __main__ - INFO - Running main loop
2026-04-20 10:16:58,456 - src.cache.client - INFO - Connected to cache at redis://redis:6379/0?suppress=true
2026-04-20 10:16:58,456 - src.reconciler.scheduler - INFO - ReconcilerScheduler started with 2 tasks: ['sync_vectors', 'cleanup_queue']
2026-04-20 10:21:59,466 - src.reconciler.scheduler - INFO - Enqueued reconciler task: sync_vectors
2026-04-20 10:22:01,450 - src.reconciler.sync_vectors - INFO - Vector reconciliation cycle completed (pgvector mode)
2026-04-20 10:27:00,460 - src.reconciler.scheduler - INFO - Enqueued reconciler task: sync_vectors
2026-04-20 10:27:02,387 - src.reconciler.sync_vectors - INFO - Vector reconciliation cycle completed (pgvector mode)
  File "/app/.venv/lib/python3.13/site-packages/starlette/middleware/base.py", line 168, in call_next
    raise app_exc from app_exc.__cause__ or app_exc.__context__
  File "/app/.venv/lib/python3.13/site-packages/starlette/middleware/base.py", line 144, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/app/.venv/lib/python3.13/site-packages/starlette/middleware/cors.py", line 87, in __call__
    await self.app(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/middleware/exceptions.py", line 63, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    raise exc
  File "/app/.venv/lib/python3.13/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/app/.venv/lib/python3.13/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/routing.py", line 716, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/routing.py", line 736, in app
    await route.handle(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/routing.py", line 290, in handle
    await self.app(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 119, in app
    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
  File "/app/.venv/lib/python3.13/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    raise exc
  File "/app/.venv/lib/python3.13/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/app/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 105, in app
    response = await f(request)
               ^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 431, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
    )
    ^
  File "/app/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 313, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/routers/messages.py", line 115, in create_messages_for_session
    "message_id": message.id,
                  ^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/attributes.py", line 569, in __get__
    return self.impl.get(state, dict_)  # type: ignore[no-any-return]
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/attributes.py", line 1096, in get
    value = self._fire_loader_callables(state, key, passive)
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/attributes.py", line 1126, in _fire_loader_callables
    return state._load_expired(state, passive)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/state.py", line 828, in _load_expired
    self.manager.expired_attribute_loader(self, toload, passive)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/loading.py", line 1674, in load_scalar_attributes
    result = load_on_ident(
        session,
    ...<4 lines>...
        no_autoflush=no_autoflush,
    )
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/loading.py", line 510, in load_on_ident
    return load_on_pk_identity(
        session,
    ...<11 lines>...
        is_user_refresh=is_user_refresh,
    )
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/loading.py", line 695, in load_on_pk_identity
    session.execute(
    ~~~~~~~~~~~~~~~^
        q,
        ^^
    ...<2 lines>...
        bind_arguments=bind_arguments,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 2351, in execute
    return self._execute_internal(
           ~~~~~~~~~~~~~~~~~~~~~~^
        statement,
        ^^^^^^^^^^
    ...<4 lines>...
        _add_event=_add_event,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 2239, in _execute_internal
    conn = self._connection_for_bind(bind)
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 2108, in _connection_for_bind
    return trans._connection_for_bind(engine, execution_options)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 2, in _connection_for_bind
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/state_changes.py", line 101, in _go
    self._raise_for_prerequisite_state(fn.__name__, current_state)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.13/site-packages/sqlalchemy/orm/session.py", line 971, in _raise_for_prerequisite_state
    raise sa_exc.PendingRollbackError(
    ...<6 lines>...
    )
sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (builtins.ValueError) expected 1536 dimensions, not 768
[SQL: INSERT INTO public.message_embeddings (content, embedding, message_id, workspace_name, session_name, peer_name, sync_state, last_sync_at, sync_attempts) SELECT p0::VARCHAR, p1::VECTOR(1536), p2::VARCHAR, p3::VARCHAR, p4::VARCHAR, p5::VARCHAR, p6::VARCHAR, p7::TIMESTAMP WITH TIME ZONE, p8::INTEGER FROM (VALUES (%(content)s::VARCHAR, %(embedding)s, %(message_id)s::VARCHAR, %(workspace_name)s::VARCHAR, %(session_name)s::VARCHAR, %(peer_name)s::VARCHAR, %(sync_state)s::VARCHAR, %(last_sync_at)s::TIMESTAMP WITH TIME ZONE, %(sync_attempts)s::INTEGER)) AS imp_sen(p0, p1, p2, p3, p4, p5, p6, p7, p8, sen_counter) ORDER BY sen_counter RETURNING public.message_embeddings.id, public.message_embeddings.created_at, public.message_embeddings.id AS id__1]
[parameters: [{'message_id': 'PA8aY5ZeA-QEXwUvZxm0t', 'sync_state': 'pending', 'peer_name': 'owner', 'content': 'embedding test', 'workspace_name': 'main', 'session ... (16907 characters truncated) ... 544067144394, 0.007806772831827402, -0.0038657912518829107, -0.05361740663647652, -0.009963896125555038, 0.0022806485649198294], 'last_sync_at': None}, {'message_id': '0DOenfRYdieb0L2aPl-fo', 'sync_state': 'pending', 'peer_name': 'agent-main', 'content': 'Sure thing! What would you like to embed?', 'w ... (16963 characters truncated) ... 470942467451096, 0.04498995468020439, -0.03708455711603165, 0.0024789117742329836, 0.004620757419615984, -0.051529645919799805], 'last_sync_at': None}]] (Background on this error at: https://sqlalche.me/e/20/7s2a)
2026-04-20 10:32:01,461 - src.reconciler.scheduler - INFO - Enqueued reconciler task: sync_vectors
2026-04-20 10:32:03,464 - src.reconciler.sync_vectors - INFO - Vector reconciliation cycle completed (pgvector mode)
```

Image below : 
openclaw (self hosted synapse/element/matrix) chat, 
i say "embedding test" , then openclaw use LM Studio with Honcho,
honcho sends text-embedding-3-small request to LM Studio, 
LM Studio responds with an other embedding model loaded manually (text-embedding-nomic-embed-text-v1.5@q8_0),
honsho crashes : "... Original exception was: (builtins.ValueError) expected 1536 dimensions, not 768 ..."
(honcho does not respect the config : VECTOR_STORE_DIMENSIONS=768 + custom embedding model)

```
sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (builtins.ValueError) expected 1536 dimensions, not 768
[SQL: INSERT INTO public.message_embeddings (content, embedding, message_id, workspace_name, session_name, peer_name, sync_state, last_sync_at, sync_attempts) SELECT p0::VARCHAR, p1::VECTOR(1536), p2::VARCHAR, p3::VARCHAR, p4::VARCHAR, p5::VARCHAR, p6::VARCHAR, p7::TIMESTAMP WITH TIME ZONE, p8::INTEGER FROM (VALUES (%(content)s::VARCHAR, %(embedding)s, %(message_id)s::VARCHAR, %(workspace_name)s::VARCHAR, %(session_name)s::VARCHAR, %(peer_name)s::VARCHAR, %(sync_state)s::VARCHAR, %(last_sync_at)s::TIMESTAMP WITH TIME ZONE, %(sync_attempts)s::INTEGER)) AS imp_sen(p0, p1, p2, p3, p4, p5, p6, p7, p8, sen_counter) ORDER BY sen_counter RETURNING public.message_embeddings.id, public.message_embeddings.created_at, public.message_embeddings.id AS id__1]
[parameters: [{'message_id': 'PA8aY5ZeA-QEXwUvZxm0t', 'sync_state': 'pending', 'peer_name': 'owner', 'content': 'embedding test', 'workspace_name': 'main', 'session ... (16907 characters truncated) ... 544067144394, 0.007806772831827402, -0.0038657912518829107, -0.05361740663647652, -0.009963896125555038, 0.0022806485649198294], 'last_sync_at': None}, {'message_id': '0DOenfRYdieb0L2aPl-fo', 'sync_state': 'pending', 'peer_name': 'agent-main', 'content': 'Sure thing! What would you like to embed?', 'w ... (16963 characters truncated) ... 470942467451096, 0.04498995468020439, -0.03708455711603165, 0.0024789117742329836, 0.004620757419615984, -0.051529645919799805], 'last_sync_at': None}]] (Background on this error at: https://sqlalche.me/e/20/7s2a)
```

Image

Also it seems there is multiple tickets related to the same embedding hardcoded problem open, and they are not adressed ? 
Not sure if my problem is exactly the same, so i did separate ticket.

(also i see many repos managed by AI, problems not solved, developpers unable to help because wibe coding degrading their skills, many dark code, and in some cases AI directly responding to requests and closing tickets without correcting bugs, hope this will not happen to this very promising repo/project)

Best regards :)

### LINKED ISSUE #590: Allow non-1536 VECTOR_DIMENSIONS with pgvector via opt-in flag
## Context

Self-hosted Honcho deployment using a local embedding model that outputs non-1536 dimensions (e.g., `nomic-embed-text-v1.5` @ 768, `embeddinggemma` @ 768, matryoshka variants, etc.).

Upstream currently enforces a hard validation in `src/config.py` (around line 1282):

```python
if self.EMBEDDING.VECTOR_DIMENSIONS != 1536 and (
    self.VECTOR_STORE.TYPE == "pgvector" or not self.VECTOR_STORE.MIGRATED
):
    raise ValueError(
        "EMBEDDING.VECTOR_DIMENSIONS must remain 1536 while pgvector is "
        + "active or vector-store migration is incomplete"
    )
```

This effectively locks pgvector deployments to 1536 dims (OpenAI `text-embedding-3-small` shape).

## Pain point

- pgvector **does** support arbitrary fixed vector dimensions — the constraint is application-level, not storage-level.
- Self-hosters running local embedders can't choose a non-1536 model without either (a) running a 1536-dim projection layer server-side (e.g. `Orange/orange-nomic-v1.5-1536`), or (b) carrying a local patch that removes the check and manually running `ALTER TABLE ... ALTER COLUMN embedding TYPE vector(<dim>)` on both `documents` and `message_embeddings`.
- Option (b) works perfectly in practice — tested on a live install with 40k+ vectors re-embedded and reconciler drain completing cleanly — but forces an out-of-tree patch that conflicts on every rebase.

## Proposed solution

Add an opt-in boolean flag in `VectorStoreSettings`, e.g.:

```python
# src/config.py inside VectorStoreSettings
PGVECTOR_DIM_OVERRIDE: bool = False
```

Then gate the validation:

```python
if (
    self.EMBEDDING.VECTOR_DIMENSIONS != 1536
    and (self.VECTOR_STORE.TYPE == "pgvector" or not self.VECTOR_STORE.MIGRATED)
    and not self.VECTOR_STORE.PGVECTOR_DIM_OVERRIDE
):
    raise ValueError(...)
```

- Default `False` → zero behavior change, fully backwards compatible.
- Operator sets `VECTOR_STORE__PGVECTOR_DIM_OVERRIDE=true` after migrating the embedding columns to the new dim (explicit opt-in, they take responsibility for coherence).
- Error message could be extended to mention the override flag for discoverability.

## Alternative considered

A dynamic check reading the actual column dim via `pg_attribute` at boot. More robust (catches silent drift), but adds a DB query on startup. Acceptable as a follow-up if opt-in flag proves insufficient.

## Setup for reference

- Honcho @ origin/main (post-v3.0.6), self-hosted
- pgvector, `embedding` column migrated via `ALTER COLUMN ... TYPE vector(<new_dim>)` on both `documents` and `message_embeddings`
- Non-1536 embedder served via `text-embeddings-inference` (OpenAI-compatible endpoint)
- Verified end-to-end: reconciler drain clean on ~40k vectors, dialectic / recall behave as expected.

### LINKED ISSUE #601: embedding_client.py: OpenAI path doesn't send dimensions, silently ignoring EMBEDDING_VECTOR_DIMENSIONS

### LINKED ISSUE #625: OpenAI transport embeddings.create() missing dimensions parameter

VVoruganti:
## Summary by CodeRabbit

* **New Features**
  * Configurable embedding vector dimensions with optional forwarding to compatible providers
  * CLI to safely change embedding column dimensions and migrate schemas
  * Runtime startup validator preventing serving on embedding/store dimension mismatches
  * Vector-store namespace probing to detect stored embedding dimensions

* **Documentation**
  * Expanded embedding configuration, migration recipe, troubleshooting, and deprecation guidance

* **Chores**
  * Tightened cloudevents dependency range
  * Deprecated legacy vector-store dimension setting in examples/templates

* **Tests**
  * Added coverage for config, migration CLI, startup validator, probing, and embedding client behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/678)


scripts/configure_embeddings.py
```diff
@@ -0,0 +1,445 @@
...
+    args = parser.parse_args(argv)
+    try:
+        return asyncio.run(_async_main(args))
+    finally:
+        # Best-effort cleanup; the script is short-lived so a leak here is
+        # harmless but keeping the dispose explicit avoids "engine not
+        # disposed" warnings during tests.
+        asyncio.run(engine.dispose())
+
+
```
Rajat-Ahuja1997: nit: can we just call `await engine.dispose() at the end of _async_main?

scripts/configure_embeddings.py
```diff
@@ -0,0 +1,445 @@
...
+        for table in _EMBEDDING_TABLES:
+            counts[table] = await _count_non_null_embeddings(conn, plan.schema, table)
+        populated = {t: n for t, n in counts.items() if n > 0}
+        if populated:
+            detail = ", ".join(f"{t}: {n} rows" for t, n in sorted(populated.items()))
+            raise SystemExit(
+                f"error: refusing to ALTER populated embedding tables ({detail})."
+                + " This script only configures empty tables. Re-embed out-of-band"
+                + " into a fresh deployment, then cut over."
+            )
```
Rajat-Ahuja1997, line 215: error message seems like it got cut off

src/startup/embedding_validator.py
```diff
@@ -0,0 +1,219 @@
...
+    dims = await _introspect_pgvector_dims_with_retry(engine, schema)
+    _assert_pgvector_dims_match(dims, schema=schema, target_dim=target_dim)
+
+    if s.VECTOR_STORE.TYPE in ("turbopuffer", "lancedb"):
+        await _sample_external_namespaces(engine, target_dim=target_dim)
+
+
+async def _introspect_pgvector_dims_with_retry(
+    engine: AsyncEngine, schema: str
+) -> dict[str, int]:
```
Rajat-Ahuja1997, line 85: nit: could use tenacity here

src/startup/embedding_validator.py
```diff
@@ -0,0 +1,219 @@
...
+    store = get_external_vector_store()
+    if store is None:
+        # Settings said TYPE != pgvector but the store could not be created.
+        # That is its own problem and not for this validator to swallow.
+        return
+
+    mismatches: list[tuple[str, int]] = []
+    for workspace_name in workspace_names:
+        namespace = store.get_vector_namespace("message", workspace_name)
+        actual_dim = await _probe_namespace_dim(store, namespace)
```
Rajat-Ahuja1997, line 218: should we "probe" the documents namespaces as well?

Rajat-Ahuja1997 (approved):

PR #686, opened 2026-05-14
fix: reframe peer card prompts as stable identity markers

VVoruganti:
## Summary by CodeRabbit

* **Improvements**
  * Stricter peer-card rules and validation: allowed marker types, per-entry length cap, normalized deduplication, and clearer rejection feedback when entries are invalid.
  * Role-based peer-card behavior: one specialist type can no longer write to the peer card; another uses tightened identity-marker framing when writing.
  * OpenAI backend now accepts a reasoning token budget and forwards it to providers.

* **Bug Fixes**
  * Anthropic backend no longer errors on an unsupported thinking parameter (it is ignored).

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/686?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Rajat-Ahuja1997 (commented):
generally looks good but i'm concerned about valid information in an existing peer card being dropped because it doesn't have one of the required prefixes. should we add a section to the system prompt or deduction user prompt that tells the model to re-emit existing entries with a correct prefix? like `name: Rajat` to `IDENTITY: Name: Rajat`?

Rajat-Ahuja1997 (approved):

PR #692, opened 2026-05-15
feat: LLM model fallback on failure — fast failover to secondary provider/model

phuongvm:
## Summary

Introduces a robust **model fallback mechanism** for Honcho's LLM layer. When a primary LLM provider/model returns a retryable error (HTTP 429/5xx, timeout, connection error), the system automatically fails over to a configured fallback model on the **very next attempt** — no need to exhaust all retries on a dead provider first.

## Problem

When an LLM provider is rate-limited (429), experiencing server errors (5xx), or unreachable (timeout/connection error), Honcho would retry on the same provider up to `retry_attempts` (default 3) before giving up entirely. This meant:

1. **Wasted time** — retrying a dead provider is useless
2. **Wasted tokens** — partial failures still consume quota
3. **No graceful degradation** — no fallback to a local/secondary model

### Fast Fallback on First Failure

A `force_fallback` ContextVar tracks whether the current LLM call series has hit a retryable error. On the **first** detectable failure:

1. The `before_retry_callback` in `honcho_llm_call()` detects retryable errors via `_is_retryable_error()`
2. It sets `force_fallback = True` immediately
3. The `plan_attempt()` function selects the fallback `ModelConfig` on all subsequent attempts

### Configurable via Environment Variables

Each agent can define its fallback model independently using `__`-delimited Pydantic env vars:

```bash
# Example: LMStudio local model as fallback for the Deriver agent
DERIVER_MODEL_CONFIG__FALLBACK__TRANSPORT=lmstudio
DERIVER_MODEL_CONFIG__FALLBACK__MODEL=qwen/qwen3.5-9b
DERIVER_MODEL_CONFIG__FALLBACK__API_KEY=lmstudio-key
DERIVER_MODEL_CONFIG__FALLBACK__BASE_URL=http://localhost:1234/v1
```

The same pattern works for all agents:
- `DERIVER_MODEL_CONFIG__FALLBACK__*` (memory worker)
- `DIALECTIC_MODEL_CONFIG__FALLBACK__*` (query/chat)
- `DREAMER_MODEL_CONFIG__FALLBACK__*` (memory consolidation)
- `MCP_MODEL_CONFIG__FALLBACK__*` (MCP server)

### Observability: Langfuse Signal

When fallback is active, the single generation span (not a separate failed-attempt span) is updated with:

- `is_fallback: true` in metadata
- The fallback `provider` and `model` names appear on the span

This keeps traces clean — no pollution from failed-attempt spans.

### Configurable HTTP Timeout

Added `LLMSettings.DEFAULT_TIMEOUT` (env var `LLM_DEFAULT_TIMEOUT`, default `180.0`) replacing the hardcoded `600.0` for Anthropic and SDK defaults (~60s) for OpenAI/LMStudio clients. All provider clients now use a consistent, configurable timeout.

## Architecture

```
honcho_llm_call()
  │
  ├─ _get_attempt_plan()
  │   └─ plan_attempt()
  │       ├─ Normal:    returns primary ModelConfig (attempt 1 of N)
  │       ├─ Fallback:  returns fallback ModelConfig (attempt 2..N)
  │       └─ ContextVar: force_fallback controls which path
  │
  ├─ before_retry_callback()
  │   ├─ Checks _is_retryable_error() on the exception
  │   ├─ Sets force_fallback = True on first retryable failure
  │   └─ Updates Langfuse observation with fallback metadata
  │
  └─ honcho_llm_call_inner() — calls the selected provider
```

### Fallback Decision Logic

```
Attempt 1: Primary model → succeeds? Done.
Attempt 1: Primary model → 429/5xx/timeout?
  → force_fallback = True
  → Attempt 2: Fallback model → succeeds? Done.
  → Attempt 2: Fallback model → also fails?
  → Attempt 3: Fallback model → ...
  → Exhausted → error propagated to caller
```

### Retryable Error Detection

`_is_retryable_error()` classifies these as triggers:

| Error | Status/Type | Triggers Fallback |
|-------|-------------|:---:|
| Rate Limited | HTTP 429 | ✅ |
| Server Error | HTTP 5xx | ✅ |
| Timeout | `TimeoutError` | ✅ |
| Connection Error | `ConnectionError` | ✅ |
| SDK API Timeout | `APITimeoutError` | ✅ |
| SDK Connection Error | `APIConnectionError` | ✅ |
| SDK Server Error | `InternalServerError` | ✅ |
| SDK Unavailable | `ServiceUnavailableError` | ✅ |
| SDK Rate Limit | `RateLimitError` | ✅ |
| All Others | — | ❌ (pass through) |

### Core Logic

| File | Change |
|------|--------|
| `src/llm/runtime.py` | Added `_force_fallback` ContextVar, `force_fallback` parameter to `select_model_config_for_attempt()`, `is_fallback` param to `update_current_langfuse_observation()` |
| `src/llm/api.py` | Added `_is_retryable_error()`, `before_retry_callback` with first-failure fallback trigger, `_get_attempt_plan()` with `is_fallback` computation |
| `src/llm/tool_loop.py` | Reset `force_fallback.set(False)` per tool iteration |
| `src/llm/registry.py` | Replaced hardcoded `timeout=600.0` / SDK defaults with `settings.LLM.DEFAULT_TIMEOUT` (env var `LLM_DEFAULT_TIMEOUT`) |
| `src/config.py` | Added `LLMSettings.DEFAULT_TIMEOUT: float = 180.0` |

### Configuration

| File | Change |
|------|--------|
| `.env.template` | Added `LLM_DEFAULT_TIMEOUT` env var documentation |

### Tests

| File | Tests | Scope |
|------|-------|-------|
| `tests/llm/test_fallback.py` | 23 unit tests | `select_model_config_for_attempt`, `plan_attempt`, `_is_retryable_error`, cross-provider fallback (openai→anthropic, nous→openai), ContextVar lifecycle |
| `tests/llm/test_fallback_integration.py` | 3 integration tests | Fallback triggers on first failure, both primary+fallback fail → error propagated, cross-provider fallback |

### Documentation

| File | Change |
|------|--------|
| `CLAUDE.md` | Added LLM fallback architecture, config patterns, and verification commands |

## Usage

```bash
# Configure a fallback model (LMStudio local)
export DERIVER_MODEL_CONFIG__FALLBACK__TRANSPORT=lmstudio
export DERIVER_MODEL_CONFIG__FALLBACK__MODEL=qwen/qwen3.5-9b
export DERIVER_MODEL_CONFIG__FALLBACK__API_KEY=your-key
export DERIVER_MODEL_CONFIG__FALLBACK__BASE_URL=http://localhost:1234/v1

# Configure timeout (default 180s)
export LLM_DEFAULT_TIMEOUT=300.0
```

## Testing

```bash
# Unit tests
uv run pytest tests/llm/test_fallback.py -v

# Integration tests
uv run pytest tests/llm/test_fallback_integration.py -v

# Full LLM suite
uv run pytest tests/llm/ -v
```

## Verification

- Fallback triggers on **first retryable failure**, not on the final retry (default was changed from legacy behavior)
- Only the **successful generation span** is kept in Langfuse (no failed-attempt span pollution)
- Works with all agent types: Deriver, Dialectic, Dreamer, MCP
- Zero runtime overhead when no retries occur (ContextVar starts at `False`)
- All existing tests continue to pass (114 passed, 2 pre-existing failures unrelated: `test_update_env_key_creates_file_if_missing`, `test_refresh_nous_credentials_success`)

## Summary by CodeRabbit

* **New Features**
  * Immediate LLM model fallback on first retryable failure; added Nous and LMStudio provider support.

* **Configuration**
  * New global LLM timeout and expanded env placeholders for per-agent fallback and LLM settings.

* **Improvements**
  * Generation-level Langfuse observability and clearer summarizer trace names; Nous credential auto-refresh for resilient retries.

* **Infrastructure**
  * Multi-stage container build and tightened Docker ignore rules.

* **Documentation & Tests**
  * Added specs, design/proposal docs, verification plans, and tests for fallback and Nous refresh.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/692?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Rajat-Ahuja1997:
hey @phuongvm thanks for the contribution. it looks like the commits/files changes are adding a lot of changes that are not in the scope of the PR title/description. could you refactor the PR to be focused on the `LLM model fallback on failure`?

phuongvm:
# PR #692 Code Review: LLM Model Fallback on Failure

## Overall Assessment: **Approve with comments** ✅

This is a well-architected feature with strong test coverage (950+ lines of tests). The fast-failover design is the correct choice — retrying a dead provider wastes time and tokens. The implementation is production-ready with the caveats below.

---

## ✅ Strengths

1. **Fast failover on first failure** — Correct architectural decision. No point exhausting retries on a 429'd provider.
2. **ContextVar-based state management** — Thread-safe for async Python, properly scoped.
3. **Context variable cleanup** — `try/finally` with `_restore_contextvars()` prevents ContextVar leaks across calls.
4. **Backward compatibility preserved** — Legacy behavior (fallback on final attempt) still works when `force_fallback` is False.
5. **Comprehensive error classification** — `_is_retryable_error()` covers HTTP status codes, built-in exceptions, and SDK-specific exceptions.
6. **Per-agent configuration** — Independent fallback per agent via `__`-delimited env vars.
7. **Cross-provider support** — Fallback can use a different transport (e.g., OpenAI → LMStudio).
8. **Test coverage** — 4 new test files with unit tests for `select_model_config_for_attempt`, integration tests with mock backends, Nous auto-refresh tests.
9. **Observability** — Langfuse metadata with `is_fallback: true`, WARNING-level logs with primary→fallback details.
10. **Configurable timeout** — `LLM_DEFAULT_TIMEOUT` replaces hardcoded 600s (Anthropic) and ~60s (OpenAI).
11. **Summarizer telemetry fix** — Removing outer `@conditional_observe` and using `track_name` directly produces cleaner Langfuse traces.

---

## 🔴 Critical Issues

### 1. DRY Violation: Auth-refresh logic duplicated 3 times in `OpenAIBackend`

**File:** `src/llm/backends/openai.py`

The `_call_with_autorefresh` method was added (lines ~113-139) but the existing `complete()` method still has the same auth-refresh logic duplicated inline in both the `parse()` path and the `create()` path. This means the same 15-line block appears 3 times.

**Impact:** Maintenance burden — any change to the refresh flow must be made in 3 places. If one copy drifts, silent bugs follow.

**Fix:** Refactor `complete()` to delegate to `_call_with_autorefresh` instead of duplicating the logic:
```python
# Instead of inline try/except AuthenticationError blocks:
if use_parse:
    response = await self._call_with_autorefresh(use_parse=True, params=params)
else:
    response = await self._call_with_autorefresh(use_parse=False, params=params)
```

### 2. Race condition in concurrent Nous OAuth refresh

**File:** `src/llm/nous_refresh.py`

If 5 concurrent requests all hit 401 simultaneously, all 5 will call `refresh_nous_credentials()` in parallel. Each will exchange the refresh_token and mint a new agent key. The Nous API may invalidate the refresh_token after first use, causing 4 of the 5 refreshes to fail.

**Fix:** Add an `asyncio.Lock` to serialize refresh attempts:
```python
_refresh_lock = asyncio.Lock()

async def refresh_nous_credentials() -> str | None:
    async with _refresh_lock:
        # ... existing refresh logic
```

---

## 🟡 Medium Issues

### 3. `NousAuthError` breaks exception hierarchy

**File:** `src/exceptions.py`

`NousAuthError` inherits from `Exception` directly, while all other Honcho exceptions inherit from `HonchoException`. This means a `except HonchoException` catch won't trap Nous auth failures.

**Fix:**
```python
class NousAuthError(HonchoException):
    """Raised when Nous OAuth operations fail."""
    pass
```

### 4. `settings.LLM.NOUS_API_KEY=***

**File:** `src/llm/backends/openai.py`

Pydantic models are often frozen or have validators. Silently swallowing exceptions here means if the settings update fails, the in-memory key is stale for the next request.

**Fix:** Log at WARNING level if the update fails, and consider whether the settings object even supports mutation at runtime.

### 5. `_is_retryable_error` uses string-based type matching

**File:** `src/llm/api.py`

```python
return type(exc).__name__ in (
    "APIConnectionError", "APITimeoutError", "InternalServerError",
    "ServiceUnavailableError", "RateLimitError",
)
```

This avoids importing SDK-specific exceptions (good for decoupling), but is fragile to SDK renames. If OpenAI renames `RateLimitError` to `RateLimitExceededError`, this silently stops matching.

**Recommendation:** Add a comment documenting this tradeoff and the SDK versions these names were verified against. Consider adding a periodic CI assertion that these class names still exist in the installed SDK versions.

### 6. `force_fallback` resets per tool-loop iteration

**File:** `src/llm/tool_loop.py`

```python
current_attempt.set(1)
force_fallback.set(False)
```

Each tool-loop iteration resets `force_fallback` to False. If the primary provider is persistently down, every tool iteration will waste one attempt on the primary before falling back. This is correct for transient failures but suboptimal for sustained outages.

**Recommendation:** Consider an optional "sticky fallback" mode where once fallback activates, it persists for the entire tool loop. Could be a parameter like `sticky_fallback: bool = False`.

---

## 🟢 Minor / Style

### 7. Double blank line after `_update_langfuse_usage`

**File:** `src/llm/api.py` (tool-less path)

There's a double blank line between `_update_langfuse_usage(result)` and the `if trace_name` block. Should be single blank line per PEP 8.

### 8. `lmstudio` and `nous` use `get_openai_override_client` (LRU cached)

This works correctly since LMStudio/Nous are OpenAI-compatible. The `@lru_cache(maxsize=128)` keys on `(base_url, api_key)` so different configs get different clients. No issue, just confirming this was reviewed.

### 9. PR scope: OpenSpec files inflate diff

The PR includes ~30 OpenSpec archive/workspace files (~1500+ lines). While these document the design process, they significantly increase review burden. Consider separating design documentation from implementation in future PRs.

---

## Test Coverage Assessment

| Area | Coverage | Notes |
|------|----------|-------|
| `select_model_config_for_attempt` with force_fallback | ✅ Thorough | All attempt/fallback combinations tested |
| `_is_retryable_error` classification | ✅ Good | HTTP codes, built-in exceptions, SDK exceptions |
| Integration: primary fails → fallback succeeds | ✅ Tested | Mock backends verify call counts |
| Integration: both fail → error propagated | ✅ Tested | |
| Nous auto-refresh | ✅ Tested | `test_nous_refresh.py` + `test_backends/test_nous_autorefresh.py` |
| Streaming fallback | ⚠️ Not tested | Integration tests only cover non-streaming path |
| Tool loop fallback reset | ⚠️ Not tested | No test verifies `force_fallback` resets per iteration |
| Langfuse metadata | ⚠️ Not tested | `is_fallback` metadata not asserted in tests |

**Recommendation:** Add at least one streaming fallback test and one tool-loop iteration test.

---

## Summary

| Category | Count |
|----------|-------|
| Critical (must fix before merge) | 2 |
| Medium (should fix) | 4 |
| Minor (nice to have) | 3 |
| Test gaps | 3 |

The core fallback architecture is sound. The two critical issues (DRY violation in auth-refresh, race condition in Nous refresh) should be addressed before merge. The medium issues are improvements that can follow in a subsequent PR if timeline is tight.

phuongvm (commented):
## Hermes Agent Code Review -- PR #692

**Verdict: Changes Requested** (3 critical issues, 4 warnings, 5 suggestions)

---

### What is well done

- **ContextVar scoping** -- `force_fallback` is properly saved with a reset token (`_force_fallback_token`) and restored in a `finally` block. This prevents cross-call leakage.
- **First-failure fast failover** -- The design correctly switches to fallback on the *first* retryable error, not after exhausting all retries. This is the right UX for rate limits and outages.
- **Langfuse span metadata** -- Annotating the single generation span with `is_fallback` instead of creating separate failed-attempt spans is clean and avoids trace pollution.
- **Backward compatibility** -- When no fallback is configured, `force_fallback` stays `False` and the existing retry path is unchanged.
- **Timeout consolidation** -- Replacing hardcoded `600.0` (Anthropic) / SDK defaults (~60s) with a single `LLM_DEFAULT_TIMEOUT` env var is good. 180s default is reasonable.
- **Test coverage breadth** -- 23 unit tests + 3 integration tests cover the core logic: force_fallback True/False, no-fallback config, cross-provider, final-attempt fallback.

---

### CRITICAL

1. **`_is_retryable_error()` catches `OSError` too broadly**
   - `src/llm/api.py:308`: `isinstance(exc, (TimeoutError, ConnectionError, OSError))` -- `OSError` includes many non-retryable errors (permission denied, disk full, etc.). `ConnectionError` is already a subclass of `OSError`, so the `OSError` catch-all is redundant and dangerous.
   - **Fix**: Remove `OSError` from the tuple. `TimeoutError` and `ConnectionError` are sufficient for network errors. SDK-specific exceptions cover the rest.

2. **Missing agent name in fallback log (spec violation)**
   - `src/llm/api.py:339-343`: The fallback log message includes transport/model but **does not include the agent name** (deriver, dialectic, dreamer, etc.). The spec (`openspec/specs/llm-model-fallback/spec.md`, Requirement 3) explicitly requires: `"agent=deriver, primary=<provider>/<model>, fallback=<provider>/<model>, reason=<error>"`.
   - **Fix**: Thread agent name through `honcho_llm_call` and include it in the log message.

3. **`nous_refresh.py` writes to `.env` file at runtime**
   - `src/llm/nous_refresh.py`: The `update_env_key()` function modifies the `.env` file on disk after refreshing credentials. In containerized deployments (Docker, k8s), the `.env` is baked into the image or mounted read-only -- this write will fail silently or cause issues.
   - **Fix**: Write to a runtime state file (like the OAuth state file) and have the credential loader check that file before falling back to `.env`. Or make this a no-op when running in a container.

---

### WARNINGS

4. **No test for partial fallback config**
   - The spec (Requirement 8, last scenario) requires that when only `FALLBACK__TRANSPORT` is set without `FALLBACK__MODEL`, the system ignores the incomplete config. No test verifies this edge case.
   - **Fix**: Add a test case with `fallback=ModelConfig(transport="anthropic", model=None)` and verify it returns primary.

5. **`settings.LLM.NOUS_API_KEY` mutation in `_call_with_autorefresh`**
   - `src/llm/backends/openai.py:~143`: `settings.LLM.NOUS_API_KEY = new_key` -- mutating a Pydantic `BaseSettings` field at runtime is fragile. If the settings class is re-instantiated or validated elsewhere, the mutation is lost.
   - **Fix**: Consider storing the refreshed key in a dedicated module-level variable or the OAuth state file, and have the settings getter check that source first.

6. **Stream path does not get `_update_langfuse_usage`**
   - `src/llm/api.py:~456-462`: `_update_langfuse_usage(result)` is only called for `HonchoLLMCallResponse` (non-stream). Streaming responses return `AsyncIterator[HonchoLLMCallStreamChunk]` and skip usage reporting entirely.
   - **Fix**: Hook usage reporting into the stream consumer or document this as a known gap.

7. **`honcho_llm_call` ContextVar reset in try/finally**
   - The `_restore_contextvars()` function is defined but I did not see a `finally` block calling it in the visible diff. If the ContextVars are not reset on exit (exception or normal return), subsequent calls may inherit stale `force_fallback=True` state.
   - **Verify**: Ensure the `try/finally` wrapping the call body actually calls `_restore_contextvars()`.

---

### SUGGESTIONS

8. **`select_model_config_for_attempt()` docstring says "last attempt" but now also handles force_fallback** -- the docstring was updated but could be clearer about precedence: `force_fallback=True` overrides attempt number logic entirely.

9. **`_is_retryable_error()` uses `type(exc).__name__` string matching** -- this is fragile. If SDK exception class names change, the check silently stops working. Consider importing the actual SDK exception types when available.

10. **LMStudio and Nous share `OpenAIBackend`** -- this is correct (they are OpenAI-compatible), but the `is_nous=True` flag on `OpenAIBackend` is the only distinguishing feature. Consider whether a `NousBackend` subclass would be cleaner as the Nous-specific logic grows (OAuth refresh, error handling).

11. **PR scope: 4,342 additions in 58 files** -- this PR includes not just fallback logic but also: Langfuse generation traces, MCP QA verification, Nous OAuth refresh, Docker/MCP infrastructure, and OpenSpec archival. Consider splitting into smaller PRs for easier review.

---

*Reviewed by Hermes Agent*

phuongvm:
⚠️ **PR has been split** into focused PRs per maintainer request:

| PR | Scope | URL |
|---|---|---|
| #736 | LLM model fallback on failure | https://github.com/plastic-labs/honcho/pull/736 |
| #737 | Nous Research provider support | https://github.com/plastic-labs/honcho/pull/737 |
| #738 | MCP Dockerfile improvements | https://github.com/plastic-labs/honcho/pull/738 |

All tests passing (37/37 total). Branches pushed to `phuongvm/honcho`.

VVoruganti:
Thanks @phuongvm — closing this in favor of the focused stack you split it into (#736 fallback core, #737 Nous provider, #738 MCP Docker), per the note above. We'll review those individually. Appreciate breaking it up.

PR #696, opened 2026-05-17
fix: make embedding batch size configurable

### LINKED ISSUE #687: [Bug] OpenAIEmbeddingClient.max_batch_size hardcoded to 2048 breaks DashScope embeddings (cap is 10)
# 🐞 Bug Report

## Describe the bug

`OpenAIEmbeddingClient.max_batch_size` is hardcoded to `2048` for every backend using `transport = "openai"`. That's OpenAI's per-request input limit, but the openai transport is documented (in `.env.template` and the config docs) as supporting OpenAI-compatible providers — "OpenRouter, Together, Fireworks, vLLM, Ollama, LiteLLM" — several of which have much lower limits:

* Alibaba DashScope `text-embedding-v4` — **10 inputs per request**
* DashScope `text-embedding-v3` — 25 inputs per request

When Honcho's deriver embeds a batch, it sends 2048 inputs and the provider returns 400 every time:

```
400 - InternalError.Algo.InvalidParameter: Value error, batch size is
invalid, it should not be larger than 10.: input.contents
```

Effect: Qwen embeddings via DashScope (a sensible budget choice that pairs naturally with Qwen text-gen for the same key) cannot be used with Honcho without patching the source.

---

### Is this a regression?

Not that I can tell — `max_batch_size` has been hardcoded in `embedding_client.py` since the openai transport landed. The issue surfaces as the openai transport is increasingly used as a generic OpenAI-compatible adapter, not just for native OpenAI.

---

### To Reproduce

1. In `config.toml` set the embedding block to DashScope:
   ```toml
   [embedding.model_config]
   transport = "openai"
   model = "text-embedding-v4"
   [embedding.model_config.overrides]
   base_url = "https://dashscope-intl.aliyuncs.com/compatible-mode/v1"
   ```
2. Put a valid DashScope API key in `LLM_OPENAI_API_KEY`.
3. Bring up the stack (`docker compose up -d`).
4. POST a session message that triggers embedding.
5. Tail the deriver logs:
   ```
   docker compose logs deriver | grep "batch size"
   ```
6. Every embedding call fails with `400 ... batch size is invalid, it should not be larger than 10`.

---

### Expected behaviour

Either:

* **Auto-detect a lower per-provider default** — e.g. when `overrides.base_url` is set and not pointing at `api.openai.com`, fall back to a conservative limit like 10 (or, ideally, look up known providers); or
* **Expose `max_batch_size` as a config field** under `[embedding.model_config]` (or a sibling key), so it can be set per provider.

The second option is simpler, matches the pattern of other per-module overrides already in the config, and would unblock not just DashScope but any other strict OpenAI-compatible embedding provider.

---

### Your environment

* OS: Ubuntu 24.04 (Oracle ARM, docker-compose deployment)
* Honcho Server Version: **v3.0.6**
* Provider: Alibaba Cloud DashScope International — `https://dashscope-intl.aliyuncs.com/compatible-mode/v1`
* Embedding model: `text-embedding-v4`

---

### Additional context

**Affected line:** `src/embedding_client.py:66`

```python
self.max_batch_size = 2048  # OpenAI batch limit
```

**Workaround:** Local source patch lowering the constant. Bind-mounted into the container via docker-compose so image rebuilds don't lose it, but `git pull` reverts it. Not a sustainable long-term solution.

**Related:** This is one of two DashScope-compatibility bugs I hit running Honcho with `transport = "openai"` against the DashScope endpoint. The other is the JSON-mode keyword check tracked in #663 (still open after a previous fix in #453 was apparently lost in the v3 rewrite). Same root cause class: the openai transport works against native OpenAI but not always against strict OpenAI-compatible providers.

pntech20:
## Summary
- add optional `max_batch_size` to embedding model config and runtime embedding config
- use the configured batch cap when splitting OpenAI/Gemini embedding requests
- document the env/TOML setting for OpenAI-compatible providers such as DashScope

Fixes #687.

## Verification
- `uv run ruff check src\config.py src\embedding_client.py tests\llm\test_embedding_client.py`
- `uv run ruff format --check src\config.py src\embedding_client.py tests\llm\test_embedding_client.py`
- `uv run basedpyright src\config.py src\embedding_client.py tests\llm\test_embedding_client.py`
- direct runtime check that `max_batch_size=2` splits `simple_batch_embed(["a", "b", "c"])` into two OpenAI embedding calls and that embedding settings parse `max_batch_size=10`

`uv run pytest tests\llm\test_embedding_client.py -q` is blocked in this Windows environment before the test file runs because the repo-level `tests/conftest.py` imports app startup, which imports `src.telemetry.reasoning_traces`, which imports Unix-only `fcntl`.

## Summary by CodeRabbit

* **New Features**
  * Added optional `max_batch_size` configuration for embedding providers, enabling customization of per-request input limits. Defaults to provider-specific values (e.g., 2048 for OpenAI, 100 for Gemini).

* **Documentation**
  * Updated embedding configuration guide with batch size setting examples and guidance for providers with smaller request limits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/696?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


oliver-mee:
Amazing, thank you!

pntech20:
Thanks! Glad this was useful.

WizisCool:
FYI: opened #983 reviving this approach (same design, rebased onto current main — the test file had drifted into conflict). Credited in the PR body; happy to close it if you'd rather merge this one.

akattelu:
Closing this as superseded by #983

Issue #394, opened 2026-02-21
Security: Session isolation bypass, rate limiting, and input validation gaps


SocialMDev:
## Security Audit Findings — Honcho API

Security audit identified multiple vulnerabilities in the Honcho API server.

---

### Session Isolation Bypass with Incomplete JWT

**File:** `src/routers/sessions.py` (get_or_create_session)

JWT without `w` (workspace) claim bypasses workspace isolation for non-admin users:

```python
if not jwt_params.ad and jwt_params.w is not None and jwt_params.w != workspace_id:
    raise AuthenticationException(...)
```

If `jwt_params.w` is `None`, the check passes. A JWT with `{"ad": false}` and no workspace claim can access any workspace.

**Fix:** Require `w` claim for non-admin JWTs:
```python
if not jwt_params.ad:
    if jwt_params.w is None or jwt_params.w != workspace_id:
        raise AuthenticationException("Workspace not specified or mismatched")
```

---

### No Rate Limiting on Any Endpoint

**File:** `src/main.py`

No rate limiting middleware exists. An attacker with a valid token can create unlimited messages, exhaust storage, or DoS the database.

**Fix:** Add `slowapi` rate limiting middleware.

---

### Prompt Injection via search_query

**File:** `src/routers/sessions.py` (get_session_context)

`search_query` parameter passed directly to LLM context without sanitization or length limits. Allows prompt injection attacks.

**Fix:** Add 500-char length limit, sanitize control characters.

---

### JWT Potentially Logged in Error Traces

JWT tokens may be captured by logging systems or error handlers in stack traces when passed to client initialization.

**Fix:** Mask token in all log output.

---

### Overpermissive CORS Configuration

**File:** `src/main.py`

`allow_methods=["*"]` and `allow_headers=["*"]` are overpermissive.

**Fix:** Explicitly whitelist methods and headers.

---

### No Content Length Validation on Messages

**File:** `src/routers/messages.py`

No size limit on message content in `MessageBatchCreate` schema. Can cause storage/memory exhaustion.

**Fix:** Add `max_length` constraint to message content field.

PR #697, opened 2026-05-17
feat(config): make CORS allowed origins configurable via env

raulanatol:
## Summary
- Adds a new `CORSSettings` block (env prefix `CORS_`) and exposes it as `settings.CORS.ORIGINS`.
- `src/main.py` now feeds `allow_origins` from `settings.CORS.ORIGINS` instead of a hardcoded list. Defaults preserve the prior values (`http://localhost`, `http://127.0.0.1:8000`, `https://api.honcho.dev`) so existing deployments behave identically.
- Documented under a new **CORS Settings** section in `.env.template`.

Motivation: self-hosted deployments behind custom domains currently have to patch `src/main.py` to whitelist their frontend. With this change a single env var is enough:

```
CORS_ORIGINS=["https://app.example.com","https://staging.example.com"]
```

Scope is intentionally limited to `origins`. The `allow_methods` / `allow_headers` wildcards flagged in #394 are left as-is to keep this PR focused on the env-configurability hole; happy to follow up in a separate PR if maintainers want them tightened too.

## Test plan
- [x] `uv run ruff check src/`
- [x] `uv run ruff format --check src/`
- [x] `uv run basedpyright src/main.py src/config.py` — 0 errors
- [x] Verified defaults load unchanged: `uv run python -c "from src.config import settings; print(settings.CORS.ORIGINS)"`
- [x] Verified env override: `CORS_ORIGINS='["http://foo.test","https://bar.test"]' uv run python -c ...`
- [ ] Maintainers: smoke test that CORS preflight from a custom origin works once `CORS_ORIGINS` is set

## Summary by CodeRabbit

* **New Features**
  * Allowed origins for cross-origin requests are now configurable at runtime via environment settings instead of using a built-in hardcoded list.

* **Chores**
  * Environment template updated with a CORS section and example allowlist entries for local and production setups.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/697?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/config.py
```diff
@@ -1195,6 +1195,18 @@ def _validate_specialist_token_budgets(self) -> "DreamSettings":
         return self
 
 
+class CORSSettings(HonchoSettings):
```
Rajat-Ahuja1997: instead of adding a separate `CORSSetting` block, it would be simpler to just add a `CORS_ORIGINS` field to `AppSettings`. if we ever see the need to add other `CORS_*` values then we can promote this to a nested class.

Rajat-Ahuja1997 (commented):
looks good, one suggestion regarding placement

src/config.py
raulanatol: Good call, thanks! I just pushed a commit that drops the dedicated `CORSSettings` block and adds `CORS_ORIGINS` directly to `AppSettings`. The env var (`CORS_ORIGINS`) keeps working as-is since `AppSettings` has no env prefix. If more `CORS_*` knobs come up later we can promote it back to a nested class.

Rajat-Ahuja1997 (approved):

Rajat-Ahuja1997 (approved):

Issue #494, opened 2026-04-05
Self-hosted/local Honcho ingests messages but derived memory does not appear automatically; peer card/context/search remain empty


qaqcvc:
## Bug Description

In a local/self-hosted Honcho deployment, messages are successfully written into Honcho and queue items are created, but derived memory does not appear automatically unless the deriver is started manually. Even after manually running the deriver and processing queued work, peer card / profile-style surfaces still remain empty.

This makes the local deployment appear partially functional:
- message/session ingestion works
- queueing works
- manual deriver execution works
- but `peer card`, `peer context`, and profile/search surfaces remain empty or incomplete

## Environment

- Honcho version: 3.0.5
- Deployment: local/self-hosted
- API server running locally on `http://localhost:8000`
- Hermes configured to use Honcho as memory provider
- Hermes Honcho host config:
  - `memoryMode: hybrid`
  - `recallMode: hybrid`
  - `writeFrequency: async`
  - `saveMessages: true`
- Deriver config observed from local runtime:
  - `DERIVER_ENABLED = True`
  - `DERIVER_PROVIDER = custom`
  - `DERIVER_MODEL = gpt-5-mini`
  - `OPENAI_COMPATIBLE_API_KEY = set`

## What we observed

1. Messages are being logged into Honcho
- workspace, peers, and sessions are created successfully
- local API endpoints are reachable
- `messages`/session structures exist in the backend

2. Queue items are created, but were not being processed automatically
Before manually starting the deriver, queue status showed:
- `completed_work_units = 0`
- `pending_work_units = 33`

3. No separate deriver worker process was running
The API server was running, but there was no separate `python -m src.deriver` process alive.

4. Manually starting the deriver processed work successfully
Running the deriver manually (`python -m src.deriver`) processed queue items and increased completed work units, e.g. later queue status showed:
- `completed_work_units = 23`
- `pending_work_units = 15`

This suggests model/API configuration is functional enough for derivation, and that the major operational issue is that the deriver was not running automatically in our local deployment.

5. Even after manual derivation, peer-card/profile surfaces remained empty
Direct API checks still returned empty results:

- `GET /v3/workspaces/myHoncho/peers/hermes/card?target=ctx`
  - returned `{"peer_card": null}`

- `GET /v3/workspaces/myHoncho/peers/hermes/context?target=ctx`
  - returned something like:
  - `{"peer_id":"hermes","target_id":"ctx","representation":"","peer_card":null}`

[...]

PR #703, opened 2026-05-20
feat(deriver): add stale batch timeout to prevent indefinite queue stall

maxxqf-ai:
## Problem

When `REPRESENTATION_BATCH_MAX_TOKENS` is configured and `FLUSH_ENABLED` is `false`, work units that never accumulate enough tokens remain stuck in the queue indefinitely. This is especially problematic for short conversational sessions where individual session message tokens never reach the batch threshold.

In our deployment (Hermes agent with FLUSH at ~2K tokens but batch threshold at 4K), short sessions would produce ~2K tokens per flush — never reaching the 4K threshold. This resulted in **245 queue items accumulated over 10 days** with zero processing.

### Why not just use FLUSH_ENABLED=true?

FLUSH mode processes every message immediately, which produces fragmented observations lacking context. The extracted information is broken into tiny pieces with high redundancy and poor quality. The token-batch threshold exists to group related messages together for richer, more coherent observations.

The trade-off: batch mode gives better quality, but work units that never reach the threshold are **never processed at all**.

## Solution

Add a `STALE_BATCH_HOURS` configuration option (default `0` = disabled) to DeriverSettings. When set to a positive value, work units whose oldest queue item exceeds this age are force-processed regardless of token count.

This provides a safety net that:
- Preserves batch quality for active sessions (messages are still grouped up to the token threshold)
- Ensures no work unit is abandoned forever (stale work units get processed after the timeout)
- Is fully opt-in with a default of 0 (no behavior change for existing deployments)

### Changes

- **`src/config.py`**: Add `STALE_BATCH_HOURS` field to `DeriverSettings` with validation (0-168 hours)
- **`src/deriver/queue_manager.py`**: Extend `get_and_claim_work_units()` to include stale work units via a subquery on `MIN(QueueItem.created_at)` when `STALE_BATCH_HOURS > 0`

### Configuration Example

```toml
[deriver]
REPRESENTATION_BATCH_MAX_TOKENS = 2048
FLUSH_ENABLED = false
STALE_BATCH_HOURS = 5  # Force-process work units older than 5 hours
```

### Related

Possibly related to #494 (messages ingested but derived memory does not appear automatically).

## Summary by CodeRabbit

* **Configuration**
  * Added `STALE_BATCH_HOURS` setting (0-168 hours) to control work unit processing thresholds.

* **Improvements**
  * Enhanced queue management to process items based on age when batching conditions are applied.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/703?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -780,6 +780,9 @@ class DeriverSettings(HonchoSettings):
     # When enabled, bypasses the batch token threshold and processes work immediately
     FLUSH_ENABLED: bool = False
 
+    # Force-process representation work units whose oldest queue item is older than this many hours
+    STALE_BATCH_HOURS: Annotated[int, Field(default=0, ge=0, le=168)] = 0
+
     @model_validator(mode="before")
     @classmethod
     def _merge_model_config_defaults(cls, data: Any) -> Any:
--- a/src/deriver/queue_manager.py
+++ b/src/deriver/queue_manager.py
@@ -310,16 +310,51 @@ class QueueManager:
             )
 
             # Apply batch threshold filter (skip if FLUSH_ENABLED is True)
+            # Also force-process work units whose oldest item exceeds STALE_BATCH_HOURS
             if not settings.DERIVER.FLUSH_ENABLED and batch_max_tokens > 0:
-                query = query.where(
-                    or_(
-                        ~work_units_subq.c.work_unit_key.startswith(
-                            representation_prefix
-                        ),
-                        func.coalesce(token_stats_subq.c.total_tokens, 0)
-                        >= batch_max_tokens,
+                stale_hours = settings.DERIVER.STALE_BATCH_HOURS
+                if stale_hours > 0:
+                    stale_cutoff = datetime.now(timezone.utc) - timedelta(
+                        hours=stale_hours
                     )
-                )
+                    oldest_item_subq = (
+                        select(
+                            models.QueueItem.work_unit_key,
+                            func.min(models.QueueItem.created_at).label(
+                                "oldest_created"
+                            ),
+                        )
+                        .where(~models.QueueItem.processed)
+                        .group_by(models.QueueItem.work_unit_key)
+                        .subquery()
+                    )
+                    query = query.outerjoin(
+                        oldest_item_subq,
+                        work_units_subq.c.work_unit_key
+                        == oldest_item_subq.c.work_unit_key,
+                    ).where(
+                        or_(
+                            ~work_units_subq.c.work_unit_key.startswith(
+                                representation_prefix
+                            ),
+                            func.coalesce(token_stats_subq.c.total_tokens, 0)
+                            >= batch_max_tokens,
+                            func.coalesce(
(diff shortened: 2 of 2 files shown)


menhguin:
+1, would love to see this merged. We're hitting the exact symptom on the managed instance (`api.honcho.dev`).

**Our context:**
- Hosted workspace, OpenClaw integration ingesting messages from chat sessions across ~6 conversational surfaces (Slack, Telegram, Signal, etc.)
- Many sessions are short — single-digit message counts is common for one-off threads or thread-replies that route into their own session
- Deriver processes sporadically. Over the past 24h: long stretches of zero derivations punctuated by short bursts. Some sessions <10 messages have been queued for days without a representation update.

**What we tried before finding this PR:**
1. Lowered `MESSAGE_THRESHOLD` to 2 (suspecting per-session message count was gating). Briefly spiked throughput to ~70 derivations/hour, then stalled again — consistent with this PR's root cause: lowering the per-session threshold gets *more* work units onto the queue, but they still never reach `REPRESENTATION_BATCH_MAX_TOKENS`, so the queue manager never claims them.
2. Considered toggling `FLUSH_ENABLED=true` but that gives up the token-batching efficiency — not what we want as a permanent answer.
3. Verified the deriver worker process is running and not crashed.

**Why this PR is the right shape:**
- Gated behind `STALE_BATCH_HOURS > 0` (default `0` = current behavior) — strictly additive, no regression risk for existing deployments.
- Composes with `FLUSH_ENABLED` cleanly: token-threshold-first, stale-timeout as a backstop.
- The `0–168` upper bound is a sensible safeguard.

**Requests, if helpful:**
1. Once merged, please consider exposing `STALE_BATCH_HOURS` as a per-workspace or per-instance setting on the managed `api.honcho.dev` deployment — currently we have no knob for this from the hosted side.
2. Sensible default for hosted: something like `2–6` hours would catch our case without much queue churn.
3. Slightly orthogonal but related: it would help operators if `honcho.queue_status(...)` surfaced the oldest unprocessed `created_at` per work unit, so stalled work units are diagnosable without DB access.

Happy to test against a self-hosted build with this patch applied if it helps validate before merge.

pbeatty84:
> Second production Hermes-agent deployment confirming this. `REPRESENTATION_BATCH_MAX_TOKENS=512`, flush disabled: short-lived agent sessions strand their tail messages permanently — we accumulated 58 unprocessed `representation` work units over 13 days (1-2 rows each across dozens of churned sessions), and a fresh cohort began accumulating within days of cleanup. Confirmed on v3.0.8/v3.0.9 that deriver restarts do **not** recover aged rows. Even a small threshold strands session tails, so this isn't avoidable by tuning — an age-based flush like the proposed timeout (or the purge endpoint in #799/#801) would close it. Interim mitigation for anyone else: we monitor `SELECT count(*) FROM queue WHERE processed=false AND created_at < now()-interval '2 days'` and manually purge with a CSV backup.

VVoruganti:
Superseded by #826

PR #693, opened 2026-05-15
feat: Langfuse LLM observability — generation tracing with token usage reporting

phuongvm:
## Summary

Integrates Langfuse as an observability layer for Honcho's LLM calls. Every LLM generation is automatically traced as a Langfuse generation span with model name, provider, input/output tokens, and reasoning effort metadata — enabling cost tracking, latency analysis, and debugging across all Honcho agents.

## Problem

Honcho's LLM calls (Deriver, Dialectic, Dreamer, MCP agents) had no centralized observability. Teams could not:
- Track which models/providers were called and how often
- Monitor per-call latency and token usage
- Debug failed or slow LLM interactions in production
- Measure cost attribution per agent or workspace

### Automatic Generation Tracing

Every call through `honcho_llm_call()` is automatically traced as a Langfuse generation span via the `@conditional_observe` decorator. The span is updated with runtime metadata: model name, provider, input/output tokens, and is_fallback status.

### Token Usage Reporting

Token counts are reported on the generation span after each call completes, working for both standard providers (Anthropic, OpenAI) and custom model configurations.

### Refactored Summarizer Telemetry

Outer summarizer span decorators were replaced with direct `track_name` passing to `honcho_llm_call()`, enabling properly nested traces under correct parent traces and eliminating orphaned/duplicate generation spans.

### Configuration

Langfuse is configured via environment variables:
- `LANGFUSE_PUBLIC_KEY`
- `LANGFUSE_SECRET_KEY`
- `LANGFUSE_HOST`

The `@conditional_observe` decorator is a no-op with zero overhead when Langfuse is not configured.

## Architecture

```mermaid
flowchart TD
    A[honcho_llm_call] --> B[@conditional_observe]
    B --> C[Create Langfuse generation span]
    C --> D[resolve_runtime_model_config]
    D --> E[plan_attempt / select provider+model]
    E --> F[honcho_llm_call_inner]
    F --> G[update_current_generation with tokens]
```

### Trace Flow

```
Trace: "smart-summarizer"
  Generation: "LLM Call"
    model: claude-sonnet-4-20250514
    provider: anthropic
    usage.input: 2048
    usage.output: 512
```

## Files Changed

| File | Change |
|------|--------|
| `src/llm/api.py` | Added `@conditional_observe` decorator, post-call Langfuse token usage reporting |
| `src/llm/runtime.py` | Added `update_current_langfuse_observation()` for span metadata updates |
[...]

PR #707, opened 2026-05-21
feat: support Langfuse trace grouping via LiteLLM proxy metadata

DanielMaly:
## Summary

Support Langfuse trace grouping for Honcho's agent loops by passing `session_id` through to LiteLLM proxy via request metadata. When LLM calls are routed through a LiteLLM proxy with `callbacks: ["langfuse"]`, all calls from the same agent operation (dream cycle, dialectic request, deriver batch) are now grouped under one Langfuse session.

## Problem

When Honcho's LLM calls are routed through a LiteLLM proxy with `callbacks: ["langfuse"]`, each completion request creates a separate Langfuse trace with no correlation between them. A single dream cycle (deduction specialist → induction specialist, potentially 10–20 LLM calls) produces 10–20 orphaned traces. The same applies to dialectic tool loops and deriver batches.

This makes it difficult to:
- Understand the full lifecycle of an agent operation
- Attribute cost and latency to a specific dream/dialectic/deriver run
- Debug multi-step agent behaviour in context

## Solution

Add `langfuse_session_id` to `LLMTelemetryContext`. Agent entry points set it, and the OpenAI backend passes it as `extra_body={metadata: {session_id: ...}}` to LiteLLM. LiteLLM's Langfuse callback reads `metadata.session_id` to group traces under one session.

### Data flow

```
Agent entry point (e.g. run_dream)
  → LLMTelemetryContext(langfuse_session_id="dream-abc123")
    → honcho_llm_call(telemetry=ctx)
      → honcho_llm_call_inner
        → extra_params["langfuse_session_id"] = "dream-abc123"
          → execute_completion / execute_stream
            → OpenAIBackend._build_params
              → params["extra_body"] = {"metadata": {"session_id": "dream-abc123"}}
                → LiteLLM proxy
                  → Langfuse callback: trace.session_id = "dream-abc123"
```

### Session ID format

| Agent | Format | Example | Scope |
|-------|--------|---------|-------|
| Dreamer specialist | `dream-{run_id}` | `dream-V1StGXR8` | One dream cycle |
| Dialectic | `dialectic-{run_id}` | `dialectic-3KjR9mN2` | One chat request |
| Deriver | `deriver-{batch_id}` | `deriver-K7pWxQ4z` | One message batch |

Each agent already generates a unique ID for internal telemetry (the dreamer and dialectic use `run_id` from nanoid). The deriver now generates a `batch_id` nanoid per batch in `process_representation_tasks_batch`, matching the same pattern.

### Changes

- **`src/llm/types.py`**: Add `langfuse_session_id: str | None = None` to `LLMTelemetryContext`
- **`src/llm/executor.py`**: Propagate `langfuse_session_id` from telemetry to `extra_params` dict
- **`src/llm/backends/openai.py`**: When `extra_params` contains `langfuse_session_id`, emit `extra_body` with `metadata.session_id`
- **`src/dreamer/specialists.py`**: Set `langfuse_session_id` from `run_id`
- **`src/dialectic/core.py`**: Set `langfuse_session_id` from `run_id`
- **`src/deriver/deriver.py`**: Generate `batch_id` nanoid per batch; set `langfuse_session_id` from it
- **`src/llm/tool_loop.py`**: Preserve `langfuse_session_id` in per-iteration telemetry context copies
- **Tests**: Verify `extra_body` is present when `langfuse_session_id` is set, absent when not

### Why session_id, not trace_id?

LiteLLM supports both `session_id` (groups traces into a session) and `trace_id`/`existing_trace_id` (merges calls into a single trace). I chose `session_id` because:

1. **Robustness**: If the first LLM call fails, subsequent calls still appear under the session. With `existing_trace_id`, orphaned spans result if the parent trace was never created.
2. **Simplicity**: No need to distinguish "first call" vs "subsequent calls" — every call just carries the same `session_id`.
3. **Semantics**: A dream cycle or dialectic request IS a session — multiple related traces that should be viewed together.

If tighter single-trace grouping is desired in the future, the same plumbing can carry `trace_id`/`existing_trace_id` with minimal changes.

## Complements #693

This PR is complementary to #693 (Langfuse generation tracing). That PR makes individual LLM calls visible as proper generation observations via `@conditional_observe`. This PR ensures that when those calls are routed through a LiteLLM proxy, they're also grouped together by agent operation.

## Related

- #693 — Langfuse generation tracing (active PR)
- #565 — Langfuse metadata (namespace, model, provider)
- #253 — Langfuse rework introducing conditional_observe

## Summary by CodeRabbit

* **New Features**
  * Added per-operation session IDs across components so related interactions are consistently tagged and grouped in observability traces and propagated to upstream requests for improved trace correlation.

* **Tests**
  * Added tests ensuring session IDs are included in backend requests when provided and omitted when not present.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/707?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -1,6 +1,8 @@
 import logging
 import time
 
+from nanoid import generate as generate_nanoid
+
 from src import crud
 from src.config import ConfiguredModelSettings, settings
 from src.crud.representation import RepresentationManager
@@ -62,6 +64,8 @@ async def process_representation_tasks_batch(
         return
 
     overall_start = time.perf_counter()
+
+    batch_id = generate_nanoid()
 
     messages.sort(key=lambda x: x.id)
     latest_message = messages[-1]
@@ -159,7 +163,7 @@ async def process_representation_tasks_batch(
             call_purpose=CallPurpose.DERIVER_REPRESENTATION.value,
             parent_category="representation",
             observed=observed,
-            langfuse_session_id=f"deriver-{latest_message.workspace_name}-{observed}",
+            langfuse_session_id=f"deriver-{batch_id}",
         ),
     )
     llm_duration = (time.perf_counter() - llm_start) * 1000
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -159,6 +159,7 @@ async def process_representation_tasks_batch(
             call_purpose=CallPurpose.DERIVER_REPRESENTATION.value,
             parent_category="representation",
             observed=observed,
+            langfuse_session_id=f"deriver-{latest_message.workspace_name}-{observed}",
         ),
     )
     llm_duration = (time.perf_counter() - llm_start) * 1000
--- a/src/dialectic/core.py
+++ b/src/dialectic/core.py
@@ -316,6 +316,7 @@ class DialecticAgent:
             agent_type="dialectic",
             run_id=self._run_id,
             peer_name=self.observed,
+            langfuse_session_id=f"dialectic-{self._run_id}" if self._run_id else None,
         )
 
     def _log_response_metrics(
--- a/src/dreamer/specialists.py
+++ b/src/dreamer/specialists.py
@@ -277,6 +277,7 @@ If you update it, send the full deduplicated list and remove stale entries.
                     run_id=run_id,
                     observer=observer,
                     observed=observed,
+                    langfuse_session_id=f"dream-{run_id}" if run_id else None,
                 ),
             )
 
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -324,9 +324,10 @@ class OpenAIBackend:
         # LiteLLM's Langfuse callback reads metadata.session_id to group
         # traces from the same agent operation (dream cycle, dialectic
         # request, etc.) under one Langfuse session.
-        if extra_params and "langfuse_session_id" in extra_params:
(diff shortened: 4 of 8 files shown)


VVoruganti:
Is the idea here to add support so that you can centralize the tracing via a litellm proxy rather than via Honcho? A bit confused on the goal if Honcho already has langfuse tracing built into it. 

Would this work without litellm?

DanielMaly:
@VVoruganti 

> Is the idea here to add support so that you can centralize the tracing via a litellm proxy rather than via Honcho? A bit confused on the goal if Honcho already has langfuse tracing built into it.
> 
> Would this work without litellm?

The issue here is that if traces go from Honcho directly to Langfuse, any routing and cost-tracking information from LiteLLM (or any other proxy) is lost because the trace bypasses it. This change allows a session id to propagate through the OpenAI-compatible endpoint which gives a lot more flexibility in observability integrations.

akattelu:
Hi @DanielMaly, thanks for opening this.

We've just landed #814, which adds that session grouping natively in the Langfuse SDK path: one session per dialectic call, with workspace as the user id, so the multi-call correlation you're describing is now handled there. The piece this PR adds on top is routing `session_id` through `extra_body.metadata` for the LiteLLM proxy path specifically, and that's not something we're prioritizing right now, so I'm going to close this.

Thanks again for the clear writeup of the problem.

PR #708, opened 2026-05-21
Add OpenRouter app attribution headers

andyylin:
## Summary
- add configurable OpenRouter attribution headers for OpenAI-compatible LLM clients
- cover embedding client OpenRouter calls too
- document LLM_OPENROUTER_APP_URL and LLM_OPENROUTER_APP_TITLE
- add unit coverage for attribution behavior

## Validation
- rebuilt Honcho api/deriver Docker images locally
- force-recreated live api/deriver containers
- verified /health returns ok
- verified live container resolves OpenRouter headers:
  - HTTP-Referer: https://local.honcho.andy
  - X-OpenRouter-Title: Andy Honcho Memory
  - X-Title: Andy Honcho Memory
- ran targeted ruff/compile/runtime smoke checks locally

Note: full pytest was not run locally because the production image excludes dev dependencies and the Pi host lacks the full Honcho test environment; CI should run the full suite.

## Summary by CodeRabbit

* **New Features**
  * Auto-applies OpenRouter attribution headers when an OpenRouter-compatible base URL is used; supports configurable app URL/title.
  * Added CLI utilities for re-embedding database vectors and running semantic benchmarks.

* **Configuration**
  * Example config includes commented OpenRouter app attribution keys.
  * Added a full example config.toml for self-hosted deployment.

* **Docker**
  * Added docker-compose for self-hosted deployment and a local compose file for embedding tests.

* **Tests**
  * Added tests for OpenRouter attribution and updated embedding client test fakes.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -13,12 +13,12 @@ NAMESPACE = "honcho"
 
 [embedding]
 VECTOR_DIMENSIONS = 1024
-# Ollama's mxbai tokenizer can count some Hermes/Honcho messages higher than
-# Honcho's cl100k estimate. Keep chunks comfortably below the provider context
-# and keep each provider request to one large chunk so batched input does not
-# blow Ollama's shared request context.
-MAX_INPUT_TOKENS = 768
-MAX_TOKENS_PER_REQUEST = 768
+# Ollama's mxbai tokenizer can count some Hermes/Honcho messages much higher than
+# Honcho's cl100k estimate. 768 still caused `input length exceeds context length`
+# on normal Hermes memory snapshots (~695 cl100k tokens), so keep chunks well below
+# the provider context and one large chunk per provider request.
+MAX_INPUT_TOKENS = 384
+MAX_TOKENS_PER_REQUEST = 384
 
 [db]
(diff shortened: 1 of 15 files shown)


akattelu:
Hey @andyylin, thanks for opening this, and apologies for the slow turnaround. The OpenRouter attribution idea is a good one and the core of it (detecting an OpenRouter base URL and injecting `HTTP-Referer` / `X-OpenRouter-Title`) is exactly right.

We just merged #805, which lands the same attribution-header feature in a focused form, wired into the registry and embedding clients. So I'm going to close this as superseded to avoid duplicating the change.

Two things worth carrying over if you're up for a follow-up:
- #805 sends `HTTP-Referer` and `X-Openrouter-Title` but not the `X-Title` variant you included here. If that header is meaningful for OpenRouter attribution, a small PR adding it on top of #805 would be welcome.
- The rest of this branch (the `config.toml`, `docker-compose` files, and the watchdog / re-embed / benchmark scripts) is the part that made this hard to merge as one unit. If any of those are generally useful, splitting them into their own focused PRs would let us review them on their own merits.

Thanks again for the contribution.

PR #709, opened 2026-05-21
fix(openai-backend): add json_object fallback for compatible providers

liannnix:
## Problem

Many OpenAI-compatible providers (Z.AI GLM, Ollama, vLLM, Together, etc.) support `{"type": "json_object"}` but **not** the full native structured-output API — specifically `chat.completions.parse()` with a Pydantic model and `{"type": "json_schema"}`.

When Honcho's OpenAI backend calls `parse()` against these providers, it fails with `ValidationError` or `LengthFinishReasonError`. The current fallback path uses `_create_structured_response` which sends `json_schema` — these providers also ignore it and return plain text. `repair_json` cannot parse plain text, so the deriver produces **zero observations**.

This is especially problematic with reasoning models (e.g. Z.AI GLM-5-turbo) that spend tokens on `reasoning_content`, sometimes leaving `content` empty when `max_tokens` is insufficient.

## Solution

Add a `_json_object_fallback` method to `OpenAIBackend` that:

1. Uses `{"type": "json_object"}` instead of `json_schema`
2. Injects a system message with the Pydantic model's JSON schema so the model returns the correct key names
3. Parses the response with `repair_response_model_json` for robust handling of minor formatting issues

The fallback is triggered in three cases:
- `parse()` → `LengthFinishReasonError` with empty `content`
- `parse()` → `BadRequestError` / `ValidationError` / `JSONDecodeError`
- `parse()` succeeds but `parsed is None` with no refusal

## Testing

Self-hosted Honcho deployment with Z.AI GLM-5-turbo as the deriver LLM:

| Metric                  | Before   | After      |
| ----------------------- | -------- | ---------- |
| Observations per batch  | 0        | 4–12       |
| Queue processed (2 min) | ~2 items | ~150 items |

Also verified with `max_output_tokens=8000` to account for reasoning token overhead.

## Impact

- **No behavior change** for providers that support native structured output (OpenAI, Anthropic, Gemini) — the `parse()` path runs first, fallback only triggers on errors
- **Improves compatibility** with OpenAI-compatible providers out of the box
- **No new dependencies** — uses existing `repair_response_model_json` and `json` stdlib

## Release Notes

* **Bug Fixes**
  * Improved reliability of structured output handling when using OpenAI-compatible API providers. Enhanced error recovery mechanisms to gracefully handle edge cases where responses fail validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/709?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -152,28 +152,22 @@ class OpenAIBackend:
             except LengthFinishReasonError as exc:
                 truncated = exc.completion
                 raw_content = truncated.choices[0].message.content or ""
-                content = repair_response_model_json(
-                    raw_content,
-                    response_format,
-                    model,
-                )
-                return self._normalize_response(
-                    truncated,
-                    content_override=content,
+                if raw_content:
+                    content = repair_response_model_json(
+                        raw_content,
+                        response_format,
+                        model,
+                    )
+                    return self._normalize_response(
+                        truncated,
+                        content_override=content,
+                    )
+                return await self._json_object_fallback(
+                    params, response_format, model,
                 )
             except (BadRequestError, json.JSONDecodeError, ValidationError):
-                fallback_response = await self._create_structured_response(
-                    params=params,
-                    response_format=response_format,
-                )
-                content = self._parse_or_repair_structured_content(
-                    fallback_response,
-                    response_format,
-                    model,
-                )
-                return self._normalize_response(
-                    fallback_response,
-                    content_override=content,
+                return await self._json_object_fallback(
+                    params, response_format, model,
                 )
             parsed = response.choices[0].message.parsed
             raw_content = response.choices[0].message.content or ""
@@ -191,7 +185,9 @@ class OpenAIBackend:
                         response,
                         content_override=refusal,
                     )
-                raise ValidationException("No parsed content in structured response")
+                return await self._json_object_fallback(
+                    params, response_format, model,
+                )
             return self._normalize_response(
                 response,
                 content_override=validate_structured_output(parsed, response_format),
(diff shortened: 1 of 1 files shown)


akattelu:
Hey @liannnix, thanks for opening this. The bug is real: on OpenAI-compatible providers like Z.AI GLM and many Ollama/vLLM deployments that ignore or reject `json_schema`, the deriver was silently producing zero observations, and your before/after numbers on GLM-5-turbo (0 to 4-12 observations per batch) capture exactly why this class of fix matters.

We've since landed #820, which handles this at the config layer instead of in the backend. It adds a per-model-config `STRUCTURED_OUTPUT_MODE` option, so a self-hoster on one of these providers sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` and the deriver sends `{"type":"json_object"}`, injects the schema into the prompt, and parses tolerantly via `json_repair`. The prompt instruction also deliberately includes the literal word "JSON", which satisfies the Alibaba/Qwen "messages must contain 'json'" contract. Since both changes share the same root cause and remedy, I'm going to consolidate onto #820 and close this one.

The one ergonomic difference I think is genuinely worth keeping in mind is your automatic, error-triggered fallback: catch the failed `parse()` and retry transparently, so self-hosters don't need to discover the config flag at all. We went with the explicit opt-in for predictability, but the auto-detect idea is a nice follow-up if we see people tripping over the setting.

If `json_object` mode doesn't resolve it on your provider once you try #820, please reopen or ping me here. Thanks again for the careful writeup and the real-world metrics.

PR #710, opened 2026-05-21
Add Queue Work Unit Breakdown Endpoint

VVoruganti:
## Summary by CodeRabbit

* **New Features**
  * Added queue work-unit inspection for workspaces and sessions, with cursor-based pagination and per-work-unit threshold/envelope details.
  * Expanded queue status to split pending work units into **stalled** vs **ready** counts.
  * Added a new `GET /health` endpoint.
* **Bug Fixes**
  * Queue status pending counters now reliably include stalled/ready values (defaulting to `0`) and keep totals consistent.
* **Documentation**
  * Updated queue-status guides and the API overview to reflect the new stalled/ready counters and work-unit pagination behavior.


docs/v3/documentation/features/advanced/queue-status.mdx
```diff
@@ -43,6 +43,17 @@ class QueueStatus(BaseModel):
     pending_work_units: int
     """Work units waiting to be processed"""
 
+    pending_stalled_work_units: int
+    """Pending representation work units waiting to accumulate enough tokens to
+    hit DERIVER_REPRESENTATION_BATCH_MAX_TOKENS. Always 0 when
```
eisene, line 48: this will need to be changed after [this PR](https://github.com/plastic-labs/honcho/pull/889) goes in - worth stacking one PR behind the other?

eisene: other examples of this below, as well

sdks/python/src/honcho/aio.py
```diff
@@ -443,6 +445,64 @@ async def queue_status(
         )
         return QueueStatusResponse.model_validate(data)
 
+    async def queue_work_units(
```
eisene, line 448: ❯ the file `sdks/python/src/honcho/aio.py` seems to contain a lot of duplicates of similar logic, how hard is it to clean this up?

---

What the duplication actually is

1. The structural one: aio.py mirrors the sync classes almost line-for-line. Every method in HonchoAio/PeerAio/SessionAio/ConclusionScopeAio re-implements the request-building and response-parsing logic of its sync twin in client.py, peer.py, session.py, conclusions.py. Compare Honcho.peers() at client.py:344 with HonchoAio.peers() at aio.py:203 — identical except await, the client attribute, and SyncPage vs AsyncPage. That's ~1,700 async lines shadowing ~2,600 sync lines, and every new endpoint must be written twice (drift risk, not just aesthetics).

2. Intra-file repeats inside aio.py:
- The pagination boilerplate (build query, define transform, define a fetch_next closure that rebuilds the same query, wrap in AsyncPage) is copy-pasted 7 times (aio.py:203, 307, 359, 703, 1124, 1569, plus the cursor variant in queue_work_units twice).
- Summary(...) is hand-built field-by-field 3 times (aio.py:1279, 1312, 1323) — likely collapsible to Summary.model_validate(s) if the field names align.
- queue_status/queue_work_units exist in both HonchoAio and SessionAio (aio.py:422 vs 1445) differing only by a pinned session_id.
- The AsyncMetadataConfigMixin plumbing is the clearest slop: each class implements ~9 hook methods (_get_fetch_route, _parse_response, _set_metadata, …) to satisfy the mixin, then overrides most of the mixin's public methods anyway (get_metadata, get_configuration, set_configuration, refresh) to add _ensure_workspace_async and typed returns. In PeerAio/SessionAio, effectively only set_metadata still flows through the mixin — ~120 lines of hook plumbing supporting one inherited method.

How hard is cleanup?

Pass A — intra-file, low risk (a few hours): an AsyncPage closure factory, a shared queue-query builder, Summary.model_validate, and collapsing the mostly-bypassed mixin hooks. No public API change; would drop aio.py by roughly 350–450 lines. This is safe, mechanical deletion-first work.

Pass B — sync/async unification, medium effort (about a day): extract pure "request spec" builders (route + body + query construction, response parsing) shared by both transports, leaving sync/async methods as thin await-or-not shells. Touches ~6 files but is mechanical; kills the write-everything-twice problem. The alternative is unasync-style codegen (write async once, generate sync), but this SDK is inverted — sync classes are primary and .aio wraps them — so that'd be a bigger architectural decision, not a cleanup.

Risk caveat (this is what makes it "medium" rather than "easy"): there are no unit tests under sdks/python/ — the aio surface is exercised only by the unified integration runner (tests/unified/runner.py), which needs a live server + DB + Redis. Behavior lock for Pass A is achievable with basedpyright + a unified run; Pass B should add a thin unit layer (mock HTTP client, assert route/body/query per method) before refactoring, which adds maybe half a day.

Verdict: Pass A is easy and worth doing anytime. Pass B is moderately hard — not because the code is subtle, but because the regression net is integration-only. Recommended follow-up order: (1) add mocked request-shape unit tests, (2) Pass A deletions, (3) shared request-builder extraction. Say the word and I'll start with Pass A (or the tests first, if you want Pass B).

---

❯ can tests be added prior to pass B to derisk?

---

Yes — and the seams to do it cheaply already exist, so this doesn't require any refactoring of the SDK first. What I verified:

- Both HonchoHTTPClient and AsyncHonchoHTTPClient accept an injected httpx.Client / httpx.AsyncClient (http/client.py:46, http/async_client.py:46). That means tests can use httpx.MockTransport — a fake transport that records every request and returns canned JSON — with no live server, no DB, no monkeypatching.
- Honcho.__init__ already exposes http_client= for the sync side. The async client is lazily built from _http_config (client.py:153), which doesn't accept an async client today, so tests would set honcho._async_http directly (private-attr poke in a test fixture — fine) or we add a small async_http_client= passthrough as part of the test PR.

The test design that actually de-risks pass B is a wire-shape parity table: one entry per SDK method with the expected (HTTP method, route, query, body), and each entry executed twice — once through the sync method, once through .aio — asserting both produce the identical request and parse the canned response into the same SDK object. That table encodes exactly the invariant pass B must preserve ("same wire behavior, shared builder underneath"), so after the refactor the same table keeps passing unchanged. Pagination gets a couple of extra assertions (calling fetch_next sends page=2; the cursor variant sends cursor=...).

Two mechanics to account for:

1. Almost every method calls _ensure_workspace_async() first, which POSTs to the workspaces route — the mock transport handles it, and the tests should assert it happens exactly once (it's cached via _workspace_ensured), which locks that behavior too.
2. There's no sdks/python/tests/ today and the SDK's dev deps are just ruff — but the root uv environment already imports the SDK (the unified runner does from honcho import ...), so a sdks/python/tests/test_request_shapes.py runnable via root uv run pytest is the low-friction placement, mirroring how the TS SDK tests are driven from root pytest.

Effort: roughly half a day. The surface is ~60 methods but the tests are table-driven, so most of it is transcribing expected routes/bodies from the current code — which doubles as a manual audit of the exact drift pass B is meant to eliminate.

sdks/python/src/honcho/pagination.py
```diff
@@ -248,3 +255,234 @@ async def get_next_page(self) -> "AsyncPage[T, U] | None":
             return None
 
         return await self._fetch_next(current_page + 1)
+
+
+class SyncCursorPage(Generic[T, U]):
```
eisene, line 260: worth adding some of these tests? See Claude's analysis below

---

What's not covered

- The stale-cursor case: no test mints a next_page token, then marks the remaining items processed (or deletes them) before fetching with that token. The "you just get a blank page and next_page: None" behavior we walked through is currently relying on fastapi-pagination/sqlakeyset semantics, not asserted anywhere.
- The mutable-sort-key caveat: nothing exercises a work unit's oldest_item_at shifting across a cursor boundary (the duplicate/skip case).
- Backwards navigation: previous_page / fetch_previous is wired in the SDK but never traversed in a test.
- SyncCursorPage unit tests: tests/sdk/test_pagination.py only covers the offset-based SyncPage/AsyncPage (page-number get_next_page, transform preservation, etc.) — there are no unit tests for the cursor page class itself, including its auto-chasing __iter__.

So your degradation question found a genuine gap: the behavior is sound by construction (keyset comparison against a value tuple can't dangle), but a regression — say, someone changing the query so an empty result 500s, or the SDK mishandling an empty items — wouldn't be caught for the stale-cursor path specifically. A server-side test that gets page 1 of 3 with size=2, marks everything processed, then fetches with the stale body1["next_page"] and asserts items == [] / next_page is None would close the main gap cheaply.

sdks/python/src/honcho/aio.py
eisene: there's just generally kind of a lot of boilerplate in these clients and routes, might be worth a refactor at some point

sdks/python/src/honcho/pagination.py
```diff
@@ -248,3 +255,234 @@ async def get_next_page(self) -> "AsyncPage[T, U] | None":
...
+    def items(self) -> list[U] | list[T]:
+        """Get all transformed items on the current page (snapshot only)."""
+        if self._transform_func is not None:
+            return [self._transform_func(item) for item in self._raw_items]
+        return list(self._raw_items)
+
+    @property
+    def total(self) -> int | None:
+        """Total items across all pages, when the server populates it."""
+        return self._data.get("total")
```
eisene, line 348: i believe it is impossible for `total`, `current_page` and `current_page_backwards` to not be present, it doesn't seem necessary to allow a `None` return type for those. Would tighten up the static signal from pylance.

This seems to be defensive coding against FastAPI pagination schemas changing. I would consider throwing an exception instead instead of letting these return `None`.

sdks/typescript/src/pagination.ts
```diff
@@ -181,3 +193,200 @@ export class Page<T, TOriginal = T> implements AsyncIterable<T> {
...
+ * Cursor-paginated result wrapper. Uses opaque tokens (`nextPage`/`previousPage`)
+ * instead of page numbers. Stable across concurrent mutations of the
+ * underlying data, unlike offset pagination.
+ *
+ * **Warning:** The underlying server data may still mutate between page fetches.
+ * Iterating across pages snapshots what the server returns per page, but pages
+ * may be inconsistent with each other under concurrent processing. Use
+ * `.items` to read just the current page if you need a stable view.
+ */
+export class CursorPage<T, TOriginal = T> implements AsyncIterable<T> {
```
eisene, line 207: I believe both comments on the Python version apply here as well

src/crud/deriver.py
```diff
@@ -2,15 +2,23 @@
...
 from src import models, schemas
+from src.config import settings
 
 logger = getLogger(__name__)
 
 
+# Task types surfaced by the queue status endpoint.
+_TRACKED_TASK_TYPES = ("representation", "summary", "dream")
+
+# Only this task type is gated by DERIVER_REPRESENTATION_BATCH_MAX_TOKENS.
```
eisene, line 18: another instance of changes after my PR, and more in this file

src/crud/deriver.py
```diff
@@ -86,56 +147,182 @@ def _build_queue_status_query(
...
+            models.ActiveQueueSession,
+            models.QueueItem.work_unit_key == models.ActiveQueueSession.work_unit_key,
+        )
+        .outerjoin(
+            models.Message,
+            models.QueueItem.message_id == models.Message.id,
+        )
+    )
+
+    inner = inner.where(models.QueueItem.workspace_name == workspace_name)
```
eisene, line 194: nit: why not combine these with the definition of `inner`?

src/crud/deriver.py
```diff
@@ -2,15 +2,23 @@
 from logging import getLogger
```
eisene, line 2: this logic is complex, I would consider making unit tests for this stuff

eisene: Claude's analysis:

---

Five of the ten src/crud/ modules have no counterpart in tests/crud/: deriver.py (as we established — only via /queue/* route tests), message.py, peer.py, collection.py, and webhook.py. Those are exercised through the corresponding files in tests/routes/ (test_messages.py, test_peers.py, test_webhooks.py, etc.), which is real coverage but black-box: it verifies the HTTP contract, and internal branches only get hit if some route scenario happens to reach them — the same "counters never nonzero" blind spot we found in deriver.py.

Beyond those two layers, crud functions also get incidental exercise from tests/integration/ (enqueue, representation, embeddings), tests/deriver/, and tests/dreamer/, and several test conftests use crud as setup helpers — but that's scaffolding, not coverage you should count on.

So the honest summary: the repo clearly has the convention of direct crud tests (tests/crud/ mirrors src/crud/), but it's been applied selectively — roughly, modules with subtle internal logic that's awkward to reach via HTTP (documents/vector storage, peer cards, the representation manager) got direct tests, while modules whose behavior is mostly "what the route does" (peers, messages, webhooks) lean on route tests.

eisene: I would consider making direct unit tests for this - it's cheap and sends a clean signal that's easy to understand.

src/crud/deriver.py
```diff
@@ -86,56 +147,182 @@ def _build_queue_status_query(
...
-    stmt = stmt.outerjoin(
-        models.ActiveQueueSession,
-        models.QueueItem.work_unit_key == models.ActiveQueueSession.work_unit_key,
+        func.count(case((is_stalled, 1)))
+        .over(partition_by=inner_subq.c.session_id)
+        .label("session_pending_stalled"),
+        func.count(case((is_pending_ready, 1)))
+        .over(partition_by=inner_subq.c.session_id)
+        .label("session_pending_ready"),
     )
```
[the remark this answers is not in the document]
eisene: this complaint appears to just be about the use of the words "work units" instead of something like "queued items" in the docstring for `get_queue_status` and the API field names

eisene (commented):
a few minor comments

eisene:
is this safe to close?

VVoruganti:
Yeah i think this work has gone stale enough where it's worth just throwing it away and restarting later

VVoruganti:
Closing for staleness

PR #720, opened 2026-05-23
fix: flush session before upserting session_peers to prevent FK violation

TipKnuckle:
## Description

With `autoflush=False` in `src/db.py`, Core-level SQL statements like `pg_insert` do not automatically flush pending ORM objects before executing. In `get_or_create_session`, `db.add(honcho_session)` marks the session as pending, but when `_get_or_add_peers_to_session` executes the `pg_insert` upsert on `session_peers` (which has a foreign key to `sessions`), the session row doesn't exist in PostgreSQL yet, causing a `ForeignKeyViolation`.

## Fix

Added `await db.flush()` after the session creation and before the peer upsert, ensuring the session row exists in the database before the FK-dependent upsert executes.

## Testing

Verified working with the OWUI sync script that was previously hitting this error (500 from server after a few imports).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition during session creation that could cause related participant/peer records to fail to attach reliably, improving session setup consistency and reducing intermittent errors when joining or initializing sessions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/720?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/crud/session.py
+++ b/src/crud/session.py
@@ -226,11 +226,12 @@ async def get_or_create_session(
         try:
             async with db.begin_nested():
                 db.add(honcho_session)
-            # Flush so the session row exists in PostgreSQL before
-            # _get_or_add_peers_to_session tries to upsert session_peers
-            # (which has a FK to sessions). Without this flush, the Core-level
-            # pg_insert executes before the ORM flush (autoflush=False in db.py).
-            await db.flush()
+                # Flush inside the savepoint so any IntegrityError (race condition)
+                # is confined to the nested transaction and the retry logic stays valid.
+                # Without this flush, the Core-level pg_insert in
+                # _get_or_add_peers_to_session executes before the ORM flush
+                # (autoflush=False in db.py), causing a FK violation on session_peers.
+                await db.flush()
             needs_cache_update = True
             created = True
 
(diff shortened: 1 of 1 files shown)


TipKnuckle:
Good catch — moved `db.flush()` inside the `begin_nested()` context in 4b85fcb.

VVoruganti:
Thanks for the careful root-cause @TipKnuckle — the FK-violation chain is real, but it's already prevented on main. get_or_create_session adds the new session, and the pending row is flushed before the session_peers upsert runs — without needing this explicit flush.

Per the SQLAlchemy docs (Using SAVEPOINT, https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#using-savepoint): "When Session.begin_nested() is called, the Session first flushes all currently pending state to the database; this occurs unconditionally, regardless of the value of the Session.autoflush parameter." Two begin_nested savepoints cover this in get_or_create_session: get_or_create_peers (called before the upsert) opens one, and the session's own savepoint (#399, for concurrent-create handling) is another — either flushes the session before the session_peers FK insert.

So the explicit db.flush() here is redundant with documented behavior. Closing as already-resolved. You most likely hit this on a build predating #399's savepoint; if you can still reproduce a ForeignKeyViolation on main, please reopen with the Honcho version and we'll dig in — that'd point to a different gap.

TipKnuckle:
Yeah this was awhile ago before 399. Thanks.

PR #730, opened 2026-05-27
Fix delete_observations crash with stringified observation_ids

### LINKED ISSUE #719: [Bug] delete_observations can crash when observation_ids is a stringified list
## Summary

`delete_observations` can crash during agentic dream/dialectic tool execution when the LLM/tool-call payload provides `observation_ids` as a string or stringified list instead of an actual JSON array.

Example problematic payload shapes seen from tool-call output:

```json
{"observation_ids": "[doc_abc123]"}
```

or:

```json
{"observation_ids": "[id:doc_abc123]"}
```

The schema asks for `list[str]`, but in practice model/tool-call output can still produce list-ish strings. When that string reaches the delete path, SQLAlchemy's `.in_()` handling rejects it because it is not an iterable/list of IDs in the expected form.

I searched existing open issues for `delete_observations`, `observation_ids`, `Column.in_`, `IN expression list`, `ArgumentError observation_ids`, and related terms. I found adjacent deriver/dream/provider robustness issues, but not this exact `delete_observations` input-normalization failure.

## Impact

A malformed-but-plausible `delete_observations` tool argument can cause a dream/dialectic/tool loop to fail instead of gracefully returning a tool error or normalizing the argument.

This is especially likely in agentic loops where the model is copying IDs from earlier tool output such as `[id:...]` strings or bracketed ID lists.

## Sanitized error signature

The failure is in the delete-observations path when a string/list-ish value is passed toward the CRUD delete operation and ultimately into SQLAlchemy's `in_()` expression.

Representative sanitized error shape:

```text
sqlalchemy.exc.ArgumentError: IN expression list, SELECT construct, or bound parameter object expected, got '[doc_abc123]'
```

No real observation IDs, workspace names, peer names, message contents, or deployment identifiers are included here.

## Minimal reproducer shape

At the tool-handler level, a payload shaped like this should be enough to exercise the edge case:

```python
await _handle_delete_observations(
    ctx,
    {"observation_ids": "[doc_abc123]"},
)
```

Expected behavior:

- The handler should either normalize the value to `['doc_abc123']` and proceed, or reject it with a controlled `ERROR: ...` tool response.
- It should not pass a raw string into a DB `.in_()` path and crash the loop.

## Local mitigation tested

A local hotfix was tested successfully by normalizing `observation_ids` in `_handle_delete_observations` before calling the CRUD layer.

The mitigation approach:

- Accept `None` as an empty list.
- Accept an actual list/tuple/set as list input.
- If a string is provided:
  - trim whitespace,
  - try `json.loads()` first,
  - if that fails, treat it as a single/list-ish string,
  - strip surrounding brackets,
  - split simple comma/whitespace-separated IDs,
  - strip quotes/brackets,
  - strip a leading `id:` prefix if present,
  - validate IDs against a conservative ID charset such as `[A-Za-z0-9_-]+`,
  - deduplicate while preserving order.
- If the normalized list is empty, return the existing controlled empty-list error.

A regression test was added locally for this shape:

```python
result = await _handle_delete_observations(
    ctx,
    {"observation_ids": f"[{doc_id}]"},
)
assert "Deleted 1 observations" in result
```

That test passed locally against the self-hosted source tree after the fix.

## Notes

This issue is intentionally filed as a report rather than a PR because there may be broader design expectations around how strict tool argument validation should be in agentic dream/dialectic loops. If others are seeing similar failures, it would be useful to collect those cases here before deciding whether the best fix belongs in:

- the `delete_observations` handler only,
- a shared tool-argument normalization layer,
- stricter schema/tool-call validation before dispatch,
- or the lower-level CRUD boundary.

## Environment, generalized

- Self-hosted Honcho Docker deployment
- Recent v3.0.x source build
- Failure occurred in an agentic dream/dialectic/tool execution path
- Backend database: Postgres/SQLAlchemy path

I omitted deployment-specific names, observation content, real IDs, peer/workspace names, hostnames, and IPs.

thakoreh:
## Problem

`delete_observations` crashes when LLM tool-call output sends `observation_ids` as a string or stringified list instead of a proper JSON array. This triggers `sqlalchemy.exc.ArgumentError` when the raw string hits `.in_()`.

Reproducer: any agentic dream/dialectic loop where the model copies IDs from earlier tool output in bracketed or `id:` prefixed format.

Closes #719

## Fix

Added `_normalize_observation_ids()` helper that normalizes input before the CRUD call:

- **None** → empty list (returns controlled error)
- **list/tuple/set** → pass through
- **JSON string** (e.g. `"[\"abc\"]"`) → parse and extract
- **Bracketed string** (e.g. `"[abc123]"`) → strip brackets, split
- **id: prefix** (e.g. `"[id:abc123]"`) → stripped
- **Comma/whitespace separated** → split
- **Validation** → conservative `[A-Za-z0-9_-]+` charset filter
- **Dedup** → preserves order, removes duplicates

## Tests

12 unit tests covering: None, empty string, list passthrough, JSON string, bracketed single ID, id: prefix, comma-separated, deduplication, invalid character filtering, tuple/set, single string, quoted+bracketed IDs.

## Files changed

- `src/utils/agent_tools.py` — added `_normalize_observation_ids()`, updated `_handle_delete_observations` to use it
- `tests/utils/test_agent_tools.py` — added `TestNormalizeObservationIds` class

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced robustness of the observation deletion tool to handle multiple input formats for observation IDs, including JSON-encoded strings, bracketed formats, and comma-separated values.
  * Added validation and deduplication of observation IDs.

* **Tests**
  * Added comprehensive test coverage for observation ID normalization across various input formats and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/730?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -1,4 +1,6 @@
 import asyncio
+import json
+import re
 import logging
 import weakref
 from collections.abc import Callable
@@ -2064,11 +2066,59 @@ async def _handle_get_peer_card(ctx: ToolContext, tool_input: dict[str, Any]) ->
     )
 
 
+
+def _normalize_observation_ids(raw: Any) -> list[str]:
+    """Normalize observation_ids to a clean list of strings.
+
+    Handles cases where LLM tool-call output provides observation_ids as
+    a string or stringified list instead of a proper JSON array.
+    See: https://github.com/plastic-labs/honcho/issues/719
+    """
+    if raw is None:
+        return []
+    if isinstance(raw, list | tuple | set):
+        ids = list(raw)
+    elif isinstance(raw, str):
+        s = raw.strip()
+        if not s:
+            return []
+        # Try JSON parse first (handles '["a","b"]')
+        try:
+            parsed = json.loads(s)
+            if isinstance(parsed, list):
+                ids = parsed
+            else:
+                ids = [str(parsed)]
+        except (json.JSONDecodeError, ValueError):
+            # Strip surrounding brackets, split by comma/whitespace
+            s = s.strip("[]")
+            ids = re.split(r"[,\s]+", s) if s else []
+    else:
+        ids = [str(raw)]
+
+    cleaned: list[str] = []
+    seen: set[str] = set()
+    for item in ids:
+        if not isinstance(item, str):
+            item = str(item)
+        # Strip quotes and brackets
+        item = item.strip().strip('"\'[]')
+        # Strip leading "id:" prefix
+        item = re.sub(r"^id:", "", item)
+        # Validate against conservative charset
+        if item and re.fullmatch(r"[A-Za-z0-9_-]+", item) and item not in seen:
+            cleaned.append(item)
+            seen.add(item)
+    return cleaned
+
+
 async def _handle_delete_observations(
     ctx: ToolContext, tool_input: dict[str, Any]
 ) -> "str | ToolResult":
     """Handle delete_observations tool."""
-    observation_ids = tool_input.get("observation_ids", [])
+    observation_ids = _normalize_observation_ids(
+        tool_input.get("observation_ids")
+    )
     if not observation_ids:
         return "ERROR: observation_ids list is empty"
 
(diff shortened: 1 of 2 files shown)


thakoreh:
Closing this while I pause open-source contributions. Apologies for the noise.

PR #746, opened 2026-05-29
fix(agent_tools): normalize stringified observation_ids in delete_observations (#719)

### LINKED ISSUE #719: [Bug] delete_observations can crash when observation_ids is a stringified list

KennyUMN:
## Problem

`_handle_delete_observations` (`src/utils/agent_tools.py`) passes `tool_input["observation_ids"]` straight into `crud.delete_documents(document_ids=...)`, which feeds SQLAlchemy's `.in_()`. The tool schema asks for `list[str]`, but in agentic dream/dialectic loops the model/tool-call output sometimes emits a **stringified list** instead of a real array:

```json
{"observation_ids": "[doc_abc123]"}
{"observation_ids": "[id:doc_abc123]"}
```

When a bare string reaches `.in_()` it either raises or silently iterates character-by-character, crashing (or corrupting) the tool loop. This is especially likely when the model copies IDs from earlier tool output formatted as `[id:...]`.

Fixes #719.

## Fix

Add `_normalize_observation_ids()` to coerce the argument into `list[str]` before it reaches the delete path. It handles:

- real `list`/`tuple`/`set` (unchanged passthrough)
- JSON-encoded array strings (`'["a","b"]'`)
- bracketed non-JSON strings (`"[a]"`, `"[a, b]"`)
- bare single-id strings (`"a"`)
- `id:` prefixes and stray quotes/brackets
- empty / `None` -> `[]`

The handler then calls `_normalize_observation_ids(tool_input.get("observation_ids", []))`. Real-list callers are unaffected.

## Tests

- `TestNormalizeObservationIds`: parametrized unit tests covering all input shapes above (including the two exact payloads from the issue).
- `TestDeleteObservationsStringifiedIds`: handler-level regression test asserting a `"[<doc_id>]"` string soft-deletes the same row a real list would.

Both files pass `python -m py_compile`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced observation deletion to reliably accept and process observation IDs in multiple formats, improving robustness when handling edge cases, malformed inputs, and various string representations.

* **Tests**
  * Added comprehensive test coverage for observation ID parsing and deletion operations, validating support for various input formats and soft-deletion behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/746?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import logging
 import weakref
 from collections.abc import Callable
@@ -2064,11 +2065,56 @@ async def _handle_get_peer_card(ctx: ToolContext, tool_input: dict[str, Any]) ->
     )
 
 
+def _normalize_observation_ids(raw: Any) -> list[str]:
+    """Coerce a delete_observations ``observation_ids`` argument into ``list[str]``.
+
+    The tool schema asks for ``list[str]``, but LLM/tool-call output sometimes
+    provides a stringified list (e.g. ``"[doc_abc123]"``, ``"[id:doc_abc123]"``,
+    or a JSON-encoded array). A bare string passed into SQLAlchemy's ``.in_()``
+    either raises or silently iterates character-by-character, so it is
+    normalized here before reaching the delete path. See issue #719.
+    """
+    if raw is None:
+        return []
+    if isinstance(raw, (list, tuple, set)):
+        items: list[Any] = list(raw)
+    elif isinstance(raw, str):
+        text = raw.strip()
+        if not text:
+            return []
+        try:
+            parsed = json.loads(text)
+        except (ValueError, TypeError):
+            parsed = None
+        if isinstance(parsed, list):
+            items = parsed
+        elif isinstance(parsed, str):
+            items = [parsed]
+        else:
+            inner = text
+            if inner.startswith("[") and inner.endswith("]"):
+                inner = inner[1:-1]
+            items = list(inner.split(","))
+    else:
+        items = [raw]
+
+    normalized: list[str] = []
+    for item in items:
+        if not isinstance(item, str):
+            item = str(item)
+        cleaned = item.strip().strip("\"'").strip()
+        if cleaned.lower().startswith("id:"):
+            cleaned = cleaned[3:].strip()
+        if cleaned:
+            normalized.append(cleaned)
+    return normalized
+
+
 async def _handle_delete_observations(
     ctx: ToolContext, tool_input: dict[str, Any]
 ) -> "str | ToolResult":
     """Handle delete_observations tool."""
-    observation_ids = tool_input.get("observation_ids", [])
+    observation_ids = _normalize_observation_ids(tool_input.get("observation_ids", []))
     if not observation_ids:
         return "ERROR: observation_ids list is empty"
 
--- a/tests/utils/test_agent_tools.py
+++ b/tests/utils/test_agent_tools.py
@@ -21,6 +21,7 @@ from src.utils.agent_tools import (
     ToolContext,
     _handle_create_observations,  # pyright: ignore[reportPrivateUsage]
     _handle_delete_observations,  # pyright: ignore[reportPrivateUsage]
+    _normalize_observation_ids,  # pyright: ignore[reportPrivateUsage]
     _handle_extract_preferences,  # pyright: ignore[reportPrivateUsage]
     _handle_finish_consolidation,  # pyright: ignore[reportPrivateUsage]
     _handle_get_messages_by_date_range,  # pyright: ignore[reportPrivateUsage]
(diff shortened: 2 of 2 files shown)


TcDrozd:
Production data point supporting this fix, specifically the `id:` prefix stripping:

In our self-hosted deployment (dream specialists on a local 14B model via Ollama), we hit a variant of #719 where the model passes a well-formed `list[str]` — so no crash — but copies the display format verbatim from `[id:xxx]` tool output, calling e.g. `delete_observations({"observation_ids": ["id:3-bwp1hxCRkRbUh_nrqn0", ...]})` despite the schema description saying to pass the bare ID. Over a 24h window, **16 of 16 failed deletes had the `id:` prefix** (zero other failure causes); each one silently no-ops with `Failed to delete observation id:...: Document ... not found`, so the model believes it cleaned up observations that are still there.

Two notes that may be useful while you're in here:

1. The same model behavior corrupts other ID inputs that this PR doesn't cover: `source_ids` on `create_observations_*` (we've seen `source_ids: ['id:wIXfBLbxZzZtoZPrZi6Dx', ...]` — stored silently as broken provenance links in document metadata) and `observation_id` on `get_reasoning_chain`. Applying `_normalize_observation_ids` (or a scalar variant) at those entry points would close the gap. Happy to follow up with a complementary PR for those paths if that's easier than expanding this one.

2. Since nanoid's default alphabet includes `-` and `_`, IDs like `3-bwp...` or trailing-`_` are legitimate — so normalization beyond prefix/quote stripping (e.g. trimming `-`/`_`) would risk mangling real IDs. This PR gets that right; just flagging it for any future reviewer tempted to strip more aggressively.

VVoruganti:
Heads-up: #795 just merged and added a `_normalize_observation_id` helper in `agent_tools.py` for the same `id:`-prefix problem (covering `source_ids` on `create_observations_*` and `observation_id` on `get_reasoning_chain`). To avoid two near-identical helpers, this PR should reuse that one rather than introducing a separate `_normalize_observation_ids` — e.g. map the existing singular helper over the `observation_ids` list. Worth a quick rebase onto main first.

gcs8:
Additional verification after `v3.0.11`:

- The `v3.0.11` tag (`60a15e6`) still has `_handle_delete_observations()` reading `tool_input.get("observation_ids", [])` directly before calling the delete path.
- The merged #795 helper covers `source_ids` / `get_reasoning_chain` display-format `id:` prefixes, but does not normalize `delete_observations.observation_ids`.
- A local/self-hosted `v3.0.11` image with an equivalent normalizer patch smoke-tested the relevant payload shapes successfully:
  - `[doc_abc123]` -> `['doc_abc123']`
  - `[id:doc_abc123]` -> `['doc_abc123']`
  - `["doc_abc123", "doc_def456"]` -> `['doc_abc123', 'doc_def456']`

So #719 is still not fixed by the `v3.0.11` release/tag; this PR (or an equivalent reuse/extension of the #795 normalization helper for `observation_ids`) is still needed.

VVoruganti:
## Review: still relevant, but rework to reuse the merged helper

The core bug this targets is **still unfixed on `main`** — `_handle_delete_observations` remains `tool_input.get("observation_ids", [])` with no normalization, so a stringified list like `"[doc_abc123]"` still reaches SQLAlchemy `.in_()` and crashes/iterates char-by-char. So this PR's central purpose is not stale.

However, after this PR was opened, **#795 merged** and added `_normalize_observation_id` (**singular**) for the same `[id:xxx]` display-prefix problem — applied to `source_ids` and `get_reasoning_chain`, but **not** to the delete handler. That creates overlap:

| Concern | On `main` today | This PR |
|---|---|---|
| Strip `id:` prefix from a single ID | ✅ `_normalize_observation_id` | 🔁 re-implements it |
| Delete handler normalizes at all | ❌ | ✅ |
| **Stringified whole-list `"[...]"` → items** | ❌ | ✅ (the real value-add) |

The genuinely new capability here is **splitting a stringified list into items**; the per-item `id:`/quote stripping duplicates the helper #795 already landed.

### Suggested rework
1. Rebase on `main` (branch is ~65 commits behind).
2. Delegate per-item cleaning to the existing `_normalize_observation_id` instead of reimplementing it — keep only the list-splitting:

```python
def _normalize_observation_ids(raw: Any) -> list[str]:
    if raw is None:
        return []
    if isinstance(raw, (list, tuple, set)):
        items = list(raw)
    elif isinstance(raw, str):
        text = raw.strip()
        if not text:
            return []
        try:
            parsed = json.loads(text)
        except (ValueError, TypeError):
            parsed = text[1:-1] if text.startswith("[") and text.endswith("]") else text
        items = parsed if isinstance(parsed, list) else str(parsed).split(",")
    else:
        items = [raw]
    return [nid for i in items if (nid := _normalize_observation_id(str(i).strip().strip("\"'")))]
```

Same behavior, no duplicated prefix logic, reuses #795.

(For reference: #730 was an earlier duplicate of this fix, closed unmerged.)

akattelu:
Closing this as per comment in https://github.com/plastic-labs/honcho/issues/719

PR #751, opened 2026-05-31
fix(openai-backend): fallback json_schema→json_object for providers without structured output support

### LINKED ISSUE #716: [Bug] Ollama/OpenAI-compatible providers: deriver silently produces zero observations (reasoning-content fallback + json_schema ignored)
*Type:** Bug

Disclaimer: I am not a developer; this is an issue I ran into using ollama cloud with the honcho deriver failing, and my hermes agent was able to resolve by editing the honcho code. This is the writeup that my agent produced, in case its helpful.  It seems to be at least partially an issue with how ollama cloud structures its output, so perhaps from the honcho side it should be considered more of a compensation rather than a bug in honcho.  If it is a sin to post ai-generated bug reports then I apologize.

**Environment:** Self-hosted Honcho with Ollama Cloud as the
OpenAI-compatible provider, models `glm-5.1:cloud` and
`deepseek-v4-flash:cloud`

## Description

When using Ollama (Cloud or local) as an OpenAI-compatible provider
for the deriver, two independent bugs cause the deriver to silently
produce zero observations from message batches — no errors, no
warnings, just empty `PromptRepresentation(explicit=[])`.

Both bugs share the same blast radius: any self-hosted deployment
using reasoning models or Ollama-based providers will see deriver
output that appears to work (no crashes, no exceptions) but actually
produces nothing.

---

## Bug 1: Reasoning models return content in `reasoning` field, not
`message.content`

**Affected method:** `_parse_or_repair_structured_content()` in
`src/llm/backends/openai.py`

Reasoning models served via Ollama (e.g. `glm-5.1:cloud`) return their
output in the `reasoning` field of the response message instead of
`message.content` when using `json_schema` response format. The
current code only checks `message.content`:

```python
if raw_content:
    return repair_response_model_json(raw_content, response_format, model)
if refusal:
    return refusal
raise ValidationException("No raw content available for structured
output repair")
```

When `message.content` is empty (because the content landed in
`reasoning`), this falls through to the `ValidationException`. The
retry logic in `clients.py` catches this and retries, but every
attempt hits the same wall. The deriver logs show successful LLM calls
but zero observations stored.

The file already has `extract_openai_reasoning_content()` used in the
streaming path — the fix is to reuse it in the non-streaming path as a
fallback:

```python
if not raw_content:
    reasoning = extract_openai_reasoning_content(response) or ""
    if reasoning:
        raw_content = reasoning
if raw_content:
    return repair_response_model_json(raw_content, response_format, model)
```

This is distinct from #676 (which covers missing `choices`/`usage`
causing crashes) — here the response object is well-formed, just with
content in an unexpected field.

---

## Bug 2: Ollama ignores `json_schema` response format, prompt
reinforcement needed

**Affected file:** `src/deriver/prompts.py`, function `minimal_deriver_prompt`

Ollama Cloud models do not reliably enforce `json_schema` response
format. When the deriver sends a `json_schema` constraint, Ollama
often ignores it and returns JSON with its own key names (e.g.
`{"name": "John", "age": 35}` instead of `{"explicit": [{"content":
"...", "salience": N}]}`).

The `repair_response_model_json` function then successfully parses
this into a valid `PromptRepresentation(explicit=[])` — schema-valid
but empty. No crash, no exception, no retry. Just silently zero
observations.

This is the same class of problem as #453 / #663 (Qwen requiring
"json" keyword for `json_object` mode), but for `json_schema` mode
specifically, and the fix is at the prompt level rather than the API
call level. Adding explicit JSON output format instructions to the
deriver prompt ensures the model produces the correct schema by
instruction, not just by API constraint:

```
OUTPUT FORMAT: Return a JSON object with a single key "explicit"
containing an array of objects. Each object must have a "content"
field (the observation text) and a "salience" field (integer 1-10
indicating importance). Example: {"explicit": [{"content": "John is 25
years old", "salience": 7}, {"content": "John has a dog", "salience":
5}]}

EXAMPLES:
- EXPLICIT: "I just had my 25th birthday last Saturday" → {"explicit":
[{"content": "{peer_id} is 25 years old", "salience": 7}, {"content":
"{peer_id}'s birthday is June 21st", "salience": 5}]}
- EXPLICIT: "I took my dog for a walk in NYC" → {"explicit":
[{"content": "{peer_id} has a dog", "salience": 6}, {"content":
"{peer_id} lives in NYC", "salience": 7}]}
```

The examples section also changes from plain text to JSON-shaped
examples, which reinforces the schema in the model's output.

---

## Reproduction

1. Configure Honcho to use Ollama Cloud as a custom OpenAI-compatible provider
2. Set the deriver model to `glm-5.1:cloud` or `deepseek-v4-flash:cloud`
3. Send messages and watch the deriver process them
4. Check conclusions: empty despite successful LLM call logs

## Expected behavior

The deriver should extract observations from message batches, falling
back to the `reasoning` field when `message.content` is empty, and
producing the correct `PromptRepresentation` schema regardless of
whether the provider enforces `json_schema` natively.

## Workaround

Both patches must be manually applied to `src/llm/backends/openai.py`
and `src/deriver/prompts.py` and re-applied after each `git pull` that
touches those files.

## Related issues

- #676 — OpenAI-compatible responses with missing `choices`/`usage`
(adjacent robustness issue, different root cause)
- #524 — PromptRepresentation should tolerate string-valued explicit
observations (schema normalization, not prompt reinforcement)
- #453 / #663 — Qwen providers requiring "json" keyword for
`json_object` mode (same class of problem: OpenAI-compatible providers
with JSON mode quirks, but different provider and different mode)

tgtjam:
## Problem

Honcho uses `openai.chat.completions.parse()` for structured output, which sends `response_format` as a Pydantic model. The OpenAI SDK converts this to `{"type": "json_schema", "json_schema": {...}}`.

DeepSeek, Ollama, vLLM, Together, Qwen, and most OpenAI-compatible providers **do not support** `response_format: {"type": "json_schema"}`. Only native OpenAI models support this.

Result: all deriver tasks requiring structured output fail with `400 BadRequestError: "This response_format type is unavailable now"`. The deriver silently produces zero observations on these providers.

Closes #716.

## Fix

When `_create_structured_response()` gets a `BadRequestError` from `json_schema`, it falls back to:
1. `response_format: {"type": "json_object"}`
2. Injects the JSON Schema into the last user message as a prompt instruction

This is transparent — no config changes needed, no breaking changes. Providers that support `json_schema` (native OpenAI) keep working as before.

## Testing

- ✅ Fallback triggers correctly: `json_schema not supported by provider, falling back to json_object for PromptRepresentation`
- ✅ Queue processes without 400 errors
- ✅ 6 embeddings generated successfully after fix
- ✅ All existing tests pass (164 tests)
- ✅ No regressions for native OpenAI flows (json_schema attempted first)

## References
- Issue #716: Ollama/OpenAI-compatible providers deriver silently produces zero observations
- PR #670: Previous attempt (closed) — feat(openai): extra_body passthrough + json_object default

## Summary by CodeRabbit

* **New Features**
  * Improved structured-response handling by embedding schema instructions to produce stricter JSON outputs and auto-parsing/repairing responses.

* **Bug Fixes**
  * Added a provider-fallback to ensure structured responses when strict schema formats are rejected, improving compatibility and reliability.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -142,6 +142,31 @@ class OpenAIBackend:
         )
 
         if isinstance(response_format, type):
+            if extra_params and extra_params.get("json_mode"):
+                # Some providers (e.g. DeepSeek, Ollama, vLLM) don't support
+                # json_schema structured output. Skip directly to json_object
+                # with the schema injected into the prompt.
+                params["response_format"] = {"type": "json_object"}
+                schema_json = response_format.model_json_schema()
+                schema_prompt = (
+                    f"You MUST respond with valid JSON that conforms to this JSON Schema:\n"
+                    f"{json.dumps(schema_json, indent=2)}\n\n"
+                    f"Return ONLY valid JSON, no markdown, no explanation."
+                )
+                messages = list(params.get("messages", []))
+                if messages and messages[-1]["role"] == "user":
+                    messages[-1]["content"] = (
+                        str(messages[-1]["content"]) + "\n\n" + schema_prompt
+                    )
+                else:
+                    messages.append({"role": "user", "content": schema_prompt})
+                params["messages"] = messages
+                response = await self._client.chat.completions.create(**params)
+                content = self._parse_or_repair_structured_content(
+                    response, response_format, model,
+                )
+                return self._normalize_response(response, content_override=content)
+
             params["response_format"] = response_format
             try:
                 response = await self._client.chat.completions.parse(**params)
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -381,14 +381,40 @@ class OpenAIBackend:
         response_format: type[BaseModel],
     ) -> Any:
         structured_params = dict(params)
+        schema_json = response_format.model_json_schema()
         structured_params["response_format"] = {
             "type": "json_schema",
             "json_schema": {
                 "name": response_format.__name__,
-                "schema": response_format.model_json_schema(),
+                "schema": schema_json,
             },
         }
-        return await self._client.chat.completions.create(**structured_params)
+        try:
+            return await self._client.chat.completions.create(**structured_params)
+        except BadRequestError:
+            # Some providers (e.g. DeepSeek) don't support json_schema type.
+            # Fall back to json_object + inject schema into the system prompt.
+            logger.info(
+                "json_schema not supported by provider, falling back to json_object for %s",
+                response_format.__name__,
+            )
+            schema_prompt = (
+                f"You MUST respond with valid JSON that conforms to this JSON Schema:\n"
+                f"{json.dumps(schema_json, indent=2)}\n\n"
+                f"Return ONLY valid JSON, no markdown, no explanation."
+            )
+            json_params = dict(params)
+            json_params["response_format"] = {"type": "json_object"}
+            messages = list(json_params.get("messages", []))
+            # Inject schema into the last user message or append a system instruction
+            if messages and messages[-1]["role"] == "user":
+                messages[-1]["content"] = (
+                    str(messages[-1]["content"]) + "\n\n" + schema_prompt
+                )
+            else:
+                messages.append({"role": "user", "content": schema_prompt})
+            json_params["messages"] = messages
+            return await self._client.chat.completions.create(**json_params)
 
     @staticmethod
     def _parse_or_repair_structured_content(


tgtjam:
## Update: `json_mode` flag added

Added a `json_mode=true` option in `extra_params` (via `provider_params` in model config) that **skips the `json_schema` attempt entirely** and goes straight to `json_object` + schema injection.

For DeepSeek users: set `DERIVER_MODEL_CONFIG__OVERRIDES__PROVIDER_PARAMS__json_mode=true` in `.env` to avoid the wasted round-trip.

This is **backward compatible**: callers that don`t set `json_mode` keep the existing behavior (try json_schema → fallback to json_object on BadRequestError).

akattelu:
Hey @tgtjam, thanks for opening this. Your diagnosis of #716 is exactly right: OpenAI-compatible providers that reject `response_format: {"type": "json_schema"}` cause the deriver to silently produce zero observations, and that is a real bug worth fixing.

We ended up landing the fix for this whole cluster in #820 (now merged), so we're going to consolidate onto that. It adds a per-model-config option, `STRUCTURED_OUTPUT_MODE`, and a self-hoster on one of these providers sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` (the deriver is the only feature that uses structured output). In `json_object` mode it sends `response_format: {"type": "json_object"}`, injects the schema into the prompt with a no-fences instruction that deliberately includes the literal word "JSON" (which is what satisfies the Qwen/Alibaba "messages must contain 'json'" contract), and routes failed parses through `json_repair` so fenced or prefixed output is still recovered. We went with an explicit opt-in flag rather than an automatic `BadRequestError` fallback partly because the catch-and-retry path is hard to scope safely, which lines up with the two correctness notes CodeRabbit raised here.

That said, the auto-detect idea in your PR is the one thing #820 does not do: a self-hoster still has to know to set the flag. I think surfacing a log hint on first provider rejection (rather than silently producing nothing) is worth folding in as a follow-up. Closing this as superseded by #820, and thanks again for chasing down the root cause.

PR #752, opened 2026-05-31
fix(llm): use adaptive thinking for Opus 4.7+ Anthropic models

Couiz:
## Summary

Claude **Opus 4.7 and 4.8** removed the legacy extended-thinking API. The Anthropic backend (`src/llm/backends/anthropic.py`) only ever emitted the legacy `thinking: {"type": "enabled", "budget_tokens": N}` shape, so **every dream/dialectic call on Opus 4.8 fails with HTTP 400**:

```
"thinking.type.enabled" is not supported for this model.
Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
```

Reproduced on a self-hosted **v3.0.7** deployment: deriver dream (`dream.deduction`, `claude-opus-4-8`, `thinking_budget_tokens=16000`) → 400 on all 3 retries.

## Root cause

Both the `complete` and `stream` paths built the thinking parameter unconditionally as the legacy budget shape:

```python
if thinking_budget_tokens:
    params["thinking"] = {"type": "enabled", "budget_tokens": thinking_budget_tokens}
```

Opus 4.7/4.8 are **adaptive-thinking only** and reject that shape.

## The fix

A new `_build_thinking_params(model, thinking_budget_tokens, thinking_effort)` helper chooses the right shape per model:

- **Models that require adaptive** (Opus 4.7+): emit `thinking: {"type": "adaptive"}` plus `output_config: {"effort": ...}`. The effort level comes from an explicit `thinking_effort` when set, otherwise it is bucketed from the existing `thinking_budget_tokens` so current configs keep a comparable thinking depth:

  | budget | effort |
  |---|---|
  | `< 4096` | `low` |
  | `< 16000` | `medium` |
  | `16000–31999` | `high` |
  | `>= 32000` | `xhigh` |

  (The deployment's real values map `16000 → high` and `32000 → xhigh`.)

- **Models that still accept the legacy format** (Opus 4.6, Sonnet 4.6, Sonnet 4.5, Opus 4.5, Haiku, …): unchanged `{"type": "enabled", "budget_tokens": N}`.

The previously-`del`-eted `thinking_effort` parameter is now used on the adaptive path.

## Model-detection approach

Rather than a blanket switch, `_requires_adaptive_thinking()` parses the model version (`claude-<tier>-<major>-<minor>`) and returns `True` for **Opus ≥ 4.7** (and any future Opus, e.g. `claude-opus-5-0`). This is the precise set of models that return a 400 on the legacy shape — so older models behave exactly as before, and the next Opus release won't reintroduce the bug. Opus 4.6 / Sonnet 4.6 still accept the legacy format (deprecated but functional), so they intentionally stay on the legacy path to avoid changing their behavior or prompt-cache breakpoints.

## API verification

The adaptive shape (`thinking: {"type": "adaptive"}` + `output_config: {"effort": "low|medium|high|xhigh|max"}`) and the per-model availability were verified against the official Anthropic docs — https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking and https://platform.claude.com/docs/en/build-with-claude/effort — which confirm Opus 4.7/4.8 are adaptive-only and that the legacy `enabled` shape returns a 400 there.

## Notes

- Applied to **both** the `complete` and `stream` paths.
- The assistant-prefill gate now keys off whether thinking is enabled (`"thinking" not in thinking_params`) instead of the raw `thinking_budget_tokens`, keeping the existing "no prefill while thinking" invariant correct on adaptive models too.
- No change to `_normalize_response` / thinking-block replay: on Opus 4.7/4.8 `display` defaults to `"omitted"` (empty `thinking` text), but the `signature` is still captured for multi-turn replay.
- Sampling-parameter handling (`top_p`/`top_k`) is deliberately untouched to keep the diff focused on the thinking-format bug.

## Test plan

- `uv run ruff check src/` — clean
- `uv run basedpyright` — 0 errors
- `uv run pytest tests/llm/test_backends/test_anthropic.py` — 18 passed

New unit tests assert the adaptive shape for Opus 4.7/4.8 (and a future `opus-5-0` / date-suffixed variant), the unchanged legacy shape for Opus 4.6 / Sonnet 4.6 / Sonnet 4.5 / Haiku 4.5, the budget→effort bucketing, explicit-effort override, and the streaming path.

## Summary by CodeRabbit

* **New Features**
  * Adaptive "thinking" signals sent for Opus model variants; LLM retry behavior centralized for all call paths.

* **Bug Fixes**
  * Consistent thinking parameter formatting for streaming and non‑streaming requests and corrected JSON prefill detection.
  * Deterministic validation failures now surface immediately as HTTP 422 and are not retried.

* **Tests**
  * Expanded coverage for adaptive vs legacy thinking, effort overrides, streaming/live behavior, and validation/regression cases.

### DIFF
--- a/src/llm/api.py
+++ b/src/llm/api.py
@@ -17,10 +17,15 @@ from typing import Any, Literal, TypeVar, cast, overload
 
 from pydantic import BaseModel
 from sentry_sdk.ai.monitoring import ai_track
-from tenacity import retry, stop_after_attempt, wait_exponential
+from tenacity import (
+    retry,
+    retry_if_not_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+)
 
 from src.config import ConfiguredModelSettings, ModelConfig
-from src.exceptions import ValidationException
+from src.exceptions import HonchoException, ValidationException
 from src.telemetry.logging import conditional_observe
 from src.telemetry.reasoning_traces import log_reasoning_trace
 
@@ -290,6 +295,11 @@ async def honcho_llm_call(
         decorated = retry(
             stop=stop_after_attempt(retry_attempts),
             wait=wait_exponential(multiplier=1, min=4, max=10),
+            # HonchoExceptions are deterministic input/config errors (e.g. an
+            # invalid thinking budget/effort raised by a backend): fail fast so
+            # they propagate with their own status_code instead of being
+            # retried, wrapped in RetryError, and surfaced as a generic 500.
+            retry=retry_if_not_exception_type(HonchoException),
             before_sleep=before_retry_callback,
         )(decorated)
 
@@ -403,6 +413,7 @@ async def honcho_llm_call(
                 wrapped = retry(
                     stop=stop_after_attempt(retry_attempts),
                     wait=wait_exponential(multiplier=1, min=4, max=10),
+                    retry=retry_if_not_exception_type(HonchoException),
                     before_sleep=before_retry_callback,
                 )(wrapped)
             result: (
--- a/src/llm/api.py
+++ b/src/llm/api.py
@@ -17,15 +17,9 @@ from typing import Any, Literal, TypeVar, cast, overload
 
 from pydantic import BaseModel
 from sentry_sdk.ai.monitoring import ai_track
-from tenacity import (
-    retry,
-    retry_if_not_exception_type,
-    stop_after_attempt,
-    wait_exponential,
-)
 
 from src.config import ConfiguredModelSettings, ModelConfig
-from src.exceptions import HonchoException, ValidationException
+from src.exceptions import ValidationException
 from src.telemetry.logging import conditional_observe
 from src.telemetry.reasoning_traces import log_reasoning_trace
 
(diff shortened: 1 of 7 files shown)


akattelu (changes requested):
Hey @Couiz, thanks for opening this. The 400 is real, and I think the Anthropic backend is the right layer. `_build_thinking_params` plus actually using `thinking_effort` instead of `del`ing it is the shape I would want.

A few things before this can land:

- Need to fix merge conflicts.
- The adaptive-required table is already stale against Anthropic's current troubleshooting page. `_ADAPTIVE_THINKING_MIN_VERSION = {"opus": (4, 7)}` would still send `thinking.type.enabled` to Sonnet 5 / Fable 5 / Mythos 5, which now 400 the same way. Opus 5 happens to pass the opus >= 4.7 check; the others do not. https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting
- I would lean toward dropping the `with_llm_retry` / `HonchoException` fail-fast refactor from this PR. Reasonable idea, but it is a second change across `api.py` / `tool_loop.py` / `runtime.py` on files that have moved a lot. Happy to take that as a follow-up.

Suggested shape: rebase onto current `main`, keep the thinking-format helper + tests, update the adaptive-only set to match the current Anthropic table, and leave the retry plumbing for a separate PR.

akattelu:
Closing due to inactivity

PR #755, opened 2026-05-31
fix: disable PyJWT built-in exp validation for ISO string exp

hyuan:
## Problem

Honcho stores JWT expiry (`exp`) as an ISO 8601 string (e.g. `'2036-01-01T00:00:00Z'`) in `JWTParams`, and already has a manual expiry check via `parse_datetime_iso()`.

However, PyJWT's `jwt.decode()` defaults to `verify_exp=True`, which validates `exp` as a Unix integer per RFC 7519. When it encounters an ISO string, it raises:

```
DecodeError: Expiration Time claim (exp) must be an integer.
```

This exception was caught and re-raised as `AuthenticationException('Invalid JWT')`, making **all scoped tokens with an expiry unverifiable** — i.e., any token generated with an `--expires` flag by the `/v1/keys` endpoint or `generate_jwt.py` would always be rejected.

## Fix

Pass `options={'verify_exp': False}` to `jwt.decode()` to skip PyJWT's built-in check. Honcho's own manual expiry check (already present immediately after the `jwt.decode()` call) continues to enforce expiration correctly.

## Testing

Verified manually:
- Valid token with future ISO expiry → accepted ✓
- Token with past ISO expiry → rejected with `AuthenticationException: JWT expired` ✓  
- Token with no expiry → accepted ✓

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated JWT token expiration validation behavior to ensure proper handling of expired authentication tokens.

### DIFF
--- a/src/security.py
+++ b/src/security.py
@@ -88,7 +88,10 @@ def verify_jwt(token: str) -> JWTParams:
         if not settings.AUTH.JWT_SECRET:
             raise ValueError("AUTH_JWT_SECRET is not set, cannot verify JWT.")
         decoded = jwt.decode(
-            token, settings.AUTH.JWT_SECRET.encode("utf-8"), algorithms=["HS256"]
+            token,
+            settings.AUTH.JWT_SECRET.encode("utf-8"),
+            algorithms=["HS256"],
+            options={"verify_exp": False},
         )
         if "t" in decoded:
             params.t = decoded["t"]


Joe-Kneeland:
Ran into exactly this in production today (2026-07-04) on a self-hosted instance running `ghcr.io/plastic-labs/honcho:latest` built from `602347d` (current `main`, so this is still live).

Independent minimal repro against that build, same root cause as described above:

```python
import jwt
payload = {"t": "", "w": "myworkspace", "exp": "2027-07-04T22:00:00Z"}
token = jwt.encode(payload, "secret", algorithm="HS256")
jwt.decode(token, "secret", algorithms=["HS256"])
# DecodeError: Expiration Time claim (exp) must be an integer.
```

Confirmed the practical impact matches the PR description exactly: any `generate_jwt.py` token created with `--expires` (admin or scoped) is rejected with a generic `AuthenticationException: Invalid JWT` from `verify_jwt`'s broad `except jwt.PyJWTError`, while the identical scope with no `--expires` works fine. This masked the real cause for a while on my end since the 401 gives no hint it's expiry-related.

The fix here (`options={"verify_exp": False}`, deferring to the existing manual `parse_datetime_iso` check) looks correct and is exactly what I'd have proposed independently. Would be great to see this merged — happy to help verify further if useful.

Babakkamali:
Hit this on a self-hosted deployment (`main`, `4d3ab1c`). Can confirm the fix — without it, every expiring token is dead on arrival.

`python scripts/generate_jwt.py --admin --expires 1y` produces a token the API rejects with `401 {"detail":"Invalid JWT"}`. Drop `--expires` and it works. So there's currently no way to issue a credential that expires.

The reason is that `verify_jwt()` can't satisfy both halves of the code at once:

- ISO string exp (what `generate_jwt.py` writes) → PyJWT raises `DecodeError: Expiration Time claim (exp) must be an integer`
- numeric exp → Honcho's own check raises `ValueError: Invalid ISO 8601 datetime format: '1817219199'`

PyJWT's check runs first, so `parse_datetime_iso` below it never executes. It's dead code right now.

I looked at whether changing `generate_jwt.py` to emit a numeric exp would be the better fix, and I don't think so. The ISO format is deliberate — `JWTParams` documents it (``` `exp`: a string timestamp of when the JWT expires (optional) ```) and `verify_jwt` already implements the comparison. Going numeric would hand expiry to PyJWT and leave that documented path unreachable. Turning off PyJWT's duplicate check keeps Honcho's own logic in charge.

Checked that expiry still actually works after patching, since disabling a validation flag looks alarming out of context:

```
1y token : ACCEPTED
expired  : rejected -> JWT expired
```

One thing that tripped me up while testing: `src/` is baked into the image rather than mounted, so patching the file on the host does nothing until `docker compose build api deriver`. I briefly thought the fix hadn't worked.

Would be good to get this in — `--expires` is in the script's own `--help` examples but can't currently produce a usable token.

akattelu:
Hey @hyuan, thanks for opening this, and thanks @Joe-Kneeland / @Babakkamali for the independent prod confirmations. The bug is real and still on `main`: any token minted with `--expires` or `expires_at` lands an ISO string in the reserved `exp` claim, PyJWT raises `DecodeError: Expiration Time claim (exp) must be an integer`, and the blanket `except jwt.PyJWTError` turns that into a generic `Invalid JWT`. The manual `parse_datetime_iso` check underneath is dead code today.

I think the fix is at the wrong layer though. Turning off `verify_exp` keeps a non-standard string inside a claim name the rest of the JWT ecosystem treats as integer seconds (RFC 7519 §4.1.4). It would unblock `--expires`, but it freezes the non-standard wire format and permanently opts out of a library check we could be using.

Suggested shape: fix it at mint time. Type `JWTParams.exp` as `datetime | None`, let PyJWT encode a NumericDate, catch `jwt.ExpiredSignatureError` before `PyJWTError` so expired tokens still report `JWT expired`, and cover it with tests. That is the direction in #1025 against the canonical writeup in #1016 (Maintainer Approved). Going to close this one in favor of that shape rather than land the decode-side workaround.

#1016 stays open until a NumericDate (or rename-the-claim) fix actually merges. Appreciate the clean diagnosis here. It made the later writeups much faster.

PR #756, opened 2026-05-31
Improve local Ollama handling for Honcho

gravyflex:
This patch tightens Honcho's local-model handling in three places:\n\n- deriver prompts now require JSON-only explicit facts\n- dialectic prompts are stricter about evidence and avoid invented source-tree details\n- the OpenAI backend falls back to Ollama's native /api/chat with think:false for local models, and uses json_object for local structured calls\n\nValidation:\n- python3 -m compileall on the three modified modules passed\n\nContext: this is intended to make local Ollama behavior more deterministic for self-hosted deployments.

## Summary by CodeRabbit

* **New Features**
  * Added routing to support local Ollama-style models for inference.

* **Improvements**
  * Prompt for explicit-fact extraction now enforces a strict JSON-only OUTPUT FORMAT.
  * Agent system prompt tightened: concise context synthesis, evidence-only answers, explicit notice when evidence is weak/missing, and brevity-by-default.

* **Tests**
  * Added tests verifying native Ollama request/response behavior and merged request options.

### DIFF
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -67,6 +67,12 @@ RULES:
 - Extract ALL observations from {peer_id} messages, using others as context.
 - Contextualize each observation sufficiently (e.g. "Ann is nervous about the job interview at the pharmacy" not just "Ann is nervous")
 
+OUTPUT FORMAT:
+Return ONLY a valid JSON object with this exact shape:
+{{"explicit":[{{"content":"self-contained explicit fact"}}]}}
+If there are no explicit facts, return {{"explicit":[]}}.
+Do not include Markdown, prose, code fences, or hidden reasoning text.
+
 EXAMPLES:
 - EXPLICIT: "I just had my 25th birthday last Saturday" → "{peer_id} is 25 years old", "{peer_id}'s birthday is June 21st"
 - EXPLICIT: "I took my dog for a walk in NYC" → "{peer_id} has a dog", "{peer_id} lives in NYC"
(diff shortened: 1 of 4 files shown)


akattelu:
Hey @gravyflex, thanks for opening this. You're right that the deriver silently producing zero observations against local OpenAI-compatible providers is a real bug, and your instinct to force JSON-only output is the correct one.

We ended up consolidating that fix at the config layer in #820, which adds a `STRUCTURED_OUTPUT_MODE` per-model-config option. A self-hoster on Ollama/vLLM/GLM/etc. sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object`, which sends `response_format={"type":"json_object"}` instead of `json_schema`, injects the schema into the prompt with a no-fences instruction (and deliberately includes the literal word "JSON", which also satisfies the Alibaba/Qwen "messages must contain 'json'" contract), then runs the result through `json_repair` to recover JSON from fenced or prefixed output. Because it's transport-agnostic rather than a parallel `/api/chat` path, I'd lean toward that as the single tracked fix here, so I'm going to close this in favor of #820.

Two ideas in here I think are worth keeping separately, though. The `think:false` flag to suppress reasoning output at the source is a clean complement to `json_repair` (which is largely robust to chain-of-thought but not guaranteed), and the dialectic prompt hardening in `src/dialectic/prompts.py` (anti-fabrication, evidence-only, brevity) is an independent prompt-quality win unrelated to structured output. Happy to see either of those as a focused follow-up PR. Thanks again for digging into this.

PR #760, opened 2026-06-01
fix(llm): pass OpenAI provider extra options

ShapeShifter499:
## Summary

- forward configured `extra_body`, `thinking`, and `extra_headers` from `OpenAIBackend` into chat-completions calls
- add focused coverage proving those provider-specific options reach the OpenAI-compatible SDK call

## Why

Some OpenAI-compatible providers expose useful or required request options outside Honcho's current hard-coded parameter allowlist. Honcho can carry these provider settings in config, but the OpenAI backend currently drops them before making the SDK call.

This PR keeps the behavior generic and provider-neutral: if a provider config supplies `extra_body`, `thinking`, or `extra_headers`, pass them through to the OpenAI-compatible request.

## Testing

- `uv run ruff check src/llm/backends/openai.py tests/llm/test_backends/test_openai.py`
- `uv run pytest tests/llm/test_backends/test_openai.py`

## Disclosure

Assisted-by: OpenClaw:gpt-5.5

## Summary by CodeRabbit

* **New Features**
  * OpenAI backend now accepts provider-specific request customizations: custom request body content, a thinking directive, and custom headers.

* **Tests**
  * Added test coverage to verify provider-specific request customizations are passed through correctly.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -322,6 +322,14 @@ class OpenAIBackend:
             ):
                 if key in extra_params:
                     params[key] = extra_params[key]
+            if "extra_body" in extra_params:
+                params.setdefault("extra_body", {}).update(extra_params["extra_body"])
+            if "thinking" in extra_params:
+                params.setdefault("extra_body", {})["thinking"] = extra_params[
+                    "thinking"
+                ]
+            if "extra_headers" in extra_params:
+                params["extra_headers"] = extra_params["extra_headers"]
         return params
 
     def _normalize_response(
--- a/tests/llm/test_backends/test_openai.py
+++ b/tests/llm/test_backends/test_openai.py
@@ -228,6 +228,52 @@ async def test_openai_backend_skips_extra_body_when_thinking_budget_zero() -> No
 
 
 @pytest.mark.asyncio
+async def test_openai_backend_passes_provider_extra_body_headers_and_thinking() -> None:
+    client = Mock()
+    client.chat.completions.create = AsyncMock(
+        return_value=SimpleNamespace(
+            choices=[
+                SimpleNamespace(
+                    finish_reason="stop",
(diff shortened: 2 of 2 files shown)


ShapeShifter499:
Related prior work: I found closed PR #670 after opening this split PR. #670 included an `extra_body` passthrough as one part of a larger mixed OpenAI-compatible provider compatibility PR (embedding dimensions and structured-output fallback too), and it appears to have been closed because it was opened against the wrong base.

This PR is intentionally narrower: it only forwards configured provider-specific OpenAI-compatible request options from the backend (`extra_body`, `thinking`, and `extra_headers`) and adds focused test coverage for that behavior.

Assisted-by: OpenClaw:gpt-5.5

ShapeShifter499:
Closing this in favor of #722 for the overlapping `extra_body` pass-through work.

After reviewing the existing PRs more carefully, #722 already handles the most important part of this PR better than this branch does:

- it forwards caller/provider-supplied `extra_body` from `extra_params` into the OpenAI-compatible request;
- it preserves existing `reasoning.max_tokens` populated from `thinking_budget_tokens`;
- it deep-merges `extra_body.reasoning` instead of clobbering it;
- it guards against malformed non-dict `extra_body` values.

This PR also attempted to pass through two additional provider options that #722 does not cover:

- `extra_headers`, for OpenAI-compatible providers that require custom request headers;
- explicit top-level `thinking`, folded into `extra_body["thinking"]`, for providers that expose that request shape.

Those remaining pieces are smaller and should not block or duplicate #722. If they still seem useful upstream, they can be proposed as a narrower follow-up after #722 lands or alongside it.

Leaving #761 open separately because the Kimi coding `User-Agent: KimiCLI/1.5` shim is unrelated to #722.

Assisted-by: OpenClaw:gpt-5.5

PR #766, opened 2026-06-02
feat: embed messages immediately on create with reconciler fallback

VVoruganti:
Adds embed_messages_now background task so newly created messages are searchable within seconds instead of waiting up to the reconciler interval. Three-phase claim/lease → embed → persist never holds a DB session across the embedding call; the reconciler remains the fallback for failures and stragglers.


Rajat-Ahuja1997 (approved):
lgtm

PR #768, opened 2026-06-02
fix(dedup): reinforce times_derived on duplicate detection

erosika:
## Summary

`times_derived` was never incremented — it stayed pinned at `1` for every conclusion. The two write paths in `is_rejected_duplicate()` both dropped the reinforcement:

- **reject-new**: incoming duplicate discarded, surviving doc's count never bumped
- **new-wins**: replacement created fresh at `times_derived=1`, accumulated count from the deleted doc lost

With every row equal to `1`, `ORDER BY times_derived DESC` (`_query_documents_most_derived`, `src/crud/representation.py`) had no signal to sort on and fell through to heap/insertion order ≈ oldest-first. So `include_most_frequent=true` wasn't surfacing reinforced memory — it silently degraded to "oldest conclusion first." That's the stale-context stickiness.

The dedup logic itself was working correctly the whole time; only the reinforcement tally it was meant to maintain was broken.

## Changes

`src/crud/document.py`
- reject-new → `existing_doc.times_derived += 1`
- new-wins → `doc.times_derived = max(doc.times_derived, existing_doc.times_derived + 1)` (carry the count forward through a replacement instead of resetting to 1)

`src/crud/representation.py` + `src/crud/document.py`
- add `created_at DESC` tiebreaker to both most-derived queries, so a flat/tied `times_derived` column degrades gracefully to recency instead of arbitrary heap order

## Tests

`tests/crud/test_document.py` — 3 regression tests, each fails on pre-fix code:
- most-derived ties break toward recency, not insertion order
- rejecting a duplicate reinforces the surviving doc
- a winning duplicate inherits the replaced doc's count + 1

## Notes

- No backfill: the historical reinforcement events left no trace (dedup correctly deleted/rejected the duplicates as they arrived), so `times_derived` can only accrue forward from this fix. With the recency tiebreak, ordering is sensible immediately while the column has no spread yet.
- Follow-ups (not in this PR): client-side blend config on `context()`; `source_ids`-as-event so the count is self-backfilling; align `include_most_frequent` defaults across the peer/session/representation endpoints.

Closes DEV-1851

## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate consolidation now preserves and carries forward derivation counts reliably when replacing or rejecting incoming documents.
  * Result ordering is now deterministic: primary by derivation count, then by recency, and finally by a stable tie-breaker to ensure consistent lists.

* **Tests**
  * Added regression and async tests covering duplicate-handling behaviors and tie-break ordering by recency.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -176,7 +176,8 @@ async def query_documents_most_derived(
         limit: Maximum number of documents to return
 
     Returns:
-        Sequence of documents ordered by times_derived descending
+        Sequence of documents ordered by times_derived descending,
+        ties broken by created_at descending (most recent first)
     """
     stmt = (
         select(models.Document)
@@ -189,6 +190,9 @@ async def query_documents_most_derived(
         .order_by(
             models.Document.times_derived.desc(),
             models.Document.created_at.desc(),
+            # created_at is the transaction timestamp, so documents created in
+            # the same batch share it -- id keeps the order deterministic.
+            models.Document.id,
         )
         .limit(limit)
     )
@@ -983,7 +987,13 @@ async def is_rejected_duplicate(
     If the document is not a duplicate, returns False.
 
     If the document is a duplicate AND the new document is superior,
-    deletes the existing document and returns False.
+    deletes the existing document and returns False. In this case
+    ``doc.times_derived`` is updated in place to carry the replaced
+    document's reinforcement count forward.
+
+    If the document is a duplicate AND the existing document is superior,
+    increments the existing document's ``times_derived`` to record the
+    reinforcement, then returns True.
     """
     # Step 1: Find potential duplicates using cosine similarity
     similar_docs = await query_documents(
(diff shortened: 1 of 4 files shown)


src/crud/document.py
```diff
@@ -1014,12 +1017,18 @@ async def is_rejected_duplicate(
...
         # Soft-delete the existing document - reconciliation will clean up vectors and hard-delete
         existing_doc.deleted_at = datetime.datetime.now(datetime.timezone.utc)
         await db.flush()
         return False  # Don't reject the new document
 
-    # Existing document has more information, reject the new one
+    # Existing document has more information, reject the new one but record the
+    # reinforcement: a semantic duplicate was derived again.
+    existing_doc.times_derived += 1
+    await db.flush()
```
Rajat-Ahuja1997, line 1043: the `times_derived` field can be set by concurrent writers. should make this atomic in SQL, something like 
```
await db.execute(
     update(models.Document)
     .where(models.Document.id == existing_doc.id)
     .values(times_derived=models.Document.times_derived + 1)
)
````

Rajat-Ahuja1997 (commented):
one comment regarding atomicity but otherwise lgtm

VVoruganti (approved):

PR #775, opened 2026-06-06
fix: pass custom_instructions to session summaries

### LINKED ISSUE #748: Session summaries ignore workspace custom_instructions and default to English
### Description

The `custom_instructions` field in workspace configuration is correctly threaded into the Deriver prompt (`src/deriver/prompts.py` via `minimal_deriver_prompt()`), but the Summarizer (`src/utils/summarizer.py`) has its own hardcoded English prompts and does **not** receive or apply `custom_instructions` at all.

This means that even when `custom_instructions` instructs Honcho to respond in a non-English language (e.g. German), session summaries are always generated in English, regardless of the language of the conversation.

### Steps to reproduce

1. Set `custom_instructions` on a workspace:
```json
{
  "configuration": {
    "reasoning": {
      "custom_instructions": "Always write all outputs in German."
    }
  }
}
```
2. Have a conversation entirely in German.
3. Wait for a short or long summary to be generated.
4. Observe: the summary is in English.

### Expected behavior

`custom_instructions` should be passed to the summarizer prompts in `src/utils/summarizer.py`, the same way it is already used in `minimal_deriver_prompt()` in `src/deriver/prompts.py`.

### Suggested fix

Thread `custom_instructions` through to the summarizer, analogous to the existing pattern in the deriver:

```python
# src/utils/summarizer.py
def summarize(..., custom_instructions: str | None = None):
    # include custom_instructions_section in the prompt
    ...
```

### Environment

- Using selfhostet API
- All conversation messages are in German
- Workspace `custom_instructions` confirmed set correctly via API

Oxygen56:
## Summary

Fixes session summaries ignoring workspace `custom_instructions` (e.g., language preferences). Even when `custom_instructions` instructed Honcho to respond in a non-English language, summaries were always generated in English.

The deriver already handles `custom_instructions` correctly via `_custom_instructions_section()` in `src/deriver/prompts.py`. This fix applies the **exact same proven pattern** to the summarizer.

## What Changed

1. **Import** `_custom_instructions_section` from `src.deriver.prompts` (reuse, don't duplicate)
2. **Thread** `custom_instructions` through the call chain:
   * `_create_and_save_summary()` → extracts from `configuration.reasoning.custom_instructions`
   * `_create_summary()` → new `custom_instructions` parameter
   * `create_short_summary()` / `create_long_summary()` → new `custom_instructions` parameter
   * `short_summary_prompt()` / `long_summary_prompt()` → renders via `_custom_instructions_section()`

## Design Decisions

* **Reuses the deriver's** `_custom_instructions_section()` — no code duplication, same normalization and formatting
* **Backward compatible** — `custom_instructions=None` (the default) produces identical output to before
* **Minimal change** — 1 file, +29/−2 lines, surgical fix

## Test Plan

- [X] Prompt functions accept `custom_instructions=None` (backward compat)
- [X] `_custom_instructions_section(None)` returns empty string (no prompt pollution)
- [X] `_custom_instructions_section("Write in German")` renders `CUSTOM INSTRUCTIONS:` block
- [X] All 12 structural validation checks pass
- [ ] E2E: requires live Honcho instance with non-English workspace

Closes plastic-labs/honcho#748

## Summary by CodeRabbit

* **New Features**
  * Per-summary custom instructions can be configured and are injected into short- and long-form summary generation to guide output and LLM calls.
  * Token estimation now accounts for optional custom instructions when calculating prompt size.
* **Bug Fixes / Validation**
  * Custom-instruction length/budget is validated and enforced when resolving summary configuration.
* **Documentation**
  * Minor docstring clarifications for summary-related helpers.
* **SDK**
  * SDK configuration now exposes an optional custom_instructions field for summaries.

### DIFF
--- a/src/schemas/configuration.py
+++ b/src/schemas/configuration.py
@@ -61,6 +61,15 @@ class SummaryConfiguration(BaseModel):
         ge=20,
         description="Number of messages per long summary. Must be positive, greater than or equal to 20, and greater than messages_per_short_summary.",
     )
+    custom_instructions: str | None = Field(
+        default=None,
+        description="Optional custom instructions for session summaries. Rejected if they exceed the summarizer custom-instruction token cap.",
+    )
+
+    @field_validator("custom_instructions")
+    @classmethod
+    def validate_custom_instructions(cls, value: str | None) -> str | None:
+        return _validate_custom_instructions_budget(value)
 
     @model_validator(mode="after")
     def validate_summary_thresholds(self) -> Self:
@@ -172,6 +181,12 @@ class ResolvedSummaryConfiguration(BaseModel):
     enabled: bool
     messages_per_short_summary: int
     messages_per_long_summary: int
+    custom_instructions: str | None = None
+
+    @field_validator("custom_instructions")
+    @classmethod
+    def validate_custom_instructions(cls, value: str | None) -> str | None:
+        return _validate_custom_instructions_budget(value)
 
 
 class ResolvedDreamConfiguration(BaseModel):
--- a/src/schemas/configuration.py
+++ b/src/schemas/configuration.py
@@ -63,7 +63,7 @@ class SummaryConfiguration(BaseModel):
     )
     custom_instructions: str | None = Field(
         default=None,
-        description="Optional custom instructions for session summaries. Rejected if they exceed the summarizer custom-instruction token cap.",
+        description="Optional custom instructions for session summaries. Validated against DERIVER.MAX_CUSTOM_INSTRUCTIONS_TOKENS.",
     )
 
     @field_validator("custom_instructions")
--- a/src/schemas/configuration.py
+++ b/src/schemas/configuration.py
@@ -135,7 +135,7 @@ class WorkspaceConfiguration(BaseModel):
     )
     dream: DreamConfiguration | None = Field(
         default=None,
-        description="Configuration for dream functionality. If reasoning is disabled, dreams will also be disabled and these settings will be ignored.",
+        description="Configuration for dream functionality. If reasoning is disabled, dreams will also be disabled and this setting will be ignored.",
     )
 
 
@@ -159,6 +159,10 @@ class MessageConfiguration(BaseModel):
     reasoning: ReasoningConfiguration | None = Field(
         default=None,
         description="Configuration for reasoning functionality.",
+    )
+    summary: SummaryConfiguration | None = Field(
+        default=None,
+        description="Configuration for summary functionality.",
     )
 
 
--- a/src/utils/summarizer.py
+++ b/src/utils/summarizer.py
@@ -58,6 +58,7 @@ class Summary(TypedDict):
 
 
 def to_schema_summary(s: Summary) -> schemas.Summary:
+    """Convert a Summary TypedDict to a Pydantic Summary schema object."""
     return schemas.Summary(
         content=s["content"],
         message_id=s["message_id"],
@@ -82,6 +83,7 @@ __all__ = [
 
 
 def _get_summary_model_config() -> ConfiguredModelSettings:
+    """Return the configured model settings for summary generation."""
     return settings.SUMMARY.MODEL_CONFIG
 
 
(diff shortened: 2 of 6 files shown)


Rajat-Ahuja1997 (changes requested):
hey @Oxygen56 thanks for the contribution! i think the motivation here makes sense but we don't want the summarizer using custom instructions that are intended for the deriver. I think adding custom_instructions to SummaryConfiguration and ResolvedSummaryConfiguration is the correct path forward here

Oxygen56:
Hey @Rajat-Ahuja1997, thanks for the review! That makes total sense — the summarizer definitely shouldn't be borrowing the deriver's config field.

I've pushed the changes:

- Added `custom_instructions` to `SummaryConfiguration` so it has its own field
- Switched `_create_and_save_summary` to read from `configuration.summary`
- Added `estimate_short_summary_prompt_tokens_with_custom_instructions` and `estimate_long_summary_prompt_tokens_with_custom_instructions` following the deriver pattern, so token estimates account for custom instructions when they're set
- Filled in the missing params in the docstring

Let me know if anything else needs adjusting!

Oxygen56:
Hey @Rajat-Ahuja1997, pushed the changes per your feedback — added `custom_instructions` to `SummaryConfiguration` and switched `_create_and_save_summary` to read from `configuration.summary`. Ready for another look when you have time.

Rajat-Ahuja1997:
> Hey @Rajat-Ahuja1997, pushed the changes per your feedback — added `custom_instructions` to `SummaryConfiguration` and switched `_create_and_save_summary` to read from `configuration.summary`. Ready for another look when you have time.

not seeing any new changes on the branch?

Rajat-Ahuja1997 (changes requested):
couple things here:

1) ResolvedSummaryConfiguration is missing custom_instructions
2) you updated SummaryConfiguration in the SDK but not in the honcho server code itself (in src/schemas/configuration.py)
3) the two `_with_custom_instructions` helpers are unnecessary. you can collapse these into an optional param on the original methods, like `def estimate_short_summary_prompt_tokens(custom_instructions: str | None = None)`
4) revert the uv.lock change.

Oxygen56:
@Rajat-Ahuja1997  thanks for the detailed feedback, really appreciate it. I've addressed all four points:

  - Added custom_instructions to ResolvedSummaryConfiguration in src/schemas/configuration.py
  - Added custom_instructions to the server-side SummaryConfiguration as well
  - Merged the two _with_custom_instructions helpers into optional params on the original estimate_*_summary_prompt_tokens functions
  - Reverted uv.lock

  Also added a TODO for the _custom_instructions_section import for a future shared-utility refactor, and switched the custom_instructions check to is not None. Let me
  know if anything else needs adjusting.

src/utils/summarizer.py
```diff
@@ -164,45 +174,60 @@ def long_summary_prompt(
...
-    """Estimate tokens for the long summary prompt (without messages/previous_summary)."""
+def estimate_long_summary_prompt_tokens(custom_instructions: str | None = None) -> int:
+    """Estimate tokens for the long summary prompt, optionally including custom instructions."""
     try:
         return estimate_tokens(
             long_summary_prompt(
                 formatted_messages="",
                 output_words=0,
                 previous_summary_text="",
+                custom_instructions=None,
```
Rajat-Ahuja1997: should be `custom_instructions=custom_instructions`

src/schemas/configuration.py
```diff
@@ -61,6 +61,15 @@ class SummaryConfiguration(BaseModel):
         ge=20,
         description="Number of messages per long summary. Must be positive, greater than or equal to 20, and greater than messages_per_short_summary.",
     )
+    custom_instructions: str | None = Field(
+        default=None,
+        description="Optional custom instructions for session summaries. Rejected if they exceed the summarizer custom-instruction token cap.",
```
Rajat-Ahuja1997: this is validating against the deriver's generic `MAX_CUSTOM_INSTRUCTIONS_TOKENS` cap, not a summarizer-specific one. i think that's fine, but make that clear in the description

Rajat-Ahuja1997 (changes requested):

Oxygen56:
@Rajat-Ahuja1997 updated the field description to clarify it uses `DERIVER.MAX_CUSTOM_INSTRUCTIONS_TOKENS`. The `custom_instructions` forwarding and `lru_cache` switch are also in. The `None`-resolution issue CodeRabbit flagged looks like a pre-existing behavior in `deep_update()` — happy to address in a follow-up if you'd like.

Oxygen56:
Hi @Rajat-Ahuja1997, checking if you could take another look when you have a chance. I believe the requested changes have been addressed: custom_instructions is forwarded into summaries, the config description clarifies the shared DERIVER token cap, and the estimator cache is bounded. Happy to adjust anything else.

leandrosampaio:
We’re experiencing the same issue in our self-hosted deployment with Brazilian Portuguese: workspace custom_instructions are respected by the Deriver, but session summaries still default to English. Looking forward to the fix.

PR #776, opened 2026-06-06
fix(llm): guard OpenAI backend against malformed provider responses

### LINKED ISSUE #676: [Bug] OpenAI-compatible chat responses can crash backend when choices or usage are missing
## Describe the bug

While looking into a self-hosted/OpenAI-compatible provider setup, I noticed the OpenAI chat backend is pretty trusting about the shape of responses coming back from the SDK/proxy.

There are a few places in `src/llm/backends/openai.py` where Honcho assumes the response is a normal OpenAI `ChatCompletion` with at least one choice and a `usage` attribute. If an OpenAI-compatible gateway returns an empty or odd response shape, Honcho can crash with a raw `AttributeError`, `TypeError`, or `IndexError` instead of turning that into a controlled provider error or trying the next fallback.

The main spots I noticed:

```py
# complete(), structured parse path
parsed = response.choices[0].message.parsed
```

```py
# _normalize_response()
usage = response.usage
finish_reason = response.choices[0].finish_reason
message = response.choices[0].message
```

```py
# _parse_or_repair_structured_content()
```

This seems especially easy to hit with OpenAI-compatible proxies/gateways, where the HTTP call may technically succeed but the object coming back is missing `usage`, has `choices=[]`, has `choices=None`, or is not quite the SDK object Honcho expected.

## To Reproduce

I don't have a neat public reproduction server for this, but this should be easy to cover with unit tests by mocking `chat.completions.create()` or `chat.completions.parse()` to return one of these shapes:

1. `SimpleNamespace(choices=[], usage=None)`
2. `SimpleNamespace(choices=None, usage=None)`
3. `SimpleNamespace(choices=[SimpleNamespace(message=None)], usage=None)`
4. `SimpleNamespace(choices=[...])` with no `usage` attribute at all
5. a plain dict or other non-`ChatCompletion` object from an OpenAI-compatible adapter

For structured output, another useful repro is:

1. `parse(response_model)` fails or returns a malformed shape
2. fallback `create(response_format=json_schema)` also fails or returns malformed content
3. Honcho currently does not appear to fall back further to `json_object` or plain `create()`

## Expected behaviour

I would expect the OpenAI backend to treat malformed/empty provider responses as a provider failure with a clear `ValidationException` or similar controlled error, not a raw attribute/index crash.

For structured output, it would also be nice if the fallback chain were a bit more defensive:

```text
parse(response_model)
  -> create(json_schema)
  -> create(json_object)
  -> create(no response_format)
```

Maybe not every provider should get every fallback, but right now the code only seems to cover part of that path.

## Why this matters

This is mostly a robustness issue. The current code is fine when the provider behaves exactly like OpenAI, but self-hosted setups often go through OpenRouter, vLLM, new-api, DashScope/Qwen, Groq-compatible layers, etc. Those paths are close enough to work most of the time, but they are not always faithful about response shape.

When they misbehave, the backend should fail in a way that points at the provider response instead of crashing somewhere like `response.choices[0]` or `response.usage`.

## Suggested fix

A small helper to extract the first choice/message safely would probably clean this up:

```py
def _first_message(response: Any) -> tuple[Any, Any | None]:
    usage = getattr(response, "usage", None)
    choices = getattr(response, "choices", None) or []
    if not choices:
        raise ValidationException("OpenAI response did not include any choices")
    choice = choices[0]
    message = getattr(choice, "message", None)
    if message is None:
        raise ValidationException("OpenAI response choice did not include a message")
    return choice, usage
```

Then `_normalize_response()` and the structured repair path could use `getattr(..., 0)` for token counts and catch `AttributeError`, `TypeError`, and `IndexError` around structured-response parsing/fallbacks.

Tests that would be worth adding:

* empty `choices`
* missing `usage`
* `choices=None`
* `message=None`
* non-`ChatCompletion` object
* `json_schema` structured fallback fails, then `json_object` or plain `create` is attempted

## Related issues

This is adjacent to, but not exactly the same as:

* plastic-labs/honcho#663
* plastic-labs/honcho#453
* plastic-labs/honcho#641
* plastic-labs/honcho#634

Those are more provider-specific compatibility problems. This one is about the OpenAI backend being defensive when an OpenAI-compatible provider returns a weird but possible response shape.

sanidhyasin:
## Description

Closes #676.

The OpenAI chat backend (`src/llm/backends/openai.py`) assumed every response is a well-formed `ChatCompletion` with at least one choice and a `usage` attribute. OpenAI-*compatible* gateways (OpenRouter, vLLM, DashScope, LiteLLM, new-api, Groq-compatible layers, ...) sometimes return a technically-successful HTTP response whose body is missing `choices`, carries an empty list, or has a null `message`/`usage`. In those cases the backend crashed with a raw `AttributeError`/`IndexError`/`TypeError` deep inside `response.choices[0].message` / `response.usage` instead of surfacing a controlled provider error.

## Changes

- Add two small helpers — `_first_choice()` and `_first_message()` — that raise a controlled `ValidationException` when `choices` is `None`/empty or the first choice has no `message`.
- Route the three trusting spots through them:
  - `complete()` structured `parse()` path (`.message.parsed` / `.content` / `.refusal`)
  - `_normalize_response()` (`response.usage`, `choices[0].finish_reason`, `choices[0].message`)
  - `_parse_or_repair_structured_content()`
- Missing `usage` now degrades to zero token counts via `getattr(...)` rather than raising (the `CompletionResult` already handled a falsy `usage`).

The malformed/empty case now fails as a `ValidationException` that points at the provider response, so callers can treat it as a provider failure (and fall back) instead of crashing.

### Scope

This intentionally covers the robustness/crash-defense half of #676. The broader structured-output fallback-chain expansion (`json_schema → json_object → no response_format`) overlaps with the JSON-mode work in #663/#751 and is left out to keep this change focused and easy to review.

## Tests

Added cases to `tests/llm/test_backends/test_openai.py` covering the response shapes called out in the issue:

- `choices=[]`, `choices=None`, no `usage` attribute, and `message=None` → controlled `ValidationException`
- response missing `usage` entirely → returns a result with token counts defaulted to `0`
- malformed structured (`parse()`) response with empty `choices` → controlled `ValidationException`

`uv run pytest tests/llm/test_backends/test_openai.py` → 27 passed. `ruff check`, `ruff format --check`, and `basedpyright` on the touched files are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened parsing of OpenAI-style responses to reliably detect missing or empty choices/messages and raise controlled validation errors.
  * Preserved existing fallback/repair behavior when parsed output is absent but raw content exists; improved safe extraction of finish reasons and content overrides.
  * Improved structured-output repair to handle malformed inputs more robustly.

* **Tests**
  * Added tests covering malformed responses, missing usage/token defaults, and structured-output validation error paths.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -18,6 +18,36 @@ from src.llm.structured_output import (
 logger = logging.getLogger(__name__)
 
 
+def _first_choice(response: Any) -> Any:
+    """Return the first choice from an OpenAI(-compatible) chat response.
+
+    OpenAI-compatible gateways (OpenRouter, vLLM, DashScope, LiteLLM, ...)
+    sometimes return a technically-successful HTTP response whose body is
+    missing ``choices`` or carries an empty list. Indexing
+    ``response.choices[0]`` directly turns that into a raw
+    ``AttributeError``/``IndexError``/``TypeError`` deep in the backend; raise a
+    controlled ``ValidationException`` instead so callers treat it as a provider
+    failure (and can fall back) rather than crashing.
+    """
+    choices = getattr(response, "choices", None)
+    if not choices:
+        raise ValidationException(
+            "OpenAI-compatible response did not include any choices"
+        )
+    return choices[0]
+
+
+def _first_message(response: Any) -> Any:
+    """Return the ``message`` of the first choice, guarding against malformed
+    OpenAI-compatible responses (see :func:`_first_choice`)."""
+    message = getattr(_first_choice(response), "message", None)
+    if message is None:
+        raise ValidationException(
+            "OpenAI-compatible response choice did not include a message"
+        )
+    return message
+
+
 def _uses_max_completion_tokens(model: str) -> bool:
     """OpenAI reasoning models (gpt-5 family + o-series) require
     ``max_completion_tokens`` instead of the classic ``max_tokens`` parameter.
@@ -171,8 +201,9 @@ class OpenAIBackend:
                     fallback_response,
                     content_override=content,
                 )
-            parsed = response.choices[0].message.parsed
-            raw_content = response.choices[0].message.content or ""
+            message = _first_message(response)
+            parsed = getattr(message, "parsed", None)
+            raw_content = getattr(message, "content", None) or ""
             if parsed is None and raw_content:
                 content = repair_response_model_json(
                     raw_content,
@@ -181,7 +212,7 @@ class OpenAIBackend:
                 )
                 return self._normalize_response(response, content_override=content)
             if parsed is None:
-                refusal = getattr(response.choices[0].message, "refusal", None)
+                refusal = getattr(message, "refusal", None)
                 if refusal:
                     return self._normalize_response(
                         response,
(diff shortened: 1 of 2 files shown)


sanidhyasin:
Hi @VVoruganti, mind taking a look when you have a moment? It guards the OpenAI backend against malformed provider responses so a bad payload doesn't take things down. Tests included. Happy to adjust anything.

PR #784, opened 2026-06-08
docs: adding honcho-memory skill

ajspig:
Reorganizes the agent-skills tree and corrects memory-concept documentation.
- Add honcho-memory (concepts/strategy) and honcho-mcp (MCP path) skills, split out of the former single memory skill.
- Restructure honcho-integration: slim SKILL.md, extract core-patterns.md + agent-patterns.md references.
- Move all skills to a canonical top-level skills/ directory with a .claude/skills symlink, so one tree serves both Claude Code discovery and npx skills add distribution.
- Fix core language: observe_me (peer-level) vs observe_others (SessionPeerConfig, models other participants) scope; reserve observe_me=False for deterministic bots; README honcho-memory line.


.claude/skills/honcho-memory/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+| Tool | When to use |
+| --- | --- |
+| `search` | Semantic search across past messages (scope by peer or session). |
+| `get_representation` | The user's representation as text — lightweight. |
+| `get_peer_context` / `get_session_context` | Fuller context (representation + peer card, or LLM-ready history). Note: `get_session_context` is session-local — for cross-session memory, pass a peer target so it folds in that peer's representation. |
+| `get_peer_card` / `set_peer_card` | Read or correct compact biographical facts. |
+| `create_conclusions` | Store a fact directly instead of waiting for background reasoning. |
+| `list_conclusions` / `query_conclusions` | Review what's known (check before storing duplicates) or find one to delete. |
+| `delete_conclusion` | Remove an incorrect or outdated fact. |
+| `schedule_dream` | Trigger background consolidation after a long conversation. |
```
VVoruganti: This endpoint is more of a manual escape hatch to force a dream to run, shouldn't in practice ever be necessary. Would not recommend it including it here

.claude/skills/honcho-memory/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+- **Representation / peer card** — the synthesized understanding of a peer, queryable via `chat`. A peer's representation **accumulates across every session** it appears in — that's the cross-conversation memory. Session-scoped data (recent messages, summaries) stays local to one session.
+
+Reasoning happens **asynchronously**. After you record a turn, don't poll or wait — the representation updates in the background and is richer next time you ask.
+
+## The loop: recall → respond → record
+
+Do this every conversation. It's the whole skill.
+
+1. **Once per conversation** — make sure there's a session with you and the user as peers (observe the user, don't observe yourself).
+2. **Before responding, when personalization helps** — ask Honcho what it knows about the user (`chat`), or pull relevant past context (`search`). Skip it for trivial turns; it costs a few seconds.
```
VVoruganti: I think `context` would potentially make more sense for the 2nd option

VVoruganti: .chat takes a few sessions but .context should be near instantaneous. May also want to discuss the different things that .context() returns. 

Might also want a section that discussions the different reasoning levels of the dialectic and what they are good for

.claude/skills/honcho-memory/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+**Reuse the same `session_id` for the whole continuous conversation** (don't mint a new one per turn) — that's what lets the user's messages accumulate past the ~1,000-token reasoning threshold. And use **one stable `peer_id` per real person**, reused across every session and channel; a fresh or per-channel ID (`user-web` vs `user-discord`) builds separate, weaker representations instead of one.
+
+**Other useful tools:**
+
+| Tool | When to use |
+| --- | --- |
+| `search` | Semantic search across past messages (scope by peer or session). |
+| `get_representation` | The user's representation as text — lightweight. |
```
VVoruganti: frame as a contextualized snapshot of a peer's representation that can be inserted into a system prompt.

.claude/skills/honcho-memory/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+| `get_peer_context` / `get_session_context` | Fuller context (representation + peer card, or LLM-ready history). Note: `get_session_context` is session-local — for cross-session memory, pass a peer target so it folds in that peer's representation. |
+| `get_peer_card` / `set_peer_card` | Read or correct compact biographical facts. |
+| `create_conclusions` | Store a fact directly instead of waiting for background reasoning. |
+| `list_conclusions` / `query_conclusions` | Review what's known (check before storing duplicates) or find one to delete. |
+| `delete_conclusion` | Remove an incorrect or outdated fact. |
+| `schedule_dream` | Trigger background consolidation after a long conversation. |
+
+`chat` accepts an optional `reasoning_level` (`minimal` → `max`). Use `minimal`/`low` for quick lookups, `high`/`max` when depth genuinely matters — higher is slower and costs more.
+
+### Path B — `honcho` CLI
```
VVoruganti: There's already a skill for the Honcho CLI by itself. Does it make sense to reuse that skill and make it a resource file under this as a new top level skill? Then the MCP one can be it's own resource as well depending on the path of usage

.claude/skills/honcho-memory/SKILL.md
```diff
@@ -0,0 +1,157 @@
...
+- **Check before you store.** Background reasoning derives most conclusions automatically. Store a conclusion manually only for a durable fact you want available immediately; `list`/`query` first to avoid duplicates.
+- **One workspace per app/user-context.** Don't scatter the same user's memory across multiple workspaces.
+
+## Resources
+
+- All integrations & plugins: <https://honcho.dev/docs/v3/guides/overview>
+- MCP server & client setup: <https://honcho.dev/docs/v3/guides/integrations/mcp>
+- Full MCP usage walkthrough: <https://raw.githubusercontent.com/plastic-labs/honcho/refs/heads/main/mcp/instructions.md>
+- Agent development overview: <https://honcho.dev/docs/v3/documentation/introduction/vibecoding>
+- CLI reference: <https://honcho.dev/docs/v3/documentation/reference/cli>
```
VVoruganti: Could point to the llm.txt versions of each of these

VVoruganti (changes requested):

.claude/skills/honcho-memory/SKILL.md
ajspig: maybe. The science of skills feels imprecise. I think it makes sense to keep them not nested for **discoverability** purposes. As resource files, neither path is independently findable, but an MCP-connected agent (for example, screenpipe) searching "how do I use these Honcho tools" should hit a skill directly, and so should someone debugging a deployment via CLI.

Did add a MCP skill.

docs/v3/documentation/introduction/vibecoding.mdx
```diff
@@ -122,7 +122,9 @@ npx skills add plastic-labs/honcho
 ```
 
 ```bash Install as Claude Skill Manually
```
akattelu, line 124: Not related to this line, but could be worth referencing the `honcho-integration` skill from https://honcho.dev/docs/v3/guides/overview 

Something where people can just ask their agent about honcho or how to integrate it into their codebase via running this skill.

docs/v3/documentation/introduction/vibecoding.mdx
```diff
@@ -139,6 +141,16 @@ curl -o ~/.claude/skills/honcho-integration.md https://raw.githubusercontent.com
 
 Invoke with `/honcho-integration` in your coding agent.
 
+#### honcho-memory
+
+**Concepts & strategy for using Honcho at runtime.** The hub skill: it teaches the recall → respond → record loop, session and peer design, and token-batching (so reasoning actually fires) — the durable model behind using Honcho as memory, independent of how you're connected. It routes to the path skill for your connection (`honcho-mcp` or `honcho-cli`). Use this when your agent already has Honcho available and you want it to remember the user across conversations — as opposed to `honcho-integration`, which adds the SDK to a codebase.
+
+#### honcho-mcp
```
akattelu: I have mixed feelings on having a separate skill for the mcp server. 

MCP servers can return an instructions string at initialize that clients inject into the system prompt whenever the server is connected (TS SDK: new McpServer({ name, version, instructions })). It's always-on, no discovery step needed. We already have the content in mcp/instructions.md, but mcp/src/server.ts constructs the server without the instructions field. 

Though the skill is useful because it can help with setup and discovery. One thing we could do is split this so the guidelines on how to use `honcho-mcp` are sent by the server itself, and the parts that are about how to get connected move to `honcho-integration`. Thoughts on that?

skills/migrate-honcho-py/SKILL.md
```diff
@@ -21,6 +21,8 @@ This skill migrates code from `honcho` Python SDK v1.6.0 to v2.1.1 (required for
 
 ## Quick Migration
 
+The four changes below are the structural breaks you hit first — apply them inline. The remaining renames and smaller changes are one-line lookups in the [Quick Reference Table](#quick-reference-table); the full before/after for every change lives in [DETAILED-CHANGES.md](DETAILED-CHANGES.md).
```
akattelu: Thoughts on removing these migration skills from the repo altogether? Perhaps users who still need this could find it on an old tag in the repo.

ajspig: Yeah I'm on the same page with this. Its really only relevant for small percent of users.

docs/v3/documentation/introduction/vibecoding.mdx
ajspig: onboard. We also have the [mcp page in the docs](https://honcho.dev/docs/v3/guides/integrations/mcp) for setup. but something more condensed could totally get moved to honcho-integration. I'll update.

akattelu (dismissed):

akattelu (approved):

Issue #568, opened 2026-04-17
Misleading 422 from /search hides the real embedding error


lucharo:
`src/utils/search.py:384` catches every `ValueError` from `embedding_client.embed()` and re-raises it with a fixed message:

```python
try:
    query_embedding = await embedding_client.embed(query)
except ValueError as e:
    raise ValidationException(
        f"Query exceeds maximum token limit of {settings.MAX_EMBEDDING_TOKENS}."
    ) from e
```

`embedding_client.embed()` raises `ValueError` for several different reasons:

- token limit exceeded (which the wrapper correctly describes)
- missing API key, e.g. `OpenAI API key is required` or `Gemini API key is required`
- empty embedding response from the provider (`No embedding returned from Gemini API`)

All of them surface to the caller as "Query exceeds maximum token limit of 8192", which sends them down a completely wrong debugging path. I just spent a while chasing what I thought was a token-counting bug before reading `embedding_client.py` and discovering my embedding provider was misconfigured.

The reported number is also wrong for non-default providers. `embedding_client.py:44` caps `max_embedding_tokens` at `min(settings.MAX_EMBEDDING_TOKENS, 2048)` for Gemini, so a genuine token-limit error from a Gemini setup still reports "8192" while the actual cap is 2048.

## Repro

Set `LLM_EMBEDDING_PROVIDER` to a provider whose API key/credentials aren't configured, then `POST /v3/workspaces/{ws}/search` with a short query. The 422 will claim the query exceeds the token limit even though it's a few words.

## Suggested fix

`embed()` already raises with descriptive messages, so the wrapper can just pass them through:

```python
except ValueError as e:
    raise ValidationException(str(e)) from e
```

Same pattern applies in `src/crud/document.py:351` and `src/utils/agent_tools.py:1266`.

Happy to open a PR.

PR #791, opened 2026-06-10
fix: stop mislabeling transient embedding errors as token-limit errors

kweiner:
## Summary

- Several call sites (`save_representation`, `query_documents`, `create_observations`) catch a generic `ValueError`/`Exception` from the embedding client and unconditionally rewrite the message to "...exceeds maximum token limit of N", regardless of the actual cause.
- `simple_batch_embed` (used by the deriver's `save_representation`) had no token pre-check at all, unlike `embed()`, which already does `len(self.encoding.encode(text)) > self.max_embedding_tokens` before calling the provider.
- As a result, transient provider errors — e.g. some OpenAI-compatible providers (observed via OpenRouter) occasionally return HTTP 200 with an empty `data: []` body, which the OpenAI SDK's response parser turns into `ValueError("No embedding data received")` — get relabeled as token-limit errors. This makes it impossible to distinguish "your content is too long" from "the embedding provider had a transient hiccup, please retry" from logs/error messages alone.

## Changes

- Add `EmbeddingTokenLimitError(ValueError)` in `src/embedding_client.py` — a distinguishable subclass (still a `ValueError` for backwards compatibility) raised only when content genuinely exceeds the configured token limit.
- `simple_batch_embed` now performs the same pre-check as `embed()` (count tokens via `self.encoding.encode`, compare to `self.max_embedding_tokens`) *before* calling the provider, raising `EmbeddingTokenLimitError` with the actual token count. Any other provider error now propagates unchanged.
- Narrow the `except` clauses in `src/crud/representation.py` (`save_representation`) and `src/crud/document.py` (`query_documents`, `create_observations`) from `except ValueError` to `except EmbeddingTokenLimitError`, so non-token-limit errors surface with their real message/type instead of being mislabeled.

## Test plan

- [x] `uv run pytest tests/llm/test_embedding_client.py -q` — 15 passed (added 2 new regression tests: pre-check raises `EmbeddingTokenLimitError` without a provider call when content is oversized; a non-token `ValueError` such as `"No embedding data received"` propagates unchanged through `simple_batch_embed`)
- [x] `uv run pytest tests/crud/test_representation_manager.py tests/deriver/test_deriver_processing.py -q` — 18 passed, no regressions
- [x] `uv run ruff check src/` — clean
- [x] `uv run basedpyright` on changed files — 0 errors, 0 warnings

## Release Notes

* **Bug Fixes**
  * Improved handling of oversized embedding inputs and search queries.
  * Users now receive clearer guidance when content or queries exceed the configured token limit.
  * Other embedding and provider errors are now distinguished from input-size validation issues.

* **Tests**
  * Added coverage for token-limit detection across individual and batch embedding workflows.
  * Verified that unrelated embedding failures continue to be reported appropriately.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -16,7 +16,7 @@ from src.crud.collection import get_or_create_collection
 from src.crud.peer import get_peer
 from src.crud.session import get_session
 from src.dependencies import tracked_db
-from src.embedding_client import embedding_client
+from src.embedding_client import EmbeddingTokenLimitError, embedding_client
 from src.exceptions import (
     ResourceNotFoundException,
     ValidationException,
@@ -350,7 +350,7 @@ async def query_documents(
     if embedding is None:
         try:
             embedding = await embedding_client.embed(query)
-        except ValueError as e:
+        except EmbeddingTokenLimitError as e:
             raise ValidationException(
                 "Query exceeds maximum token limit of "
                 + f"{settings.EMBEDDING.MAX_INPUT_TOKENS}."
@@ -798,7 +798,7 @@ async def create_observations(
     contents = [obs.content for obs in observations]
     try:
         embeddings = await embedding_client.simple_batch_embed(contents)
-    except ValueError as e:
+    except EmbeddingTokenLimitError as e:
         raise ValidationException(str(e)) from e
 
     # Create document objects and track embeddings for vector store
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -18,7 +18,7 @@ from src.crud.collection import get_or_create_collection
 from src.crud.peer import get_peer, reject_scope_observed
 from src.crud.session import get_session
 from src.dependencies import tracked_db
-from src.embedding_client import embedding_client
+from src.embedding_client import EmbeddingTokenLimitError, embedding_client
 from src.exceptions import (
     ResourceNotFoundException,
     ValidationException,
@@ -362,7 +362,7 @@ async def query_documents(
     if embedding is None:
         try:
             embedding = await embedding_client.embed(query)
-        except ValueError as e:
+        except EmbeddingTokenLimitError as e:
             raise ValidationException(
                 "Query exceeds maximum token limit of "
                 + f"{settings.EMBEDDING.MAX_INPUT_TOKENS}."
@@ -987,7 +987,7 @@ async def create_observations(
         embeddings = await embedding_client.simple_batch_embed(
             contents, on_oversize="truncate"
         )
-    except ValueError as e:
+    except EmbeddingTokenLimitError as e:
         raise ValidationException(str(e)) from e
 
     # Create document objects and track embeddings for vector store
--- a/src/crud/representation.py
+++ b/src/crud/representation.py
@@ -13,7 +13,7 @@ from src import crud, exceptions, models, schemas
 from src.config import settings
 from src.dependencies import tracked_db
 from src.dreamer.dream_scheduler import check_and_schedule_dream
-from src.embedding_client import embedding_client
+from src.embedding_client import EmbeddingTokenLimitError, embedding_client
 from src.schemas import ResolvedConfiguration
 from src.telemetry.events import EmbeddingCallPurpose
 from src.telemetry.logging import accumulate_metric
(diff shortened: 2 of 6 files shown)


akattelu (changes requested):
Hey @kweiner, thanks for opening this. The mislabeling is real, and a dedicated `EmbeddingTokenLimitError` is the right way to split "content too long" from provider failures.

A few things have moved since June 10:
- #704 already added the per-text pre-check in `simple_batch_embed` and dropped the `"token" in str(e)` remapping inside the client. That half of the diff now conflicts with a rewritten batching pipeline.
- #938 fixed the OpenRouter empty-`data: []` case this PR uses as the motivating example (`encoding_format="float"`).
- The caller remaps are still on `main`. `src/utils/search.py` and `src/utils/agent_tools.py` still catch every `ValueError` and rewrite it as a token-limit 422, which is the original report in #568. This PR does not touch those two.

Suggested shape: rebase onto current `main`, raise `EmbeddingTokenLimitError` from the existing pre-checks in `embed()` / `simple_batch_embed`, and narrow the remaps in `search.py`, `agent_tools.py`, `document.py`, and `representation.py`. Leave the batching pipeline alone.

kweiner:
@akattelu Thanks — that review was exactly the steer I needed, and you were right that most of the original diff had been overtaken. I've rebased onto current `main` (`cf07068d`) and reshaped it to what you described. Force-pushed as `765e1cb`; the PR now reports `MERGEABLE`.

## What changed

**Dropped**, per your notes:
- The `simple_batch_embed` pre-check and the `"token" in str(e)` removal — #704 already landed both, so that half is gone. The batching pipeline is untouched.
- The empty-`data: []` motivation — #938 fixed that with `encoding_format="float"`.

**Kept, narrowed to your suggested shape:**
- `EmbeddingTokenLimitError(ValueError)` in `src/embedding_client.py`, raised *only* from the two pre-flight token checks that already exist in `embed()` and `simple_batch_embed()`. Provider and configuration failures keep raising plain `ValueError`.
- Remaps narrowed in `src/utils/search.py`, `src/crud/document.py` (two sites), and `src/crud/representation.py` — these now catch `EmbeddingTokenLimitError` and let genuine provider faults propagate.

Subclassing `ValueError` means any handler I haven't touched keeps behaving as before, so this shouldn't be a breaking change for callers outside these four files.

**One deliberate deviation.** `src/utils/agent_tools.py` returns error *strings* into the dialectic tool loop rather than raising, so letting a provider error propagate there would break the loop mid-iteration. It now catches `EmbeddingTokenLimitError` for the real token-limit message and keeps a `ValueError` catch-all that reports accurately instead of blaming the query:

```python
except EmbeddingTokenLimitError:
    return "ERROR: Query exceeds maximum token limit of ... Please use a shorter query."
except ValueError as e:
    return f"ERROR: Embedding the query failed: {e}"
```

Happy to make that one propagate too if you'd rather keep it uniform.

## Something that turned up while narrowing

Both `simple_batch_embed` remap sites — `document.py:987` and `representation.py:109` — pass `on_oversize="truncate"`. Oversized input is truncated there, never raised, so those two handlers could not have been catching a token-limit error under any input. They only ever caught provider failures and relabelled them as token-limit validation errors. That's the clearest instance of #568 in the tree.

The same class of bug is what makes a dimension mismatch or an empty provider response surface to a user as "your query is too long", which is what sent me down this path originally.

## Verification

| check | result |
| --- | --- |
| `pytest tests/crud tests/llm` | 369 passed |
| `pytest tests/llm/test_embedding_client.py` | 46 passed (42 existing + 4 new) |
| `ruff check` (changed files) | clean |
| `ruff format --check` | clean |
| `basedpyright` (changed files) | 0 errors, 0 warnings |

Four regression tests added: the exception subclasses `ValueError`; `embed()` and `simple_batch_embed()` each raise it *before any provider call*; and a dimension mismatch raises `ValueError` but explicitly **not** `EmbeddingTokenLimitError`.

I have not run `tests/live_llm` or the unified runner — happy to if you think this warrants either, though I don't believe it moves the queue or SDK contract.

## One ask

#568 is currently labelled `observability` only, without `maintainer-approved`. Since that label gates PRs from being auto-closed, could you add it if you're happy with this direction? Would rather not have this one time out a second time.

akattelu (approved):

Issue #792, opened 2026-06-10
MCP server advertises OAuth in 401 response, causing spurious client-registration prompt in JetBrains AI Assistant


texas697:
## Summary

Every time a JetBrains IDE (WebStorm, IntelliJ) is opened with the Honcho MCP server configured, the AI Assistant panel shows the following popup:

> **The authorization server 'https://mcp.honcho.dev' does not support automatic client registration.**
> Do you want to proceed by manually providing a client registration (client ID)?

The connection still works — hitting **Cancel** dismisses the popup and Honcho functions normally via Bearer token auth. But the prompt appears on every IDE launch.

## Root cause (hypothesis)

The MCP server at `https://mcp.honcho.dev` appears to include an OAuth `WWW-Authenticate` challenge in its 401 response. JetBrains AI Assistant's MCP client sees this, follows the OAuth discovery flow, and hits dynamic client registration — which Honcho doesn't support.

Honcho's own docs only document Bearer token / API key auth. The OAuth hint in the 401 seems unintentional and is confusing OAuth-aware MCP clients.

## Config used

`.ai/mcp/mcp.json` (standard JetBrains AI project config):

```json
{
  "mcpServers": {
    "honcho": {
      "url": "https://mcp.honcho.dev",
      "headers": {
        "Authorization": "Bearer <api-key>",
        "X-Honcho-User-Name": "<username>",
        "X-Honcho-Workspace-ID": "<workspace>"
      }
    }
  }
}
```

JetBrains reads the `headers` field and passes the Bearer token correctly — the green "connected" indicator confirms the connection succeeds. The OAuth popup is a side effect of the server's 401 advertising OAuth before the headers are applied.

## Expected behavior

If the server authenticates successfully via Bearer token, it should not emit an OAuth `WWW-Authenticate` challenge that triggers client-registration flows in compliant MCP clients.

## Suggested fix

- Suppress the OAuth `WWW-Authenticate` header on authenticated requests, or
- Don't emit it at all if Honcho doesn't intend to support a full OAuth client-registration flow

## Environment

- JetBrains WebStorm (AI Assistant MCP Beta)
- `https://mcp.honcho.dev` (hosted)
- macOS 14+

— Jean "Claude"

akattelu:
@texas697 is this still an issue on the latest jetbrains versions?

akattelu:
Hey @texas697, thanks for opening this. I don't think stripping WWW-Authenticate from every createMcpHandler response is the right fix.

[...]

PR #796, opened 2026-06-10
fix(mcp): strip WWW-Authenticate header to prevent spurious OAuth popups in JetBrains

texas697:
## Problem

Reported in #792: every time a JetBrains IDE (WebStorm, IntelliJ) opens a project with the Honcho MCP server configured, the AI Assistant panel shows:

> **The authorization server 'https://mcp.honcho.dev' does not support automatic client registration.**

The connection works fine (Bearer token auth succeeds), but the popup fires on every IDE launch because `createMcpHandler` from the `agents` package emits a `WWW-Authenticate: Bearer` header on 401 responses as part of the MCP OAuth spec. JetBrains' MCP client is RFC-compliant — it sees that header, follows OAuth discovery, and hits dynamic client registration, which Honcho doesn't implement.

## Fix

Add a `stripOAuthChallenge` helper that removes `WWW-Authenticate` from any response returned by `createMcpHandler`. Honcho uses static Bearer token auth only — the header is an unintentional side effect of the agents package, not an advertised capability.

The helper is a no-op on responses that don't carry the header, so there's no impact on normal responses.

## Changes

- `mcp/src/index.ts` — wrap the handler response in `stripOAuthChallenge`; no other files touched

## Testing

Verified the fix by checking that the `WWW-Authenticate` header is stripped on 401 responses. JetBrains AI Assistant should no longer trigger the OAuth registration popup on project open.

— Jean "Claude"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response handling by filtering authentication-related headers from responses before they are returned to the caller.

### DIFF
--- a/mcp/src/index.ts
+++ b/mcp/src/index.ts
@@ -13,6 +13,17 @@ const CORS_HEADERS = {
   "Access-Control-Allow-Headers": CORS_ALLOWED_HEADERS,
 };
 
+function stripOAuthChallenge(response: Response): Response {
+  if (!response.headers.has("WWW-Authenticate")) return response;
+  const headers = new Headers(response.headers);
+  headers.delete("WWW-Authenticate");
+  return new Response(response.body, {
+    status: response.status,
+    statusText: response.statusText,
+    headers,
+  });
+}
+
 export default {
   async fetch(
     request: Request,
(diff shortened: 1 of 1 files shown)


akattelu:
Closing per comment on #792

PR #798, opened 2026-06-11
fix: use json_object instead of json_schema in structured output fallback

### LINKED ISSUE #797: openai backend _create_structured_response retries with json_schema after parse() failure — providers that don't support json_schema loop forever
## Problem

When `chat.completions.parse()` fails (e.g., `ValidationError` from providers that return non-JSON content), the exception handler at line 160 calls `_create_structured_response()` as a fallback. This method retries with `json_schema` response format — the same format that providers like Z.AI (GLM-5.1) do not support.

## Impact

Providers that silently ignore `json_schema` (Z.AI GLM-5.1/5-Turbo/4.7, and likely other OpenAI-compatible proxies) return plain text or markdown instead of JSON. The downstream `repair_response_model_json()` then fails with `Expecting value: line 1 column 1 (char 0)`, and the deriver generates zero observations on every cycle.

## Reproduction

1. Self-host Honcho v3.0.9 with Z.AI as the OpenAI-compatible LLM provider
2. Configure deriver to use any GLM model
3. Ingest messages into any workspace
4. Observe deriver logs: `❌ Repair failed: Expecting value: line 1 column 1 (char 0)` followed by `Deriver generated zero observations`

## Root Cause

The exact failure path:

1. `chat.completions.parse()` sends `json_schema` response format to Z.AI
2. Z.AI returns **markdown-formatted text** (e.g., `- EXPLICIT: user lives in Helsinki, Finland...`). The model understands the schema and extracts correct facts, but does not wrap them in JSON.
3. OpenAI SDK raises `ValidationError` — cannot parse markdown into the Pydantic model
4. Exception handler calls `_create_structured_response()` which retries with `json_schema` (same format, same failure mode)
5. Z.AI returns markdown again. `repair_response_model_json()` tries `json.loads()` on markdown → fails
6. The `except ValidationError` in `repair_response_model_json` returns `PromptRepresentation(explicit=[])` — an empty model
7. Deriver sees zero observations

## Verification

Tested all 5 Z.AI models:

| Mode | Result |
|---|---|
| `json_schema` (structured output) | ❌ All 5 models return plain text/markdown |
| `json_object` (plain JSON mode) | ✅ All 5 models return valid JSON |

## Proposed Fix

In `_create_structured_response()`, change from `json_schema` to `json_object` mode and inject the Pydantic schema as a system message so the model knows the expected JSON shape:

```python
# Before:
structured_params["response_format"] = {
    "type": "json_schema",
    "json_schema": {
        "name": response_format.__name__,
        "schema": response_format.model_json_schema(),
    },
}

# After:
structured_params.pop("response_format", None)
structured_params["response_format"] = {"type": "json_object"}
schema = response_format.model_json_schema()
schema_msg = {"role": "system", "content": f"Respond in JSON matching this schema: {json.dumps(schema)}"}
structured_params["messages"] = [schema_msg] + structured_params.get("messages", [])
```

`json_object` mode is supported by all OpenAI-compatible providers. The existing `repair_response_model_json()` already handles parsing JSON text into Pydantic models, so no changes are needed downstream.

## Environment

- Honcho v3.0.9
- LLM: Z.AI GLM-5.1 (OpenAI-compatible coding plan endpoint)
- Self-hosted on ARM64 (Rock 5B+)

aigent9:
## Summary

Fixes #797

Changes `_create_structured_response()` in `src/llm/backends/openai.py` to use `json_object` mode instead of `json_schema` when retrying after `chat.completions.parse()` failure.

## Problem

When `chat.completions.parse()` fails (e.g., provider returns non-JSON), the fallback `_create_structured_response()` retries with `json_schema` format. Providers that silently ignore `json_schema` (Z.AI, and likely other OpenAI-compatible proxies) return plain text/markdown, causing the deriver to produce zero observations indefinitely.

## Fix

- Pop the stale `response_format` (Pydantic model class) from params before setting `json_object`
- Set `response_format` to `{"type": "json_object"}` (universally supported)
- Inject the Pydantic schema as a system message so the model returns the correct JSON shape
- The existing `repair_response_model_json()` handles parsing into the Pydantic model

## Testing

Tested on self-hosted Honcho v3.0.9 with Z.AI GLM-5.1:

- **Before patch**: Deriver generates zero observations on every cycle (markdown returned for `json_schema`)
- **After patch**: Deriver correctly extracts 6 observations from a 2-message test conversation, stored and retrievable via the representation/search API

## Changes

- `src/llm/backends/openai.py`: `_create_structured_response()` method (7 lines removed, 6 added)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAI structured output request formatting for enhanced compatibility and reliability.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -381,13 +381,12 @@ class OpenAIBackend:
         response_format: type[BaseModel],
     ) -> Any:
         structured_params = dict(params)
-        structured_params["response_format"] = {
-            "type": "json_schema",
-            "json_schema": {
-                "name": response_format.__name__,
-                "schema": response_format.model_json_schema(),
-            },
-        }
+        structured_params.pop("response_format", None)
+        structured_params["response_format"] = {"type": "json_object"}
+        # Inject schema hint so the model returns the correct JSON shape
+        schema = response_format.model_json_schema()
+        schema_msg = {"role": "system", "content": f"Respond in JSON matching this schema: {json.dumps(schema)}"}
+        structured_params["messages"] = [schema_msg] + structured_params.get("messages", [])
         return await self._client.chat.completions.create(**structured_params)
 
     @staticmethod


zsxh1990:
@aigent9 Good fix on the structured output fallback. The `json_object` approach is more robust than `json_schema` for providers with inconsistent support.

@plastic-labs — This PR has been sitting without any maintainer acknowledgment. Contributors spend their time debugging and fixing issues for your project. A simple "Thanks, merged!" or "Looks good, appreciate the fix!" takes 5 seconds and makes contributors feel valued.

Silent merges signal to the community that their effort doesn't matter. Please consider adding a brief comment when merging external contributions. 🙏

akattelu:
The way this is written right now, it is a pretty big breaking change that also has its own set of compatibility issues. This seems unnecessary at the moment but would visit this in the future if it becomes an issue. Closing this.

PR #802, opened 2026-06-12
fix: handle Gemini transport in simple_batch_embed individually

### LINKED ISSUE #764: [Bug] Gemini transport in `simple_batch_embed` returns 1 embedding per batch
# **🐞 Bug Report**

## **Note**
My agent found this bug. I had my agent opencode/big pickle write it up. I don't understand the issue, but I want you to know about it. We did a fresh install  of the code **on** 6/2/2026 and the issue still exists. 
---
## Summary

When using Honcho with `EMBEDDING_MODEL_CONFIG__TRANSPORT=gemini`, the `simple_batch_embed()` method in `src/embedding_client.py` passes a list of texts to the Google genai SDK `embed_content()` call. The SDK ignores the multiple items and returns only one embedding, causing every subsequent`save_observations()` call to fail.

## Affected Versions

Honcho self-hosted (main branch, 2026-05-30). The Gemini transport was added recently and the `simple_batch_embed` path was not tested with multi-item batches.

## Root Cause

`src/embedding_client.py` line ~130:

```python
response = await self.client.aio.models.embed_content(
    model=self.model,
    contents=batch,  # ["text1", "text2", "text3"]
    config={"output_dimensionality": self.vector_dimensions},
)
```

The Google genai Python SDK's `embed_content()` method (singular) treats a `list[str]` as a single multi-part content input and collapses it into one embedding vector. This is **not** a batch operation.

For batch embedding, the SDK would need one of:
- The `embed_content` endpoint called individually per string, OR
- A separate endpoint (there is no `batchEmbedContents` exposed in the genai SDK in the same way as the REST API)

The OpenAI path (`AsyncOpenAI().embeddings.create(input=batch)`) correctly returns one embedding per input string — the Gemini path does not.

## Symptoms

- Deriver runs without LLM errors but `save_observations()` crashes with `ValueError` from `zip(..., strict=True)` because embedding count (1) does not match observation count (>1).
- Observations never persist to the `documents` table.
- Logs show `WARNING  Embedding count 1 < observation count N; truncating observations to match` after the safety fix in `representation.py`.

## Suggested Fix

Replace the batch call with individual embedding calls when using the Gemini transport:

```python
if isinstance(self.client, genai.Client):
    for text in batch:
        emb_vec = await self.embed(text)
        embeddings.append(emb_vec)
```

where `self.embed()` already correctly calls:

```python
response = await self.client.aio.models.embed_content(
    model=self.model,
    contents=query,  # single string
    config={"output_dimensionality": self.vector_dimensions},
)
```

This is slightly slower (N API calls instead of 1) but produces correct results. A future optimization could batch through the REST `batchEmbedContents` endpoint directly.

## Related Issue

Even with the fix above, it's worth also adding a safety check in `save_observations()` (or keeping `strict=False` in the zip) so that a partial embedding failure doesn't lose all observations. See

`src/crud/representation.py` line ~157:

```python
if len(embeddings) < len(all_observations):
    logger.warning(
        "Embedding count %d < observation count %d; truncating observations to match",
        len(embeddings),
        len(all_observations),
    )
    all_observations = all_observations[: len(embeddings)]
```

Oxygen56:
## Summary
- Fixed simple_batch_embed() to process Gemini transport items individually instead of as a batch
- Added regression test for multi-item batch with Gemini transport

Fixes #764

## Root Cause
The Google genai Python SDK `embed_content()` method treats a `list[str]` as a single multi-part content input and collapses it into one embedding vector. This is not a batch operation. The OpenAI path correctly returns one embedding per input string.

## Fix
In `_process_batch()`, when the Gemini transport is detected, iterate over each item individually instead of passing all texts at once.

## Test Plan
- [x] `uv run pytest tests/llm/test_embedding_client.py` passes (17 tests)
- [x] Multi-item batch with Gemini transport returns correct number of embeddings

## Summary by CodeRabbit

* **Bug Fixes**
  * Gemini embedding provider now sends and validates individual embedding requests per item, preventing empty-item results and improving consistency.

* **Tests**
  * Added test verifying per-item embedding requests and that results are returned in the correct order.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -422,17 +422,18 @@ class _EmbeddingClient:
             item in analytics."""
             result: dict[str, dict[int, list[float]]] = defaultdict(dict)
             if isinstance(self.client, genai.Client):
-                response = await self.client.aio.models.embed_content(
-                    model=self.model,
-                    contents=[item.text for item in batch],
-                    config={"output_dimensionality": self.vector_dimensions},
-                )
-                if response.embeddings:
-                    for item, embedding in zip(batch, response.embeddings, strict=True):
-                        if embedding.values:
-                            result[item.text_id][item.chunk_index] = (
-                                self._validate_embedding_dimensions(embedding.values)
+                for item in batch:
+                    response = await self.client.aio.models.embed_content(
+                        model=self.model,
+                        contents=item.text,
+                        config={"output_dimensionality": self.vector_dimensions},
+                    )
+                    if response.embeddings and response.embeddings[0].values:
+                        result[item.text_id][item.chunk_index] = (
+                            self._validate_embedding_dimensions(
+                                response.embeddings[0].values
                             )
+                        )
             else:  # openai
                 openai_kwargs: dict[str, Any] = {
                     "model": self.model,
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -428,12 +428,16 @@ class _EmbeddingClient:
                         contents=item.text,
                         config={"output_dimensionality": self.vector_dimensions},
                     )
-                    if response.embeddings and response.embeddings[0].values:
-                        result[item.text_id][item.chunk_index] = (
-                            self._validate_embedding_dimensions(
-                                response.embeddings[0].values
-                            )
+                    if not response.embeddings or not response.embeddings[0].values:
+                        raise ValueError(
+                            f"Gemini returned empty embeddings for text_id={item.text_id!r},"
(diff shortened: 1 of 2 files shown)


akattelu:
Hey @Oxygen56 thanks for opening this. I think the problem described in the issue makes sense, but we shouldn't be making N calls here. Instead, we should try to move this to the correct gemini batch endpoint so we can still maintain this as a single API call for performance reasons. 

Also I think the testing here could be more robust. The unit test here doesn't prove much, so might be worth adding a `live_llm/` test.

PR #805, opened 2026-06-12
feat: send OpenRouter app-attribution headers on OpenAI-compatible clients

kweiner:
## Summary
- Adds an `OPENROUTER_ATTRIBUTION_HEADERS` constant (`HTTP-Referer: https://honcho.dev`, `X-Title: Honcho`) in `src/llm/registry.py`.
- Passes these as `default_headers` on every `AsyncOpenAI` client construction: `get_openai_client()`, `get_openai_override_client()`, the module-level `CLIENTS["openai"]`, and the embedding client in `src/embedding_client.py`.

This lets [OpenRouter's app attribution](https://openrouter.ai/docs/app-attribution) recognize requests as coming from Honcho, surfacing the app name (and icon, derived from the `HTTP-Referer` domain's favicon) in OpenRouter's dashboard/analytics. Other OpenAI-compatible providers (OpenAI, vLLM, Ollama, etc.) ignore unrecognized headers, so this is safe to send unconditionally regardless of which provider is configured.

## Test plan
- [x] Updated `FakeOpenAIClient` test doubles in `tests/llm/test_embedding_client.py` to accept the new `default_headers` kwarg
- [x] `uv run pytest tests/llm/ -q` — 141 passed

## Summary by CodeRabbit

* **New Features**
  * Improved compatibility with OpenAI-compatible API providers through enhanced configuration support.
  * Enabled automatic app-attribution headers for supported services.

* **Tests**
  * Added test coverage for header configuration functionality.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -12,7 +12,7 @@ from google.genai import types as genai_types
 from openai import AsyncOpenAI
 
 from .config import EmbeddingModelConfig, resolve_embedding_model_config, settings
-from .llm.registry import OPENROUTER_ATTRIBUTION_HEADERS
+from .llm.registry import _openrouter_headers
 
 logger = logging.getLogger(__name__)
 
@@ -176,7 +176,7 @@ class _EmbeddingClient:
             self.client = AsyncOpenAI(
                 api_key=config.api_key,
                 base_url=config.base_url,
-                default_headers=OPENROUTER_ATTRIBUTION_HEADERS,
+                default_headers=_openrouter_headers(config.base_url),
             )
             self.max_embedding_tokens = max_input_tokens
             self.max_batch_size = 2048  # OpenAI batch limit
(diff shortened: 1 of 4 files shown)


akattelu:
Hey @kweiner thanks for opening this. This change looks good but a couple of concerns:
- It would be ideal if we could only send these attribution headers to OpenRouter, instead of to all providers. Perhaps we could make a map of attribution headers, and then inject those based on if the base url is the OpenRouter base url? Then we could also update the tests to match that. 
- CI is failing, might be fixable via merging some of the latest changes. 

also nit: the docs suggest we should use x-openrouter-title for this https://openrouter.ai/docs/app-attribution#x-openrouter-title

kweiner:
Thanks for the review! Addressed all three points in the latest commit: attribution headers are now conditional on the base URL starting with `https://openrouter.ai`, switched to `X-Openrouter-Title`, and the branch was already up to date with main so CI should be clean.

akattelu (approved):

PR #811, opened 2026-06-14
fix(parser): handle empty or markdown-only LLM responses

louis-cai:
When certain LLMs (like `gemini-3-flash` or other lightweight models) generate zero observations, they frequently drift from the required JSON schema. Instead of returning a valid empty JSON array `[]`, they often return:
1. Completely empty strings or pure whitespace.
2. Empty markdown code blocks (e.g., ```json\n```).
3. Pure-text hallucinated explanations (e.g., "*Based on the messages provided, there are no facts...*").

Previously, `validate_and_repair_json` would strip the string but then attempt to parse the invalid result, crashing with `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`.

**This PR:**
- Adds a pre-processing guard using `re.fullmatch` to intercept truly empty outputs (including empty markdown fences), gracefully returning `[]`.
- Intercepts non-JSON hallucinated text (outputs not starting with `[`, `{`, or ```) with a WARNING log, coercing to `[]` rather than silently swallowing or crashing.
- Replaces `repr()` with payload length in error logging to prevent PII (raw model outputs) from leaking into persistent logs on parse failures, addressing CodeRabbit review feedback.

### DIFF
--- a/src/utils/json_parser.py
+++ b/src/utils/json_parser.py
@@ -355,11 +355,26 @@ def simple_bracket_repair(json_str: str) -> str:
 def validate_and_repair_json(json_str: str) -> str:
     """Main function with comprehensive repair strategies"""
     raw = json_str.strip() if json_str else ""
-    # L1: Basic empty check
-    if not raw or not (raw.startswith("[") or raw.startswith("{") or raw.startswith("```")):
-        logger.info(f"[PARSER-GUARD] Non-JSON content detected, bypassing: {repr(raw)[:200]}...")
+    # Detect empty markdown fence-only responses (e.g. "```json\n```" or "``` ```")
+    empty_fence_only = bool(
+        re.fullmatch(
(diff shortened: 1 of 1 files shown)


akattelu:
Hi @louis-cai thanks for opening this. Generally speaking I'm surprised that gemini 3 flash is producing results like this. Our `validate_and_repair_json` is meant to be generic, so I don't think returning `[]` makes sense in cases outside the deriver. Also the function has some other comprehensive checks downstream, and I think this behavior isn't fully consolidated alongside that. I wonder if the better fix here is to rely on gemini API native json repair / structured output features.

> Previously, validate_and_repair_json would strip the string but then attempt to parse the invalid result, crashing with json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0).

Perhaps we could catch this error downstream instead?

PR #814, opened 2026-06-17
Track user and session ID on Langfuse traces

akattelu:
Adds better metadata tracking to langfuse traces.
- There are no changes to the case where langfuse is not used
- Adds a session ID to agentic loops. This doesn't map to the honcho internal session ID. Instead its supposed to be 1 session per dialectic call
- Adds a user ID, which is the workspace name 

Testing
- Added unit tests and a new unified test
- Ran unified tests locally against a testing langfuse instance. Sessions and users are being represented correctly.
Screenshot 2026-06-17 at 11 29 19 AM

Helpful to read before reviewing:
- https://langfuse.com/docs/observability/data-model
- https://langfuse.com/docs/observability/features/sessions
- https://langfuse.com/docs/observability/features/users

## Release Notes

* **New Features**
  * Improved Langfuse tracing for LLM calls, multi-turn agent runs, per-iteration steps, and tool execution with clearer run/step visibility.
  * Added optional trace naming via `track_name` and richer input/output annotation for both runs and steps (including streaming scenarios).

* **Bug Fixes**
  * Improved Langfuse shutdown reliability by flushing buffered telemetry during shutdown.

* **Tests**
  * Added/updated coverage for trace/session propagation and run/step I/O annotation, plus telemetry naming behavior.

* **Chores**
  * Added a commented `LANGFUSE_SECRET_KEY` entry to the environment template.


src/llm/runtime.py
```diff
@@ -25,39 +27,215 @@
...
+    `run_span` is None; input and output can be set independently.
+    """
+    if run_span is None:
+        return
+    try:
+        kwargs: dict[str, Any] = {}
+        if input is not None:
+            kwargs["input"] = input
+        if output is not None:
+            kwargs["output"] = output
```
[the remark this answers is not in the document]
akattelu: This didn't trigger for me when running ruff locally so skipping this.

tests/llm/test_langfuse_trace_annotation.py
```diff
@@ -0,0 +1,529 @@
...
+    monkeypatch.setattr(langfuse, "get_client", lambda: FakeClient())
+    return captured
+
+
+@pytest.fixture
+def langfuse_enabled(monkeypatch: pytest.MonkeyPatch):
+    """Turn the integration on with a known NAMESPACE (the tenant / user_id)."""
+    monkeypatch.setattr(settings, "LANGFUSE_PUBLIC_KEY", "pk-test")
+    monkeypatch.setattr(settings, "NAMESPACE", "acme-tenant")
+
```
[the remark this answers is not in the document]
akattelu: This is the only test file where these fixtures are used, so skipping this

src/llm/runtime.py
```diff
@@ -25,39 +27,215 @@
...
+
+        # Only a trace-root generation stamps trace attrs; nested ones inherit
+        # them from the run root.
+        if not inside_run:
+            metadata = (
+                _step_metadata(telemetry)
+                if telemetry is not None
+                else {"namespace": str(settings.NAMESPACE)}
+            )
+            metadata["provider"] = str(provider)
```
VVoruganti: **Regression: multi-turn agent observations lose `provider`/`model`.** These are written only in the `if not inside_run:` branch. For Dialectic/Dreamer (`run_id` set, `_in_agent_run=True`), each per-iteration generation takes the `inside_run` branch — which only calls `update_current_generation(name=...)`, no `model=` — and `_step_metadata` (run root + step spans) omits provider/model too. Net: no observation for a multi-turn agent carries provider/model, so cost can no longer be aggregated by provider/model in Langfuse for the dominant agents. (Single-call deriver/summarizer still get it.) The deleted `update_current_langfuse_observation` stamped them on every call. Fix: add provider/model to `_step_metadata`, or pass `model=` to `update_current_generation`.

src/llm/api.py
```diff
@@ -181,6 +181,15 @@ async def honcho_llm_call(
     """
     runtime_model_config = resolve_runtime_model_config(model_config)
 
+    # Default telemetry.track_name to track_name so per-call traces are named
+    # per agent rather than after the bare function.
+    if (
+        telemetry is not None
+        and track_name is not None
+        and telemetry.track_name is None
```
VVoruganti: **`track_name` is silently dropped when `telemetry is None`.** This default only copies `track_name` into telemetry when `telemetry is not None`, and `annotate_current_langfuse_trace` derives the trace name from `telemetry.track_name`. So a caller passing `track_name` with `telemetry=None` gets an unnamed trace, no error. The deleted `update_current_langfuse_observation(..., name=track_name)` named the span unconditionally. Latent today (all 4 prod callers pass telemetry) but a real footgun — either default `track_name` onto telemetry unconditionally, or drop the kwarg.

tests/utils/test_clients.py
```diff
@@ -910,8 +912,10 @@ async def test_retry_disabled(self):
 
             assert response.content == "No retry response"
 
-    async def test_track_name_updates_langfuse_span_name(self):
-        """track_name should rename the top-level Langfuse span."""
+    async def test_track_name_names_langfuse_trace_and_generation(self):
```
VVoruganti: This replaces `test_track_name_updates_langfuse_span_name`, which called `honcho_llm_call(track_name=...)` with **no telemetry**. The new test always passes `telemetry=LLMTelemetryContext(...)`, so the `track_name`-without-telemetry naming path (see the api.py comment) lost its only coverage — that regression would now pass CI silently. Consider keeping one case with `telemetry=None`.

src/llm/api.py
```diff
@@ -429,30 +433,40 @@ async def _toolless_call() -> (
...
+            max_input_tokens=max_input_tokens,
+            get_attempt_plan=_get_attempt_plan,
+            before_retry_callback=before_retry_callback,
+            stream_final=stream_final_only,
+            iteration_callback=iteration_callback,
+            telemetry=telemetry,
+        )
+        # Stamp run output; stream_final has no content yet (drains later in the
+        # caller, where that generation self-stamps).
+        if isinstance(result, HonchoLLMCallResponse):
```
VVoruganti: **Streaming run-root trace always shows blank output (confirm this is acceptable).** `annotate_langfuse_run_io(output=...)` is guarded by `isinstance(result, HonchoLLMCallResponse)`, but the dominant path (`stream_final_only=True`) returns `StreamingResponseWithMetadata` — so the run-root span gets `input` but never `output`. The streamed answer self-stamps onto a *separate* session-joined trace (`session_id=run_id`), and that generation's `@observe` span closes when the generator is returned (before chunks drain), so it carries no output text / zero tokens (`executor.py`). Accurate totals live on `LLMCallCompletedEvent`, so this is intended — but the run-trace preview for the primary path is the query with no answer. Flagging for a conscious sign-off.

src/llm/types.py
```diff
@@ -75,6 +75,10 @@ class LLMTelemetryContext:
     # agent — dialectic/deduction/induction. Used by agent iteration
     # event and tool call event.
     agent_type: str | None = None
+    # Human-readable name for the Langfuse trace (e.g. "Dialectic Agent",
+    # "Minimal Deriver"). Populated by honcho_llm_call from its track_name
+    # kwarg so honcho_llm_call_inner can name each per-call trace.
+    track_name: str | None = None
```
VVoruganti, line 83: **Redundant dual state.** `track_name` now lives both here and as a `honcho_llm_call` kwarg, but no caller sets this field — the `replace(..., track_name=track_name)`-only-when-None default in api.py always fires, making it pure derived state. Combined with `trace_name` (for `log_reasoning_trace`), there are three overlapping name params carrying near-identical strings that can drift. Prefer one home for the name.

src/llm/runtime.py
```diff
@@ -25,39 +27,215 @@
...
 logger = logging.getLogger(__name__)
 
 # ContextVar tracking the current retry attempt for provider switching.
 current_attempt: ContextVar[int] = ContextVar("current_attempt", default=0)
 
+# True while inside a run-level trace (opened by `langfuse_agent_run`). Lets a
+# nested generation (run owns the trace attrs → stay silent) be told apart from
+# one that escaped its run root — e.g. the streamed final answer, drained after
+# the run trace closed, which must self-stamp and rejoin the session by run_id.
+_in_agent_run: ContextVar[bool] = ContextVar("_in_agent_run", default=False)
```
VVoruganti, line 42: **Fragile ordering invariant (currently correct).** The escaped-stream self-stamp depends on `honcho_llm_call` exiting the `langfuse_agent_run` block (resetting `_in_agent_run→False`) *before* the lazily-evaluated `stream_final_response` generator is drained by the caller — so the final generation runs with `inside_run=False` and rejoins the session by `run_id`. If anyone later makes the stream eager, pre-fetches a chunk inside the run span, or wraps stream consumption, the generation flips to `inside_run=True` and silently stops self-stamping. Threading the already-captured `run_span` into `stream_final_response` would express the intent explicitly instead of via ambient ContextVar state.

tests/conftest.py
```diff
@@ -34,10 +34,17 @@
 from src.db import Base
 from src.dependencies import get_db, get_read_db
 from src.exceptions import HonchoException
-from src.main import app
 from src.models import Peer, Workspace
 from src.security import JWTParams, create_admin_jwt, create_jwt
 
+# Disable Langfuse for the whole suite before importing src.main: @conditional_observe
+# binds to settings.LANGFUSE_PUBLIC_KEY at import time, so blanking it here keeps mocked
+# test calls from shipping 0-duration traces to the live project. Tests that exercise
```
VVoruganti: Comment references "the live project" — runtime/ops context in a public-repo comment. Suggest behavior-only wording, e.g. "...from emitting traces to a configured Langfuse backend."

src/llm/runtime.py
```diff
@@ -25,39 +27,215 @@
...
+        # clean by-name cost aggregation.
+        if telemetry is not None and telemetry.track_name:
+            get_client().update_current_generation(
+                name=f"{telemetry.track_name} LLM call"
+            )
+    except Exception as exc:  # pragma: no cover - best-effort telemetry
+        logger.debug("Failed to update Langfuse trace metadata: %s", exc)
+
+
+def _step_metadata(telemetry: LLMTelemetryContext) -> dict[str, str]:
```
VVoruganti: **Minor efficiency.** `_step_metadata` is rebuilt (and `str()`-coerced) at run open, each step open, and each generation annotation (~2N times per run), though only `iteration` varies. Compute the static portion once at run open and add `iteration` per step.

src/llm/tool_loop.py
```diff
@@ -337,217 +339,234 @@ async def execute_tool_loop(
...
-                plan.thinking_budget_tokens,
-                stream=False,
-                client_override=plan.client,
-                tools=tools,
-                tool_choice=effective_tool_choice,
-                messages=conversation_messages,
-                selected_config=plan.selected_config,
-                plan=plan,
-                telemetry=_telemetry_for_iteration(telemetry, iteration_for_call),
+        with langfuse_agent_step(
```
VVoruganti: **Re-indentation churn.** Wrapping the entire ~190-line loop body in this `with langfuse_agent_step(...)` is most of the file's diff (pure indentation) and pushes the body 4+ levels deep. The step span only needs to scope the LLM call + tool execution; generations/tool observations auto-parent to the current span anyway, so a narrower wrap would shrink both the diff and the nesting and ease future edits.

src/llm/runtime.py
```diff
@@ -25,39 +27,215 @@
...
+                if telemetry is not None
+                else {"namespace": str(settings.NAMESPACE)}
+            )
+            metadata["provider"] = str(provider)
+            metadata["model"] = str(model)
+            trace_name = telemetry.track_name if telemetry is not None else None
+
+            # Entering propagate_attributes stamps the active @observe root span;
+            # the empty body is intentional (no child spans to wrap here).
+            with propagate_attributes(
```
VVoruganti, line 88: The empty-body `with propagate_attributes(...): pass` (stamping the active `@observe` root via a context-manager entry side-effect) reads as a no-op and invites an accidental "dead code" deletion that would silently drop all single-call / escaped-stream trace stamping. If the SDK exposes `update_current_trace`/`update_current_span` (the sibling helpers already use `update_current_span`), prefer that to express intent directly.

VVoruganti (changes requested):
Review of the Langfuse session/trace wiring. The instrumentation is well-guarded (every Langfuse call is best-effort in `try/except`, so telemetry can't crash a request) and the big `tool_loop.py` restructure preserves control flow correctly. Verified false alarms while reviewing: `_finish_tool_observation` is in a `finally` (always closes), `level=None` is a safe Langfuse default, the `_in_agent_run` token reset is correct, and `dataclasses.replace` is actually *more* correct than the old hand-rolled copy.

Two items I'd fix before merge: the multi-turn `provider`/`model` metadata regression and the `track_name`-without-telemetry gap (which also lost its only test). One confirmed-real-but-documented tradeoff (blank run-root output on the streaming path) is flagged for a conscious sign-off. The rest are altitude/cleanup.

**Cross-file cleanup (not inlined — spans 8 sites):** the `from langfuse import get_client` + `if not LANGFUSE_PUBLIC_KEY: return` + `except Exception: logger.debug(...)` skeleton is hand-copied across 8 functions (`runtime.py` ×5, `agent_tools.py` ×2, `logging.py` `flush_langfuse`), with the disabled-guard written two different ways. `langfuse` is already a top-level import (`from langfuse import observe` in `logging.py`), so the lazy in-function imports buy zero deferral and add a `sys.modules` lookup per tool call / per iteration on the synchronous dialectic hot path. Worth a `_langfuse_client()` helper next to `conditional_observe`.

VVoruganti (dismissed):

VVoruganti (approved):

PR #824, opened 2026-06-19
fix: strip <think> blocks and markdown fences before JSON parsing

KarimMakhlas:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON parsing to handle edge cases, malformed input, and unexpected formatting, increasing overall system reliability.
  * Enhanced output format specifications to ensure consistent generation of properly structured responses.

### DIFF
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -76,6 +76,10 @@ EXAMPLES (using `alice` as the target peer id):
 - EXPLICIT: "I took my dog for a walk in NYC" → "alice has a dog", "alice lives in NYC"
 - EXPLICIT: "alice attended college" + general knowledge → "alice completed high school or equivalent"
 
+OUTPUT FORMAT:
+Always return a valid JSON object. Never wrap in markdown fences or <think> blocks.
+If nothing to extract, return: {"explicit": []}
+
 {custom_instructions_section}
 
 Target peer:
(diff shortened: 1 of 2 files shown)


akattelu:
Hey @KarimMakhlas, thanks for opening this. The underlying bug is real: on providers that wrap deriver structured output in `<think>` blocks or markdown fences, JSON parsing fails and the deriver silently produces zero observations.

We ended up consolidating this whole cluster onto #820, which handles it at the config layer with a new per-model-config option. A self-hoster sets `DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object`, which switches to `response_format={"type":"json_object"}` and injects the schema into the prompt with a fence-forbidding instruction. Parsing routes failures through `json_repair`, which pulls the JSON object out of fenced, prefixed, and most reasoning-model output. Because of that I'm going to close this in favor of #820.

The one piece worth keeping in mind is your dedicated `_strip_thinking_wrappers()` pre-pass in `src/utils/json_parser.py`. #820 covers `<think>` output incidentally via `json_repair` rather than with an explicit stripper, so if a real reasoning-model sample still defeats the parser after setting `json_object`, a hardened version of that helper (anchored rather than greedy/global, per CodeRabbit's note) would be worth folding in. If you hit such a sample, please ping me with it on #820. Thanks again for digging into this.

PR #825, opened 2026-06-20
feat: OpenAI OAuth device code flow (ChatGPT Plus / Codex)

kweiner:
## Summary

Adds OAuth authentication support for OpenAI using the device code flow, allowing Honcho to run LLM features against a **ChatGPT Plus subscription** instead of OpenAI API credits.

### Key insight: use the Responses API, not Chat Completions

The ChatGPT Plus OAuth token works against \`chatgpt.com/backend-api/codex/responses\` — the same endpoint used by the Codex CLI and Hermes. The \`/chat/completions\` endpoint on the same host is Cloudflare-protected for non-browser clients. \`api.openai.com/v1\` requires API billing credits that the ChatGPT Plus subscription does not include.

### Files changed

- **\`src/llm/oauth.py\`** — \`OAuthTokenManager\` manages refresh-token grant cycle with automatic background renewal; \`OAuthOpenAI\` subclasses \`AsyncOpenAI\` and overrides the bearer auth property to inject the live token per-request
- **\`src/llm/backends/openai_responses.py\`** *(new)* — direct \`httpx\` backend for the Responses API. Translates Honcho's Chat Completions message format to the Responses API wire format (\`instructions\` + \`input\` items), handles SSE streaming, tool calls via \`function_call\`/\`function_call_output\` items, and Pydantic structured output. Always sets \`stream: true\` and \`store: false\` as required by the endpoint.
- **\`src/llm/registry.py\`** — \`initialize_oauth()\` wires the token manager into \`CLIENTS["openai"]\`; \`backend_for_provider()\` routes \`OAuthOpenAI\` clients pointing at \`chatgpt.com/backend-api\` through \`OpenAIResponsesBackend\` instead of the standard OpenAI backend; uses \`settings.LLM.OPENAI_OAUTH_BASE_URL\`
- **\`src/config.py\`** — new \`LLMSettings\` fields: \`LLM_OPENAI_AUTH_MODE\` (\`api_key\` | \`oauth\`), \`LLM_OPENAI_REFRESH_TOKEN\`, \`LLM_OPENAI_CLIENT_ID\`, \`LLM_OPENAI_OAUTH_BASE_URL\` (default: \`https://chatgpt.com/backend-api/codex\`)
- **\`src/main.py\`** / **\`src/deriver/__main__.py\`** — call \`initialize_oauth()\` at startup and run the background token-refresh task for the process lifetime
- **\`scripts/honcho_oauth_setup.py\`** — self-contained device code setup script (no Codex CLI required). Drives the flow directly via \`httpx\`: POST \`/api/accounts/deviceauth/usercode\` → display code → poll \`/api/accounts/deviceauth/token\` → exchange at \`/oauth/token\`. The server generates and returns the \`code_verifier\` (non-standard; not client-side PKCE).
- **\`tests/llm/test_oauth.py\`** — unit tests covering token refresh, rotation, near-expiry detection, lock-based race safety, background loop, and bearer auth dynamic read

### Behavior

- Default behavior (\`LLM_OPENAI_AUTH_MODE=api_key\`) is unchanged — no OAuth code runs
- When \`LLM_OPENAI_AUTH_MODE=oauth\`: \`initialize_oauth()\` performs an immediate token refresh at startup, then a background asyncio task calls \`refresh_if_needed()\` every 60 s
- Token rotation is handled: if the token endpoint returns a new refresh token, the manager updates its stored value
- Features with an explicit \`OVERRIDES__BASE_URL\` (e.g. embeddings via OpenRouter) continue to use \`LLM_OPENAI_API_KEY\` directly via \`get_openai_override_client()\` — unaffected by OAuth mode
- Available models on the Responses API backend: \`gpt-5.5\`, \`gpt-5.4\`, \`gpt-5.4-mini\`, \`codex-auto-review\`

## Test plan

- [ ] \`uv run pytest tests/llm/test_oauth.py -v\` — unit tests pass
- [ ] \`uv run pytest tests/llm/ -q\` — full suite passes
- [ ] Run \`python scripts/honcho_oauth_setup.py\`, complete browser login, copy \`LLM_OPENAI_REFRESH_TOKEN\` to \`.env\`
- [ ] Set \`LLM_OPENAI_AUTH_MODE=oauth\`, restart Honcho, confirm dialectic chat and deriver both work without \`LLM_OPENAI_API_KEY\`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,3 +196,4 @@ grafana-data/
 
 # Claude Code addon stuff
 .omc
+secrets/
--- a/scripts/honcho_oauth_setup.py
+++ b/scripts/honcho_oauth_setup.py
@@ -1,14 +1,11 @@
 #!/usr/bin/env python3
-"""OAuth setup for Honcho using the OpenAI Codex CLI.
+"""OAuth setup for Honcho using the OpenAI ChatGPT Plus device code flow.
 
-The Codex CLI handles the browser-based OAuth flow correctly (including
-Cloudflare challenges on auth.openai.com).  This script delegates
-authentication to it, then reads the resulting refresh token and prints
-the .env snippet to add.
+Opens a browser-based login at auth.openai.com/codex/device, polls for
+authorization, exchanges the code for tokens, and prints the .env snippet
+to add.
 
-Requirements:
-    Install the Codex CLI via snap:     sudo snap install codex
-    Or via npm:                         npm install -g @openai/codex
+Requirements: Python 3.11+ with httpx (already a Honcho dependency).
 
 Usage:
     uv run python scripts/honcho_oauth_setup.py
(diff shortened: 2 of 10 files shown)


harukaxxxx:
Thanks for the work on this. I evaluated the current PR head `1153b328eb2e32a84c628c034972402b77bb4474` against its base `4f9a41360a853e520ddc6a8aecf851ca04137da3`, and also tried to apply the 13 commits in isolated worktrees against the current upstream `main` (`226bd0d0fb4ffeae9b96278d98fb72f64cba1e0c`). Before this is merged, I believe the following issues need to be addressed.

## Merge and CI blockers

- GitHub currently reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`, and `REVIEW_REQUIRED`.
- Rebasing/applying the PR onto the current upstream `main` produces a conflict in `src/llm/registry.py`, which has changed since this PR's base. Applying it to an older local `main` also required a manual `.gitignore` resolution. This should be rebased onto current `main` rather than resolved only against the old base.
- The current GitHub checks are not green: `test-python` and `test-status` are failing. The recorded Python run ended with `379 passed, 21 skipped, 1 failed`; the failure was `tests/crud/test_document.py::TestDocumentCRUD::test_duplicate_rejection_reinforces_existing` with `ValueError: OpenAI API key is required`. This may be environment/configuration-related rather than caused by OAuth, but it still needs to be explained or fixed before merge.
- In an isolated checkout, the OAuth tests passed (`8 passed`) and the LLM test subset passed (`209 passed`), but the changed-file Ruff check still reported errors. The branch therefore needs a clean full-suite and lint result after the rebase.
- I did not find a database migration or a new Python dependency in the PR.

## Concrete runtime issues found in the current head

### 1. `tool_choice="any"` is sent unchanged to the Responses API

`OpenAIResponsesBackend._build_body()` currently does:

```python
if tool_choice is not None:
    body["tool_choice"] = tool_choice
```

Honcho's canonical tool-choice vocabulary includes `"any"`, and the existing OpenAI Chat Completions backend translates that value to `"required"`. The Responses API contract exposed by the current OpenAI SDK accepts `"none" | "auto" | "required"`; it does not accept `"any"`.

I reproduced this directly with the PR code: passing `tool_choice="any"` produces `body["tool_choice"] == "any"`. This can result in an invalid request rather than forcing a tool call. The Responses backend needs the same canonical translation as the other backends (`any -> required`), plus a regression test.

### 2. `max_tokens` / `max_output_tokens` and `stop` are dropped

Both `complete()` and `stream()` accept `max_tokens`, `stop`, and `max_output_tokens`, but `_build_body()` does not accept or emit them. The generated request body therefore contains neither `max_output_tokens` nor `stop`.

This means callers can believe they have set an output limit or stop sequences while the Codex Responses request ignores those controls. The backend should define the intended precedence between `max_tokens` and `max_output_tokens`, forward the supported value, and either translate or explicitly reject unsupported `stop` behavior.

### 3. The streaming path drops tool calls

`complete()` accumulates `response.output_item.added`, `response.function_call_arguments.delta`, and `response.output_item.done` events into `ToolCallResult` values. `stream()`, however, only yields text deltas and handles `response.completed` / `response.failed`; it does not handle function-call events and always ends with `finish_reason="stop"`.

If the generic request builder selects the streaming path for a tool-enabled request, the tool call is not exposed to the caller, so the tool loop cannot continue. Please either implement tool-call streaming consistently with the provider interface or explicitly prevent this backend from using the streaming path for tool calls, with a test covering the behavior.

### 4. `OPENAI_REFRESH_TOKEN_FILE` cannot currently be the sole credential source

The new token-file behavior is documented as allowing the most recently rotated refresh token to survive restarts. However, `initialize_oauth()` still fails immediately when `LLM_OPENAI_REFRESH_TOKEN` is unset, before `OAuthTokenManager` gets a chance to read `LLM_OPENAI_REFRESH_TOKEN_FILE`:

```python
if not settings.LLM.OPENAI_REFRESH_TOKEN:
    raise ValueError(...)
```

If the file is intended to be a valid persisted source, startup should accept either the environment token or a readable token file, and the error should only be raised when neither is available. Otherwise the documented file-only recovery path does not work.

### 5. Refresh-token coordination is only intra-process

The API server and deriver each create an independent `OAuthTokenManager`. The manager uses an `asyncio.Lock`, which protects concurrent refreshes inside one process, but there is no cross-process lock or compare-and-swap protocol around the shared refresh-token file.

Re-reading the file before a refresh helps a process pick up a token rotated by its sibling, but it does not prevent both processes from refreshing the same token at the same time. If refresh-token rotation invalidates the previous token, one process can still invalidate the token used by the other. This needs either cross-process coordination, a clearly documented single-refresh-owner design, or a retry protocol that handles this race deterministically.

## Configuration/integration concern

The OAuth client is only used for the global OpenAI path. Existing per-feature model configurations with explicit `overrides.base_url` and `api_key_env` continue to use those configured endpoints. That is reasonable, but it is easy to misunderstand as a global switch.

For example, a self-hosted deployment with explicit OpenAI-compatible proxy overrides for Deriver/Dialectic will continue to send those workloads to the proxy even when `LLM_OPENAI_AUTH_MODE=oauth` is enabled. The PR should document this precedence clearly and ideally validate or log the resolved backend so operators do not believe OAuth is active while workers are still using another provider.

My recommendation is to rebase first, fix the request-translation and credential-lifecycle issues above, add focused regression tests for the Responses wire format and streaming/tool behavior, and then rerun the complete CI matrix before merging.

kweiner:
@harukaxxxx Thank you for this — it's an unusually thorough review and I appreciate the time it took. I've verified each of the five runtime findings against the current head and they're all legitimate. Two of them I'd characterize slightly differently, with evidence below, and I have a question for the maintainers before I invest in the rebase.

## Confirmed

All five are real:

1. **`tool_choice="any"` not translated** — confirmed. `src/llm/backends/openai.py` translates `any` → `required` (with an explicit comment noting OpenAI has no `any`), and the Responses backend doesn't. Clear inconsistency between the two backends.
2. **`max_tokens` / `max_output_tokens` / `stop` dropped** — confirmed. `_build_body()` doesn't accept or emit them.
3. **`stream()` doesn't handle function-call events** — confirmed as written (see caveat below).
4. **`OPENAI_REFRESH_TOKEN_FILE` can't be the sole credential source** — confirmed. `initialize_oauth()` raises on the missing env var before `OAuthTokenManager` gets a chance to read the file, so the documented file-only recovery path doesn't work as described.
5. **Refresh-token coordination is intra-process only** — confirmed, and I'd call this the highest-priority item of the five (more below).

## Two clarifications

### The `test-python` failure is a fork-PR artifact, not a regression

You flagged this as possibly environmental. I believe it definitively is, and it isn't specific to this PR:

- This PR is from a fork (`isCrossRepository: true`).
- `.github/workflows/unittest.yml` sets `LLM_OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}`.
- GitHub Actions does not expose repository secrets to `pull_request` runs originating from forks, so that variable resolves to empty.
- `src/embedding_client.py` then raises `ValueError: OpenAI API key is required` — the exact failure you recorded.
- The failing test (`tests/crud/test_document.py::TestDocumentCRUD::test_duplicate_rejection_reinforces_existing`) exercises document dedup/embedding. This PR changes 10 files; `src/embedding_client.py` is not among them.
- `main` is green because those runs execute in-repo with secrets available.

Any fork PR touching this suite should fail identically. Happy to be corrected if the maintainers see it differently, but I don't think there's an OAuth-side fix for it.

### Finding 3 is latent rather than active

You correctly hedged this with "if the generic request builder selects the streaming path for a tool-enabled request." It currently doesn't. `src/llm/tool_loop.py` is stream-final-only by design — the tool loop runs non-streaming and the final streamed turn is issued with `tools=None`, with an inline comment stating that tool calls have already run in the loop so the streamed turn is text-only.

So this is a real gap in the backend's contract conformance, but not a reachable bug on the current code paths. I agree it should be fixed (or explicitly guarded with a test asserting the backend rejects streaming-with-tools), just noting it's lower severity than 1, 2, 4, and 5.

## On finding 5

I want to underline this one. I've been running this branch in production, and after moving my API server to multiple Uvicorn workers, the process count sharing a single refresh-token file went from 2 to 4. At startup all workers refresh within a few hundred milliseconds of each other — exactly the race you describe. It hasn't caused a failure for me, which suggests the provider tolerates some overlap in rotation, but that's an empirical accident rather than a designed guarantee, and it degrades as worker count grows.

Re-reading the file before refresh (which the current head does) narrows the window but doesn't close it, as you say. I think the honest fix is a documented single-refresh-owner design rather than trying to make N processes coordinate through a file.

## Question for maintainers

@VVoruganti — this PR has been open since June 20 without maintainer review, and it now needs a rebase onto current `main` (the conflict is in `src/llm/registry.py`) plus the fixes above and regression tests. That's a meaningful chunk of work and I'm willing to do it, but I'd rather not if there's no appetite for this feature upstream.

Is OAuth/Codex support something you'd want in Honcho? If yes, I'll rebase and work through the list. If it's out of scope or conflicts with a direction you're planning, that's a completely fine answer — I'd just like to know before starting, and I'm happy to close it out if it's not a fit.

akattelu:
Hey @kweiner, thanks for opening this and for the careful writeup. 

I'm going to close this without merging. The path the PR takes (Codex CLI OAuth client id → chatgpt.com/backend-api/codex/responses as Honcho's general LLM backend) isn't something OpenAI documents as a supported third-party integration. Their docs cover Sign in with ChatGPT for first-party Codex clients and App Server embeddings, and the Platform API for product backends. Consumer/Pro language also calls out using ChatGPT to power third-party services and programmatic extraction of output, which is uncomfortably close to how dialectic + deriver would use a single subscription token, especially in multi-tenant or hosted setups.

Supported path remains an OpenAI Platform API key (or another documented provider). If OpenAI ever publishes a real third-party contract for subscription-backed Responses access, happy to revisit.

Issue #780, opened 2026-06-06
Deriver produces zero observations when LLM is Ollama cloud ministral-3:8b (response_format.json_schema not honored)


binarybasher:
### Environment

- Honcho 3.0.9 (ghcr.io/plastic-labs/honcho:latest, pulled 2026-06-06)
- Self-hosted via Podman rootless, AUTH_USE_AUTH=false
- LLM: Ollama cloud, `ministral-3:8b`, base URL `https://ollama.com/v1`
- Embedding: local Ollama, `nomic-embed-text`, `host.containers.internal:11434/v1`

### Symptoms

The deriver processes every queued message successfully but produces **zero observations**. Logs show:

```
src.utils.json_parser - ERROR - ❌ Repair failed: Expecting value: line 1 column 1 (char 0)
src.deriver.deriver - WARNING - Deriver generated zero observations for messages 1:2 in <workspace>/<session>!
```

The queue items get marked `processed=true` and the LLM call completes in 5-15s, so it looks like the pipeline is working end-to-end. But the `documents` table stays empty.

### Root cause

`src/llm/backends/openai.py::_create_structured_response()` sends this `response_format` to the LLM:

```python
structured_params["response_format"] = {
    "type": "json_schema",
    "json_schema": {
        "name": response_format.__name__,
        "schema": response_format.model_json_schema(),
    },
}
```

**Ollama cloud's `ministral-3:8b` does not honor `response_format={"type": "json_schema", ...}`.** It ignores the schema constraint and returns free-form markdown text like:

```
"Here are the explicit atomic facts derived directly from Ryan's message:

1. **Ryan is concerned about stale reads in a database context**...
2. **Ryan has firsthand experience with Postgres replication lag**..."
```

This is then passed to `validate_and_repair_json()` → `repair_json()` from the `json_repair` library, which sees no JSON structure and returns the empty string. The fallback `comprehensive_json_repair` also fails ("Expecting value: line 1 column 1 (char 0)" — i.e. the parser received an empty string). The deriver then logs "zero observations" and marks the queue item as processed.

### Reproduction

Minimal curl replicating what Honcho sends:

```bash
curl -sS -H "Authorization: Bearer $OLLAMA_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "ministral-3:8b",
    "messages": [{"role": "user", "content": "Extract observations from: Ryan owns a dog named Max. Respond with JSON only."}],
    "stream": false,
    "response_format": {
[...]

PR #829, opened 2026-06-21
fix(llm): make structured-output extraction robust on OpenAI-compatible providers

### LINKED ISSUE #716: [Bug] Ollama/OpenAI-compatible providers: deriver silently produces zero observations (reasoning-content fallback + json_schema ignored)

### LINKED ISSUE #743: Deriver prompt should explicitly require JSON for structured-output extraction
## Summary

When Honcho Deriver runs with `response_model=PromptRepresentation` and `json_mode=True`, the minimal Deriver prompt does not explicitly tell the model to return JSON matching the expected schema. With OpenAI-compatible Ollama models that support structured output via `response_format`, this can still produce natural-language bullet output copied from the prompt examples instead of JSON. Honcho then repairs/validates the malformed output into an empty `PromptRepresentation(explicit=[])`, marks the queue item processed, and saves no observations/documents.

## Environment

- Honcho v3.0.7
- Transport: `openai`
- Base URL: OpenAI-compatible Ollama endpoint
- Ollama server: 0.24.0
- Deriver models tested: `gemma4:31b-cloud`, `deepseek-v4-flash:cloud`, `kimi-k2.6:cloud`
- Also observed similar non-JSON behavior on the original prompt with other Ollama Cloud models.

## Observed behavior

Deriver logs show processed batches with zero observations and JSON repair failures:

```text
src.utils.json_parser - ERROR - ❌ Repair failed: Expecting value: line 1 column 1 (char 0)
src.deriver.deriver - WARNING - Deriver generated zero observations for messages ...
```

The queue item is treated as processed, but no `documents` are created, so peer representation remains empty even though messages and message embeddings exist.

## Relevant code paths

- `src/deriver/deriver.py` builds the prompt and calls `honcho_llm_call(... response_model=PromptRepresentation, json_mode=True ...)`.
- `src/deriver/prompts.py` `minimal_deriver_prompt()` asks for explicit atomic facts and includes bullet examples, but does not include an explicit JSON-only output contract.
- `src/llm/structured_output.py` returns `PromptRepresentation(explicit=[])` for unrecoverable malformed `PromptRepresentation` output, causing this failure mode to become a processed zero-observation result rather than a retryable/visible failure.

## Why this appears to be a prompt issue

Ollama structured output works on simple prompts through the OpenAI-compatible API when `response_format` is supplied. The failure appears with the real minimal Deriver prompt: models return useful bullet/prose observations instead of JSON, likely because the prompt examples are natural-language bullets and there is no explicit JSON-only instruction in the user-visible prompt.

https://docs.ollama.com/capabilities/structured-outputs#tips-for-reliable-structured-outputs

Adding a workspace-level Deriver custom instruction fixes new jobs without code changes:

```text
Return only JSON matching this schema: {"explicit":[{"content":"..."}]}. Do not include markdown, prose, bullets, or code fences.
```

After applying that instruction, Deriver produced nonzero observations and documents were saved successfully.

## Expected behavior

If Deriver requests structured output, the prompt should explicitly instruct the model to return only JSON matching the `PromptRepresentation` schema, for example:

```json
{"explicit":[{"content":"..."}]}
```

Alternatively or additionally, malformed structured output for Deriver should be surfaced as a retryable/visible failure instead of being silently converted to an empty observation set.

## Suggested fixes

1. Add an explicit JSON-only output contract to `minimal_deriver_prompt()` when it is used with `PromptRepresentation` structured output.
2. Consider changing the `PromptRepresentation` repair failure behavior so malformed output does not silently become `PromptRepresentation(explicit=[])` unless the model genuinely returned a valid empty result.
3. Consider logging a short sanitized raw-output preview when structured output repair fails, to make provider/prompt issues easier to diagnose.

### LINKED ISSUE #771: [Bug]   Deriver generates zero observations on chain-of-thought models (e.g. MiniMax-M3 / api.minimax.io)
# **🐞 Bug Report**

## **Describe the bug**

 **Environment**
- Honcho v3.x
- Deriver model: MiniMax-M3 via OpenAI-compatible transport at https://api.minimax.io/v1
- Transport: openai (uses chat.completions.parse() with response_format={"type":"json_schema", ...})

**Summary**
The deriver silently returns zero observations for every processed message. The queue is consumed, the LLM call succeeds (200 OK), but the response is wrapped in a chain-of-thought block and/or markdown fence that the JSON parser can't handle, so the deriver falls through to an empty PromptRepresentation. The same root cause likely affects any chain-of-thought model accessed via the OpenAI-compatible transport that doesn't strictly honour json_schema response format.

**Reproduction**
1. Configure the deriver with MiniMax-M3 (or any chain-of-thought model) at an OpenAI-compatible endpoint.
2. POST a message to a session:
   {"messages":[{"content":"hello, my name is frank","peer_id":"frank"}]}
3. Wait ~15s.
4. Call POST /v3/workspaces/<ws>/conclusions/list â€” returns {"items":[]}.
5. Check the deriver logs: contains ERROR Repair failed: Expecting value: line 1 column 1 (char 0) followed by WARNING Deriver generated zero observations for messages N:N.

**Expected**
The model returns a PromptRepresentation with at least one explicit observation. The response lands in the conclusions endpoint and the peer's representation card.

**Actual**
The model returns content like:
    <think>
    The user wants observations about frank...
    </think>

    ```json
    {
      "explicit": [
        {"content": "frank's name is frank"}
      ],
      "deductive": []
    }
    ```
or for low-information inputs:
    **Analysis of frank's Messages:**

    - frank sent a message on June 3, 2026...

In both cases validate_and_repair_json raises JSONDecodeError, the caller in repair_response_model_json catches it and returns PromptRepresentation(explicit=[]), and the deriver logs zero observations.

**Root cause**

1. src/llm/backends/openai.py:145 passes the pydantic response_model directly to self._client.chat.completions.parse(), which sends response_format={"type":"json_schema", ...}. Many third-party OpenAI-compatible providers (and chain-of-thought models in particular) accept the parameter but still emit prose wrapped in think blocks and code fences. The SDK's auto-parse fails silently and response.choices[0].message.content is the wrapped string.

2. src/utils/json_parser.py:validate_and_repair_json does json_str.strip() and hands it to the repair library. It has no concept of think blocks or markdown fences.

3. src/deriver/prompts.py:minimal_deriver_prompt does not tell the model to always return valid JSON even when there is nothing to extract.

**Fix**

Two changes, both in the Honcho repo:

**1. src/utils/json_parser.py** â€” strip wrappers before JSON parsing:

```python
def _strip_thinking_wrappers(content: str) -> str:
    """Strip <think>...</think> blocks and markdown ```json fences that
    chain-of-thought models wrap around their JSON output. Falls back
    to finding a bare JSON object/array if no fence is present."""
    if not content:
        return content
    text = re.sub(r"<think>.*?</think>", "", content, flags=re.DOTALL)
    fence = re.search(r"```(?:json)?\s*(\{.*?\}|\[.*?\])\s*```", text, re.DOTALL)
    if fence:
        return fence.group(1).strip()
    bare = re.search(r"(\{[\s\S]*\}|\[[\s\S]*\])", text)
    if bare:
        candidate = bare.group(1).strip()
        if candidate and candidate != text.strip():
            return candidate
    return text.strip()
```

Then call it from validate_and_repair_json as the first line:

```python
def validate_and_repair_json(json_str: str) -> str:
    json_str = _strip_thinking_wrappers(json_str.strip())
    good_json = repair_json(json_str)
    ...
```

**2. src/deriver/prompts.py** â€” add an explicit always-return-JSON instruction to minimal_deriver_prompt. Three lines are enough:

- "ALWAYS return a valid JSON object matching the response schema, even if there are no observations to extract. Use \"explicit\": [] when nothing is extractable. Do not return markdown, prose, code fences, or any non-JSON output."
- Add an explicit OUTPUT FORMAT section showing the exact JSON shape:
  `{"explicit": [{"content": "<one observation as a string>"}, ...]}`
- Add a NO FACTS example returning exactly `{"explicit": []}`.

Note: PromptRepresentation only has an `explicit` field â€” no `deductive` field. Do not promise a `deductive` key in the prompt or the model will include one and the schema validation will fail in a different way.

**Verification**

After applying both changes, rebuild the deriver image and restart:

    cd /home/keith/honcho
    sudo docker compose -p honcho -f /home/keith/honcho/docker-compose.yml build deriver
    sudo docker compose -p honcho -f /home/keith/honcho/docker-compose.yml up -d --force-recreate --no-deps deriver

Post a substantive message:

    curl -X POST -H "Content-Type: application/json" \
      -d '{"messages":[{"content":"hello my name is frank and i love motorcycling","peer_id":"frank"}]}' \
      http://127.0.0.1:8000/v3/workspaces/<ws>/sessions/<s>/messages

Wait ~15s, then check:

    curl -X POST -H "Content-Type: application/json" -d '{}' \
      http://127.0.0.1:8000/v3/workspaces/<ws>/conclusions/list

Expected: a non-empty items array with one or more explicit observations attributed to frank.

**Impact**

- Affects any Honcho deployment using a chain-of-thought model through the openai transport.
- Silent failure mode: no error surfaces to the API caller; only the deriver's "zero observations" warning gives any signal, and only at DEBUG log level.
- Other agents that use the same response_model + json_mode path (dialectic tool loop, summarizer, dreamer specialists) likely suffer the same failure and would benefit from the same parser fix.

**Suggested additional follow-ups (out of scope for this fix)**

- Add a regression test in tests/deriver/ that feeds a known wrapped response (think block + fence) through validate_and_repair_json and asserts clean JSON.
- Lower the log level for "Deriver generated zero observations" from WARNING to a more visible level (or emit it at INFO so it's not buried under structured-output errors).
- Consider adding a `json_object` fallback path in src/llm/backends/openai.py for providers that reject json_schema, mirroring how the SDK's parse() falls back to create() for some errors.

---

### **Is this a regression?**

---

### **To Reproduce**

1.
2.
3.
4.

---

### **Expected behaviour**

*

---

### **Media prove**

---

### **Your environment**

* OS: 
* Browser name and version:
* Honcho Server Version: 
* Honcho Client Version: 

---

### **Additional context**

*

### LINKED ISSUE #797: openai backend _create_structured_response retries with json_schema after parse() failure — providers that don't support json_schema loop forever

sethbrasile:
## Problem

The OpenAI backend always requests structured output via native `json_schema`
(`chat.completions.parse()`, with `_create_structured_response` as the fallback).
Many OpenAI-compatible providers either reject `json_schema` outright or silently
ignore it and return text / markdown / chain-of-thought-wrapped JSON. Either way
`repair_response_model_json()` fails with `Expecting value: line 1 column 1
(char 0)` and the **deriver silently produces zero observations on every cycle**.

Reported across several environments:
- #716 — Ollama / OpenAI-compatible (GLM, Kimi): json_schema ignored + reasoning-content wrapping
- #771 — chain-of-thought models (MiniMax-M3): `<think>` + markdown fence breaks the parser
- #797 — `_create_structured_response` fallback retries with the same `json_schema` that just failed
- #743 — `json_object` prompt doesn't tell the model to emit JSON / the schema
- also #663, #780

## Fix

Keep native `json_schema` as the **first attempt** — no change for providers that
support it (OpenAI, Anthropic, capable proxies). Harden only the failure/repair
paths:

1. **Reasoning/fence stripping** — `_strip_reasoning_and_fences()` reduces a
   response to its first complete JSON value before repair. It is parse-first
   (a valid JSON object/array — even one whose string values contain `<think>`
   or stray braces — is returned untouched), then strips `<think>…</think>`
   blocks, then does a string-aware `json.JSONDecoder().raw_decode` from the
   first `{`/`[` so it ignores ```json fences and trailing prose. Applied in the
   `LengthFinishReasonError` path, the `parsed is None` path, and
   `_parse_or_repair_structured_content`.
2. **json_object fallback** — `_create_structured_response` (already the
   `BadRequestError`/`ValidationError` fallback) now retries with
   `{"type": "json_object"}` instead of the `json_schema` that just failed, and
   injects the schema into the system prompt (`_inject_schema_hint`) so the model
   still knows the target shape. The caller repairs/validates the result.

No new config. Behavior is unchanged for providers that succeed natively; the
only added cost for broken providers is one already-failing native attempt.

## Testing

- Unit tests for `_strip_reasoning_and_fences` (14 cases) and
  `_inject_schema_hint` (6 cases), covering the regressions above (string values
  containing `<think>`/braces, object-then-prose, array-first, truncated
  reasoning, multimodal system content, schema-serialize failure).
- Verified end-to-end against a self-hosted llama.cpp server (Qwen3-27B): the
  deriver went from 0 → correct observations.

## Known limitations / out of scope

- **Deeply-nested JSON followed by prose** that happens to end with a matching
  close char is a residual edge; `raw_decode` handles the common cases, a full
  balanced parse is possible follow-up.
- **Injected schema is unbounded** — very large pydantic schemas can crowd small
  context windows.
- **The streaming path still uses `json_schema`** (out of scope here; the
  non-streaming deriver/dialectic is where all four issues live).
- **The `json_object` fallback forwards the original params** (incl. any
  reasoning/thinking-budget fields); a provider that also rejects those could
  `400` in the fallback. This is pre-existing behavior, called out for awareness.

Closes #716, closes #771, closes #743, closes #797.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of structured JSON responses from OpenAI, including better processing of wrapped outputs with reasoning tags and markdown formatting.

* **Tests**
  * Added comprehensive test coverage for structured output processing helpers.

### DIFF
--- a/src/llm/backends/openai.py
+++ b/src/llm/backends/openai.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 from collections.abc import AsyncIterator
 from typing import Any, cast
 
@@ -16,6 +17,103 @@ from src.llm.structured_output import (
 )
 
 logger = logging.getLogger(__name__)
+
+_THINK_RE = re.compile(r"<think>.*?</think>", re.DOTALL | re.IGNORECASE)
+
+
+def _json_payload(text: str) -> str | None:
+    """Return ``text`` if it parses as a JSON object/array, else None."""
+    try:
+        if isinstance(json.loads(text), (dict, list)):
+            return text
+    except (json.JSONDecodeError, ValueError):
+        pass
+    return None
+
+
+def _strip_reasoning_and_fences(text: str | None) -> str:
+    """Reduce a model response to its first complete JSON value.
+
+    OpenAI-compatible providers that don't honor ``response_format`` as a hard
+    grammar (Ollama, many self-hosted llama.cpp builds, chain-of-thought models
+    like GLM / MiniMax / DeepSeek) wrap structured output in ``<think>…</think>``
+    blocks and/or ```json fences, and sometimes append a natural-language
+    explanation after the JSON. ``repair_response_model_json`` then fails with
+    ``Expecting value: line 1 column 1 (char 0)`` and the deriver produces zero
+    observations. If no JSON structure is present the text is returned
+    unchanged."""
+    text = text or ""
+    # 1. Already a JSON object/array — return untouched. Done first so a value
+    #    that legitimately contains "<think>" or stray braces is never mangled.
+    payload = _json_payload(text)
+    if payload is not None:
+        return payload
+    # 2. Strip closed <think>…</think> blocks (and the surrounding fence/prose
+    #    they sit in), then retry.
+    cleaned = _THINK_RE.sub("", text).strip()
+    if (payload := _json_payload(cleaned)) is not None:
+        return cleaned
+    # 3. Extract the first complete JSON value starting at the first { or [.
+    #    raw_decode is string-aware and stops at the end of that value, so it
+    #    ignores ```json fences, trailing prose (incl. stray braces/brackets),
+    #    and unclosed reasoning that precedes real JSON.
+    start = min(
+        (i for i in (cleaned.find("{"), cleaned.find("[")) if i != -1),
+        default=-1,
+    )
+    if start != -1:
+        try:
+            _, end = json.JSONDecoder().raw_decode(cleaned, start)
+            return cleaned[start:end]
+        except (json.JSONDecodeError, ValueError):
+            # Truncated/invalid — hand the tail to repair_response_model_json.
+            return cleaned[start:]
+    return cleaned
+
+
+def _inject_schema_hint(
+    messages: list[dict[str, Any]],
+    response_format: type[BaseModel],
+) -> list[dict[str, Any]]:
+    """Append the target JSON schema to the system message.
+
+    ``json_object`` mode (unlike native ``json_schema``) does not tell the model
+    which fields to emit, so weaker / local models return empty or free-text
+    JSON. Injecting the schema makes the fallback reliable."""
+    try:
+        schema = json.dumps(response_format.model_json_schema())
+    except Exception as exc:
+        logger.warning(
+            "Could not serialize schema for %s; json_object fallback proceeds "
+            + "without a schema hint: %s",
+            getattr(response_format, "__name__", response_format),
+            exc,
+        )
+        return messages
+    hint = (
+        "You must respond with a single JSON object that conforms exactly to "
+        + "this JSON schema. Output only the JSON object — no prose, no markdown "
+        + "code fences, no <think> tags.\n\nSchema:\n"
+        + schema
+    )
+    # Shallow copy is sufficient: we replace a message's content, never mutate
+    # the original dicts or their nested values in place.
+    messages = [dict(m) for m in messages]
+    sys_idx = next(
+        (i for i, m in enumerate(messages) if m.get("role") == "system"), None
+    )
+    if sys_idx is None:
+        messages.insert(0, {"role": "system", "content": hint})
+        return messages
+    content = messages[sys_idx].get("content", "")
+    if isinstance(content, str):
+        messages[sys_idx]["content"] = f"{content}\n\n{hint}" if content else hint
+    elif isinstance(content, list):
+        # Multi-modal content: append the hint as an additional text part.
+        messages[sys_idx]["content"] = [*content, {"type": "text", "text": hint}]
+    else:
+        messages[sys_idx]["content"] = hint
+    return messages
 
 
 def _uses_max_completion_tokens(model: str) -> bool:
@@ -149,7 +247,7 @@ class OpenAIBackend:
                 truncated = exc.completion
                 raw_content = truncated.choices[0].message.content or ""
                 content = repair_response_model_json(
-                    raw_content,
+                    _strip_reasoning_and_fences(raw_content),
                     response_format,
                     model,
                 )
@@ -175,7 +273,7 @@ class OpenAIBackend:
             raw_content = response.choices[0].message.content or ""
             if parsed is None and raw_content:
                 content = repair_response_model_json(
-                    raw_content,
+                    _strip_reasoning_and_fences(raw_content),
                     response_format,
                     model,
                 )
@@ -381,13 +479,14 @@ class OpenAIBackend:
         response_format: type[BaseModel],
     ) -> Any:
         structured_params = dict(params)
-        structured_params["response_format"] = {
-            "type": "json_schema",
-            "json_schema": {
-                "name": response_format.__name__,
-                "schema": response_format.model_json_schema(),
-            },
-        }
+        # The native attempt already failed, often because the provider rejects
+        # or ignores json_schema. Retry with json_object (widely supported) and
+        # inject the schema so the model still knows the target shape; the
+        # response is repaired/validated by the caller.
+        structured_params["response_format"] = {"type": "json_object"}
+        structured_params["messages"] = _inject_schema_hint(
+            structured_params["messages"], response_format
+        )
         return await self._client.chat.completions.create(**structured_params)
 
     @staticmethod
@@ -398,7 +497,8 @@ class OpenAIBackend:
     ) -> BaseModel | str:
         raw_content = response.choices[0].message.content or ""
         if raw_content:
-            return repair_response_model_json(raw_content, response_format, model)
+            cleaned = _strip_reasoning_and_fences(raw_content)
+            return repair_response_model_json(cleaned, response_format, model)
         refusal = getattr(response.choices[0].message, "refusal", None)
         if refusal:
             return refusal
--- a/tests/llm/test_backends/test_openai_structured_output.py
+++ b/tests/llm/test_backends/test_openai_structured_output.py
@@ -1,0 +1,124 @@
+"""Unit tests for the OpenAI-backend structured-output helpers."""
+
+from __future__ import annotations
+
+import logging
+
+import pytest
+from pydantic import BaseModel
+
+from src.llm.backends.openai import (
+    _inject_schema_hint,  # pyright: ignore[reportPrivateUsage]
+    _strip_reasoning_and_fences,  # pyright: ignore[reportPrivateUsage]
+)
+
+
+class _Obs(BaseModel):
+    facts: list[str] = []
+
+
+# --------------------------------------------------------------------------- #
+# _strip_reasoning_and_fences
+# --------------------------------------------------------------------------- #
+@pytest.mark.parametrize(
+    "raw, expected",
+    [
+        # empty <think></think> + ```json fence (typical local llama.cpp output)
+        ('<think>\n\n</think>\n\n```json\n{"a": 1}\n```', '{"a": 1}'),
+        # closed reasoning block, JSON, then trailing prose
+        ('<think>r</think>{"ok": true} done.', '{"ok": true}'),
+        # bare ``` fence without the json language tag
+        ('```\n{"z": 9}\n```', '{"z": 9}'),
+        # already-valid object / array — fast path, returned untouched
+        ('{"facts": ["x", "y"]}', '{"facts": ["x", "y"]}'),
+        ('[{"a": 1}, {"b": 2}]', '[{"a": 1}, {"b": 2}]'),
+        # brace/bracket inside a string value must not confuse extraction
+        ('{"n": "a } b ] c"}', '{"n": "a } b ] c"}'),
+        # C-1 regression: a string value literally containing "<think>" must be
+        # preserved (fast path runs before any <think> stripping)
+        ('{"note": "a <think> tag in text"}', '{"note": "a <think> tag in text"}'),
+        # C1: object followed by prose containing the OTHER bracket type
+        ('{"k": "v"} list: [1, 2]', '{"k": "v"}'),
+        # M1: object followed by prose containing the SAME closing bracket
+        ('{"items": [1,2,3]} the value is 4}', '{"items": [1,2,3]}'),
+        # array first, trailing prose containing a stray "}"
+        ('[1, 2] and a dict: {"k": "v"}', "[1, 2]"),
+        # unclosed <think> (truncated) followed by recoverable JSON
+        ('<think>partial reasoning\n{"ok": true}', '{"ok": true}'),
+        # pure prose with no JSON is returned unchanged
+        ("I cannot help with that.", "I cannot help with that."),
+        # None / empty
+        (None, ""),
+        ("", ""),
+    ],
+)
+def test_strip_reasoning_and_fences(raw: str | None, expected: str) -> None:
+    assert _strip_reasoning_and_fences(raw) == expected
+
+
+# --------------------------------------------------------------------------- #
+# _inject_schema_hint
+# --------------------------------------------------------------------------- #
+def test_inject_appends_to_existing_system_message():
+    messages = [
+        {"role": "system", "content": "Extract facts."},
+        {"role": "user", "content": "hi"},
+    ]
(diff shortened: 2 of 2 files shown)


sethbrasile:
If this is a useful PR but you would like the coderabbit issues addressed, please let me know and I'm happy to oblige. Thank you!

akattelu:
Thanks @sethbrasile I think this is useful but https://github.com/plastic-labs/honcho/pull/820 also has an implementation of something really similar that we are favoring for now. We're hoping to merge that one very soon!

PR #832, opened 2026-06-22
fix(llm): support per-request provider timeouts

alvedder:
## Summary by CodeRabbit

* **New Features**
  * Added support for per-request `timeout` settings across supported LLM providers, including typed forwarding to provider SDK request options.
  * Updated the configuration guide and example with a per-request `timeout` example and recommended usage guidance for queued/worker paths.
* **Bug Fixes**
  * Improved `timeout` validation (rejects booleans, empty/non-numeric values, non-positive, and non-finite numbers) and normalizes valid strings to numeric seconds.
* **Tests**
  * Added/extended backend and request-builder tests to verify timeout normalization, forwarding, and validation for multiple providers and request modes.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -67,32 +67,34 @@ ThinkingEffortLevel = Literal[
 StructuredOutputMode = Literal["json_schema", "json_object"]
 
 
-PROVIDER_TIMEOUT_ERROR = "provider_params.timeout must be a positive number of seconds"
+PROVIDER_TIMEOUT_ERROR_TEXT = (
+    "provider_params.timeout must be a positive number of seconds"
+)
 
 
 def coerce_provider_timeout(value: Any) -> float:
     """Coerce a `provider_params.timeout` value to positive, finite seconds.
 
     Canonical implementation shared by config-load validation (here) and
-    per-request validation (`src.llm.backend.request_timeout_from_extra_params`,
+    per-request validation (`src.llm.request_builder.request_timeout_from_extra_params`,
     which translates the ValueError into a ValidationException). Lives in
     config.py because src.exceptions imports src.config, so config validators
     cannot raise Honcho exception types.
     """
     if isinstance(value, bool):
-        raise ValueError(PROVIDER_TIMEOUT_ERROR)
+        raise ValueError(PROVIDER_TIMEOUT_ERROR_TEXT)
     if isinstance(value, int | float):
         timeout = float(value)
     elif isinstance(value, str):
         try:
             timeout = float(value.strip())
         except ValueError as exc:
-            raise ValueError(PROVIDER_TIMEOUT_ERROR) from exc
+            raise ValueError(PROVIDER_TIMEOUT_ERROR_TEXT) from exc
     else:
-        raise ValueError(PROVIDER_TIMEOUT_ERROR)
+        raise ValueError(PROVIDER_TIMEOUT_ERROR_TEXT)
 
     if not math.isfinite(timeout) or timeout <= 0:
-        raise ValueError(PROVIDER_TIMEOUT_ERROR)
+        raise ValueError(PROVIDER_TIMEOUT_ERROR_TEXT)
     return timeout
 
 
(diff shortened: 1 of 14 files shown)


akattelu:
Hey @alvedder, thanks for opening this. The configurable timeout is a good idea and the 600s default is a safe match for the SDK default.

That said, I think this is at the wrong layer. The timeout is set in the client constructor, but clients are cached by `(base_url, api_key)` ([registry.py:68](https://github.com/plastic-labs/honcho/blob/main/src/llm/registry.py#L68)), so a single global value is all that layer can express. Different callers want different ceilings though. The Dialectic is on the synchronous chat path and wants a tight timeout, while the Deriver/Dreamer run off the queue and can tolerate much longer ones.

We already have a seam for this: `provider_params` on `ModelConfig` flows through `build_config_extra_params()` ([request_builder.py:18](https://github.com/plastic-labs/honcho/blob/main/src/llm/request_builder.py#L18)) into the `.create()` call, and the SDKs accept a per-request `timeout=`. So I'd lean toward a per-`ModelConfig` timeout rather than a global one. It leaves the cached clients shared and matches how temperature/max_tokens already work.

Also worth covering Anthropic/Gemini if we keep a global knob, otherwise it's inconsistent that only OpenAI is configurable.

Suggested shape: push the timeout down to `ModelConfig`/`provider_params`, applied per-request.

alvedder:
Thank you @akattelu for quick response! I'll correct this PR following your suggestions

alvedder:
@akattelu can you please review the updated PR?

src/llm/backend.py
```diff
@@ -45,6 +48,24 @@ class StreamChunk:
     output_tokens: int | None = None
 
 
+def request_timeout_from_extra_params(
```
eisene: nit: would prefer to keep this file just dataclasses and move this to `runtime.py` or something similar

consider renaming at some point `backend.py` -> `types.py`

src/llm/backends/gemini.py
```diff
@@ -289,19 +295,37 @@ def _build_config(
...
                 existing_headers.update(
                     coerce_passthrough_mapping("extra_headers", operator_extra_headers)
                 )
+                http_options.headers = existing_headers
+
+        timeout = request_timeout_from_extra_params(extra_params)
+        if timeout is not None:
+            if http_options is None:
+                http_options = genai_types.HttpOptions()
+            http_options.timeout = int(timeout * 1000)
```
eisene, line 325: nit: worth adding a short comment here explaining gemini API doesn't natively support this so we use the httpx version, as in `configuration.mdx`? don't feel super-strongly, but i feel it's a sharp edge that differs from the other endpoints (google is always special lol)

src/llm/request_builder.py
```diff
@@ -99,6 +104,14 @@ def build_config_extra_params(config: ModelConfig) -> dict[str, Any]:
     return extra_params
 
 
+def _normalize_extra_params(extra_params: dict[str, Any]) -> dict[str, Any]:
```
eisene, line 132: nit: would prefer to make this a bit more general, like `_remove_if_nulls(extra_params: dict, to_remove_if_null: list[str]` and have it go through `to_remove_if_null` and remove those keys if they're `None`. again don't feel strongly, my sense is that this is the kind of thing Claude will just proliferate all over without logically consolidating unless we keep a lid on it

src/config.py
```diff
@@ -66,6 +67,35 @@ def load_toml_config(config_path: str = "config.toml") -> dict[str, Any]:
 StructuredOutputMode = Literal["json_schema", "json_object"]
 
 
+PROVIDER_TIMEOUT_ERROR = "provider_params.timeout must be a positive number of seconds"
```
eisene: nit: `PROVIDER_TIMEOUT_ERROR_TEXT`

eisene (dismissed):
lgtm with 4 small nits

eisene (approved):
lgtm

PR #845, opened 2026-06-25
feat(telemetry): CloudEvents + Langfuse tracing as projections over a captured LLM stream

akattelu:
## What

LLM call tracing where each call is captured once (`CapturedLLMCall`) and fanned out to multiple exporters — **"one data model, two projections":**

- **CloudEvents trace stream** (`llm.call.traced` / `trace.content`) — feeds the parquet / ML substrate.
- **Langfuse** (`LangfuseExporter`) — reconstructs trace → run → step → generation from the same captured stream.

## Highlights

- **Capture seam** (`src/llm/capture.py`): single canonicalization + content-addressed hashing point; O(N) per-span memo so repeated context isn't re-hashed/re-clipped on the hot path.
- **Session correlation**: `session_id` (the opaque `Session.id`, namespaced only at the Langfuse boundary) threaded telemetry → captured call → exporters.
- **Span identity** consolidated onto `LLMTelemetryContext`; dropped the `TRACE_ENDPOINT` knob.
- Canonical generation/step names; dreamer branches nest under one dream trace; tool calls become spans under their step.
- **Dedup registries** (`trace_session`, `langfuse_session`) bounded by an LRU so dedup/grouping survive long-running workers.
- Embedding-call tracing; deterministic high-volume sampling.

## Config / rollout

- `LANGFUSE_EXPORTER_MODE` — default `exporter`; `inline` keeps the legacy live `@observe` path one release for side-by-side validation (the env-flip rollback).
- Code defaults stay **off** (`TELEMETRY_ENABLED` / `TELEMETRY_TRACE_PAYLOADS` = false) so self-hosted doesn't auto-emit. Prod enables tracing by setting `TELEMETRY_TRACE_PAYLOADS=true` (telemetry endpoint already configured for billing).

## Testing

Added unit tests and verified unified test e2e runs in langfuse

## Summary by CodeRabbit

* **New Features**
  * Added opt-in “full-fidelity” trace payload capture for LLM and embedding calls, with purpose allowlisting and payload size limits.
  * Introduced replay-grade captured-call tracing export with trace-content deduplication (including Langfuse projection).
  * Added trace viewer configuration and a Langfuse exporter mode switch (inline vs exporter).

* **Bug Fixes**
  * Improved trace/span/session correlation across single-shot, agentic, tools, embeddings, and streamed generations; reduced trace collisions and duplicate content.

* **Tests**
  * Expanded/updated coverage for capture/export, truncation & hashing, Langfuse nesting, and cross-agent trace semantics.


src/llm/capture.py
```diff
@@ -0,0 +1,344 @@
...
+    max_bytes = settings.TELEMETRY.TRACE_MAX_BYTES
+    if max_bytes <= 0 or not isinstance(content, str):
+        return content, False
+    encoded = content.encode("utf-8")
+    if len(encoded) <= max_bytes:
+        return content, False
+    marker = "…[truncated]"
+    keep = max(0, max_bytes - len(marker.encode("utf-8")))
+    clipped = encoded[:keep].decode("utf-8", errors="ignore") + marker
+    return clipped, True
```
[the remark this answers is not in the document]
akattelu: Skipping — leaving structured content intact is intentional (preserves the replay structure and keeps `content_hash` stable), and it's locked by `test_leaves_structured_content_intact`; clipping structured payloads risks corrupting them. Tracking oversized-payload hardening as a separate follow-up.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

src/llm/types.py
```diff
@@ -81,6 +114,39 @@ class LLMTelemetryContext:
...
+        Internally `_telemetry_for_iteration` sets `parent_span_id` to the run
+        span's own id (== this span_id) so it doubles as the "inside a run"
+        signal for Langfuse nesting (see `runtime.py`). But a span that is its
+        own parent is just a root: emitting `parent_span_id == span_id` makes
+        span-tree consumers treat the root as a child of itself. So for exported
+        trace data, normalize that to None. The in-memory field is unchanged —
+        the Langfuse `inside_run` check still reads `parent_span_id` directly.
+        """
+        pid = self.parent_span_id
+        return None if pid is not None and pid == self.span_id else pid
```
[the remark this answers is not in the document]
akattelu: Skipping — no live trigger: every telemetry context in the codebase sets `span_id` (dialectic/dreamer/summarizer), so the `pid == span_id` check never misses. The `run_id` fallback only matters if `span_id` is `None`, which doesn't occur today. Happy to switch to `span_identity()` for consistency if we ever add a `span_id`-less path.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

src/telemetry/events/trace.py
```diff
@@ -0,0 +1,185 @@
...
+    # --- Accounting copy ---
+    # v1: input tokens are a tiktoken ESTIMATE (no authoritative provider count is
+    # plumbed yet); output tokens are always 0 (embeddings produce none).
+    provider_input_tokens: int = 0
+    provider_output_tokens: int = 0
+    input_count: int = 0
+    was_truncated: bool = False
+
+    def get_resource_id(self) -> str:
+        return f"{self.span_id}:embedding:{self.call_purpose}:{self.input_count}"
```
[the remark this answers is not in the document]
akattelu: The suggested fix doesn't apply: embeddings don't run in a tool loop, so there's no meaningful `iteration`/`step_seq` for `embedding_client` to supply (both stay at defaults), and adding them to the key wouldn't change anything. The real concern (multiple same-shape embeddings in one span conflating) is valid but needs a per-run embedding sequence counter — filing as a follow-up rather than a no-op key change.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

src/dialectic/core.py
```diff
@@ -81,9 +82,12 @@ def __init__(
             observed_peer_card: Biographical information about the observed peer
             metric_key: Optional key for logging metrics (if provided, agent won't log separately)
             reasoning_level: Level of reasoning to apply
+            session_id: Opaque Session.id (nanoid PK, NOT session_name) used as the
+                trace session grouping key. None for global queries.
```
Rajat-Ahuja1997: nit: i would just clean up some of these generated comments/docstrings. there are some other instances in this PR, but a lot of these comments reflect the diff relative to a prior state instead of describing what the code does for the next reader.

src/llm/capture.py
```diff
@@ -0,0 +1,344 @@
+"""Single-capture point for replay-grade LLM call content + the exporter seam.
+
+The executor boundary (`honcho_llm_call_inner`) is the one place that holds the
+normalized `messages` array and the provider response together. Historically
+three consumers each re-serialized that content their own way (Langfuse, the
+JSONL reasoning trace, and — new in the CloudEvents tracing spec — the
+`llm.call.traced` / `trace.content` events). Divergent canonicalization would
+produce divergent `content_hash` values and silently break the O(N)
```
Rajat-Ahuja1997: same point here: re cleaning up the docstring

.env.template
```diff
@@ -276,6 +276,12 @@ LLM_OPENAI_API_KEY=your-api-key-here
 # TELEMETRY_MAX_BUFFER_SIZE=10000
 # TELEMETRY_NAMESPACE=honcho  # Inherits from NAMESPACE if not set
 
+# Full-fidelity payload tracing (llm.call.traced / trace.content). Default-off
+# and additive — leaving TRACE_PAYLOADS unset changes nothing.
+# TELEMETRY_TRACE_PAYLOADS=false  # Trace events ship to TELEMETRY_ENDPOINT (Xatu), on a separate buffer
```
Rajat-Ahuja1997: nit: maybe TELEMETRY_TRACE_PAYLOADS_ENABLED?

src/embedding_client.py
```diff
@@ -121,6 +122,29 @@ def _publish_embedding_event(
...
+        # embedding to the driving agent run: dialectic sets run_id == trace_id
+        # == span_id, so an embedding made in its prefetch nests under that
+        # run's trace and carries its session.
+        if settings.TELEMETRY.TRACE_PAYLOADS:
+            from src.telemetry.events import EmbeddingCallTracedEvent, emit_trace
+
+            run_id = get_embedding_run_id()
+            emit_trace(
+                EmbeddingCallTracedEvent(
+                    trace_id=run_id,
```
Rajat-Ahuja1997: we should make the parent_span_id here the run_id and then make span_id a new nanoid. an agent run can create several different embeddings and all would have the same run_id

src/telemetry/events/__init__.py
```diff
@@ -184,6 +215,18 @@ async def initialize_telemetry_events() -> None:
     from src.config import settings
     from src.telemetry.emitter import initialize_emitter
 
+    # Langfuse as a projection over the captured LLM stream. Registered
+    # independently of CloudEvents telemetry (it historically ran inline,
+    # decoupled from the CloudEvents emitter) — so it works even when
+    # TELEMETRY.ENABLED is off. Registering the exporter makes has_exporters()
```
Rajat-Ahuja1997: "Works even when TELEMETRY.ENABLED" is not currently true per the caller of initialize_telemetry_events so this is a little misleading

src/telemetry/trace_session.py
```diff
@@ -0,0 +1,83 @@
...
+linear: the content-addressed event id folds no timestamp, but a naive producer
+still ships one ``trace.content`` per message per iteration. In an agentic loop
+iteration N's context window *contains* iteration N−1's, so that's O(N²) on the
+wire. This module tracks which hashes a run has already shipped so each unique
+message ships exactly once per run.
+
+The set is bounded by an LRU over runs: once more than `_MAX_RUNS` runs are
+tracked, the least-recently-touched one is evicted (almost always a run that has
+already finished), so dedup keeps working for active runs no matter how many the
+process has handled. A single run exceeding `_MAX_HASHES_PER_RUN` unique messages
```
Rajat-Ahuja1997: would just make this more readable

src/telemetry/trace_session.py
```diff
@@ -0,0 +1,83 @@
...
+            _record_overflow("max_hashes")
+            return True
+        seen.add(content_hash)
+        return True
+
+
+def end_run(run_key: str) -> None:
+    """Free a run's hash set. Safe to call for an unknown key."""
+    with _lock:
+        _runs.pop(run_key, None)
```
Rajat-Ahuja1997: currently unused

src/telemetry/trace_exporter.py
```diff
@@ -0,0 +1,176 @@
...
+                tool_call_id=tool_call_id,
+                honcho_authored=honcho_authored,
+            )
+        )
+
+    @staticmethod
+    def _emit_trace(event: LLMCallTracedEvent | TraceContentEvent) -> None:
+        from src.telemetry.events import emit_trace
+
+        emit_trace(event)
```
Rajat-Ahuja1997: nit: don't think we need this helper

src/telemetry/trace_exporter.py
```diff
@@ -0,0 +1,176 @@
...
+                finish_reason=call.finish_reason,
+                provider_input_tokens=call.input_tokens,
+                provider_output_tokens=call.output_tokens,
+                cache_read_tokens=call.cache_read_tokens,
+                cache_creation_tokens=call.cache_creation_tokens,
+                was_truncated=was_truncated,
+            )
+        )
+
+    def _emit_derived_content(
```
Rajat-Ahuja1997: make `_emit_hashed_content`? considering the deriver is a honcho primitive, i initially thought that this only emitted content related to the deriver

src/llm/types.py
```diff
@@ -214,14 +294,36 @@ async def _iterate_with_usage_capture(
...
-                handle.end(output="".join(accumulated_text) or None)
+                handle.end(output=text or None)
+            # Finalize the content capture with the full streamed text. Even a
+            # partial/aborted stream captures, tagged with the right outcome.
+            finalizer = self._capture_finalizer
+            if finalizer is not None:
+                self._capture_finalizer = None
+                finish_reason = (
+                    "stop"
+                    if stream_error is None
```
Rajat-Ahuja1997, line 284: should we capture the actual finish reason from the llm chunks?

src/llm/capture.py
```diff
@@ -0,0 +1,344 @@
...
+        key = id(message)
+        cached = memo.get(key) if memo is not None else None
+        if cached is not None:
+            captured.append(cached)
+            any_truncated = any_truncated or cached.truncated
+            continue
+        role = str(message.get("role", ""))
+        raw_content = message.get("content")
+        tool_call_id = message.get("tool_call_id")
+        content, truncated = clip_for_trace(raw_content)
```
Rajat-Ahuja1997, line 252: should we just store the full `message` object and truncate `message.content`? per claude, we're currently excluding tool calls made by openai and gemini (which don't record tool calls in the content field).

haven't tested this personally but let's look into it before shipping

Rajat-Ahuja1997 (commented):
generally looks pretty good - i think the overall structure makes sense

src/llm/capture.py
akattelu: I ended up verifying this issue and just pushed another fix by unifying them all into a standard format

Rajat-Ahuja1997 (approved):
lgtm

Issue #843, opened 2026-06-24
Bug: get_openai_override_client() creates AsyncOpenAI without explicit timeout - intermittent ConnectTimeout errors when upstream provider is briefly unreachable


FabioBonfiglio:
## Summary

`get_openai_override_client()` in `src/llm/registry.py` creates an `AsyncOpenAI` client **without an explicit `timeout` parameter**. The OpenAI SDK v2.x default is `Timeout(connect=5.0, read=600.0, write=600.0, pool=600.0)`.

When the upstream LLM provider (any OpenAI-compatible endpoint, including self-hosted proxies, vLLM, OpenRouter, etc.) becomes briefly unreachable — even for just 5–10 seconds — the **5-second connect timeout** fires, producing `httpcore.ConnectTimeout` -> `openai.APITimeoutError`. Honcho then retries via tenacity (3 attempts with exponential backoff), but if the provider is still unreachable for ~15s total, all retries fail and the call surfaces as an unhandled `RetryError`.

## Environment

- Honcho: v3.0.7 (self-hosted, Docker)
- LLM provider: OpenAI-compatible proxy on localhost:8787 (forwarding to a TEE-encrypted inference endpoint)
- OpenAI SDK: v2.36.0
- Python: 3.13

## Root cause

In `src/llm/registry.py`, lines 69-73:

```python
@lru_cache(maxsize=128)
def get_openai_override_client(
    base_url: str | None, api_key: str | None
) -> AsyncOpenAI:
    """OpenAI client for a specific (base_url, api_key) pair. Cached by key."""
    return AsyncOpenAI(api_key=api_key, base_url=base_url)  # no timeout param
```

The same applies to `get_openai_client()` (line 47) and the module-level `CLIENTS["openai"]` entry (line 132).

The OpenAI SDK default `connect=5.0s` is too short for OpenAI-compatible providers that may have higher latency than `api.openai.com` itself (self-hosted proxies, regional providers, providers behind CDN/load balancers with cold-start behavior).

## Observed impact

In a self-hosted deployment using a local proxy to an OpenAI-compatible provider, we observed **749 `APITimeoutError` instances in a 12-hour window**, all caused by `httpcore.ConnectTimeout` (connect timeout, not read timeout). The pattern is diurnal and correlates with upstream provider load peaks.

The failures cascade through Honcho:
- `src.dialectic.core` - `Failed to prefetch observations: Request timed out.`
[...]

PR #853, opened 2026-06-27
Fix #839 and #843: add explicit timeout to AsyncOpenAI and emit struc…

### LINKED ISSUE #839: Deriver silently marks work units processed=true when LLM response fails to parse — need louder observability
## Is your enhancement request related to a problem? Please describe.

When the deriver parses its LLM response and the result is empty, the failure<br>mode is **completely silent from the operator's perspective**. You only see<br>two logs and they look like routine behaviour:

```
src.utils.json_parser - ERROR - ❌ Repair failed: Expecting value: line 1 column 1 (char 0)
src.deriver.deriver - WARNING - Deriver generated zero observations for messages 1:2 in <workspace>/<session>!
```

That's it. The queue item gets `processed=true`, the LLM call returns in 5-15s<br>with HTTP 200, and the `documents` table stays empty. No metric, no trace<br>span error, no alertable signal. An operator only finds out something is<br>wrong by:

1. Querying Postgres and noticing the `documents` table hasn't grown
2. Searching logs for the WARNING string
3. Manually correlating the WARNING with the ERROR above it
4. Realising the ERROR's "empty string" + the WARNING's "zero observations"<br>together mean the entire pipeline just binned a message silently

This is exactly the failure mode from plastic-labs/honcho#780 — but the silent-zero outcome also<br>affects *any* future structured-output regression, not just the<br>`json_schema`/`json_object` mismatch. The current signal is easy to miss<br>during normal log volume, and on a busy Honcho deployment the WARNING gets<br>buried.

For a background worker whose entire job is to produce observations, "zero<br>observations, queue item processed" is the most important failure signal in<br>the system, and it should never be a WARNING.

## Describe the solution you'd like

Two complementary improvements (both small, both follow the "harvestable<br>improvement" framing from the plastic-labs/honcho#820 closing comment):

### 1\. Structured "deriver produced zero observations from a parse failure" event

Today the path is: `validate_and_repair_json` → `repair_json` returns `""` →<br>`comprehensive_json_repair` raises → that gets swallowed somewhere upstream<br>(seen in logs as a single ERROR, then a separate WARNING). Replace the<br>swallowed exception + free-text WARNING with a single structured log record<br>(or counter increment, depending on what's preferred for metrics scraping)<br>that includes:

* `work_unit_id` and `peer_id` / `session_id`
* The transport that was used (`openai` / `anthropic` / `gemini`) and the<br>model name
* The structured_output_mode that was requested (`json_schema` /<br>`json_object`)
* The parse failure reason (`empty_after_repair` /<br>`validation_error` / `bad_request`) — distinguishable, not just "ERR"
* A truncated excerpt of the raw LLM response (first \~500 chars) so the<br>operator can see *what* the model actually returned without enabling debug<br>logs

That record is alertable, count-able, and grep-able in a way that the<br>current WARNING + swallowed ERROR isn't.

### 2\. Don't mark the work unit as `processed=true` when parse failed

If the LLM call returned HTTP 200 but the parse failed, leave the work unit<br>in `processed=false` (or move it to a `failed_parse` state if that's easier<br>to model). That way:

* The work unit will be retried naturally (or held for inspection)
* An operator who restarts the deriver sees the stuck queue immediately
* The bug becomes self-reporting instead of silent

This is the actual "loud" part — the queue itself becomes the alarm.

## Describe alternatives you've considered

* **Bump the WARNING to ERROR and add a** `failed=true` **counter** — minimal<br>change, gets 80% of the value, but operators still have to grep for the<br>string and it doesn't fix the "marked processed but actually broken"<br>problem.
* **Auto-fallback to** `json_object` **on** `BadRequestError` (the other<br>follow-up mentioned in the plastic-labs/honcho#820 closing comment) — also valuable but a<br>bigger design discussion about how to detect "this provider ignores<br>structured output" generically. Worth its own issue. I'd file it as one<br>if you want.
* **Just re-emit the LLM response into the work unit's metadata** — useful<br>for forensics but doesn't fix the "no signal" problem.

## Additional context

* Originating report: plastic-labs/honcho#780 (closed in plastic-labs/honcho#820, partial fix landed). The<br>closing comment from @plastic-labs/akattelu explicitly called this out<br>as a follow-up worth filing.
* Repro is one message through any OpenAI-compatible provider that ignores<br>`json_schema`. The current `latest` image at `ghcr.io/plastic-labs/honcho:latest`<br>on a fresh Honcho deployment against Ollama cloud `ministral-3:8b` produces<br>zero observations with no signal after several minutes of activity.
* After upgrading to the plastic-labs/honcho#820 build, switching to<br>`DERIVER_MODEL_CONFIG__STRUCTURED_OUTPUT_MODE=json_object` does fix the<br>*production* outcome (observations flow again), but the *silent* outcome<br>remains: a future structured-output regression on any provider would<br>reproduce plastic-labs/honcho#780's silent-failure mode without anyone noticing until they<br>manually check the documents table.
* Self-hosted stack: Podman rootless, 4 containers (`api`, `deriver`,<br>`database`, `redis`), all data on host volumes, deriver set to batch<br>(`DERIVER_FLUSH_ENABLED=false`). Pre-fix baseline is `documents` row count<br>stalling — exact same diagnostic the original plastic-labs/honcho#780 report described.

jashwanth-reddy-g:
…tured logs on unparseable LLM output

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of structured output parsing failures with clearer errors and better logging details.
  * Requests that exceed the provider’s structured-output support now fail explicitly instead of silently falling back.
  * Added longer timeout handling for OpenAI-based requests to reduce premature timeouts.

### DIFF
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -6,6 +6,7 @@ from src.config import ConfiguredModelSettings, settings
 from src.crud.representation import RepresentationManager
 from src.dependencies import tracked_db
 from src.llm import honcho_llm_call
+from src.llm.structured_output import StructuredOutputError
 from src.llm.types import LLMTelemetryContext
 from src.models import Message
 from src.schemas import ResolvedConfiguration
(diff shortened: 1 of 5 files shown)


akattelu:
Hey @jashwanth-reddy-g , since #832 is merged I believe you should be able to set timeout on a per-model basis now. The observability on `main` should also be better now. Closing this.

Issue #569, opened 2026-04-17
Deriver loses an entire batch of observations when one exceeds the embedding token limit


lucharo:
`src/crud/representation.py` collects all observations from a single deriver pass, then calls `embedding_client.simple_batch_embed(observation_texts)`. If even one observation in that batch exceeds the embedding provider's token cap, `simple_batch_embed` raises a `ValueError` and `save_representation` re-raises a `ValidationException`. The deriver catches it at `src/deriver/deriver.py:205` and logs `Failed to save representation for observer X: ...`. **None** of the observations from that pass are saved — not just the long one.

`simple_batch_embed` itself doesn't have a chunking path (only the more elaborate `batch_embed` does, which is used for messages but not observations).

This bites in practice as soon as one observation contains a long pasted artifact — a PR review, a verbose tool output, a stack trace. With OpenAI's 8192 cap it's relatively rare; with Gemini, where `embedding_client.py:44` caps `max_embedding_tokens` at 2048, it's frequent. I see it in production on a self-hosted instance.

The error message also reports `settings.MAX_EMBEDDING_TOKENS` (8192) instead of the provider's actual cap, which compounds the related bug in plastic-labs/honcho#568.

## Suggested fix

Add a `truncate=True` knob to `simple_batch_embed` for storage callers:

```python
async def simple_batch_embed(
    self,
    texts: list[str],
    *,
    truncate: bool = False,
) -> list[list[float]]:
    if truncate:
        texts = [self._truncate_to_token_limit(t) for t in texts]
    ...

def _truncate_to_token_limit(self, text: str) -> str:
    tokens = self.encoding.encode(text)
    if len(tokens) <= self.max_embedding_tokens:
        return text
    margin = 16
    keep = max(1, self.max_embedding_tokens - margin)
    return self.encoding.decode(tokens[:keep])
```

Then update the storage callers to pass `truncate=True`:

* `src/crud/representation.py` — the most critical one, every deriver run
* `src/crud/document.py:743`
* `src/reconciler/sync_vectors.py:191` and `:305`

Query callers (`src/utils/agent_tools.py:971`) keep the default so users still see a real error if their query is too long.

Truncation loses the tail of very long observations but keeps them indexed; failing the whole save loses everything. The trade-off favors truncation for storage paths, especially for the deriver where individual observation length is unpredictable.

Happy to open a PR. I'm running this exact change locally and the deriver no longer drops observations.

PR #854, opened 2026-06-27
fix(promotion): intent-aware chunking + per-chunk edges for oversized observations

hermes-tmw:
## What
- Detects oversized observations before embedding and splits them at intent-aware boundaries (paragraph → sentence → clause → word) instead of truncating.
- Embeds each chunk independently and runs vector-similarity search per chunk.
- Merges chunk-level candidates so a multi-intent observation can form edges to several semantically distinct clusters without duplicate edges.
- Adds a summary-embedding fallback for single dense blocks that survive all intent-aware splits (telemetry purpose `SUMMARY_SHORT`; full LLMinal L1 compression remains Fix 5).
- Keeps `_embed_observation` as a backward-compatible single-vector alias.

## Verification
- `tests/test_promotion_worker.py` passes (7/7), including the new multi-topic chunking test.
- Anti-pattern scanner reports no findings in changed files.
- Targeted schema + scheduler suites pass.

## Summary by CodeRabbit

* **New Features**
  * Added graph memory capabilities, including recall, contexts, thread bindings, pinning, verification, cold storage, and rehydration.
  * Introduced promotion handling so observations can be automatically evaluated and promoted into longer-term memory.
  * Added support for scheduled promotion and access-log compaction in the background.

* **Documentation**
  * Added setup, operations, troubleshooting, and API guides for the new graph memory workflows.

* **Tests**
  * Added extensive coverage for graph memory, recall, promotion, and scheduler behavior.


akattelu:
Hey @hermes-tmw, thanks for the deep work here. This grew well past the bug it references: roughly 8k added lines with new tables, a router, migrations, and a promotion worker, and it doesn't touch src/crud/representation.py or src/embedding_client.py, so #569 reproduces identically after merging it. The chunking and graph-memory direction might have merit on its own, but that scale of change needs a design discussion rather than a bugfix PR. Closing this one; #864 is the targeted fix for #569. Happy to discuss the graph-memory idea in an issue if you want to write it up.

PR #860, opened 2026-06-30
fix(deriver): surface representation save failures instead of silent loss (#728)

### LINKED ISSUE #728: Deriver silently marks representation work units `processed` when the embedding save fails on a transient 429 (no retry, no fail-loud → silent memory loss)
## Summary

When `save_representation` fails because the embedding call hits a transient `429 RESOURCE_EXHAUSTED`, the deriver **swallows the exception, logs it, and then marks the queue work unit `processed` with zero documents saved**. The memory for that turn is permanently lost with no error state, no retry, and no alert — the unit looks successfully processed.

We hit this during a re-derivation burst: ~11 representation units logged `Failed to save representation for observer X: 429 RESOURCE_EXHAUSTED`, were marked `processed`, and saved nothing. Document count dropped ~300 → ~121 with no errored units and no crash.

## Root cause (three compounding factors)

1. **No retry on the deriver's embedding path.** `RepresentationManager.save_representation` (`src/crud/representation.py:106`) calls `embedding_client.simple_batch_embed(...)`. That path (`_EmbeddingClient.simple_batch_embed` in `src/embedding_client.py`) has **no retry**, unlike the bulk `_process_batch` path which retries 3×. The surrounding `try` in `save_representation` only catches `ValueError` (token-limit) — a `google.genai.errors.APIError` with `code == 429` propagates straight up.

2. **The deriver swallows the failure.** In `process_representation_tasks_batch` (`src/deriver/deriver.py`), the per-observer save loop catches `except Exception`, logs `"Failed to save representation for observer %s: %s"`, and **returns normally**. `successful_observer_count` is computed but used only for telemetry — never to decide success. Control returns to `queue_manager` (`src/deriver/queue_manager.py`), which calls `mark_queue_items_as_processed`. The unit is now **processed with nothing saved**.

3. **Errored ≠ retried, so there's no recovery either.** `mark_queue_item_as_errored` (`queue_manager.py`) sets `processed=True` **and** `error=…`; the polling loop selects `WHERE processed=false`. So even if the unit *were* errored, it would be dead-lettered (retained `QUEUE_ERROR_RETENTION_SECONDS`), **never auto-retried** — there is no queue-level retry / attempt counter. The "reconciliation job" only re-embeds docs that were *committed* with a pending vector sync; it does not cover an embedding failure that happened *before* any document was created.

Net: a transient, server-side embedding 429 (generic `"Resource exhausted"`, no `QuotaFailure` — i.e. capacity, not a quota cap) causes silent, permanent memory loss.

## Impact

Transient Gemini embedding rate limits are not rare. Each one can silently drop a turn's derived memory while the work unit reports success. There is no signal (no error row, no exception, no metric divergence) that memory went missing.

## Proposed fix (layered)

1. **Retry the embedding call on transient 429 / RESOURCE_EXHAUSTED** on the `simple_batch_embed` path (tenacity, exponential backoff + jitter, ~5 attempts, `reraise=True`), so a momentary capacity blip self-heals and a persistent failure still propagates.
2. **Fail loud on total save failure** in the deriver: after the observer loop, if `successful_observer_count == 0` (and there were observations), `raise` so the unit is marked **errored** (recorded/alertable) rather than silently `processed`. Partial success (≥1 observer) can still be accepted to avoid duplicate re-derivation.
3. *(Optional, larger)* a **queue-level retry / attempt counter** for errored representation units, since today an errored unit is dead-lettered with no automatic retry.

We are running (1) + (2) as a downstream patch and would be happy to open a PR if useful.

## Environment

- Honcho `v3.0.7` (`7470866`), host-native install, Postgres + pgvector.
- Embeddings: `gemini-embedding-001` @ 1536 dims, direct to `generativelanguage.googleapis.com` (paid key). The 429 was transient server capacity, confirmed by a follow-up probe (60 concurrent embeds → all 200).

leo922oel:
## Description

When every observer's representation save failed (e.g. embedding `429 RESOURCE_EXHAUSTED` past the in-batch retries), `process_representation_tasks_batch`  <br>(`src/deriver/deriver.py`) caught the exception, logged it, and returned normally —  <br>so the queue marked the work unit `processed` with **zero documents saved**:  <br>silent, permanent memory loss with no error state or alertable signal. (The errored  <br>path `mark_queue_item_as_errored` existed but was never reached.)

This makes a save failure either loud or recorded — never a silent success:

* **Total failure → errored.** Collect per-observer errors; raise when no observer  <br>succeeded, so the queue marks the unit `errored`.
* **Partial failure → processed + recorded.** Observers are independent  <br>`(observer, observed)` collections, so the saved ones are kept (re-running would  <br>duplicate; the queue has no requeue). The failure is recorded via a new additive  <br>`failed_observer_count` on `RepresentationCompletedEvent` (schema unchanged).

Not a retry fix: the embedding client already retries transient errors; the bug is  <br>the swallow after they're exhausted.

**Quantified** (`tests/bench/bench_deriver_silent_loss.py` — real deriver path with  <br>a faked failing save, trials=20):

| Metric | Before (`main`) | After |
| -- | -- | -- |
| `silent_loss_rate` | 100% | 0% |
| `failure_visibility_rate` | 0% | 100% |

plastic-labs/honcho#728 reported docs dropping < />300 → < />121 (< />60%) in a 429 burst. Supersedes [#660](<https://github.com/plastic-labs/honcho/issues/660>)  <br>(raised on any failure, no tests); coordinates with plastic-labs/honcho#839 (@akattelu).

## Linked issues

Fixes plastic-labs/honcho#728

## Type of change

- [X] `fix` — bug fix
- [X] `test` — adding or correcting tests

## Test plan

New / updated tests (RED→GREEN):

* `tests/deriver/test_deriver_processing.py`
  * `test_all_observer_saves_failing_surfaces_failure` — every observer save  <br>raises ⇒ `process_representation_tasks_batch` raises (queue marks errored)  <br>instead of returning silently.
  * `test_partial_observer_failure_is_processed_and_surfaced` — one observer  <br>succeeds, one fails ⇒ no raise (unit stays processed), and the emitted  <br>`RepresentationCompletedEvent` reports `observer_count == 1` and  <br>`failed_observer_count == 1`.
* `tests/telemetry/test_representation_v2_fields.py` — asserts the new  <br>`failed_observer_count` defaults to 0 (additive; schema stays at v2).

Quantitative before/after (a standalone harness, not a CI test):

```bash
# AFTER (this branch)
PYTHONPATH=. uv run python tests/bench/bench_deriver_silent_loss.py
# BEFORE (main source)
git checkout main -- src/deriver/deriver.py src/telemetry/events/representation.py
PYTHONPATH=. uv run python tests/bench/bench_deriver_silent_loss.py
git checkout HEAD -- src/deriver/deriver.py src/telemetry/events/representation.py
```

## Checklist

- [X] Tests added/updated for the change
- [X] Existing tests pass (`tests/deriver/` + `tests/telemetry/` green; 4 unrelated  <br>failures in `test_auth_route_policy.py` / `test_document.py` dedup require a  <br>local embedding API key and fail identically on `main`)
- [X] `ruff` and `basedpyright` clean on every changed file
- [X] Commit messages follow Conventional Commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented representation processing from completing silently when all observer saves fail.
  * When every observer save fails, the batch now surfaces an explicit error; partial failures continue to emit accurate outcomes.
* **New Features**
  * Added a new completion telemetry field, `failed_observer_count`, to report observer save failures.
* **Tests**
  * Added coverage for all-failure vs partial-failure behavior and telemetry expectations.
  * Added a benchmark to measure “silent loss” and failure visibility rate.

### DIFF
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -196,6 +196,7 @@ async def process_representation_tasks_batch(
 
     agg_representation_result = crud.CreateDocumentsResult()
     successful_observer_count = 0
+    save_errors: list[str] = []
     if observations.is_empty() or not message_ids:
         logger.warning(
             "Deriver generated zero observations for messages %s:%s in %s/%s!",
@@ -240,6 +241,7 @@ async def process_representation_tasks_batch(
                 logger.error(
                     "Failed to save representation for observer %s: %s", observer, e
                 )
+                save_errors.append(f"{observer}: {e.__class__.__name__}: {e}")
 
     # Log metrics
     overall_duration = (time.perf_counter() - overall_start) * 1000
@@ -339,3 +341,12 @@ async def process_representation_tasks_batch(
             semantic_dup_replaced_count=agg_representation_result.semantic_dup_replaced_count,
         )
     )
+
+    # If every observer's save failed, surface the failure to the queue manager so the
+    # work unit is marked errored instead of silently processed with zero documents saved
+    # (#728). Raised after telemetry so metrics still record the attempt.
+    if save_errors and successful_observer_count == 0:
+        raise RuntimeError(
+            f"save_representation failed for all {len(save_errors)} observer(s): "
+            + "; ".join(save_errors)
+        )
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -7,6 +7,7 @@ from src import crud
 from src.config import ConfiguredModelSettings, settings
 from src.crud.representation import RepresentationManager
 from src.dependencies import tracked_db
+from src.exceptions import RepresentationSaveError
 from src.llm import honcho_llm_call
 from src.llm.types import LLMTelemetryContext
 from src.models import Message
@@ -196,7 +197,7 @@ async def process_representation_tasks_batch(
 
     agg_representation_result = crud.CreateDocumentsResult()
     successful_observer_count = 0
-    save_errors: list[str] = []
+    save_errors: list[tuple[str, Exception]] = []
     if observations.is_empty() or not message_ids:
         logger.warning(
             "Deriver generated zero observations for messages %s:%s in %s/%s!",
@@ -241,7 +242,7 @@ async def process_representation_tasks_batch(
                 logger.error(
                     "Failed to save representation for observer %s: %s", observer, e
                 )
-                save_errors.append(f"{observer}: {e.__class__.__name__}: {e}")
+                save_errors.append((observer, e))
 
     # Log metrics
     overall_duration = (time.perf_counter() - overall_start) * 1000
@@ -347,7 +348,11 @@ async def process_representation_tasks_batch(
     # work unit is marked errored instead of silently processed with zero documents saved
     # (#728). Raised after telemetry so metrics still record the attempt.
     if save_errors and successful_observer_count == 0:
-        raise RuntimeError(
-            f"save_representation failed for all {len(save_errors)} observer(s): "
-            + "; ".join(save_errors)
+        details = "; ".join(
+            f"{observer}: {exc.__class__.__name__}: {exc}"
+            for observer, exc in save_errors
         )
+        raise RepresentationSaveError(
+            f"save_representation failed for all {len(save_errors)} observer(s): "
+            + details
+        ) from save_errors[0][1]
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -239,8 +239,8 @@ async def process_representation_tasks_batch(
                 )
                 successful_observer_count += 1
             except Exception as e:
-                logger.error(
-                    "Failed to save representation for observer %s: %s", observer, e
+                logger.exception(
+                    "Failed to save representation for observer %s", observer
                 )
                 save_errors.append((observer, e))
 
--- a/src/deriver/deriver.py
+++ b/src/deriver/deriver.py
@@ -339,6 +339,7 @@ async def process_representation_tasks_batch(
             exact_dup_in_batch_count=agg_representation_result.exact_dup_in_batch_count,
             semantic_dup_rejected_count=agg_representation_result.semantic_dup_rejected_count,
             semantic_dup_replaced_count=agg_representation_result.semantic_dup_replaced_count,
+            failed_observer_count=len(save_errors),
         )
     )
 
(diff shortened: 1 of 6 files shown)


theohsiung (approved):
@VVoruganti LGTM! 

Two small **non-blocking** notes for a future follow-up (not this PR's scope):

- The raise uses a bare `RuntimeError`; a typed exception from `src/exceptions.py` would read a bit better in logs/Sentry. Purely cosmetic since the queue layer treats all exceptions uniformly.
- Heads-up: total-save failures now flow into the queue manager's existing incremental-retry loop (`_handle_processing_error` marks only `items[0]`). Under a *sustained* 429, a multi-item batch will re-run the LLM call once per remaining item as it drains. That's a property of the pre-existing batch-error handling, not this change — just worth tracking as a potential follow-up if 429 bursts prove costly.

Nice, well-tested fix for the silent-loss bug.

karaokedjodua:
Finn-loop review of a46cf2b

CI: not configured (zero required checks) — the single non-required check (CodeRabbit) passed
Mergeability: structurally clean (`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` = stale branch-protection rule referencing a non-existent check)

## Review
Summary: Fixes silent representation save failures in the deriver — when every observer's `save_representation` fails (e.g. embedding 429 exhaustion), the batch now raises `RepresentationSaveError` so the queue marks the work unit as `errored` instead of silently returning `processed` with zero documents saved. Partial failures are kept (saved observers preserved) and surfaced via a new additive `failed_observer_count` telemetry field.

## 1. Must fix before merge
None.

## 2. Should fix soon
None.

## 3. Safe to merge
Yes from code review — the diff is minimal, targeted, and well-tested. The fix correctly separates total-failure (raise) from partial-failure (record + continue), telemetry fires before the raise so metrics aren't lost, and the new `failed_observer_count` field defaults to 0 (backward compatible). Tests cover both all-failure and partial-failure paths with assertions on the raise, telemetry emission order, and field counts.

**CI gap:** `mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE` and zero required checks is a stale branch-protection rule — an admin must update or remove the rule. GitHub merge button will be blocked despite the branch being structurally clean.

**Labels to apply:** `needs-human-review` (for CI configuration gap), `loop-approved` (code is clean)

akattelu:
Thanks for opening @leo922oel! I combined your fix in #864 and this one into #1019 so that they both go out together. Thanks for the issues and the fixes. I also addressed some other issues in there and added some tests. Closing this.

PR #861, opened 2026-06-30
feat: add exact content deduplication in document creation

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document creation deduplication by normalizing text (case-insensitive and whitespace-trimmed) before matching.
  * Exact duplicates within the same batch are collapsed to a single inserted record.
  * Exact matches against existing live documents now reinforce the existing record (updating `times_derived` atomically) instead of creating a new row.
  * When enabled, semantic deduplication continues to work after the exact-dedup step, preserving reinforcement outcomes.
* **Tests**
  * Added async coverage for normalized exact-duplicate collapsing, live-record reinforcement, `times_derived` max-style behavior, and exact/semantic dedup interaction.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -531,7 +531,16 @@ async def create_documents(
             #    the re-derivation as reinforcement on the existing row.
             existing_match = existing_by_normalized.get(normalized_content)
             if existing_match is not None:
-                existing_match.times_derived = models.Document.times_derived + 1
+                # Reinforce the existing row. greatest(...) keeps the bump atomic
+                # server-side (concurrent workers can't lose an increment) while
+                # still honoring an incoming doc that already carries accumulated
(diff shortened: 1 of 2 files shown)


src/crud/document.py
```diff
@@ -456,8 +474,76 @@ async def create_documents(
...
+    # duplicates within a single inference call collapse to one document.
+    seen_in_batch: set[str] = set()
+
     for doc in documents:
         try:
+            normalized_content = _normalize_content(doc.content)
+
+            # Exact-match dedup, always on:
+            # 1) collapse exact duplicates within this batch (drop silently).
+            if normalized_content in seen_in_batch:
```
ajspig, line 526: Do we want to check for dups earlier? pre-embedding? I know it won't impact the embedding quality so it would just be an optimization change.

VVoruganti (approved):

PR #864, opened 2026-07-01
fix(deriver): truncate oversize observations so one can't drop the batch (#569)

### LINKED ISSUE #569: Deriver loses an entire batch of observations when one exceeds the embedding token limit

leo922oel:
## Description

`src/crud/representation.py` collects every observation from a deriver pass and  <br>embeds them in one call — `embedding_client.simple_batch_embed(observation_texts)`.  <br>If a **single** observation exceeded the provider's per-input token cap,  <br>`simple_batch_embed` raised `ValueError`, `save_representation` re-raised  <br>`ValidationException`, and the deriver caught it — so **none** of that pass's  <br>observations were saved, not just the long one (plastic-labs/honcho#569). One long pasted artifact  <br>(a PR review, verbose tool output, a stack trace) dropped the whole batch. Rare  <br>under OpenAI's 8192 cap; routine under Gemini's smaller embedding cap.

`simple_batch_embed` has no sub-chunking path (only `batch_embed`, used for  <br>messages, does), and truncation is the right call here: an observation is a short  <br>derived fact, so a token-capped prefix still carries the meaningful content.

This adds an opt-in truncation mode and points the deriver at it:

* `simple_batch_embed` **gains** `on_oversize: "raise" | "truncate"` (keyword-only,  <br>default `"raise"`). `"raise"` preserves the exact historical contract for every  <br>existing caller. `"truncate"` embeds a token-capped prefix of the oversize input  <br>and logs a warning; the one-input→one-vector mapping is preserved.
* `RepresentationManager` **opts into** `on_oversize="truncate"`, so one oversize  <br>observation is truncated instead of failing the whole observer's batch.

Complements plastic-labs/honcho#860 (surface representation save failures instead of silent loss):  <br>plastic-labs/honcho#569 keeps the common oversize case from ever failing; plastic-labs/honcho#860 makes a genuinely  <br>total save failure loud instead of silent. Neither masks the other.

**Quantified** (`tests/bench/bench_embedding_oversize_survival.py` — real  <br>`save_representation` path with a faked provider + faked DB write, `batch_n=10`,  <br>exactly one over-length observation):

| Metric | Before (`main`) | After |
| -- | -- | -- |
| `batch_survival_rate` | 0% | 100% |

## Linked issues

Fixes plastic-labs/honcho#569

## Type of change

- [X] `fix` — bug fix
- [ ] `feat` — new feature
- [ ] `refactor` — neither fixes a bug nor adds a feature
- [ ] `perf` — performance improvement
- [ ] `docs` — documentation only
- [X] `test` — adding or correcting tests
- [ ] `chore` — build process or tooling

## Test plan

New / updated tests:

* `tests/llm/test_embedding_client.py` — `simple_batch_embed` raises on oversize  <br>when `on_oversize="raise"` (default) and truncates the oversize prefix (keeping  <br>1 vector per input) when `on_oversize="truncate"`. (18 passed)
* `tests/crud/test_representation_manager.py::TestRepresentationManagerSave::test_save_representation_embeds_with_truncate_on_oversize`  <br>— the manager passes `on_oversize="truncate"` through. (7 passed; needs a local  <br>pgvector DB)
* `tests/conftest.py` — the autouse embedding mock accepts the new `on_oversize`  <br>kwarg (via `**_kwargs`) so existing tests keep passing.

Quantitative before/after (standalone harness, not a CI test):

```bash
# AFTER (this branch)
PYTHONPATH=. uv run python tests/bench/bench_embedding_oversize_survival.py
# BEFORE (main source)
git checkout main -- src/embedding_client.py src/crud/representation.py
PYTHONPATH=. uv run python tests/bench/bench_embedding_oversize_survival.py
git checkout HEAD -- src/embedding_client.py src/crud/representation.py
```

## Checklist

- [X] Tests added/updated for the change
- [X] Existing tests pass (`tests/llm/test_embedding_client.py` 18/18;  <br>`tests/crud/test_representation_manager.py` 7/7 against a local pgvector DB)
- [X] `ruff` clean; `basedpyright` 0 errors on every changed file (4 pre-existing  <br>`conftest.py` header warnings, present identically on `main`)
- [X] Commit messages follow Conventional Commits
- [ ] Documentation update — n/a (internal behavior; new param defaults to the  <br>prior raise-on-oversize contract, so no caller changes)

## Summary by CodeRabbit

* **New Features**
  * Added automatic truncation for oversized observation text during embedding, preventing hard failures on long inputs.
* **Bug Fixes**
  * Oversized inputs no longer raise errors when truncation mode is enabled.
  * Ensured returned embeddings stay aligned with the original input order and count.
* **Tests**
  * Added benchmark coverage for “oversize survival.”
  * Expanded unit tests for truncation success, edge-case tokenization drift, and default error behavior.

### DIFF
--- a/src/crud/representation.py
+++ b/src/crud/representation.py
@@ -105,7 +105,7 @@ class RepresentationManager:
                 parent_category="representation",
             ):
                 embeddings = await embedding_client.simple_batch_embed(
-                    observation_texts
+                    observation_texts, on_oversize="truncate"
                 )
         except ValueError as e:
             raise exceptions.ValidationException(
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -663,9 +663,19 @@ class EmbeddingClient:
         """Embed a single query string."""
         return await self._get_client().embed(query)
 
-    async def simple_batch_embed(self, texts: list[str]) -> list[list[float]]:
-        """Batch embed a list of text strings (each must fit token limit)."""
-        return await self._get_client().simple_batch_embed(texts)
+    async def simple_batch_embed(
+        self,
+        texts: list[str],
+        *,
+        on_oversize: Literal["raise", "truncate"] = "raise",
+    ) -> list[list[float]]:
+        """Batch embed a list of text strings, one vector per input.
+
+        See ``_EmbeddingClient.simple_batch_embed`` for ``on_oversize`` semantics.
+        """
+        return await self._get_client().simple_batch_embed(
+            texts, on_oversize=on_oversize
+        )
 
     def prepare_chunks(self, id_resource_dict: dict[str, str]) -> dict[str, list[str]]:
         """Chunk texts using the same rules as `batch_embed` (no network)."""
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -274,39 +274,69 @@ class _EmbeddingClient:
             fn=_call_openai,
         )
 
-    async def simple_batch_embed(self, texts: list[str]) -> list[list[float]]:
+    async def simple_batch_embed(
+        self,
+        texts: list[str],
+        *,
+        on_oversize: Literal["raise", "truncate"] = "raise",
+    ) -> list[list[float]]:
         """
-        Batch-embed a list of text strings. Each input must already fit within
-        `max_embedding_tokens`; this method does not sub-chunk oversized inputs.
+        Batch-embed a list of text strings, one vector per input (1:1).
+
+        Each input must fit within `max_embedding_tokens`. When an input exceeds
+        the cap, behaviour depends on `on_oversize`:
+
+        - ``"raise"`` (default): raise ``ValueError`` — preserves the historical
+          contract for existing callers.
+        - ``"truncate"``: embed a token-capped prefix of the input and log a
+          truncation warning, so one oversize input does not fail the whole batch.
+          The one-input-to-one-vector mapping is preserved.
 
         Internally goes through the same token-aware batching pipeline as
         `batch_embed()` so the per-request token cap is respected.
 
         Args:
             texts: List of text strings to embed
+            on_oversize: How to handle an input over the per-input token cap.
 
         Returns:
             List of embedding vectors, one per input text (in order)
 
         Raises:
-            ValueError: If any text exceeds token limits
+            ValueError: If any text exceeds token limits and ``on_oversize`` is
+                ``"raise"``.
         """
         if not texts:
             return []
 
-        # Validate per-input token limit and collect token counts for batching
+        # Validate per-input token limit and collect the (possibly truncated)
+        # texts plus their token counts for batching.
+        prepared_texts: list[str] = []
         token_counts: list[int] = []
         for idx, text in enumerate(texts):
-            tokens = len(self.encoding.encode(text))
-            if tokens > self.max_embedding_tokens:
+            token_ids = self.encoding.encode(text)
+            if len(token_ids) > self.max_embedding_tokens:
+                if on_oversize == "truncate":
+                    text = self.encoding.decode(token_ids[: self.max_embedding_tokens])
+                    logger.warning(
+                        "truncated oversize embedding input at idx %d: %d->%d tokens",
+                        idx,
+                        len(token_ids),
+                        self.max_embedding_tokens,
+                    )
+                    prepared_texts.append(text)
+                    token_counts.append(self.max_embedding_tokens)
+                    continue
                 raise ValueError(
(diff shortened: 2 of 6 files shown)


karaokedjodua:
Finn-loop review of 96a66c4

CI: not configured — no required checks reported on the `fix/embedding-truncate-oversize` branch
Mergeability: clean (mergeable: MERGEABLE, but mergeStateStatus: BLOCKED — stale branch-protection rule)

## Review

Summary: Fixes #569 — a single oversize observation (e.g. a pasted PR review exceeding the embedding token cap) previously caused the deriver to drop the entire batch of observations via `ValueError` in `simple_batch_embed`. This PR adds an opt-in `on_oversize: "raise" | "truncate"` parameter (default `"raise"` — no breaking change) and points the `RepresentationManager` at `"truncate"`, so oversize observations get a token-capped prefix embedding instead of failing the whole deriver pass. The `"truncate"` path includes a re-encode loop to handle BPE/pre-tokenizer boundary drift where `decode(ids[:cap])` re-encodes to more than `cap` tokens.

## 1. Must fix before merge

- **[CI]** No required CI checks are configured on this branch. `gh pr checks --required` returns "no required checks reported on the branch." Additionally, `mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE` indicates a stale branch-protection rule referencing a check that no longer exists or doesn't match. A repo admin must either add required CI or update/remove the stale rule before this can merge. 

## 2. Should fix soon

None.

## 3. Safe to merge

Yes on code merits — the fix is exactly scoped to the problem. The new `on_oversize` parameter defaults to `"raise"` (zero behavioral change for existing callers), the `"truncate"` path has a proper re-encode loop for BPE boundary drift edge cases, and tests are thorough:

- 3 new unit tests in `test_embedding_client.py`: truncate success, re-encode drift loop, backward-compatible default raise
- 1 new integration test in `test_representation_manager.py`: manager passes `on_oversize="truncate"` through
- Existing tests updated to expect the new kwarg (`conftest.py` mock, existing `test_save_representation_filters_blank_*` assertions)
- Standalone benchmark harness for quantitative before/after measurement

No security concerns, no scope creep, no regressions. The stale branch-protection rule and missing CI are repo-configuration issues, not code defects.

**Note:** No Paperclip KAR-NNN was linked — `Fixes #569` references a GitHub issue. Reviewed against the PR body as the contract.

**Labels that should be applied:** `needs-human-review`, `loop-approved`

akattelu:
Thanks for opening @leo922oel! I combined your fix in #728 and this one into #1019 so that they both go out together. Thanks for the issues and the fixes. I also addressed some other issues in there and added some tests. Closing this.

PR #868, opened 2026-07-02
feat: add ChromaDB vector store backend

VVoruganti:
## Summary

Adds `chromadb` as a `VECTOR_STORE_TYPE`, with three client modes selected by `VECTOR_STORE_CHROMA_CLIENT_MODE`:

- `persistent` — local embedded storage (dev / small deployments)
- `http` — self-hosted Chroma server
- `cloud` — Chroma Cloud (requires `VECTOR_STORE_CHROMA_API_KEY`; tenant/database optional, resolved from the key)

Linear: DEV-1973

## Design notes

- **One sync code path.** Chroma's persistent and cloud clients have no async variants, so all three modes use the sync client offloaded through `asyncio.to_thread`.
- **Namespace → collection mapping.** Honcho namespace strings aren't valid Chroma collection names (base64url hashes are mixed-case), so collections are named `ns-{sha256hex(namespace)}` with the original namespace stored in collection metadata.
- **Per-collection index isolation.** Chroma gives every collection its own physically separate vector index (hnswlib locally, SPANN + per-collection object-storage WAL on Cloud), so the pgvector problem of one global HNSW index spanning all collections does not apply. Embedded mode still serializes writes through a shared SQLite file and keeps a bounded number of collection indices warm, hence the dev-scale positioning in the docs.
- **Contract parity with existing backends.** Cosine distance scores, lazy collection creation, missing-namespace reads/deletes as empty/no-op, `VectorStoreError` on transient write failures (reconciler retries), empty results on transient query failures. Write batches split at `min(get_max_batch_size(), 300)` for the Chroma Cloud per-request cap; `None`-valued metadata keys are stripped for Cloud portability.
- **Optional dependency.** Ships as an extra: `uv sync --extra chromadb`. Default installs and the Docker image are unchanged (verified via `uv sync --frozen --no-group dev --dry-run` — chromadb and its 57-package tree drop out). The factory raises an actionable error if the SDK is missing. Follow-up DEV-1974 tracks giving turbopuffer/lancedb the same treatment.

No call-site changes — everything goes through the existing `VectorStore` ABC.

## Testing

- 22 new mocked unit tests (filter translation, batching, metadata sanitization, not-found paths, projection, error mapping)
- Real embedded round-trip test against a tmp-dir `PersistentClient` (no mocks), plus real dim-probe tests
- Behaviors verified live against chromadb 1.5.9 before writing mocks — caught that single-clause `$and` is rejected (bare clause passed instead) and that the collection model's `dimension` field is unreliable (probe peeks a stored embedding)
- Full suite: 1488 passed, 9 skipped; basedpyright and ruff clean

## Summary by CodeRabbit

- **New Features**
  - Added ChromaDB as a supported vector store.
  - Supports persistent, HTTP, and cloud deployment modes with configurable connection, storage, authentication, and database settings.
  - Added vector search, batching, filtering, deletion, namespace management, and dimension detection.
  - Cloud mode now validates required API key configuration.
  - Provides clear installation guidance when the optional ChromaDB support is unavailable.

- **Documentation**
  - Added ChromaDB setup, configuration, deployment, and operational guidance.
  - Expanded embedding backend documentation, including collection dimension behavior and rebuild requirements.


src/vector_store/chroma.py
```diff
@@ -0,0 +1,515 @@
...
+                    )
+                )
+
+            logger.debug(
+                f"Query returned {len(query_results)} results from namespace {namespace}"
+            )
+            return query_results
+
+        except Exception as exc:
+            if _is_transient_error(exc):
```
eisene, line 393: we don't want a retry here?

pyproject.toml
```diff
@@ -46,6 +46,9 @@ lancedb = [
 qdrant = [
     "qdrant-client>=1.18.0",
 ]
+chromadb = [
```
eisene, line 49: Do you intend to not change the `Dockerfile` here? There's an `INSTALL_LANCEDB` but you're not adding an equivalent `INSTALL_CHROMADB`. That's fine, just wanted to check it's intended.

src/vector_store/chroma.py
```diff
@@ -0,0 +1,515 @@
...
+
+    def _create_client(self) -> "ClientAPI":
+        """Construct the sync Chroma client for the configured mode.
+
+        Runs inside a worker thread: PersistentClient does disk I/O and the
+        http/cloud clients issue a version heartbeat on construction.
+        """
+        mode = settings.VECTOR_STORE.CHROMA_CLIENT_MODE
+        if mode == "persistent":
+            return chromadb.PersistentClient(path=settings.VECTOR_STORE.CHROMA_PATH)
```
eisene, line 109: It looks like this never works. This `PersistentClient` is very similar to SQLite. If you have multiple processes sharing the same DB storage in this mode, they get a split-brain problem and things can go horribly wrong, for the same reason this happens in SQLite.

I think the solution here is that Honcho, because it is always multi-process, can't run in `persistent` mode, this should throw an `ArgumentError` or something like that.

eisene (changes requested):

PR #876, opened 2026-07-06
Add Milvus vector store backend

zc277584121:
## Summary
- Add a Milvus vector-store implementation using MilvusClient, explicit schema, vector AUTOINDEX, and a Milvus Lite-friendly default URI.
- Wire Milvus into vector-store settings, factory creation, startup dimension validation, configuration examples, and docs.
- Add pymilvus[milvus-lite] dependency coverage plus focused tests for namespace mapping, filters, projection behavior, deletes, and dimension probing.

## Testing
- uv run --exclude-newer 2026-06-27T20:44:32.746059Z --locked ruff check src/vector_store/milvus.py src/vector_store/__init__.py src/config.py src/startup/embedding_validator.py tests/vector_store/test_milvus.py tests/conftest.py tests/llm/test_model_config.py
- uv run --exclude-newer 2026-06-27T20:44:32.746059Z --locked basedpyright src/vector_store/milvus.py src/vector_store/__init__.py src/config.py src/startup/embedding_validator.py tests/vector_store/test_milvus.py tests/conftest.py tests/llm/test_model_config.py
- uv run --exclude-newer 2026-06-27T20:44:32.746059Z --locked pytest tests/vector_store/test_milvus.py tests/vector_store/test_namespace_dim_probes.py tests/llm/test_model_config.py -q

## Notes
- Also attempted `uv run pytest tests/startup/test_embedding_validator.py -q`; it is blocked locally because Postgres at localhost:5432 is not running.

## Summary by CodeRabbit

* **New Features**
  * Added Milvus as a supported vector store backend.
  * Added Milvus Lite local storage by default, with configurable Milvus server and Zilliz Cloud connectivity.
  * Added configurable tokens, database names, consistency levels, filtering, metadata projection, and namespace management.
  * Added startup validation for Milvus embedding dimensions.

* **Documentation**
  * Updated configuration, setup, troubleshooting, and changelog documentation with Milvus guidance.

* **Tests**
  * Added coverage for Milvus configuration, querying, filtering, metadata, distance handling, and local operation.


zc277584121:
Hi @VVoruganti @Rajat-Ahuja1997, could you take a look when you have a chance?

This PR adds a Milvus vector store backend following the existing VectorStore interface. I also pushed a follow-up for the score/distance handling and added focused coverage; the current review check is green.

src/config.py
```diff
@@ -1372,6 +1372,14 @@ class VectorStoreSettings(HonchoSettings):
     # LanceDB-specific settings (local embedded mode)
     LANCEDB_PATH: str = "./lancedb_data"
 
+    # Milvus-specific settings. The default URI uses Milvus Lite local storage.
+    MILVUS_URI: str = "./milvus.db"
+    MILVUS_TOKEN: str | None = None
+    MILVUS_DB_NAME: str | None = None
+    MILVUS_CONSISTENCY_LEVEL: (
```
VVoruganti, line 1465: **Blocking:** defaulting this to `None` means a Milvus server or Zilliz Cloud collection is created with Milvus's own default consistency level, `Bounded` — a just-upserted vector can be invisible to search for several seconds.

That's a real behavioral divergence from the other three backends: pgvector, lancedb, and turbopuffer are all read-your-writes. Honcho's reconciler writes embeddings and search reads them back, so this shows up as silently missing recall rather than an error.

It's double-masked right now:
- the default `MILVUS_URI` is Milvus Lite, which is effectively strong, so local dev never sees it;
- `test_milvus_lite_round_trip` pins `MILVUS_CONSISTENCY_LEVEL` to `"Strong"`, so the tested config isn't the default config.

There's also a second-order problem: the value is only passed to `create_collection`, so it's baked in at collection-creation time. An operator who hits stale reads in production can't fix it by setting the env var — existing collections keep whatever default they were born with, and there's no migration path short of dropping them.

Suggest defaulting to `"Session"` (read-your-writes per client, which is the semantics the other backends provide) rather than `None`.

src/vector_store/milvus.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+        return results
+
+    def _hit_cosine_distance(self, hit: dict[str, Any]) -> float:
+        """Return Honcho cosine distance from a Milvus search hit."""
+        if "distance" in hit:
+            return float(hit["distance"])
+        if "score" in hit:
+            return 1.0 - float(hit["score"])
+        return 0.0
```
VVoruganti: **Blocking (small):** the `"score"` branch is unreachable. `MilvusClient.search` returns hits shaped `{id, distance, entity}` — there is no `score` key. I confirmed against pymilvus directly:

```
[{'id': 'same', 'distance': 0.0,  ...},
 {'id': 'orth', 'distance': 1.0,  ...},
 {'id': 'opp',  'distance': 2.0,  ...}]
```

So `distance` is already true cosine distance and the first branch is the only one that ever runs — which is correct and matches lancedb/turbopuffer. The `1.0 - score` line is dead code guarding against a response shape this SDK doesn't produce.

Worth deleting along with `_ScoreOnlySearchClient`: as written, the two tests assert contradictory scoring conventions for an identical-vector hit (`0.0` in the round-trip test, `0.1` from the fake) and both pass, so a future reader has no way to tell which one describes reality. The round-trip test against real Milvus Lite already covers scoring properly.

The `return 0.0` fallback is fine to keep — lancedb and turbopuffer both default a missing distance to `0.0` the same way.

tests/vector_store/test_milvus.py
```diff
@@ -0,0 +1,211 @@
...
+                    "score": 0.9,
+                    "entity": {"message_id": "msg_1"},
+                },
+                {
+                    "id": "vec_far",
+                    "score": 0.2,
+                    "entity": {"message_id": "msg_2"},
+                },
+            ]
+        ]
```
VVoruganti: This fake emits a `score` key that `MilvusClient.search` never returns (real hits are `{id, distance, entity}`, with `distance` already being cosine distance). So `test_query_converts_milvus_score_to_cosine_distance` certifies a code path that can't execute against pymilvus.

Suggest removing this class and that test together with the dead branch in `_hit_cosine_distance` — `test_milvus_lite_round_trip` already asserts the real scoring behavior (`all_results[0].score == 0.0` for the identical vector) against actual Milvus Lite, which is the stronger check.

tests/conftest.py
```diff
@@ -80,6 +80,8 @@ def emit(self, record: logging.LogRecord):
     "tests/alembic/",
     "tests/unified/",
     "tests/live_llm/",
+    # Vector-store provider tests exercise SDK behavior directly and do not need DB setup.
+    "tests/vector_store/",
```
VVoruganti: This change isn't needed for the Milvus tests, and it affects pre-existing tests.

I verified by reverting just these two lines and re-running: `tests/vector_store/test_milvus.py` passes 6/6 without it. Since `tests/vector_store/` also holds `test_lancedb.py`, `test_turbopuffer.py`, and `test_namespace_dim_probes.py`, adding the directory to the blocklist silently changes fixture behavior for those existing tests as a side effect of a PR about Milvus.

Suggest dropping it from this PR. If skipping DB setup for the whole `tests/vector_store/` directory is worth doing on its own merits, it'd be easier to review as a separate change.

tests/llm/test_model_config.py
```diff
@@ -324,9 +324,9 @@ def test_app_settings_explicit_vector_store_dimensions_warns_and_overrides() ->
     messages = [
         str(w.message) for w in captured if issubclass(w.category, DeprecationWarning)
     ]
-    assert any(
-        "VECTOR_STORE_DIMENSIONS is deprecated" in m for m in messages
-    ), f"expected deprecation warning, got {messages!r}"
+    assert any("VECTOR_STORE_DIMENSIONS is deprecated" in m for m in messages), (
```
VVoruganti: This reformat is unrelated to Milvus — it's `ruff format` drift that's already latent on `main`. Running `ruff format` (0.15.12) against main's copy of this file produces exactly this hunk, so it isn't something your change introduced; it just got swept in.

Harmless, but it's noise in a backend PR and it'll reappear for the next contributor. Worth pulling out into its own formatting commit.

(The `Literal` widening further down in this same file *is* correctly part of this PR — no issue there.)

src/vector_store/milvus.py
```diff
@@ -0,0 +1,551 @@
...
+
+        for key, value in metadata.items():
+            if key in RESERVED_FIELDS or not _VALID_IDENTIFIER_PATTERN.match(key):
+                continue
+            row[key] = value
+
+        for key in STANDARD_METADATA_FIELDS:
+            row[key] = self._metadata_varchar_value(metadata.get(key))
+
+        return row
```
VVoruganti, line 395: Non-blocking design note: every metadata value gets stored twice — once inside the JSON `METADATA_FIELD`, and again as a top-level field (explicit column for the six `STANDARD_METADATA_FIELDS`, dynamic field for everything else).

Since the collection is created with `enable_dynamic_field=True`, the dynamic field already captures arbitrary keys, which makes the JSON `metadata` column redundant — roughly 2x metadata storage. It also creates a precedence rule in `_entity_metadata` that's invisible until the two copies disagree: JSON is merged first, then entity fields overwrite it, so the `str()`-coerced copy always wins for those six keys.

Related, two smaller things in the same area:
- The `str()` coercion in `_metadata_varchar_value` silently changes types round-trip for those six fields. Harmless today (`level`, `session_name`, and `peer_name` are all strings), but it's a latent trap for any future non-string value.
- `message_id` — the only field actually projected on the hot message-search path (`crud/message.py` and `utils/search.py` both pass `include_attributes=["message_id"]`) — isn't among the declared fields, so it lands in the dynamic field, while six mostly-unused VARCHAR columns are declared explicitly. It works, but it's backwards from an indexing standpoint. lancedb takes the other approach in `_metadata_fields_for_namespace`, declaring `message_id` for `msg` namespaces.

Also minor: `MAX_VARCHAR_LENGTH = 65535` on the primary key is generous for Honcho's ~21-char nanoid IDs.

src/vector_store/milvus.py
```diff
@@ -0,0 +1,551 @@
...
+        """Initialize the Milvus vector store."""
+        super().__init__()
+
+        client_kwargs: dict[str, Any] = {"uri": settings.VECTOR_STORE.MILVUS_URI}
+        if settings.VECTOR_STORE.MILVUS_TOKEN:
+            client_kwargs["token"] = settings.VECTOR_STORE.MILVUS_TOKEN
+        if settings.VECTOR_STORE.MILVUS_DB_NAME:
+            client_kwargs["db_name"] = settings.VECTOR_STORE.MILVUS_DB_NAME
+
+        self.client = MilvusClient(**client_kwargs)
```
VVoruganti: Non-blocking: constructing the client eagerly in `__init__` means that with a Milvus *server* / Zilliz URI, a transient outage at first use surfaces as a raw `MilvusException` propagating out of `get_external_vector_store()` — not the `VectorStoreError` that callers of this module are written to expect.

For comparison, `LanceDBVectorStore` defers connecting until `_get_db()`, and turbopuffer's httpx client connects lazily. Since `@cache` doesn't memoize exceptions this does self-heal on the next call, so it's an error-shape wart rather than a durable failure. Fine to leave; just noting it since Milvus Lite (the default) never exercises this path.

VVoruganti (commented):
Thanks for this — the backend itself is solid work. I verified parity rather than taking the tests at face value, and it holds up:

- **Score convention is correct.** I probed `MilvusClient.search` with `metric_type="COSINE"` directly: hits come back as `{id, distance, entity}` and `distance` is *true cosine distance* (0.0 / 1.0 / 2.0 for identical / orthogonal / opposite). That matches lancedb's `_distance` and turbopuffer's `$dist`, so ordering and `max_distance` thresholding behave identically to the existing backends.
- **Filters cover every caller shape.** Callers only ever pass `level` eq, `session_name` eq or `{"in": [...]}`, and `peer_name` eq (`crud/document.py:242`, `crud/message.py:582`, `utils/search.py:104`, `dreamer/surprisal.py:264`). All supported, and this backend is *stricter* than lancedb on malformed operators (clear `ValueError` vs. emitting broken SQL) — an improvement.
- `include_attributes` `True` / `False` / `["message_id"]` all behave like the other backends, the last via the dynamic field.
- Clean on ruff, ruff format, and basedpyright (0 errors); 21/21 tests in `tests/vector_store/` pass.

Two things I'd want fixed before merge, plus some cleanup. Details inline.

### Blocking

1. **`MILVUS_CONSISTENCY_LEVEL` default is unsafe on a real Milvus server** — see inline on `src/config.py`. This is the one genuine capability gap versus the other backends.
2. **The `"score"` branch in `_hit_cosine_distance` is unreachable**, and the test that covers it asserts behavior no real backend produces — see inline on `src/vector_store/milvus.py` and `tests/vector_store/test_milvus.py`.

### Cleanup

3. Two changes unrelated to Milvus (`tests/conftest.py`, `tests/llm/test_model_config.py`) — inline.
4. **Docs and changelog are half-done.** There are zero Milvus mentions in `CHANGELOG.md`, `docs/changelog/introduction.mdx`, `CLAUDE.md` (the project tree around lines 246-247 and architectural decision #8 at line 289 still say "turbopuffer or lancedb"), `docs/v3/contributing/changing-embeddings.mdx:60`, `scripts/configure_embeddings.py:19-20`, and `docs/v3/contributing/configuration.mdx:285` and `:295`. The repo has a clear convention here — "External Vector Store support for turbopuffer and lancedb" got its own CHANGELOG entry, so a third backend should too. The `.env.template` / `config.toml.example` / `configuration.mdx` env-var updates you did make are all correct; it's just these prose surfaces that were missed.

Everything else in the non-Milvus part of the diff is exactly the mechanical widening you'd expect, and declaring `pymilvus` as a non-optional core dependency is consistent with how `turbopuffer` and `lancedb` are already declared.

The remaining inline notes are non-blocking design observations, not merge conditions.

VVoruganti (changes requested):
Converting my [previous review](https://github.com/plastic-labs/honcho/pull/876#pullrequestreview-4776207681) to a formal request for changes — the inline comments there still hold and I haven't duplicated them here.

To restate the assessment: the backend is solid and has real parity with the existing stores. I verified the score convention against pymilvus directly (`distance` is already true cosine distance, matching lancedb's `_distance` and turbopuffer's `$dist`), confirmed every caller filter shape is supported, and it's clean on ruff / ruff format / basedpyright with 21/21 vector-store tests passing. Happy to approve once the two blocking items are addressed.

**Blocking:**

1. **`MILVUS_CONSISTENCY_LEVEL` defaults to `None`** ([inline](https://github.com/plastic-labs/honcho/pull/876#discussion_r4776207681)) → real Milvus/Zilliz collections get Milvus's `Bounded` default, so a just-upserted vector can be invisible to search for seconds. pgvector, lancedb, and turbopuffer are all read-your-writes, so this is a genuine capability gap rather than a preference. It's invisible in CI and local dev because the default `MILVUS_URI` is Milvus Lite (effectively strong), and `test_milvus_lite_round_trip` pins `"Strong"` — so it would surface only in production, as missing recall rather than an error. Compounding it, the value is only passed to `create_collection`, so it's fixed at collection-creation time and can't be corrected later via config. Suggest defaulting to `"Session"`.

2. **Dead `"score"` branch in `_hit_cosine_distance`, plus a test that certifies it.** `MilvusClient.search` never returns a `score` key, so that branch can't execute; `_ScoreOnlySearchClient` asserts a response shape the SDK doesn't produce. The two tests currently assert contradictory scoring for an identical-vector hit (`0.0` real vs. `0.1` from the fake) and both pass, which leaves no way to tell which describes reality. Deleting the branch and the fake is enough — `test_milvus_lite_round_trip` already covers scoring against real Milvus Lite.

**Non-blocking**, and I don't want these to hold up the merge: the docs/changelog gaps (item 4 in the previous review) are worth picking up since the repo has an established convention for new backends, and the remaining inline notes are design observations only.

zc277584121:
Rebased this branch onto the latest `main` (`2163ab1a`) and regenerated `uv.lock` against the current dependency set (`pymilvus` 3.0.1 / `milvus-lite` 3.2.0). The branch is now conflict-free.

A Zilliz Cloud run exposed and fixed one server-specific projection issue: when `include_attributes=True`, the adapter now explicitly requests the JSON `metadata` field. Milvus Lite returned that field when `output_fields` was omitted, while Zilliz did not, which caused custom attributes to disappear on the server path. Explicit list projections and ID-only queries are unchanged.

Validation completed:
- 7 focused Milvus tests passed, including the Milvus Lite round trip
- 28 vector-store configuration tests passed
- Zilliz Cloud passed create, upsert, schema probing, full metadata, selected projection, filters, distance-threshold search, and delete
- Cloud smoke collections were verified cleaned up
- `ruff check` passed for all changed Python files
- `ruff format --check` passed for the Milvus implementation and changed runtime/test files; the latest main has an existing formatting delta in `tests/llm/test_model_config.py`, which was left untouched to avoid restoring an unrelated change
- `basedpyright` completed with 0 errors, 0 warnings
- `uv lock --check` and `git diff --check` passed

The repository-wide autouse test fixture requires PostgreSQL even for the Milvus-only file. With no local PostgreSQL on port 5432, the combined run stopped during fixture setup; the Milvus file was rerun with that unrelated fixture isolated and passed.

CHANGELOG.md
```diff
@@ -94,6 +94,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Milvus vector store backend with Milvus Lite defaults, configurable server or Zilliz Cloud connectivity, and session-consistent reads (#876)
```
akattelu, line 97: Can remove this changelog portion

docker-compose.yml.example
```diff
@@ -46,6 +48,8 @@ services:
       - DB_CONNECTION_URI=postgresql+psycopg://postgres:postgres@database:5432/postgres
       - CACHE_URL=redis://redis:6379/0?suppress=true
       - CACHE_ENABLED=true
+      # Reject process-local Milvus Lite files in this API + deriver deployment.
+      - HONCHO_DOCKER_MULTI_PROCESS=true
```
akattelu, line 52: why are we turning this on?

akattelu (changes requested):

PR #881, opened 2026-07-07
fix: apply session scoping to all working-representation query paths

VVoruganti:
## Summary

- **Fixes a session-scoping leak (DEV-1994):** `session_name` was only applied to the recent-documents query in `RepresentationManager`; the semantic and most-derived paths ignored it, so `session.context` with `limit_to_session=true` leaked cross-session conclusions into the perspective. The session restriction is now applied uniformly to all three query paths and pushed down to pgvector and external vector stores.
- The internal signature now takes a session **allowlist** (`session_names: list[str]`) so the upcoming session-allowlist API (DEV-1995) reuses the same enforcement path. An empty allowlist fails closed — downstream stores drop empty `IN` clauses, which would otherwise silently widen scope.
- **Bare-list membership sugar in the filter DSL** (groundwork for DEV-1995): `{"session_id": ["s1","s2"]}` ≡ `{"session_id": {"in": [...]}}` on regular columns, generically (`peer_id`, etc.). JSONB metadata columns keep containment semantics unchanged. Previously a bare list compiled to a type-mismatched equality that matched nothing, so this is strictly additive. Same shape translated in the turbopuffer/lancedb filter builders, plus a fix for lancedb dropping empty IN clauses (fail-open → always-false).

## Behavior change

Results for `limit_to_session=true` get narrower (correct): cross-session and session-less (dream-produced) conclusions no longer appear. Needs a release-note entry framed as a bugfix.

## Tests

- New `TestRepresentationManagerSessionScoping`: per-path allowlist enforcement, NULL-session exclusion, end-to-end blend, empty-allowlist fail-closed
- New `test_bare_list_membership_sugar`: list≡in equivalence, empty-list fail-closed, JSONB containment semantics preserved
- Full affected suite green: `tests/test_advanced_filters.py`, `tests/crud/test_representation_manager.py`, `tests/routes/test_sessions.py`, `tests/routes/test_peers.py` (144 passed)

Linear: DEV-1994 (part of DEV-1970 / Scopes RFC Phase 0)

## Summary by CodeRabbit

* **New Features**
  * Working representations and peer/session context lookups now accept multiple session IDs via a session allowlist for consistent scoping.
  * Filter inputs support bare lists/tuples/sets as shorthand for membership (IN) conditions (with JSONB fields retaining containment behavior).

* **Bug Fixes**
  * Empty session allowlists now fail closed by returning no results.
  * Empty membership filters now return no matches instead of widening queries (fixed across vector store filter builders).  

* **Tests**
  * Added coverage for multi-session scoping and bare membership sugar, including fail-closed cases.


src/vector_store/turbopuffer.py
```diff
@@ -252,6 +252,9 @@ def _build_filters(self, filters: dict[str, Any]) -> Filter | None:
                 # Membership filter using "In" operator
                 in_values = cast(Sequence[Any], value["in"])
                 filter_list.append((key, "In", in_values))
+            elif isinstance(value, list | tuple | set):
+                # Bare-list sugar: same membership semantics as {"in": [...]}
```
adavyas, line 254: doesn't seem to handle the empty-list scenario as done in LanceDB with `conditions.append("1 = 0")`. this remains fail-open

src/crud/representation.py
```diff
@@ -502,17 +522,24 @@ async def _query_documents_for_level(
...
         """
         filters: dict[str, Any] = {}
 
         if level:
             filters["level"] = level
 
+        if session_names:
+            filters["session_name"] = {"in": session_names}
+
         return filters
```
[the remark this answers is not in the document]
adavyas: seems valid

tests/crud/test_representation_manager.py
```diff
@@ -232,6 +232,206 @@ async def test_query_documents_most_derived_ties_break_by_recency(
...
+        self,
+        db_session: AsyncSession,
+        sample_data: tuple[models.Workspace, models.Peer],
+    ):
+        """An empty allowlist must return an empty representation, not fall
+        back to unscoped behavior (downstream stores drop empty IN clauses)."""
+        test_workspace, test_peer = sample_data
+        _, _, manager = await self._setup(db_session, test_workspace, test_peer)
+
+        representation = await manager.get_working_representation(
```
adavyas, line 426: this test seems to be falsely passing because in `src/crud/representation.py`:

`if session_names is not None and not session_names:
      return Representation()`

seems to be returning an empty representation. we should probably call `_build_filter_conditions` directly

adavyas (commented):
looks pretty solid. just a few small wrinkles

adavyas (changes requested):

karaokedjodua:
Finn-loop review of 08afed6

CI: not configured (`gh pr checks --required` returns "no required checks reported on the 'vineeth/dev-1994' branch")
Mergeability: BLOCKED — mergeStateStatus=BLOCKED with mergeable=MERGEABLE. All individual checks pass (basedpyright, test-python, CodeQL, CodeRabbit), but required checks don't match the running ones — stale branch protection rule.

## Review
Fixes DEV-1994: a session-scoping leak where `session_name` was only applied to the recent-documents query, allowing semantic and most-derived paths to leak cross-session conclusions into `limit_to_session=true` results. The fix changes the internal API from a single `session_name` to a session allowlist (`session_names: list[str]`), pushes the filter down to all three query paths, and adds bare-list membership sugar to the filter DSL as groundwork for the upcoming session-allowlist API (DEV-1995).

## 1. Must fix before merge
**[CI]** No required CI checks are configured on this branch. `gh pr checks 881 --required` exits 1 with "no required checks reported." While all individual checks pass (basedpyright, test-python, CodeQL, CodeRabbit), none gate the merge — there are no branch protection rules enforcing them.

**[CI]** mergeStateStatus is BLOCKED with mergeable=MERGEABLE. All individual checks pass, but the branch-protection rule references checks that no longer exist or don't match. A repo admin must update or remove the stale rule before this can merge. This is a repo-configuration gap, not a code defect.

## 2. Should fix soon
None. The code is tightly scoped and well-implemented.

## 3. Safe to merge
Code is clean — the fix is complete and correct:
- All three query paths (semantic, most-derived, recent) now receive the session allowlist uniformly
- Empty-allowlist fail-closed guard short-circuits at the entry point before any query is issued, preventing downstream stores from silently widening scope by dropping empty `IN` clauses
- `_build_filter_conditions` uses `is not None` (not truthiness) so an explicit empty list correctly emits `{"session_name": {"in": []}}` for downstream fail-closed rendering
- Bare-list membership sugar is cleanly implemented: regular columns get `in` semantics, JSONB columns keep containment — correctly gated via `JSONB_COLUMNS` tuple
- Vector store builders (lancedb, turbopuffer) handle both dict `{"in": [...]}` and bare-list forms, plus fail-closed on empty lists (`1 = 0` / `And[Eq, NotEq]` contradiction)
- 7 new representation-manager tests cover each query path in isolation + end-to-end blend + empty-allowlist + NULL-session exclusion
- 1 new filter-DSL test covers bare-list sugar, dict equivalence, empty fail-closed, and JSONB containment
- 4 new vector-store unit tests cover membership forms and empty-list guards
- No security concerns — the change narrows scope (fewer observations returned), never widens it

The CI infrastructure gap (no required checks, stale branch protection) is the only blocker. A human must verify the test suite passes and address the branch-protection rule before merging.

Labels needed: `needs-human-review`, `loop-changes-requested` (CI infrastructure, not code)

PR #882, opened 2026-07-07
feat: session allowlist on dialectic and representation via filters

VVoruganti:
**Stacked on #881** — merge that first, then retarget/merge this.

## Summary

Implements the Scopes RFC Phase 1 (DEV-1995): `peer.chat` and `/representation` accept a constrained `filters` body — the same DSL search and conclusions already accept — supporting only the `session_id` key (a session id, a bare list, or `{"in": [...]}`). This is verbatim sauna.ai's ask: *sessions_ask-style behavior across only the dynamically allowed session set*.

```jsonc
POST /v3/workspaces/{w}/peers/{p}/chat
{ "query": "...", "filters": { "session_id": ["s1", "s2", "s3"] } }
```

## Semantics (fail-closed everywhere)

- **Conclusion recall** (dialectic prefetch + `search_memory`): restricted to the allowlist; dream-produced docs (`session_name IS NULL`) excluded.
- **Message tools** (search/grep/date-range/temporal/observation-context/recent-history): strict **intersection** of allowlist ∩ observer session membership.
- **`get_reasoning_chain`**: unavailable under an allowlist — chains traverse provenance across sessions and can't be scoped without exposing out-of-scope premises/conclusions.
- **Empty allowlist** short-circuits to empty results at every path (downstream stores drop empty IN clauses, which would otherwise fail open).
- Unsupported filter keys/shapes → **422**, never silently ignored (a dropped filter here would widen scope). Composes with `session_id` (must be in the allowlist when both given). Cap: 1,000 sessions/request.

## Auth

Workspace keys pass the allowlist as-given (trusted caller). Peer-scoped JWTs must be a member of **every** allowlisted session (403 otherwise) — mirrors the existing single-`session_id` membership check, same query, one round trip.

## Deliberately out of scope (follow-ups)

- `session.context` filters (single-session `limit_to_session` already covered by #881; the Phase 2 `scope` param is the right vehicle there)
- `metadata` filter key on dialectic (deferred until conclusions have public metadata via conclusion-tagging)
- SDK convenience option `sessions: [...]` compiling to `filters` (DEV-1996)

## Tests

New `tests/test_session_allowlist.py` (20 tests): parser shapes/cap/dedupe/422s, search_memory filter pushdown + empty-allowlist short-circuit, message-crud intersection incl. non-member allowlisted session excluded, chat route validation (422s), representation route end-to-end scoping (out-of-scope and sessionless docs excluded). Full affected suites green; the 3 pre-existing `test_document.py` dedup failures reproduce identically on the base branch (embedding key issue, DEV-1935) and are unrelated.

Linear: DEV-1995 (part of DEV-1970 / Scopes RFC)

## Summary by CodeRabbit

* **New Features**
  * Added session-based allowlists for chat recall, memory search, message search, and peer representations.
  * Support single or multiple `session_id` filters with deduplication and request caps, plus consistent empty-allowlist handling.
* **Security**
  * Enforced fail-closed session scoping for peer operations, including observer membership intersection.
  * Prevents cross-session reasoning/provenance traversal when an allowlist is set.
* **Documentation**
  * Updated request filter descriptions to clarify supported keys and allowlist requirements.
* **Tests**
  * Added comprehensive validation, CRUD routing enforcement, and fail-closed contract coverage.


src/crud/message.py
```diff
@@ -689,19 +689,29 @@ async def _semantic_search_messages(
...
+    intersection).
     """
     # Pre-fetch peer session scope if needed (short-lived DB session)
     allowed_session_names: list[str] | None = None
-    if observer and not session_name:
-        async with tracked_db(f"{operation_name}.peer_scope", read_only=True) as db:
-            allowed_session_names = await get_peer_session_names(
-                db, workspace_name, observer
-            )
+    if not session_name and (observer or allowed_sessions is not None):
```
adavyas: when `session_name` is set, this only scopes to that one session itself and never applies `allowed_sessions`. rest of the allowlist is dropped ignored

src/routers/peers.py
```diff
@@ -298,7 +333,7 @@ async def get_representation(
             observed=options.target if options.target is not None else peer_id,
             session_names=[options.session_id]
             if options.session_id is not None
```
adavyas, line 345: i think the allowlist is ignored here as well

src/crud/message.py
```diff
@@ -689,19 +689,29 @@ async def _semantic_search_messages(
     after_date: datetime | None = None,
     before_date: datetime | None = None,
     observer: str | None = None,
+    allowed_sessions: list[str] | None = None,
```
adavyas: nit: allowed_sessions was called session_names in other parts of the codebase and in PR #881 should probably be consistent.

adavyas (dismissed):

akattelu (commented):
The peer card fetched during these queries still could contain information from sessions that aren't limited to `session_names` right? I think that's ok but just confirming that's the intent.

Generally think we should add more documentation because there's some nuance with how this works. Not sure where https://honcho.dev/docs/v3/api-reference/endpoint/peers/chat is populated from but it should be updated.

src/crud/message.py
```diff
@@ -689,21 +689,38 @@ async def _semantic_search_messages(
...
+    intersection).
     """
     # Pre-fetch peer session scope if needed (short-lived DB session)
     allowed_session_names: list[str] | None = None
-    if observer and not session_name:
-        async with tracked_db(f"{operation_name}.peer_scope", read_only=True) as db:
-            allowed_session_names = await get_peer_session_names(
-                db, workspace_name, observer
-            )
+    if not session_name and (observer or session_names is not None):
```
eisene: this appears to be copypasta in several places in the code - makes sense to consolidate somewhere? logic seems non-trivial enough

src/routers/peers.py
```diff
@@ -294,6 +316,15 @@ async def get_representation(
     If a target is provided, we get the Representation of the target from the perspective of the Peer.
     If no target is provided, we get the omniscient Honcho Representation of the Peer.
     """
+    # Parse the session allowlist from filters (422 on unsupported keys/shapes).
+    session_allowlist = extract_session_allowlist(options.filters)
+    if (
```
eisene: can this if statement be folded into `extract_session_allowlist` as an optional `must_include` parameter? would get rid of a few copypastas

src/utils/agent_tools.py
```diff
@@ -1049,6 +1050,9 @@ async def get_recent_history(
         List of messages in chronological order
     """
     if session_name:
+        # Fail closed: a specific session outside the allowlist is not readable.
+        if session_names is not None and session_name not in session_names:
```
eisene: i don't know that i'm asking for any changes here, but the semantics here are just kinda strange - you can provide `session_name` and `session_names`, which are both filters on the session and can in principle be provided inconsistently

Claude confirms that `session_names=[session_name]` is not the same thing as providing just `session_name`. Maybe calling the arg `session_allowlist` or something would make this more transparent?

eisene: applies in other places too

src/utils/agent_tools.py
```diff
@@ -2310,6 +2360,14 @@ async def _handle_get_reasoning_chain(
     ctx: ToolContext, tool_input: dict[str, Any]
 ) -> str:
     """Handle get_reasoning_chain tool."""
+    # Reasoning chains traverse provenance across sessions by design, so a
```
eisene, line 2380: maybe the tool can be removed from the agent in this case instead of this token-consuming failure?

src/routers/peers.py
```diff
@@ -199,6 +201,24 @@ async def chat(
...
+        session_allowlist is not None
+        and options.session_id
+        and options.session_id not in session_allowlist
+    ):
+        raise ValidationException("session_id must be included in filters.session_id")
+    # A peer-scoped key may only name sessions its peer belongs to — the
+    # allowlist widens message recall the same way session_id does above.
+    if jwt_params.p is not None and session_allowlist:
```
eisene: Eugene asked me to dig into whether this gate is right. Summary: it is, but with caveats.

`jwt_params.p` is the peer name a **peer-scoped** JWT is confined to (`src/security.py:54`). This route declares `require_auth(workspace_name="workspace_id", peer_name="peer_id")` with no `allow_member_read`, so `security.py:245` guarantees `p is not None` implies `p == peer_id`. The bypassers are admin tokens (`ad=True`), workspace tokens (`p is None`), and auth-disabled mode; session-scoped tokens can't reach this route at all (`security.py:238`). So gating on `p` is correct.

Four things worth changing or flagging:

1. **This isn't load-bearing security, and the comment above it implies otherwise.** On the chat path `observer=peer_id` is threaded into every message tool, and the CRUD layer already intersects the allowlist with `get_peer_session_names(observer)`. Without this gate, an out-of-scope session in `filters` yields an empty result, not a leak. The gate buys a loud 403 instead of a silently empty answer — worth keeping, but the comment reads as though it's the only thing standing between a peer key and someone else's messages.

2. **The two gates in this handler use different definitions of membership.** The `options.session_id` check above uses `is_peer_in_session` → active only (`left_at IS NULL`). This one uses `get_peer_session_names` → *any* membership record, including sessions the peer has left (its docstring says so explicitly). So a peer that left session S is denied `session_id=S` but permitted `filters.session_id=["S"]`. These should agree. It's genuinely non-obvious which way: the CRUD recall scoping deliberately uses the loose definition, while the auth layer's `allow_member_read` uses the strict one — so this is an open question about what "visibility" means, not just an oversight.

3. **Untested.** `tests/test_session_allowlist.py` has 17 tests and none exercises a peer-scoped JWT against this gate. Every other check in the PR has one.

4. **Style:** per CLAUDE.md the convention is an explicit `jwt_params.p == peer_id` comparison rather than relying on `require_auth` transitively. Cheap to make explicit.

Minor: `and session_allowlist` is a truthiness check, so `filters={"session_id": []}` skips it. Harmless — it fails closed downstream — but inconsistent with the `is not None` discipline used everywhere else in this PR.

— Claude

src/crud/message.py
eisene: Confirming Eugene's read — it's **four byte-identical copies**:

- `src/crud/message.py:703` (`_semantic_search_messages`)
- `src/crud/message.py:891` (`grep_messages`)
- `src/crud/message.py:956` (`get_messages_by_date_range`)
- `src/utils/agent_tools.py:1176` (`get_context_messages`)

Including the `elif session_name and session_names is not None and session_name not in session_names: return []` tail in all four.

Worth one helper, something like:

```python
async def resolve_allowed_sessions(
    workspace_name: str,
    session_name: str | None,
    session_names: list[str] | None,
    observer: str | None,
    db: AsyncSession | None = None,
    operation_name: str | None = None,
) -> tuple[list[str] | None, bool]:  # (allowlist, deny)
```

`deny=True` replaces the four `return []` sites. The awkward part is that two callers already hold a `db` and two need their own short-lived `tracked_db`, hence the optional `db` — worth it to get the `None` / `[]` / populated tri-state into one tested place.

That tri-state is the reason I'd push on this rather than leave it: `None` means unrestricted and `[]` means fail-closed, and downstream stores *drop* an empty `IN` clause rather than matching nothing (lancedb), so collapsing the two fails open. Four copies of a distinction that subtle will drift.

Also, while in here: `session_names` is undocumented in the `grep_messages`, `get_messages_by_date_range`, and `get_context_messages` docstrings — only `_semantic_search_messages` describes it.

— Claude

src/routers/peers.py
eisene: To make Eugene's suggestion concrete: a `must_include: str | None = None` parameter on `extract_session_allowlist` would collapse both copies of this block — here and the identical one in `get_representation` below.

One thing to check if you do it: the raise changes from `ValidationException` to `FilterError`. Both are `status_code = 422` (`src/exceptions.py:50`, `:90`), so no API-visible status change, but the `detail` string differs.

— Claude

src/utils/agent_tools.py
eisene: This one looks cheap to do. The toolset is already selected per-query rather than being a module constant — `src/dialectic/core.py:450` and `:533`:

```python
tools = DIALECTIC_TOOLS_MINIMAL if self.reasoning_level == "minimal" else DIALECTIC_TOOLS
```

so filtering out `get_reasoning_chain` when `session_names is not None` is a couple of lines in one place. Saves the tool schema in context *and* the wasted turn where the model calls it and gets told no.

— Claude

src/routers/peers.py
eisene: Human paraphrase:

I think the important one is 2: the two adjacent membership tests in the allowlist are inconsistent. I think worth double checking that this is intended.

src/crud/message.py
```diff
@@ -689,21 +689,38 @@ async def _semantic_search_messages(
     after_date: datetime | None = None,
```
eisene, line 776: - Peer card appears to inevitably leak conclusions across sessions. This doesn't really look fixable and, depending on the use cases here, is probably fine.
- It looks to me (and Claude) like inductive conclusions can be derived from observations across sessions and are stamped with a single choice of `session_name`, which can again leak conclusions from the other sessions they were derived from. See here: https://github.com/plastic-labs/honcho/blob/8e8b74fed14e0780c196a6105068df3398a2d270/src/dreamer/dream_scheduler.py#L205 it looks like it takes the session name of the session with the most recent explicit conclusion as the session name of the dream, and it looks like the dreamer carries this session name through everything as it runs.

All this may be fine, worth confirming explicitly and documenting though I think.

eisene: In local testing it does look like inductive conclusions have session names

Image

eisene (commented):
some minor comments and questions about what is intended vs delivered

ajspig:
> The peer card fetched during these queries still could contain information from sessions that aren't limited to `session_names` right? I think that's ok but just confirming that's the intent.

☝️ same question. I could also see it being beneficial to have scoped peer cards, but doesn't need to be apart of this initial release.

src/routers/peers.py
VVoruganti: This one narrows rather than widens, so I think it's correct as written.

Lines 321-326 already 422 unless `session_id` is in the allowlist:

```python
if (
    session_allowlist is not None
    and options.session_id
    and options.session_id not in session_allowlist
):
    raise ValidationException("session_id must be included in filters.session_id")
```

So by the time we reach the ternary, `[options.session_id]` is a strict subset of
`session_allowlist` — passing the full allowlist instead would broaden the query.

Also worth flagging in case it explains the diff view: this line has read
`else session_allowlist` since 59936959 (Jul 7), so it may have rendered against
an older revision.

Resolving — happy to reopen if you're seeing a path that reaches this without the
validation above.

src/crud/message.py
VVoruganti: Good catch — this was real when you flagged it. Fixed in 5215b841, and as of 37bba910 the logic lives in one place instead of four: `crud.message.resolve_session_scope`. A pinned `session_name` outside the allowlist now returns `(None, deny=True)` and the caller returns empty without querying. When it *is* in the allowlist, scoping to that single session is a strict subset, so the narrower filter is correct.

Resolving.

src/crud/message.py
VVoruganti: Agreed on the principle, and this went further in 37bba910: the parameter is now `session_allowlist` everywhere (crud, agent tools, dialectic, routes — 141 call sites), because `session_names` was one character away from `session_name` and they mean genuinely different things.

The function-local `allowed_session_names` is deliberately still distinct — it holds the *post-intersection* result rather than the caller's input, and conflating those two is exactly the bug you flagged in the other thread.

Resolving.

src/routers/peers.py
VVoruganti: All landed in 37bba910. Point by point:

**1 — comment overstates it.** Reworded. One correction to the reasoning though: `get_recent_history` (`agent_tools.py:1067-1085`) does *not* intersect with observer membership — it filters on `workspace_name`, `peer_name == observed` and the allowlist, nothing else. So "an out-of-scope session yields empty results" doesn't hold on that path. It narrows rather than widens there, so your conclusion stands, but the PR description's "strict intersection across all message tools" claim was wrong and I'm correcting it. The underlying gap (with `target` set, a caller already reads that peer's messages across every session with no membership check) is pre-existing and I'll file it separately.

**2 — inconsistent membership definitions.** You were right that this was the important one. `get_peer_session_names` now takes `active_only`, and the gate passes `active_only=True` so it matches the adjacent `is_peer_in_session` check. There's a test asserting a peer who has left a session is denied by *both* gates.

**3 — untested.** Added: member, non-member, left-session, workspace-key bypass, and empty-allowlist. The file went 20 -> 39 tests.

**4 — explicit `jwt_params.p == peer_id`.** Skipping this one, and I think the convention doesn't apply here: it's scoped to routes with `allow_member_read=True` keyed by a sub-resource, where membership grants session access but must not reach a co-member's data. This route has no `allow_member_read`, so `require_auth` already guarantees `p is not None` implies `p == peer_id` — an explicit re-check would be dead code. Happy to add a comment saying so if you'd rather it be stated than inferred.

**Minor — truthiness vs `is not None`.** Fixed, with a test asserting `filters={"session_id": []}` reaches the gate rather than skipping it.

src/crud/message.py
VVoruganti: Consolidated in 37bba910 as `crud.message.resolve_session_scope`, returning `(allowlist, deny)`. Your tri-state argument was the reason to do it — the helper now *never* returns an empty list, so the `None` vs `[]` distinction that the external stores fail open on exists in exactly one tested place.

Two adjustments from the sketch. First, the copies weren't quite byte-identical — one opened its own `tracked_db`, two used a caller-supplied session, one had a local import. Second, `db` ended up optional after all, but for a reason worth recording: making it required forced `_semantic_search_messages` to open a session unconditionally, which broke `test_search_messages_external_lookup_happens_before_tracked_db`. It now takes `db=None` and opens its own short-lived session only when an observer lookup is actually needed, with a test pinning that so it can't regress quietly.

Docstrings for `session_allowlist` added at every entry point, including the None / [] / populated contract.

src/routers/peers.py
VVoruganti: Done in 37bba910 — `extract_session_allowlist(filters, must_include=...)`, which collapses both copies. You were right about the exception type changing to `FilterError`; I kept the `detail` string identical so the existing 422 tests and any client string-matching are unaffected.

src/utils/agent_tools.py
VVoruganti: Agreed, and you were right that they're not interchangeable. Renamed to `session_allowlist` throughout in 37bba910.

Digging into it surfaced something worth recording: `session_name` isn't purely a scoping flag. It pins the query to one session, bypasses observer scoping, *and* drives session-history injection into the dialectic system prompt. So it can't be aliased to the allowlist or removed — it's marked deprecated for scoping specifically, with the docstrings spelling out why it stays.

src/utils/agent_tools.py
VVoruganti: Done in 37bba910. Both toolset-selection sites were already duplicated, so this became `DialecticAgent._select_tools()` — one place that picks minimal vs full and filters out `get_reasoning_chain` when an allowlist is set. The runtime guard stays as a backstop. Re-enabling it is tracked on DEV-2201, since it needs the same provenance work as the dream conclusions.

src/crud/message.py
VVoruganti: Both confirmed, and they resolved differently.

**Peer card** — you're right that it isn't fixable as-is: `crud/peer_card.py` keys on `(workspace, observer, observed)` with no session dimension at all. But it's already an operator concern rather than a Honcho one — `configuration.peer_card.use` exists today and resolves at both workspace and session level, so an operator who needs strict scoping can disable cards for that session. Documenting rather than changing.

**Dream conclusions** — confirmed end to end, exactly as you and Claude read it: `dream_scheduler.py:195-205` picks the session holding the most recent explicit conclusion, `specialists.py:322` threads it into the tool executor, and `_handle_create_observations_impl` stamps every conclusion with it — while the discovery tools default to `session_only=False` and read across all sessions.

As of 37bba910, allowlisted recall is restricted to `level == "explicit"` (`ALLOWLIST_SAFE_LEVELS` in `utils/representation.py`). Explicit conclusions come from the deriver over a single session's batch, so their stamp *is* authoritative. Fails closed, no migration, no data loss — and notably it makes the "dream output excluded" claim true in a stronger form than the original `session_name IS NULL` reasoning, which as you spotted only held for dreams invoked without a session.

The cost is that scoped queries get shallower memory. The real fix is an authoritative source-session *set* per conclusion — the reasoning trees already link premises, so the data likely exists — tracked in DEV-2201.

src/utils/agent_tools.py
```diff
@@ -1221,9 +1257,16 @@ async def extract_preferences(
...
+            Deprecated for *scoping*: prefer session_allowlist, which
+            intersects with observer membership. This parameter also pins
+            the query to one session and bypasses observer scoping, so it
+            is not a drop-in equivalent and is not removed.
         observed: The peer whose preferences to extract
         observer: When provided and session_name is None, scope results
             to sessions this peer belongs to
+        session_allowlist: Optional session allowlist. None is unrestricted; an
+            empty list fails closed (empty result); a populated list is
+            intersected with the observer's session scope when observer is set
```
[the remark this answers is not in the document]
VVoruganti: Correct on the docstring, and thanks — that was a real defect I introduced. My docstring pass was pattern-matched on the `observer:` Args entry, so it added a `session_allowlist` entry to `extract_preferences`, which has no such parameter. Fixed.

One correction to the severity, though: this is not currently reachable. `extract_preferences` is only in `DREAMER_TOOLS` — it's absent from both `DIALECTIC_TOOLS` and `DIALECTIC_TOOLS_MINIMAL` — and `ctx.session_allowlist` is only ever populated from the dialectic route. The dreamer never runs with an allowlist, so no session-scoped agent can reach this tool today and nothing leaks.

The latent hazard is real, though, so I took the fix rather than the comment explaining why not: `extract_preferences` now accepts `session_allowlist`, forwards it to every `crud.search_messages` call, and `_handle_extract_preferences` passes `ctx.session_allowlist`. If anyone wires it into the dialectic later it's scoped by default instead of silently bypassing the allowlist.

I also swept for the same class of mistake — every function whose docstring documents `session_allowlist` now actually has the parameter, and vice versa.

VVoruganti:
@akattelu @ajspig — answering the peer card question you both raised, plus the docs point.

**Peer card.** Confirmed: it can contain information from sessions outside `filters.session_id`, and that's intentional for now. `crud/peer_card.py` keys the card on `(workspace, observer, observed)` — there is no session dimension in the storage at all, so scoping it is a schema change rather than a filter.

It is already an operator concern rather than a Honcho one, though: `configuration.peer_card.use` exists today and resolves at both workspace *and* session level, and the dialectic honors it (`dialectic/chat.py`). So an operator who needs strict scoping for a given session can turn cards off there without a new flag. Agreed with @ajspig that properly scoped cards shouldn't block this release — folded into DEV-2201 alongside the related dream-provenance work.

**Docs.** Good question about where that page comes from: `docs/v3/openapi.json`, referenced from `docs/docs.json:29`. It's committed by hand and nothing in CI regenerates it, so `filters` wasn't going to show up on the API reference at all. That's still outstanding on this PR and will land before merge.

Agreed the nuance needs prose beyond the field description. The three things worth stating explicitly, since none are obvious from a schema:
- the allowlist is an **intersection** with observer session membership, not a plain filter — allowlisting a session the peer isn't in yields nothing rather than erroring;
- it now serves only **explicit** conclusions, because dream-derived ones carry a session stamp we can't vouch for (DEV-2201);
- `get_reasoning_chain` is unavailable under an allowlist, by design.

src/crud/message.py
eisene: I'm satisfied with this, convo can be considered resolved, leaving open for posterity because this may influence future work.

eisene (approved):
lgtm

PR #883, opened 2026-07-07
Session-purity invariant + card_refresh dream type (DEV-2000)

VVoruganti:
Implements **DEV-2000** (prerequisites for the Scopes RFC, DEV-1970). Scope memory is built by *copying* explicit-level documents between collections, which is only sound if explicit documents stay **session-pure**.

## Audit findings

**Was session purity violable? Yes, in two ways** (plus one latent hole):

1. **Dedup/merge paths had no level or session scoping** (`src/crud/document.py`):
   - *Exact-content dedup* in `create_documents` matched on normalized content scoped only to `(workspace, observer, observed)`. An explicit doc restated in session B would be silently dropped and recorded as reinforcement (`times_derived += 1`) on session A's row — session B's derivation leaks onto another session's document, and a scope copying session B would miss the fact entirely. Cross-**level** matches were also possible (an explicit fact could reinforce a same-content deductive doc, and vice versa).
   - *Semantic dedup* (`is_rejected_duplicate`) queried candidates with no filters, so a near-duplicate explicit doc from another session (or another level) could either be reinforced cross-session or **soft-deleted** in favor of the newcomer — destroying the original session's explicit record.
   - Content itself was never rewritten and `session_name` never mutated — the violations are merge/replacement provenance, not literal rewrites.
2. **Latent explicit-creation hole in the generic `create_observations` tool** (`src/utils/agent_tools.py`): the handler only enforced "deriver ⇒ explicit only". In a non-message context (dreamer/dialectic, `current_messages=None`) a model passing `level: "explicit"` was accepted — minting explicit docs with `session_name=NULL` during dreams. (The dream specialists' own tools force `deductive`/`inductive`, so they could not do this; only the generic tool — currently unwired to any active agent — could.)

Note: the public conclusions API (`crud.create_observations`) deliberately supports session-less, user-authored explicit conclusions and does no dedup; it is left unchanged and is a decision point for the Scopes RFC.

## Change 1 — enforce the invariant (refuse, never rewrite)

- `create_documents` refuses explicit documents with `session_name=NULL` (logged, skipped)
- exact dedup now keys on `(content, level, session-for-explicit)`; derived levels keep cross-session consolidation
- `is_rejected_duplicate` scopes candidate search to the same `level`, and the same `session_name` for explicit docs; session-less explicit docs are never merged
- the `create_observations` tool rejects `level="explicit"` outside message-ingestion (deriver) context

## Change 2 — `card_refresh` dream type

Lightweight card-only dream for event-driven refreshes (scope membership changes, cold starts):

- `DreamType.CARD_REFRESH`; `process_dream` dispatches it to `run_card_refresh_dream` → a new `CardRefreshSpecialist`
- tools restricted to `get_recent_observations`, `search_memory`, `update_peer_card` — no observation create/delete tools offered; iteration cap `min(6, DREAM.MAX_TOOL_ITERATIONS)`
- `rebuild=True` (carried in the dream payload and on `POST /v3/workspaces/{id}/schedule_dream`): the existing card is NOT injected into the prompt; the specialist rebuilds the card solely from observations present in the collection
- enqueue-able via the manual `enqueue_dream` path, bypassing the volume gates; the work-unit key embeds the dream type (`dream:card_refresh:...`) so it never collides with a pending omni dream
- never advances the omni guard pair (`last_dream_at` / `last_dream_document_count`); also works via `DREAM.ENABLED_TYPES` for volume-triggered scheduling
- telemetry: `CallPurpose.DREAM_CARD_REFRESH`, `DreamRunEvent` with `specialists_run=["card_refresh"]`

## Tests

- `tests/crud/test_document.py::TestSessionPurityInvariant` — null-session refusal, cross-session/cross-level exact-dedup isolation, derived-level consolidation preserved, semantic-dedup filter scoping
- `tests/utils/test_agent_tools.py` — non-deriver explicit rejection
- `tests/dreamer/test_card_refresh.py` — payload/queue plumbing (incl. enqueue alongside a pending omni dream), `process_dream` dispatch, guard-pair non-advancement, tool restriction, iteration cap, rebuild-mode prompt omission
- `tests/routes/test_workspaces.py` — `schedule_dream` accepts `card_refresh` + forwards `rebuild`

`uv run basedpyright` and `uv run ruff check` pass with zero errors/warnings; pre-commit passes. Full affected suites green (389 passed); the only failures are the 3 pre-existing `tests/crud/test_document.py` semantic-dedup tests that require an OpenAI embedding key.

## Summary by CodeRabbit

* **New Features**
  * Added a **card refresh** dream type that updates peer cards with a lightweight flow.
  * Introduced a **rebuild** option for card refresh to regenerate the peer card from current observations only.
* **Bug Fixes**
  * Improved duplicate handling by scoping exact/semantic deduplication to the correct **level** and, for explicit items, **session**.
  * Enforced **session-purity** rules for sessionless explicit content (now refused).
* **Tests**
  * Added end-to-end and unit coverage for routing, queue payloads, tool restrictions, and rebuild behavior.
* **Documentation**
  * Updated the dream scheduling request schema to include the new fields.


adavyas (dismissed):
lgtm

PR #884, opened 2026-07-07
Scopes Phase 2a: scope-kind peers, guardrails, and scopes CRUD routes

VVoruganti:
## Summary

Foundation of the **Scopes** feature (visibility boundaries within a peer, built on observer peers behind a facade). A scope is a named grouping of sessions; under the hood a scope named `therapy` is a peer named `scope__therapy` carrying the authoritative `{"kind": "scope", "observe_me": false}` configuration flag, which observes its member sessions (`observe_others=true, observe_me=false`) and never speaks. Developers manage scopes exclusively through the new routes (and the `scopes` field at session creation) and never see the observer/observed mechanics.

- `src/utils/scopes.py` is the single source of truth for the reserved prefix, the kind flag, and the name helpers (`scope_peer_name`, `is_scope_peer_name`, `scope_name_from_peer`).
- `src/crud/scope.py` implements the facade: scope-peer create-or-get (legacy peers occupying the reserved name **without** the kind flag are refused with 409, never adopted), membership add/list/remove reusing the session-peer upsert / soft-delete paths.
- `SessionCreate` gains optional `scopes: list[str]` — the no-backfill common path: each scope is created-or-gotten and the observer membership added at session creation.

## Guardrails

| # | Guardrail | Result |
|---|-----------|--------|
| 1 | Scope peers cannot author messages (`crud.create_messages`, covers both message routes) | 422 |
| 2 | Scope peers cannot be a chat or representation `target` (path-level observer is Phase 2b) | 422 |
| 3 | `peers.list` excludes scope peers by default; `PeerGet.kind` = `"scope"` / `"all"` switches the view (JSONB `configuration @> '{"kind": "scope"}'` filter) | — |
| 4 | Generic session-peer add/set/remove routes (and the session-create `peers` mapping) reject scope peers, directing to the scopes routes | 422 |
| 5 | Peer get-or-create rejects reserved-prefix (`scope__`) names | 422 |

## Routes

All under `/v3/workspaces/{workspace_id}/scopes`, gated by `require_auth(workspace_name=...)` — workspace-level key required; peer- and session-scoped keys get 401.

| Method | Path | Behavior |
|--------|------|----------|
| POST | `""` | create-or-get (201/200); body `{id, metadata?}`; response `{id (unprefixed), metadata, created_at}` |
| POST | `/list` | paginated scope list |
| GET | `/{scope_id}` | single scope (404 if absent or legacy-collision peer) |
| POST | `/{scope_id}/sessions` | add memberships (`{session_ids: [...]}`); sessions must exist (404) |
| DELETE | `/{scope_id}/sessions/{session_id}` | end membership (soft delete, same as generic remove-peer) |
| GET | `/{scope_id}/sessions` | list member session ids from `session_peers` |

## Litmus test

`tests/routes/test_scopes.py::test_scope_peer_observes_ingested_messages` proves the end-to-end observer semantics: after creating a scope and adding a session, a message ingested from a real peer fans out a representation queue item whose `observers` list includes the scope peer. `test_scope_membership_equals_hand_built_observer` additionally asserts the membership row is byte-identical in configuration to a hand-built observer peer — facade-less equivalence.

## Follow-ups

- Backfill of pre-existing documents and reconciliation on removal land in **DEV-1999** — for now, membership only affects messages ingested after the change (noted in route docs).
- Read-side `{scope}` resolution is **DEV-1998**.

Fixes DEV-1997 (part of DEV-1970 Scopes RFC)

## Summary by CodeRabbit

* **New Features**
  * Added scope management endpoints for creating, listing, viewing, and managing scope sessions.
  * Sessions can now be associated with scopes, with idempotent membership add, remove, and list operations.
  * Peer listings support filtering by regular peers, scope peers, or all peers.

* **Bug Fixes**
  * Prevented scope peers from being used as ordinary session members, message authors, observers, or context targets.
  * Improved validation and cache consistency during concurrent peer and scope creation.
  * Strengthened scope name and membership validation with clearer error handling.


src/crud/scope.py
```diff
@@ -0,0 +1,368 @@
...
+    existing_peers: list[models.Peer] = list(result.scalars().all())
+
+    changed_peers: list[models.Peer] = []
+    for existing_peer in existing_peers:
+        if not is_scope_peer(existing_peer.name, existing_peer.internal_metadata):
+            raise ConflictException(
+                f"A peer named '{existing_peer.name}' already exists in workspace "
+                + f"{workspace_name} but is not a scope. Rename or delete that "
+                + "peer before creating this scope."
+            )
```
ulyssepence, line 123: This is like if someone manually or gets automatically named with the scope prefix?

src/crud/scope.py
```diff
@@ -0,0 +1,368 @@
...
+    if reverse:
+        return stmt.order_by(models.Peer.created_at.desc(), models.Peer.id.desc())
+    return stmt.order_by(models.Peer.created_at.asc(), models.Peer.id.asc())
+
+
+async def get_scope(
+    db: AsyncSession,
+    workspace_name: str,
+    scope_name: str,
+) -> models.Peer:
```
ulyssepence: nit: prefer like `get_scope_or_raise` to make it more obvious elsewhere why we're calling this as a side-effect and not retaining the value.

src/crud/scope.py
```diff
@@ -0,0 +1,368 @@
...
+        .where(models.Session.is_active == True)  # noqa: E712
+    )
+    found = {row[0] for row in result.all()}
+    missing = sorted(requested - found)
+    if missing:
+        raise ResourceNotFoundException(
+            f"Session(s) {missing} not found in workspace {workspace_name}"
+        )
+
+    for session_name in sorted(requested):
```
ulyssepence, line 324: This is cool I didn't know sorted worked on all iterables

src/crud/session.py
```diff
@@ -962,8 +1079,9 @@ async def set_peers_for_session(
     peers_result = await get_or_create_peers(
         db,
         workspace_name=workspace_name,
-        peers=[schemas.PeerCreate(name=peer_name) for peer_name in peer_names],
+        peers=[schemas.PeerSpec(name=peer_name) for peer_name in peer_names],
     )
+    _reject_resolved_scope_peers(peers_result.resource)
```
ulyssepence, line 1095: Do we also want need to apply this in `get_session_peers` in src/routers/sessions.py ?

ulyssepence (commented):
Not sure I saw it in the PR, but is there a ticket/todo to check for peers that might already be using the scope prefix?

src/crud/peer.py
```diff
@@ -36,12 +47,213 @@ def peer_cache_key(workspace_name: str, peer_name: str) -> str:
...
+    if offenders:
+        raise ValidationException(
+            f"Peer name(s) {offenders} must match pattern {RESOURCE_NAME_PATTERN}"
+        )
+
+
+def scope_peer_clause() -> ColumnElement[bool]:
+    """SQL form of ``is_scope_peer()``: reserved name prefix AND the internal kind flag.
+
+    Lives here rather than in ``crud/scope.py`` because that module already imports
```
eisene, line 106: code smell - something should be refactored here (not necessarily in this PR, see below, up to you)

eisene: ### Claude says:

Agreed, and the comment's stated premise is circular. It says the clause can't live in `crud/scope.py` because "that module already imports from this one" — but `crud/scope.py` imports exactly two names from `crud/peer.py` (`src/crud/scope.py:31`):

```python
from .peer import peer_cache_key, scope_peer_clause
```

One of the two *is* `scope_peer_clause`. So the clause lives in `peer.py` because `scope.py` imports from `peer.py`, and the main thing `scope.py` imports from `peer.py` is the clause. Take it out of the equation and the only edge left is `peer_cache_key` — a trivial `PEER_CACHE_KEY_TEMPLATE.format(...)` helper that isn't peer *CRUD* at all (`crud/peer_card.py` imports it too).

The broader tell: this is the third import-pressure justification in the feature. The other two are genuine cycles at `src/crud/scope.py:302` and `:357`:

```python
# Imported lazily: crud.session imports this module for the session-create
# `scopes` path, so a module-level import would be circular.
from .session import upsert_session_peers
```

**Root cause:** `src/crud/` is flat — one module per table, all mutual siblings — so any cross-table operation must pick a host module and the loser gets a misplaced helper or a lazy import. `crud/scope.py` is a *facade* composing `peer` + `session_peers`, but it sits in the same tier as the primitives it composes.

**What would remove this specific comment (3 mechanical moves, ~1h, low risk):**

1. New `src/crud/scope_predicates.py` owning `scope_peer_clause`, `scope_peer_names`, `reject_scope_peers`, `reject_scope_observed`, `_reserved_name_candidates`. Imports only `models` + `utils.scopes`; `peer.py`/`scope.py`/`session.py` all import *down* into it. The clause is neither peer CRUD nor scope CRUD — it's a predicate over `models.Peer`.
2. Move `_reject_impossible_peer_names` / `_validate_new_peer_names` / `PEER_NAME_MAX_LENGTH` to a pure-validation module. The predicates module needs the first one, and that's the dependency that would otherwise recreate the edge.
3. Move `peer_cache_key` to a cache-keys module (two consumers already).

**Tempting shortcut worth avoiding:** putting `scope_peer_clause()` in `src/utils/scopes.py` next to `is_scope_peer()` looks ideal, since those two must stay in sync and currently live in different files each pointing at the other. But `utils/scopes.py` is DB-free and is imported by `src/schemas/api.py`, so adding `from src import models` would make the schemas layer transitively import the ORM. It wouldn't cycle (`src/models.py` imports only `config`, `utils.types`, `.db`) — it would just quietly couple two currently-independent layers. Better to keep the SQL form in the crud tier and cross-reference in a docstring.

**For the real `scope ↔ session` cycle, separately:**

- `SCOPE_MEMBERSHIP_CONFIG` is a pure `SessionPeerConfig` constant that `session.py` imports from `scope.py` → move to `utils/scopes.py`. Minutes, zero risk.
- Extract `src/crud/session_peer.py` owning the `session_peers` table: membership upsert, `left_at` soft-delete, `count_observers_in_config`, observer-limit check. Then `scope.py → session_peer.py` and `session.py → session_peer.py`, both downward, and both lazy imports die. Note `upsert_session_peers` was added in this PR purely as a public wrapper around `_get_or_add_peers_to_session` so the scopes facade could reach it — that wrapper already *is* the seam, it just hasn't been given a module. ~half a day, real risk (entangled with observer-limit counting and session cache invalidation), so its own PR.
- Hoisting the session-create scope orchestration up to `routers/sessions.py` would eliminate the last edge but split session creation and scope membership across two transactions unless the route threads one session through — trading real atomicity for a layering preference. Not worth it.

Suggest folding moves 1–3 into DEV-1998 as prep, since Phase 2b's `{scope}` read resolution will add more cross-table scope logic and inherit the same pressure. Not a blocker for this PR — the comment is honest and the behavior is correct.

eisene: agreed not a blocker for this PR

one thing i've seen work well in the past is to designate places in the codebase as "if you touch this, also clean it up" and letting future PRs slowly improve things incrementally

tests/routes/test_scope_route_policy.py
```diff
@@ -0,0 +1,760 @@
...
+
+# The scopes router *is* the facade; scope peers are its whole subject.
+_SCOPES_PREFIX = "/v3/workspaces/{workspace_id}/scopes"
+
+# A builder places `peer` into one position of one route and returns the response.
+Builder = Callable[[TestClient, str, str, str], object]
+
+
+@dataclass(frozen=True)
+class Case:
```
eisene, line 83: ### From Claude:

`Case` in `tests/routes/test_scope_route_policy.py:81` has `refuse: bool` — binary. The suite enumerates every `(method, path, position)` and proves a real scope is refused and an unflagged squatter isn't, but it has **no field for the missing-name axis**. So which variant each call site picked is only spot-checked by `test_peer_card_cannot_be_preseeded_for_a_future_scope`, `test_dream_cannot_be_queued_for_a_future_scope`, and `test_representation_rechecks_after_early_check` — three of the four strict sites, chosen by hand.

Adding `refuse_missing: bool` to `Case` and asserting it per position would systematically catch a future observed-position route that reaches for the permissive guard. Given the module docstring's own account ("four review passes each found the same class of defect: a place nobody had checked"), closing that hole in the enumeration is worth more than the ~15 lines the consolidation saves.

src/crud/scope.py
```diff
@@ -0,0 +1,368 @@
...
+        return await get_or_create_scopes(
+            db,
+            workspace_name,
+            scopes,
+            _retry=True,
+            _pending_invalidation=(_pending_invalidation or [])
+            + [p.name for p in changed_peers],
+        )
+
+    _cache_keys_to_invalidate = [
```
eisene, line 165: fragile cache invalidation logic in multiple places is also code smell, could drift and create hard-to-debug prod errors

--- 

### From Claude, abridged by me:

Five distinct concerns are stacked in those 16 lines, and four are genuinely required:

1. **Deferral via `on_commit` rather than deleting inline.** If you evict before the commit lands, a concurrent reader can miss, reload the *pre-commit* row, and repopulate the cache — then your commit lands and the entry is stale with nothing left to invalidate it. After-commit is the only correct ordering.
2. **`_pending_invalidation` carry-forward.** The subtle one. `begin_nested()` autoflushes the `h_metadata` mutations *before* opening the savepoint, so the `IntegrityError` rollback neither undoes them nor expires the now-clean ORM state. The retry therefore compares already-updated values, finds no diff, and skips the purge. Drop this and you get a *missed* invalidation — indefinitely stale cache.
3. **`dict.fromkeys(...)`** — dedup. Order preservation is incidental (deletes commute); a `set` would work, this just gives deterministic iteration.
4. **`if _cache_keys_to_invalidate else None`** — avoids scheduling a no-op callback.
5. **Eager key materialization.** This one is *weaker* than the twin comment in `peer.py` claims. It says "so the closure holds plain strings, not ORM objects," implying a post-commit lazy refresh — but `src/db.py:64,82` set `expire_on_commit=False` on both sessionmakers, and `src/deriver/queue_manager.py:794` explicitly relies on that to keep attributes readable post-close. So the cited failure can't occur today. Keep the eager capture (it's free and defensive), but the justification is aspirational rather than current.

So yes, it needs to be explicit. It does **not** need to be duplicated — and that's the actual issue.

## The real problem is that it's a near-verbatim copy

`src/crud/scope.py:165-180` and `src/crud/peer.py:~370-384` implement the same five concerns with the same `_pending_invalidation` protocol. Concern #2 is exactly the kind of invariant that gets fixed in one copy and silently missed in the other. The existence of a dedicated 197-line `tests/crud/test_get_or_create_retry_invalidation.py` says the maintainers already treat this as fragile enough to pin — which is an argument for having one implementation to pin rather than two.

src/crud/session.py
```diff
@@ -273,13 +318,39 @@ async def get_or_create_session(
             fetch_after_upsert=False,
         )
 
+    # Add the session to any requested scopes: create-or-get each scope peer
+    # and record an observer membership (observe_others=true, observe_me=false).
+    # No backfill happens here — membership only affects messages ingested
+    # after this point (backfill is DEV-1999).
```
eisene: Linear ticket number

src/routers/workspaces.py
```diff
@@ -225,6 +225,10 @@ async def schedule_dream(
     observed = request.observed if request.observed is not None else request.observer
     dream_type = request.dream_type
 
+    # The authoritative observed-position check lives in enqueue_dream, in the same
```
eisene, line 228: is this comment really needed? i think it will be kinda cryptic going forward since it's not clear this is referring to the "do not run dreams on a scope-peer" check

src/schemas/api.py
```diff
@@ -337,6 +396,26 @@ class SessionCreate(SessionBase):
...
     configuration: SessionConfiguration | None = None
+    scopes: list[str] | None = Field(
+        default=None,
+        max_length=100,
+        description=(
+            "Optional list of (unprefixed) scope names to add this session to. "
+            "Each scope is created if it does not exist yet. Note: scope "
+            "membership only affects messages ingested after the session is "
+            "added to the scope; backfill of pre-existing documents lands in a "
+            "follow-up (DEV-1999)."
```
eisene: Linear ticket number

src/schemas/api.py
```diff
@@ -337,6 +396,26 @@ class SessionCreate(SessionBase):
...
+        ),
```
eisene, line 407: `Claude says: violates "no Plastic Labs internals in public repo" rule`

This is the highest-impact instance of the four. It's a Pydantic `Field(description=...)`, so `DEV-1999` lands in `components.schemas.SessionCreate.scopes.description` in the **published OpenAPI spec** — and since generated SDKs copy field descriptions into method/parameter docstrings, it would ship in the Python and TypeScript SDK source too, not just the docs site.

Verified by generating the spec from this branch:

```
components.schemas.SessionCreate.scopes:
    Optional list of (unprefixed) scope names to add this session to. Each scope
    is created if it does not exist yet. Note: scope membership only affects
    messages ingested after the session is added to the scope; backfill of
    pre-existing documents lands in a follow-up (DEV-1999).
```

Suggested replacement — states the behavior instead, which is strictly more useful to an API consumer than a ticket ID they can't open:

> Membership applies only to messages ingested after the session is added to the scope; conclusions already derived are not backfilled.

— Claude

src/routers/scopes.py
```diff
@@ -0,0 +1,160 @@
...
+):
+    """
+    Add Sessions to a Scope.
+
+    All named sessions must already exist (404 otherwise). Returns the scope's
+    full member session list after the addition.
+
+    Note: membership only affects messages ingested after this call — backfill
+    of pre-existing documents lands in a follow-up (DEV-1999).
+    """
```
eisene, line 114: `Claude says: violates "no Plastic Labs internals in public repo" rule`

This one is **published**. FastAPI harvests route docstrings into the OpenAPI `description`, so `DEV-1999` appears verbatim in the public API reference for `POST /v3/workspaces/{workspace_id}/scopes/{scope_id}/sessions`. Confirmed against a locally generated spec.

Suggested: drop the ticket, keep the behavior —

> Note: membership applies only to messages ingested after this call; conclusions already derived are not backfilled.

Related, same file: the module docstring at line 12 also names `DEV-1999`, but that one is **not** published (FastAPI only harvests route docstrings), so it falls under the lower-severity internal-comment case rather than this one.

— Claude

src/routers/scopes.py
```diff
@@ -0,0 +1,160 @@
...
+    scope_id: str = Path(...),
+    session_id: str = Path(...),
+    db: AsyncSession = db,
+):
+    """
+    Remove a Session from a Scope.
+
+    Note: documents already derived while the session was a member are left in
+    place — reconciliation on removal lands in a follow-up (DEV-1999).
+    """
```
eisene, line 140: `Claude says: violates "no Plastic Labs internals in public repo" rule`

Also **published** — this docstring becomes the OpenAPI `description` for `DELETE /v3/workspaces/{workspace_id}/scopes/{scope_id}/sessions/{session_id}`, so the public API reference currently reads:

> Remove a Session from a Scope. Note: documents already derived while the session was a member are left in place — reconciliation on removal lands in a follow-up (DEV-1999).

Suggested: end the sentence at "left in place." The fact that reconciliation is planned is internal roadmap; the fact that it doesn't happen today is the part callers need.

— Claude

src/crud/scope.py
```diff
@@ -0,0 +1,368 @@
...
+``scope.<name>`` carrying ``{"kind": "scope"}`` in ``internal_metadata`` (the
+authoritative, user-unwritable flag) and ``{"observe_me": false}`` in
+``configuration``, that observes its member sessions (``observe_others=true``)
+and never speaks.
+See ``src/utils/scopes.py`` for the namespace helpers.
+
+Membership only affects messages ingested *after* a session is added to a
+scope; backfill of pre-existing documents and reconciliation on removal land
+in a follow-up (DEV-1999).
+"""
```
eisene, line 13: `Claude says: violates "no Plastic Labs internals in public repo" rule`

`DEV-1999` in an internal comment — repo-visible but **not** published to OpenAPI, so lower blast radius than the route/schema instances. Same file also at lines 285 and 345.

Worth noting this file already states the behavior in full ("Membership only affects messages ingested *after* a session is added to a scope; backfill of pre-existing documents and reconciliation on removal land in a follow-up"), so the ticket ID is carrying no information an outside reader can act on — deleting the parenthetical loses nothing.

— Claude

src/crud/session.py
```diff
@@ -273,13 +318,39 @@ async def get_or_create_session(
...
+    scopes_result = None
```
eisene, line 325: `Claude says: violates "no Plastic Labs internals in public repo" rule`

`DEV-1999` in an internal comment. Not published, so lower severity than the `src/routers/scopes.py` and `src/schemas/api.py` instances — but the surrounding sentence already says "No backfill happens here — membership only affects messages ingested after this point," which is the whole actionable content. The ticket reference can just go.

— Claude

src/crud/session.py
```diff
@@ -1020,8 +1162,17 @@ async def _get_or_add_peers_to_session(
         result = await db.execute(select_stmt)
         return list(result.scalars().all())
 
-    # Only validate observer limit if we're adding peers with observe_others=True
-    new_observer_count = count_observers_in_config(peer_names)
+    # Scope memberships carry observe_others=True but do not count against the
+    # limit. The limit bounds per-observer deriver fan-out for real peers; a scope
+    # costs document rows, not LLM calls (Scopes RFC §5.2), and counting them would
+    # cap scopes-per-session at SESSION_OBSERVERS_LIMIT and surface as an
```
eisene, line 1179: `Claude says: violates "no Plastic Labs internals in public repo" rule`

`Scopes RFC §5.2` is arguably a worse instance than the bare ticket IDs: it cites a *section number* of a document no external reader can obtain, to justify a genuinely non-obvious decision (scopes not counting toward `SESSION_OBSERVERS_LIMIT`). To an outside contributor it reads as though there's an authority backing the choice that they have no way to consult.

The good news is that the citation is redundant — the comment already contains the actual argument inline: "a scope costs document rows, not LLM calls." Dropping `(Scopes RFC §5.2)` loses no information.

For consistency: `src/telemetry/events/trace.py:146` has a similar bare `§6.3` reference, and `tests/dreamer/test_card_refresh.py:1` names "Scopes RFC" directly — both pre-existing, not this PR.

— Claude

tests/routes/test_scope_route_policy.py
```diff
@@ -0,0 +1,760 @@
...
+        ),
+    ),
+    Case(
+        "GET",
+        f"{_W}/peers/{{peer_id}}/context",
+        "peer_id",
+        False,
+        reason=(
+            "Read-only. The read-side scope surface belongs to Phase 2b (DEV-1998), "
+            "which owns the `scope` option on the context routes."
```
eisene: `Claude says: violates "no Plastic Labs internals in public repo" rule`

Three instances in this file's policy table: `Phase 2b (DEV-1998)` here and at line 471, and `Phase 3 (DEV-2002)` at line 491.

Of everything flagged in this PR, I'd argue these are the most sensitive — not because tests are public-facing (they aren't published anywhere), but because a ticket ID is trivia while **"Phase 3 exists and owns this surface"** is unshipped product roadmap. The ticket numbers additionally make the sequence enumerable.

The policy reasons themselves are excellent and should stay; they just don't need the phase/ticket labels. E.g. "The read-side scope surface belongs to Phase 2b (DEV-1998), …" → "The read-side scope surface is not implemented yet, …". The `test_policy_entries_are_well_formed` assertion that each allow-case `reason` exceeds 30 characters still passes comfortably.

— Claude

src/routers/peers.py
```diff
@@ -122,7 +140,15 @@ async def update_peer(
     peer: schemas.PeerUpdate = Body(..., description="Updated peer parameters"),
     db: AsyncSession = db,
 ):
-    """Update a Peer's metadata and/or configuration."""
+    """Update a Peer's metadata and/or configuration.
+
+    Three-way on the reserved namespace, all enforced inside ``crud.update_peer``
+    on the resolved row so there is no check-then-use window: a real scope is
```
eisene: `Claude says: this reasoning belongs in a code comment, not a docstring`

Different issue from the ticket-number ones — no internals leak, but this docstring is **published**. FastAPI harvests it into the OpenAPI `description` for `PUT /v3/workspaces/{workspace_id}/peers/{peer_id}`, so the public API reference currently tells consumers about `crud.update_peer`, check-then-use windows, and "facade-managed" peers:

```
PUT /v3/workspaces/{workspace_id}/peers/{peer_id}
  Update a Peer's metadata and/or configuration. Three-way on the reserved
  namespace, all enforced inside ``crud.update_peer`` on the resolved row so
  there is no check-then-use window: a real scope is refused (this route
  replaces `configuration` wholesale, so it must never touch a facade-managed
  peer); an existing *unflagged* peer that merely occupies the namespace is a
  normal peer and updates fine; and a reserved-prefix name that does not exist
  is refused by create-path validation rather than being minted.
```

The content is genuinely good and worth keeping — it explains *why* the check lives on the resolved row, which is the subtle part. It just wants to be a `#` comment above the handler body (or on the guard in `crud.update_peer` itself) rather than in the string FastAPI publishes.

What a caller actually needs from the docstring is the observable contract, roughly:

> Update a Peer's metadata and/or configuration. Returns 422 if the peer is a scope — use the scopes routes to manage scopes.

Same pattern to check elsewhere in the PR: any added route docstring that explains implementation reasoning rather than observable behavior ends up in the published spec. The `/scopes` route docstrings are mostly fine on this axis; this is the one that drifted.

— Claude

tests/crud/test_get_or_create_retry_invalidation.py
```diff
@@ -0,0 +1,197 @@
+"""Regression tests for cache invalidation across the get_or_create retry path.
```
eisene, line 1: General comment not for fixing in this PR: root cause of this race appears to be that there is a DB and a Redis cache and nothing's enforcing their lockstep. This scares me, probably worth thinking about a robustness-related refactor for stuff like this at some point.

# Idea from Claude

**(a) Upsert, which deletes the retry path.** The bug exists *only* because of savepoint → `IntegrityError` → retry. Replace that with `INSERT ... ON CONFLICT DO UPDATE ... RETURNING`:

- no retry, so no lost bookkeeping, so `_pending_invalidation` disappears from both `peer.py` and `scope.py`
- "what changed" comes from `RETURNING` instead of pre-flush ORM diffing
- the concurrent-create race gets arbitrated by the unique constraint, which was already doing that job — no second coordination system

This *removes* code instead of adding a layer, and the pattern is already in the codebase. Best value per unit of risk.

**(b) The `after_commit` invalidation hook.** Kills the other class: `post_commit()` is caller-remembered at 11 sites and forgetting one is silent. This is what your instinct is actually reaching for — one place where "the transaction committed" *implies* "the affected keys are purged" — but achieved through ordering and inescapability rather than mutual exclusion.

## The residual race neither fixes

Worth naming so you can consciously accept it: reader loads the row at T1 (pre-commit), your commit lands T2, your delete lands T3, the reader's `SET` lands T4 → stale entry with nothing left to invalidate it. Mitigations cheapest-first: short TTL (already have), delayed second delete, or row-version-stamped keys with CAS on write. A write mutex doesn't help here either — the reader never takes it.

So: build the abstraction, but frame it as *"one owner of cache lifecycle, bound to commit"* rather than *"a lock around both stores."* Sequence it upsert → commit hook → measure, and only reach for coordination primitives if you still observe staleness after both. My guess is you won't.

eisene (commented):
1. `SessionCreate.scopes` bypasses the workspace-level auth gate — High

The `/scopes` router requires a workspace key. The session-create route does not, and it can do the same two privileged things.

POST `/v3/workspaces/{ws}/sessions` is a self-authorizing route using bare `require_auth()`. In `src/security.py:227-231`, with no `workspace_name/peer_name/session_name` declared, `auth()` returns the token unconditionally. The handler (`src/routers/sessions.py:287`) then only checks the workspace claim and — when `jwt_params.s is not None` — the session name. A peer-scoped token (`{w: ws, p: alice}`, `s` unset) skips the session check entirely.

So with a peer-scoped key:

POST /v3/workspaces/ws/sessions  {"id": "anything", "scopes": ["therapy"]}

reaches `src/crud/session.py:326-343` and will create-or-get `scope.therapy` and attach the session as a member. The same key gets 401 on `POST /v3/workspaces/ws/scopes` and `POST /scopes/therapy/sessions`.

Impact is write-side: creating scope peers in the workspace, and injecting a session into an existing scope so that scope's collection accumulates conclusions from messages the scope owner never authorized. Read-side extraction isn't reachable until DEV-1998, which limits blast radius today but makes this cheaper to fix now than after the read path lands.

`test_scopes_routes_require_workspace_level_key` covers only the `/scopes` router; nothing covers the scopes field, and `test_scope_route_policy.py` classifies peer positions, not this field.

Suggested fix — in the session-create handler, before `crud.get_or_create_session`:

```python
if session.scopes and not (
    jwt_params.ad or (jwt_params.p is None and jwt_params.s is None)
):
    raise AuthenticationException(
        "Scope membership requires a workspace-level key"
    )
```

and add the peer- and session-scoped cases to `test_scopes_routes_require_workspace_level_key`.

2. Scope peers leak through `GET /sessions/{id}/peers` — Medium

`crud.get_peers_from_session (src/crud/session.py:910-938)` has no scope exclusion, so once a session is in a scope, listing that session's peers returns a peer literally named `scope.therapy` with `observe_others=true`.

This directly contradicts the facade premise ("developers … never see the observer/observed mechanics") and is inconsistent with guardrail #3, where `get_peers` got exactly this treatment via `~scope_peer_clause()`. `internal_metadata` is correctly absent from the response, so the leak is the name and membership — enough to expose the mechanic.

Fix: apply `~scope_peer_clause()` in `get_peers_from_session`, ideally via the same kind: `Literal["scope", "all"] | None` parameter `get_peers` now takes, so the scopes facade can reuse it for a member listing later.

3. Unbounded membership list responses — Medium

`GET /scopes/{scope_id}/sessions` and the `POST /scopes/{scope_id}/sessions` response both return the complete member list. Additions are capped at 100 per request, but total membership is unbounded — a scope spanning a workspace's sessions returns every id in one payload. Every other list surface in the repo returns `Page[...]`.

Fix: make the GET a `POST /list`-style paginated route (matching `peers.list` / `sessions.list` convention), and have the add route return just the ids it added, or nothing.

src/crud/scope.py
eisene: pre-existing legacy peers with the namespace collision, right?

akattelu (commented):
Think we need to take another pass at removing some redundant comments and also shrinking the number of tests we add. Generally looks good besides that.

src/crud/scope.py
VVoruganti: @eisene yes, exactly — pre-existing legacy peers are the only way this happens, and that's also the answer to @ulyssepence's top-level question about an audit ticket: we don't need one, because the API contract already refuses every other route in.

`RESOURCE_NAME_PATTERN` is `^[a-zA-Z0-9_-]+$`, which excludes `.`, so no peer created through the validated API can occupy the `scope.` namespace — that exclusion is the whole reason the prefix is a safe reservation (`src/utils/scopes.py:26-31`). It's enforced on `PeerCreate.name` at the schema boundary and, new in this PR, on the crud insert path via `_validate_new_peer_names`, so request-controlled names (message authors, session peer-map keys) can't mint one either.

The one exception is peers carried over by the `users`→`peers` rename in `d429de0e5338`: those predate the pattern and were never charset-validated, so `scope.production` is a possible real name in an old workspace. That's precisely why being a scope requires **both** halves — the reserved name *and* `{"kind": "scope"}` in `internal_metadata` — why `get_or_create_scopes` returns **409** on a legacy collision instead of adopting the row, and why every guard keys off the flag rather than the prefix.

So the 409 *is* the audit: it fires at the moment of collision, on the one operation that cares, rather than a migration scan that would have to guess intent for each hit and would still race anything created afterwards. A legacy squatter keeps working as an ordinary peer in the meantime, which is the behavior you'd want — locking a tenant out of a peer they've been using would be the worse failure.

Covered by `test_create_scope_rejects_legacy_collision`, `test_legacy_prefixed_peer_without_flag_is_not_a_scope`, and `test_legacy_dotted_peer_name_is_fully_usable`. The route-policy suite additionally proves every guarded route still accepts an unflagged squatter, so the flag-not-prefix rule is enumerated rather than spot-checked.

src/crud/peer.py
akattelu: Consider moving some of the validation and sanitization logic in crud into src/utils/scopes.py

src/crud/scope.py
akattelu: filed a separate issue for cleaning up the cache logic here

src/crud/document.py
```diff
@@ -952,7 +952,25 @@ async def create_observations(
...
+    # but it must never be *observed*: scope peers carry observe_me=false and no
+    # representation is ever formed of one. Without this, a conclusion about a
+    # scope persists and a (observer, scope) collection is created for it.
+    #
+    # The strict variant because this is an observed position, though defence in
+    # depth rather than the active guard: the loop above resolves every peer, so a
+    # reserved name that does not exist yet already 404s before reaching here. If
+    # that validation ever stops covering observed_id, this still refuses the
+    # pre-seeding case instead of persisting a conclusion that a later-created
+    # scope would retroactively own.
```
akattelu, line 967: shrinking this comment

src/crud/peer.py
```diff
@@ -36,12 +47,214 @@ def peer_cache_key(workspace_name: str, peer_name: str) -> str:
...
+    _reject_impossible_peer_names(materialized)
+    return sorted({n for n in materialized if scopes_util.is_scope_peer_name(n)})
+
+
+async def reject_scope_observed(
+    db: AsyncSession,
+    workspace_name: str,
+    names: Iterable[str],
+    *,
+    action: str,
```
akattelu, line 146: consider changing name of this param

src/crud/message.py
```diff
@@ -312,7 +313,22 @@ async def create_messages(
...
+
+    Raises:
+        ValidationException: If a message is authored by a scope peer
     """
+    # Scope peers are silent observers — they can never author messages. Keyed
+    # off name+flag so a legacy peer merely occupying the reserved namespace
+    # keeps ingesting. Must stay *before* get_or_create_session below: that call
+    # would create the scope peer and add it with a default SessionPeerConfig(),
+    # clobbering its observe_others=True/observe_me=False membership config.
+    await reject_scope_peers(
```
akattelu, line 325: any duplicate logic between `reject_scope_observed` and this?

src/crud/peer.py
```diff
@@ -36,12 +47,214 @@ def peer_cache_key(workspace_name: str, peer_name: str) -> str:
...
+        raise ValidationException("Peer name(s) must not contain NUL (0x00) bytes")
+    too_long = sorted({n for n in names if len(n) > PEER_NAME_MAX_LENGTH})
+    if too_long:
+        raise ValidationException(
+            f"Peer name(s) {too_long} must be at most "
+            + f"{PEER_NAME_MAX_LENGTH} characters"
+        )
+
+
+def _validate_new_peer_names(names: list[str]) -> None:
```
akattelu, line 80: could probably move this to peer create as a validation function

src/crud/peer.py
```diff
@@ -36,12 +47,214 @@ def peer_cache_key(workspace_name: str, peer_name: str) -> str:
     )
 
 
+def _reject_impossible_peer_names(names: Collection[str]) -> None:
```
akattelu, line 50: also consider moving to peer spec

src/crud/peer.py
```diff
@@ -36,12 +47,214 @@ def peer_cache_key(workspace_name: str, peer_name: str) -> str:
...
+
+    ``autoescape=True`` is future-proofing: '.' is not a LIKE wildcard, but '_' is,
+    so under a ``scope__``-style prefix an unescaped ``startswith`` would also match
+    ``scopeXY...``. Both columns are NOT NULL with defaults, so the negation
+    ``~scope_peer_clause()`` has no NULL-semantics trap.
+    """
+    return and_(
+        models.Peer.name.startswith(scopes_util.SCOPE_PEER_PREFIX, autoescape=True),
+        models.Peer.internal_metadata.contains({"kind": scopes_util.SCOPE_KIND}),
+    )
```
akattelu, line 118: would be nice to break this up but we need a way to not hog peer names with scoped peers, and kind can be useful in the future. this sql clause could just use one. but useful as validation.

akattelu (approved):

PR #889, opened 2026-07-08
Split `REPRESENTATION_BATCH_MAX_TOKENS` into a "minimum work unit" setting on the producer side and a "maximum LLM tokens" setting on the consumer side

eisene:
# Background

The `REPRESENTATION_BATCH_MAX_TOKENS` used to play a double role:
- On the producer side of the deriver queue in `get_and_claim_work_units` it represented the minimum cumulative token length of the batch of messages to queue for the deriver.
- On the consumer side in `get_queue_item_batch` within `process_work_unit` it represented the maximum token size of a deriver request.

Now there are 2 separately tunable knobs for this: `DERIVER_REPRESENTATION_BATCH_WORK_UNIT_TARGET_TOKENS` and `DERIVER_REPRESENTATION_BATCH_LLM_MAX_TOKENS`.

# Changes

- Add the two new config items to the config in the dotenv and the PyDantic model
  - Presence of old item throws with an informative error message instructing the developer how to fix
- Feed them to the appropriate places within the deriver queue manager
- Update tests and docs

## Side quest

Add `.omc` to `.gitignore` to ignore oh-my-claude-code stuff if anyone is using it.

# Test plan

- I have a local "two LLMs talking to each other" test, use it to exercise the queue and check state in the DB.

## Summary by CodeRabbit

* **New Features**
  * Split representation batching configuration into separate controls for work-unit claiming and the maximum token limit per LLM call.
  * Added support for disabling the work-unit accumulation gate with a zero-value setting.
* **Bug Fixes**
  * Legacy batching settings are now rejected at startup to prevent silent fallback.
  * Updated batching gating and validation to use the new token-limit controls.
* **Documentation**
  * Updated configuration examples and troubleshooting guidance to reflect the new batching options.
* **Tests**
  * Adjusted and added coverage for the renamed batching token settings and validation behavior.


src/config.py
```diff
@@ -857,7 +857,20 @@ def _MODEL_CONFIG_DEFAULT() -> ConfiguredModelSettings:
...
 
-    REPRESENTATION_BATCH_MAX_TOKENS: Annotated[
+    # Minimum tokens a representation work unit must accumulate (summed over
+    # its own unprocessed messages) before it becomes claimable. Bypassed by
+    # FLUSH_ENABLED and by REPRESENTATION_BATCH_MAX_AGE_SECONDS age-flushing.
+    # 0 disables the accumulation gate entirely (equivalent to FLUSH_ENABLED
+    # for claiming): work units are claimable as soon as anything is pending.
+    REPRESENTATION_BATCH_WORK_UNIT_TARGET_TOKENS: Annotated[
+        int,
+        Field(default=512, ge=0, le=16_384),
```
eisene, line 867: It now makes sense for this to be 0. If you set it to zero it becomes effectively the same as `FLUSH_ENABLED`. I let the duplication stand since I feel like it's a little more developer-friendly this way right now, not sure what the philosophy about this kind of thing is.

src/config.py
```diff
@@ -881,11 +894,33 @@ def _merge_model_config_defaults(cls, data: Any) -> Any:
             )
         return data  # pyright: ignore[reportUnknownVariableType]
 
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_removed_batch_max_tokens(cls, data: Any) -> Any:
```
eisene, line 899: This will raise a developer-friendly error message for the old config item. Can be expanded into a config-version-migrator later if we want.

.gitignore
```diff
@@ -193,3 +193,6 @@ metrics.jsonl
 AGENTS.md
 lancedb_data/
 grafana-data/
+
+# Claude Code addon stuff
+.omc
```
eisene, line 198: Side quest

.env.template
```diff
@@ -131,7 +131,8 @@ LLM_OPENAI_API_KEY=your-api-key-here
 # DERIVER_MAX_INPUT_TOKENS=25000
 # DERIVER_MAX_CUSTOM_INSTRUCTIONS_TOKENS=2000
 # DERIVER_WORKING_REPRESENTATION_MAX_OBSERVATIONS=100
-# DERIVER_REPRESENTATION_BATCH_MAX_TOKENS=1024
+# DERIVER_REPRESENTATION_BATCH_WORK_UNIT_TARGET_TOKENS=512  # Min tokens a work unit accumulates before the deriver claims it; 0 disables the gate
```
eisene, line 134: Wdyt of these names?

Rajat-Ahuja1997 (commented):
looks good! 1 comment on naming: 

i like `DERIVER_REPRESENTATION_BATCH_WORK_UNIT_TARGET_TOKENS` but I think `LLM_MAX_TOKENS` is not a good suffix here, intuitively this makes me think it's capping the deriver llm's max output tokens. 

my suggestion is ` DERIVER_REPRESENTATION_BATCH_MAX_INPUT_TOKENS` but feel free to workshop this

eisene:
> looks good! 1 comment on naming:
> 
> i like `DERIVER_REPRESENTATION_BATCH_WORK_UNIT_TARGET_TOKENS` but I think `LLM_MAX_TOKENS` is not a good suffix here, intuitively this makes me think it's capping the deriver llm's max output tokens.
> 
> my suggestion is ` DERIVER_REPRESENTATION_BATCH_MAX_INPUT_TOKENS` but feel free to workshop this

what about `DERIVER_REPRESENTATION_BATCH_TARGET_INPUT_TOKENS`, to your point max implies a cutoff which is not what this is, wdyt?

Rajat-Ahuja1997 (approved):

PR #890, opened 2026-07-08
fix: date dreamer conclusions to latest source observation

ajspig:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved observation timestamps to align with the underlying source messages.
  * Ensured timestamps are consistently formatted in UTC.
  * Added a fallback to the current time when source timestamps are unavailable.

### DIFF
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -24,7 +24,11 @@ from src.telemetry.events import (
     emit,
 )
 from src.utils import summarizer
-from src.utils.formatting import format_new_turn_with_timestamp, utc_now_iso
+from src.utils.formatting import (
+    format_datetime_utc,
+    format_new_turn_with_timestamp,
+    utc_now_iso,
+)
 from src.utils.representation import Representation
 from src.utils.types import ToolResult, embedding_call_purpose, get_current_iteration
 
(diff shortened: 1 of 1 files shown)


src/utils/agent_tools.py
```diff
@@ -1306,6 +1311,52 @@ def _normalize_observation_id(obs_id: str) -> str:
...
+    latest: datetime | None = None
+    async with tracked_db("create_observations.source_ts", read_only=True) as db:
+        docs = await crud.fetch_documents_by_ids(
+            db,
+            workspace_name=ctx.workspace_name,
+            observer=ctx.observer,
+            observed=ctx.observed,
+            document_ids=list(set(source_ids)),
+        )
+        # Read ORM attributes inside the session: once it closes the Document
```
eisene: this reads like an LLM self-reminder comment, i wouldn't merge this comment into main personally

src/utils/agent_tools.py
```diff
@@ -1368,10 +1419,14 @@ async def _handle_create_observations_impl(
...
         message_ids = [msg.id for msg in ctx.current_messages]
-        message_created_at = str(ctx.current_messages[-1].created_at)
+        # same ISO-8601 Z format as the dreamer path below
+        message_created_at = format_datetime_utc(ctx.current_messages[-1].created_at)
     else:
+        # Dreamer path: no current messages. Backdate the conclusion to the latest source observation.
         message_ids = []
-        message_created_at = utc_now_iso()
+        message_created_at = (
+            await _latest_source_timestamp(ctx, observations)
```
eisene, line 1430: i thought through it and i agree that this is a more correct setting - this is essentially the point at which the conclusions made by the dreamer became possible, which is the relevant thing to track, not so much when the dreamer ran.

i think it would be worth documenting that in a brief comment here

eisene (commented):
two minor comments about documentation

src/utils/agent_tools.py
ajspig: okay will add. To clarify, the conclusion made by the dreamer is still captured & stored in Document.created_at, so we don't loose that information.

eisene (approved):
lgtm

PR #891, opened 2026-07-08
feat(cli): add device code oauth login for honcho servers

akattelu:
Adds a path to `honcho init` to login with the oauth device code flow.

This isn't supported in prod in the managed offering yet, so this shouldn't release yet, but putting it up to get early eyes on the code and test locally.

Notes:
* Passes `source`as `honcho-cli`. Doesn't interact with any configuration set by the remote part of the flow.
* Stores and refreshes access token locally.
* Checks against the remote's .well-known to decide if the option should be shown or not.

## Summary by CodeRabbit

* **New Features**
  * Enhanced guided setup to support both API key entry and OAuth device login, with richer prompts and verification instructions, plus JSON-friendly output with sensitive values redacted.
  * Added persistence and reuse of OAuth device tokens, automatically choosing the effective credential source.
  * Expanded diagnostics to include more detailed authentication status and OAuth expiry timing.

* **Bug Fixes**
  * Automatically refreshes expired OAuth tokens when possible before using diagnostics or creating a client.
  * Improved diagnostic behavior to avoid connectivity checks when credentials are incomplete.

* **Tests**
  * Added/updated coverage for OAuth device flow, token polling/refresh, and OAuth config persistence/redaction.

### DIFF
--- a/honcho-cli/src/honcho_cli/commands/setup.py
+++ b/honcho-cli/src/honcho_cli/commands/setup.py
@@ -7,6 +7,8 @@
 from __future__ import annotations
 
 import json
+import time
+import webbrowser
 
 import typer
 from honcho import (
@@ -19,13 +21,14 @@ from honcho import (
 from rich.console import Console
 from rich.panel import Panel
 
-from honcho_cli import __version__
+from honcho_cli import __version__, oauth
 from honcho_cli.branding import BANNER, BRAND, ICON_FAIL, ICON_OK, ICON_RUN
-from honcho_cli.common import get_resolved_config
+from honcho_cli.common import get_resolved_config, maybe_refresh_token
 from honcho_cli.config import (
     CONFIG_FILE,
     DEFAULT_BASE_URL,
     CLIConfig,
+    OAuthTokens,
 )
 from honcho_cli.output import print_error, print_result, set_json_mode, use_json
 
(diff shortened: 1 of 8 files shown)


honcho-cli/src/honcho_cli/commands/setup.py
```diff
@@ -117,19 +120,129 @@ def init(
         _console.print()
         _console.print()
 
-    final_key = _prompt_api_key(key_val)
-    final_url = _prompt_url(url_val)
+    # Non-interactive (JSON/piped) or an explicit --api-key: manual-key path.
```
eisene, line 123: nit: starting to get a little long - worth breaking up into separate "interactive" and "non-interactive" functions?

honcho-cli/src/honcho_cli/config.py
```diff
@@ -40,6 +73,43 @@
...
+class OAuthTokens:
+    """Device-grant tokens persisted under the config ``oauth`` key."""
+
+    access_token: str = ""
+    refresh_token: str = ""
+    access_expires_at: float = 0.0  # epoch seconds
+    client_id: str = ""
+    scope: str = ""
+
+    def access_valid(self, skew: int = 60) -> bool:
```
eisene, line 99: nit: worth adding to docstring that this doesn't literally check token expiration and that's OK?

honcho-cli/src/honcho_cli/config.py
```diff
@@ -132,7 +231,9 @@ def redacted(self) -> dict[str, str]:
             if fld.name == "api_key":
                 # Show ``***<last4>`` only — enough to compare keys without
                 # leaking the header or body of the JWT.
-                d[fld.name] = "***" + val[-4:] if len(val) > 4 else "***"
+                d[fld.name] = _redact_token(val)
```
eisene: personal nit: i find single letter variables hard to read unless they're literally loop indices. up to you honestly if you want to follow that, just wanted to raise it

honcho-cli/tests/test_oauth.py
```diff
@@ -0,0 +1,231 @@
...
+        with patch("honcho_cli.oauth.httpx.get", return_value=FakeResponse(200, body)):
+            assert oauth.supports_device_login("https://api.honcho.dev") is True
+
+    def test_false_when_device_grant_absent(self):
+        body = {"grant_types_supported": ["authorization_code", "refresh_token"]}
+        with patch("honcho_cli.oauth.httpx.get", return_value=FakeResponse(200, body)):
+            assert oauth.supports_device_login("https://api.honcho.dev") is False
+
+    def test_false_on_non_200(self):
+        with patch("honcho_cli.oauth.httpx.get", return_value=FakeResponse(404, "")):
```
eisene: worth adding a 500 here or something? just something different from only 404?

honcho-cli/tests/test_config.py
```diff
@@ -104,6 +120,60 @@ def test_api_key_redaction_empty_omitted():
     assert "api_key" not in CLIConfig(api_key="").redacted()
 
 
+class TestOAuth:
```
eisene, line 149: Claude thinks it's worth adding these tests, this seems to make sense to me:

---

**1. Stale `apiKey` removal on save — the missing twin of `test_empty_oauth_is_dropped`.** `save()` pops `apiKey` from disk when `self.api_key` is empty (`config.py:199-200`). That's exactly what happens when a user with an old manual key does a device login (`CLIConfig(base_url=..., oauth=tokens).save()`). If this ever regressed to leave the old key behind, `resolved_api_key()` would prefer the dead manual key over the fresh OAuth token *forever* — every command 401s and the user can't fix it by re-logging-in. The symmetric oauth-drop case is tested; this direction isn't:

```python
def test_stale_api_key_is_dropped(self, cfg_path):
    cfg_path.write_text(json.dumps({"apiKey": "old-manual-key"}))
    CLIConfig(base_url="http://localhost:8000", oauth=self._tokens(9999999999)).save()
    assert "apiKey" not in json.loads(cfg_path.read_text())
```

**2. Garbage `accessExpiresAt` on load.** `_coerce_epoch` exists specifically because the config file is shared with sibling tools and hand-editable, and its contract is "garbage → 0.0 → treated as expired" (`config.py:53-62`). Nothing tests it. A test should seed `{"oauth": {"accessToken": "x", "accessExpiresAt": "not-a-number"}}` and assert load succeeds, `access_valid()` is `False` (i.e., it degrades to the refresh path, not a crash), plus the numeric-string case (`"12345"` → `12345.0`).

**3. Empty-string env var popping.** `load()` has a deliberate side effect: `HONCHO_API_KEY=""` is *removed from `os.environ`* because the SDK reads these vars directly and crashes on empty strings with a Pydantic ValidationError (`config.py:170-174`). This is the subtlest behavior in the file, it mutates global state, and it's untested — exactly the kind of thing a future cleanup would delete as "dead code." Test: set the env var to `""`, call `load()`, assert the var is gone from `os.environ` and the file value won through.

**4. Refresh-token rotation fallback.** `OAuthTokens.from_response(refresh_fallback=...)` keeps the prior refresh token when the server doesn't rotate one — optional per RFC 6749 §5.1 (`config.py:106`). `test_expired_token_refreshes_and_persists` in `test_common.py` only covers the rotated case. If the fallback broke, the refresh token would silently become `""`, and the failure would surface one token-lifetime *later* as a mysterious `SESSION_EXPIRED` — a delayed, hard-to-bisect logout. Either a direct `from_response` unit test or a `maybe_refresh_token` variant where the mocked `TokenResponse` has `refresh_token=""` would catch it (the latter also covers `scope_fallback`).

**5. The missing-token half of `access_valid`.** The tests at line 166-170 cover expiry and skew thoroughly, but never the `bool(self.access_token)` conjunct: `OAuthTokens(access_token="", access_expires_at=far_future).access_valid()` must be `False`. This is the same "absent ≠ near-expiry" distinction we discussed around the rename — worth pinning since it's what makes the compound predicate safe.

eisene (commented):
a few minor comments

eisene (dismissed):
lgtm

erosika (approved):
LGTM

PR #892, opened 2026-07-08
Prevent flooding the queue with embedding tasks

eisene:
# Background

Currently there's no limit to the number of message embedding tasks that can be spun up by a user through the API. This should change.

# Changes

- Add an `EmbedTaskGate` that tracks how many in-flight message embedding tasks an API process has spun up. Defer to reconciler if too many by just not scheduling.
- Prometheus metrics for the above.
- Configurable via `MAX_PENDING_EMBED_TASKS` set to 50 by default.
- Unit tests.

# Testing plan

Set `MAX_PENDING_EMBED_TASKS = 0` and try running a test convo through a self-served local Honcho docker fleet. Check the logs and DB: 
- `curl -s localhost:8000/metrics | grep embed_now_tasks_shed` increments once per shed schedule attempt
  - Should see 1.0 if you run a single session through local Honcho
- `select sync_state, count(*) from message_embeddings group by 1` in DBeaver
  - The most important is to see `synced` for all of them after the healing interval passed

## Summary by CodeRabbit

* **New Features**
  * Added limits for how many immediate embedding tasks can run at once, helping keep message creation responsive under load.
  * Improved tracking for skipped immediate embedding attempts when the system is busy.

* **Bug Fixes**
  * Message creation and file upload now fall back more reliably when immediate embedding is saturated.
  * Background embedding slots are now always released, even if a task fails.

### DIFF
--- a/src/reconciler/embed_now.py
+++ b/src/reconciler/embed_now.py
@@ -95,6 +95,8 @@ class EmbedTaskGate:
                 prometheus_metrics.record_embed_now_task_shed()
             return False
         self.in_flight += 1
+        if settings.METRICS.ENABLED:
+            prometheus_metrics.set_embed_now_tasks_in_flight(self.in_flight)
         background_tasks.add_task(self._run, message_ids)
         return True
 
@@ -103,6 +105,8 @@ class EmbedTaskGate:
             await embed_messages_now(message_ids)
         finally:
             self.in_flight -= 1
+            if settings.METRICS.ENABLED:
+                prometheus_metrics.set_embed_now_tasks_in_flight(self.in_flight)
 
 
 embed_task_gate = EmbedTaskGate()
--- a/src/reconciler/embed_now.py
+++ b/src/reconciler/embed_now.py
@@ -23,6 +23,7 @@ import logging
 from dataclasses import dataclass
 from typing import Any
 
+from fastapi import BackgroundTasks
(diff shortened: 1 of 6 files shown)


src/reconciler/embed_now.py
```diff
@@ -66,6 +68,46 @@ def reset_embed_semaphore() -> None:
...
+    on its next cycle. The count is taken at schedule time (not task start)
+    because background tasks only run after the response is sent — a request
+    burst would otherwise stack up unbounded scheduled-but-not-started tasks.
+
+    ``in_flight`` is mutated only from the event loop (request handlers and the
+    tracked task), so a plain int is race-free.
+    """
+
+    def __init__(self) -> None:
+        self.in_flight: int = 0
```
akattelu, line 86: should we export in_flight as a gauge alongside the shed counter to prometheus, so a blocked process is visible in Grafana?

akattelu (dismissed):

Rajat-Ahuja1997 (dismissed):

src/reconciler/embed_now.py
eisene: 💯

https://github.com/plastic-labs/honcho/pull/892/commits/dcde4b29075c4dfb316d08f913bc9c194b83e2f3

Rajat-Ahuja1997 (approved):

akattelu (approved):

PR #895, opened 2026-07-10
Fix `unified-tests.yml` secrets

eisene:
# Background

Currently unified-tests.yml pulls from the latest staging secrets in AWS SM. If the latest main dotenv drifts from latest staging this can cause the unified tests to break.

# Changes

## Add secrets overrides

- Load staging secret first, then load testing secrets as an override
- Make a new secret in AWS SM with the testing overrides so this can be reverted
  - Delete the old secrets in a week or so

Deleted secrets are not handled by this override and need to be deleted in the GHA yaml.

## Add a label to run unified tests GHA

- Add a trigger to also run on a `run-unified-tests` label add to the current PR
- The first thing the GHA workflow does is purge the run-unified-tests label

## Add a secret delete

- Delete `DERIVER_REPRESENTATION_BATCH_MAX_TOKENS`.

# Testing plan

Run unified tests using the new label and confirm everything works before merging.

## Summary by CodeRabbit

* **New Features**
  * Unified test suite can now be manually triggered via a designated pull request label, and the label is automatically removed afterward.

* **Bug Fixes**
  * Test configuration overrides (`HONCHO_TEST_OVERRIDE_*`) are now sourced from Secrets Manager and take precedence over previously loaded values.
  * Improved runner cleanup reliability.

* **Tests**
  * Updated configuration tests to remove coverage for legacy `REPRESENTATION_BATCH_MAX_TOKENS` rejection.

* **Chores**
  * Updated `uv` settings for dependency “newer” selection window.


.github/workflows/unified-tests.yml
```diff
@@ -122,7 +193,14 @@ jobs:
...
 
       - name: Run unified tests
-        run: uv run python -m tests.unified.run
+        run: |
+          # DERIVER_REPRESENTATION_BATCH_MAX_TOKENS was removed from config
+          # (split into ..._WORK_UNIT_TARGET_TOKENS and ..._TARGET_INPUT_TOKENS)
+          # and src/config.py fails fast if it is present in the environment,
+          # but the staging secret still carries it. Drop this once the staging
+          # secrets no longer contain the legacy key.
+          unset DERIVER_REPRESENTATION_BATCH_MAX_TOKENS
```
VVoruganti: This seems like AI comments specific to this particular set of Honcho changes. Please remove

.github/workflows/unified-tests.yml
```diff
@@ -97,11 +127,52 @@ jobs:
...
+          echo "Applied ${applied} override(s)"
+
       # Configure the test environment. Disables auth/Sentry/CloudEvents telemetry
       # (their endpoints aren't reachable from CI), and points REASONING_TRACES_FILE
       # at a shared path so the API + deriver record full LLM I/O for auditing — the
       # runner uploads it to S3. Written after the fetch steps so these win over the
-      # values loaded from Secrets Manager (last $GITHUB_ENV write wins).
+      # values loaded from Secrets Manager (last $GITHUB_ENV write wins). Note:
+      # $GITHUB_ENV can only set values (a blank is still present in os.environ);
+      # keys that must be truly *removed* are unset in the Run unified tests step.
```
VVoruganti: having a dummy extra configuration value shouldn't break anything and this requires more manual management of the CI file. Don't think this should be necessary moving forward

VVoruganti (changes requested):

.github/workflows/unified-tests.yml
eisene: https://github.com/plastic-labs/honcho/pull/895/commits/e0761c544e9e2924ee1cbcfa092b7e06b82c7ef7

.github/workflows/unified-tests.yml
eisene: https://github.com/plastic-labs/honcho/pull/895/commits/e0761c544e9e2924ee1cbcfa092b7e06b82c7ef7

Rajat-Ahuja1997 (dismissed):

Rajat-Ahuja1997 (approved):

VVoruganti (approved):

PR #896, opened 2026-07-10
feat(dialectic): optional structured outputs with limited schema for Dialectic calls

eisene:
# Background

Accept a JSON Schema in `DialecticOptions`, run the normal agentic tool loop, and format the final synthesis call’s output to conform to the schema.

# Changes

- Accept a schema and pipe it through to the agent calls
- Schema parser that converts the supplied JSON schema (appropriately restricted) to a Pydantic object
  - Enable the TS SDK to also use this, see Claude diagram:
```markdown
Zod schema ──z.toJSONSchema()──► JSON Schema dict ──HTTP body──► server
                                                                   │ json_response_schema_to_pydantic
                                                                   ▼
                                                       dynamic Pydantic model
                                                                   │ honcho_llm_call(response_model=...)
                                                                   ▼
        typed T ◄──schema.parse(JSON.parse(content))── JSON string in response.content
```
- Whole bunch of unit tests and some table tests for the schema parser

# Testing plan

Run some convos in the test app locally and call the dialectic with a variety of schemas. Do this with both the Python and the Typescript APIs, both the peer and the aio paths.

Example follows.

## Question

```markdown
[question] What are this user's main concerns, and how skeptical are they of the advice they've been given so far?
```

## Schema

```python
class SkepticismAssessment(BaseModel):
    """How skeptical the user is, with supporting evidence."""

    level: Literal["low", "moderate", "high"] = Field(
        description="How skeptical the user is of the advice received so far"
    )
    evidence: list[str] = Field(
        description="Quotes or paraphrases from the conversation showing skepticism"
    )


class DialecticReport(BaseModel):
    """Structured answer to the dialectic question about the simulated user."""

    main_concerns: list[str] = Field(
        description="The user's main concerns, one short phrase each"
    )
    skepticism: SkepticismAssessment = Field(
        description="Assessment of the user's skepticism"
    )
    salient_fact: str | int | bool | None = Field(
        default=None,
        description=(
            "One most-salient concrete fact about the user, if any stands out: "
            "a count (integer), a yes/no judgment (boolean), or a short phrase "
            "(string); null if nothing stands out"
        ),
    )
    summary: str = Field(description="One-paragraph overall answer to the question")
```

## Response

```json
{
  "main_concerns": [
    "crate training a new golden retriever puppy",
    "keeping the puppy calm during work meetings",
    "avoiding crate anxiety and making it a positive experience",
    "distinguishing anxiety from normal playfulness or whining"
  ],
  "skepticism": {
    "level": "low",
    "evidence": [
      "They asked for clarification and reassurance, but not in a distrustful way.",
      "Phrases like “I’m worried he might get anxious if he can't see me,” “I want to avoid making it a bad experience for him,” and “I want to make sure I’m responding appropriately without overreacting” show caution, not skepticism toward the advice itself."
    ]
  },
  "salient_fact": "The user is focused on a new golden retriever puppy and crate-training concerns.",
  "summary": "The user’s main concerns are crate training a newly adopted golden retriever puppy, keeping him calm during work-from-home meetings, preventing crate anxiety, and learning how to tell anxiety apart from normal puppy playfulness or whining. Their skepticism toward the advice so far appears low: they are asking follow-up questions and expressing worry about doing things correctly, but they are not dismissing or challenging the guidance. They seem more cautious and attentive than skeptical."
}
```

## Summary by CodeRabbit

* **New Features**
  * Added structured chat responses using Pydantic models, Zod schemas, or JSON Schema.
  * Python and TypeScript SDKs now support typed response parsing for chat requests.
  * Added structured response support for synchronous and streaming chat workflows.
  * Added schema validation with clear errors for unsupported or invalid formats.
* **Documentation**
  * Added structured-output guides, examples, supported schema details, and streaming guidance.
* **Bug Fixes**
  * Improved response-schema caching without retaining unused model classes.


sdks/python/src/honcho/peer.py
```diff
@@ -221,6 +234,30 @@ def __init__(
         self._configuration = configuration  # pyright: ignore[reportIncompatibleVariableOverride]
         self._created_at = created_at
 
+    @overload
```
eisene, line 237: Claude defending this in response to a challenge of "why not just make a kwarg with a union type":

---

The overloads exist because the return type depends on the argument type, and a single signature can't express that:

```python
- response_format=FoodPreferences (a Pydantic class) → returns a parsed FoodPreferences instance
- response_format={...} (a schema dict) or omitted → returns str | None
```

With just the plain signature, the declared return type is `BaseModel | str | None` for every caller. Then this breaks type-checking ergonomics in both directions:

```python
result = peer.chat("...", response_format=FoodPreferences)
result.favorite  # ❌ pyright error: "str | None" has no attribute "favorite"
                 # caller must isinstance-narrow or cast every time

answer = peer.chat("hello")  # ❌ typed as BaseModel | str | None —
                             # existing callers who treat it as str | None now get errors
```

src/utils/schema_conversion.py
```diff
@@ -0,0 +1,343 @@
+"""Convert caller-supplied JSON Schema objects into dynamic Pydantic models.
```
eisene, line 1: There are packages we could take as a dependency for this, like `json-schema-to-pydantic`, but I don't think that's really worth it.

sdks/python/src/honcho/aio.py
```diff
@@ -577,6 +577,30 @@ async def set_configuration(self, configuration: PeerConfig) -> None:  # pyright
         )
         self._peer._configuration = configuration
 
+    @overload
```
eisene, line 580: See comment on `peer.py`

.gitignore
```diff
@@ -195,4 +195,4 @@ lancedb_data/
 grafana-data/
 
 # Claude Code addon stuff
-.omc
\ No newline at end of file
+.omc
```
eisene, line 198: Blank line added for Ruff cleanup

src/dialectic/core.py
```diff
@@ -413,7 +414,9 @@ def _log_response_metrics(
             )
         )
 
-    async def answer(self, query: str) -> str:
+    async def answer(
```
akattelu, line 417: Can we double check the compatibility of this against the gemini provider? I know in the past Gemini's response_schema field rejected `anyOf` (from unions). I just want to make sure (maybe with live llm tests?) that this works on all 3 transports, even thought the default is openai

src/utils/schema_conversion.py
```diff
@@ -0,0 +1,404 @@
...
+    "$ref",
+    "$defs",
+    "definitions",
+    "allOf",
+    "not",
+    "if",
+    "then",
+    "else",
+    "patternProperties",
+)
```
akattelu, line 32: How did we decide on this limited set of schema keys? We should probably update the troubleshooting docs section or something to make a note about the limitations of what you can pass into the API. Maybe even a note in the sdk reference section?

src/dialectic/core.py
```diff
@@ -446,25 +451,38 @@ async def answer(self, query: str) -> str:
...
-            max_input_tokens=settings.DIALECTIC.MAX_INPUT_TOKENS,
-            trace_name="dialectic_chat",
-            telemetry=self._telemetry_context(track_name="Dialectic Agent"),
+        # cast: `type[BaseModel] | None` matches neither the parsed nor the
+        # plain-text overload statically, so pyright resolves the stream
+        # overload — but without stream=True the call is non-streaming.
+        response = cast(  # pyright: ignore[reportInvalidCast]
+            HonchoLLMCallResponse[Any],
+            await honcho_llm_call(
+                model_config=_get_dialectic_level_model_config(self.reasoning_level),
```
akattelu, line 460: should just double check how this interacts with json_schema vs json_object mode (PR for reference: https://github.com/plastic-labs/honcho/pull/820/). That flag is only supported by openai compatible endpoints, but it covers a lot of models and versions.

src/utils/schema_conversion.py
eisene: Re: how we decided - it was specced out this way, but on discussion we decided to re-introduce non-recursive reference schemas.

Will add docs

eisene: https://github.com/plastic-labs/honcho/pull/896/commits/cfc0ac4e45c0c3377e5463d3d8a1b98bce2be605 and https://github.com/plastic-labs/honcho/pull/896/commits/176b615cb4044546d347a3e31d2604526fe996d7

src/dialectic/core.py
eisene: This should work and was already mostly covered by existing tests since nothing really changed here, but added an explicit test for this: `test_structured_output_json_object_mode_with_dynamic_model`. Test explicitly asserts the call uses `json_object`.

https://github.com/plastic-labs/honcho/pull/896/commits/c445ef381751598018f25148404f3b2cba32bcdb

src/dialectic/core.py
eisene: Added live LLM tests for this, they all pass. Confirmed by hand includes `gemini-2.5-flash`

https://github.com/plastic-labs/honcho/pull/896/commits/c445ef381751598018f25148404f3b2cba32bcdb

akattelu (approved):

PR #900, opened 2026-07-12
fix(deriver): coerce bare-string explicit observations from json_object providers

### LINKED ISSUE #893: Deriver prompt instructs LLM to return wrong JSON shape for `explicit`, causing zero observations with OpenAI-compatible providers (DeepSeek, vLLM)
## Description

The `minimal_deriver_prompt()` function in `src/deriver/prompts.py` instructs the LLM to return explicit observations as a list of bare strings, but the Pydantic model (`PromptRepresentation.explicit`) expects a list of objects with a `content` field. When using an OpenAI-compatible provider that does **not** support `json_schema` structured outputs (e.g., DeepSeek via LiteLLM, or certain vLLM configurations), the code falls back to `json_object` mode. In that mode, the LLM has only the prompt text to infer the JSON shape — and the prompt suggests the wrong one. The result is zero observations, emitted silently with no error.

### Environment / Context

- **Project:** Self-hosted Honcho
- **LLM backend:** DeepSeek accessed via a LiteLLM proxy (OpenAI-compatible API)
- **Observation extraction path:** `minimal_deriver_prompt()` → LLM call → `PromptRepresentation` validation
- **Affected file(s):** `src/deriver/prompts.py` (primary), `src/utils/representation.py`, `src/llm/backends/openai.py`

### The Bug

The prompt's examples show bare strings as the expected output:

```
- EXPLICIT: "I just had my 25th birthday last Saturday" → "alice is 25 years old", "alice's birthday is June 21st"
```

This implicitly tells the LLM to return:
```json
{"explicit": ["fact 1", "fact 2", ...]}
```

But the actual Pydantic model expects:

```python
class ExplicitObservationBase(BaseModel):
    content: str = Field(description="The explicit observation")

class PromptRepresentation(BaseModel):
    explicit: list[ExplicitObservationBase] = Field(...)
```

The correct JSON shape is:
```json
{"explicit": [{"content": "fact 1"}, {"content": "fact 2"}, ...]}
```

### Why It's Silent

1. `complete()` first attempts `chat.completions.parse()` with `json_schema` — rejected with `400 BadRequest` by DeepSeek.
2. The `BadRequestError` is caught in `_create_structured_response()` and the code falls back to `json_object` mode.
3. In `json_object` mode the LLM has **only the prompt text** to infer the shape — and the prompt says bare strings.
4. The returned JSON is validated against `PromptRepresentation`. Each bare string fails to validate as `ExplicitObservationBase` (which requires `{"content": "..."}`).
5. `repair_response_model_json()` catches the `ValidationError` and returns `PromptRepresentation(explicit=[])` — silently discarding every observation.

No error is raised, no observation is persisted.

### When It Manifests

- ✅ **Reproducible on:** OpenAI-compatible providers that reject `json_schema` entirely — DeepSeek (via LiteLLM), some vLLM configs, Ollama, and other non-OpenAI backends.
- ❌ **Not reproducible on:** Providers with full Structured Outputs support (native OpenAI, Anthropic, etc.) — the schema itself constrains the output shape, so the prompt mismatch is irrelevant.

### Suggested Fix

Add a "Respond with" section to `minimal_deriver_prompt()` showing the correct nested object shape:

```
Respond with a JSON object in the following format:
{"explicit": [{"content": "fact 1"}, {"content": "fact 2"}, ...]}
Each item in the "explicit" list must be an object with a "content" field — not a bare string.
```

This ensures the LLM returns correctly-shaped objects even in `json_object` fallback mode.

### Notes

- Bug is present on current `main` as of 2026-07-09.
- The mismatch is harmless when providers support `json_schema` (schema constrains the output), which is likely why it hasn't been caught — it's only visible on non-OpenAI-compatible structured output paths.

### Suggested Labels

`bug`, `deriver`

faithpricejp-source:
## Linked Issue

Closes #893

## Description

`minimal_deriver_prompt()`'s examples and the `PromptRepresentation.explicit` field description both present explicit facts as **bare strings**, but the field type is `list[ExplicitObservationBase]` (objects with a `content` key). Providers without `json_schema` structured-output support (DeepSeek via LiteLLM, some vLLM configs) run in `json_object` mode and infer the JSON shape from that prompt/schema text, so they emit:

```json
{"explicit": ["alice is 25 years old", "alice has a dog"]}
```

That fails `PromptRepresentation` validation and, via the structured-output fallback in `src/llm/structured_output.py`, is silently turned into `PromptRepresentation(explicit=[])` — **zero observations, no error surfaced**.

### Fix

- Extend the existing `explicit` `mode="before"` validator to coerce bare strings into `{"content": ...}`. Dicts/objects pass through unchanged; non-string scalars (e.g. a stray `int`) still raise, so genuine shape errors aren't masked. This mirrors the precedent already in `structured_output.py`, which repairs `deductive` shape drift.
- Fix the misleading `explicit` field-description example so the schema injected in `json_object` mode no longer advertises the wrong (bare-string) shape.

Load-bearing decision: coercion lives on the model boundary (not only in the repair path) so it applies whether the payload arrives via `json_object` repair or direct validation.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Breaking Changes

N/A — objects and `None` continue to validate exactly as before; only the previously-failing bare-string shape now succeeds.

## Test Coverage

- [x] I added/updated unit tests

Added `tests/utils/test_representation.py` (pure model-validation, no DB — registered in the conftest runtime-mock blocklist alongside the other `tests/utils` unit tests). Covers: bare strings coerced, objects still accepted, mixed input normalized, `None` → `[]`, and non-string scalar still rejected.

Verified locally: with the fix all 5 pass; reverting only the validator makes the two coercion tests fail with the exact `ValidationError` that becomes a silent `explicit=[]` in production. `ruff check` and `ruff format --check` clean.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of explicit observations provided as plain text values.
  * Continued support for object-form observations and mixed input formats.
  * Empty or missing explicit observations now normalize consistently.
  * Invalid non-text values remain rejected with validation errors.

* **Documentation**
  * Updated field examples to reflect the supported object format.

### DIFF
--- a/src/utils/representation.py
+++ b/src/utils/representation.py
@@ -114,16 +114,28 @@ class PromptRepresentation(BaseModel):
     """
 
     explicit: list[ExplicitObservationBase] = Field(
-        description="Facts LITERALLY stated by the user - direct quotes or clear paraphrases only, no interpretation or inference. Example: ['The user is 25 years old', 'The user has a dog named Rover']",
+        description="Facts LITERALLY stated by the user - direct quotes or clear paraphrases only, no interpretation or inference. Example: [{'content': 'The user is 25 years old'}, {'content': 'The user has a dog named Rover'}]",
         default_factory=list,
     )
 
     @field_validator("explicit", mode="before")
     @classmethod
-    def convert_none_to_empty_list(cls, v: Any) -> Any:
-        """Convert None to empty list - handles LLMs returning null instead of []."""
+    def normalize_explicit(cls, v: Any) -> Any:
+        """Normalize the ``explicit`` payload before validation.
+
+        Handles two shapes LLMs emit in ``json_object`` mode, where the model
+        sees only the prompt/schema text and often returns the simpler form:
+
+        - ``None`` instead of ``[]`` (some models return null for empty lists).
+        - bare strings instead of ``{"content": ...}`` objects. Providers
+          without ``json_schema`` support (e.g. DeepSeek, some vLLM configs)
+          frequently emit ``{"explicit": ["fact 1", ...]}``; coerce each string
+          into the expected object so observations aren't silently dropped (#893).
+        """
         if v is None:
             return []
+        if isinstance(v, list):
+            return [{"content": item} if isinstance(item, str) else item for item in v]
         return v
 
 
--- a/src/utils/representation.py
+++ b/src/utils/representation.py
@@ -143,16 +143,28 @@ class PromptRepresentation(BaseModel):
     """
 
     explicit: list[ExplicitObservationBase] = Field(
-        description="Facts LITERALLY stated by the user - direct quotes or clear paraphrases only, no interpretation or inference. Example: ['The user is 25 years old', 'The user has a dog named Rover']",
+        description="Facts LITERALLY stated by the user - direct quotes or clear paraphrases only, no interpretation or inference. Example: [{'content': 'The user is 25 years old'}, {'content': 'The user has a dog named Rover'}]",
         default_factory=list,
     )
 
     @field_validator("explicit", mode="before")
     @classmethod
-    def convert_none_to_empty_list(cls, v: Any) -> Any:
-        """Convert None to empty list - handles LLMs returning null instead of []."""
+    def normalize_explicit(cls, v: Any) -> Any:
+        """Normalize the ``explicit`` payload before validation.
+
+        Handles two shapes LLMs emit in ``json_object`` mode, where the model
+        sees only the prompt/schema text and often returns the simpler form:
+
+        - ``None`` instead of ``[]`` (some models return null for empty lists).
+        - bare strings instead of ``{"content": ...}`` objects. Providers
+          without ``json_schema`` support (e.g. DeepSeek, some vLLM configs)
+          frequently emit ``{"explicit": ["fact 1", ...]}``; coerce each string
+          into the expected object so observations aren't silently dropped (#893).
+        """
         if v is None:
             return []
+        if isinstance(v, list):
+            return [{"content": item} if isinstance(item, str) else item for item in v]
         return v
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -85,6 +85,8 @@ _RUNTIME_MOCK_TEST_BLOCKLIST_PREFIXES = (
     # LLM transport tests mock providers directly and don't need database/runtime setup.
     "tests/utils/test_length_finish_reason.py",
     "tests/utils/test_clients.py",
+    # Pure PromptRepresentation model-validation tests — no DB needed.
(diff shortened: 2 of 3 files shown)


faithpricejp-source:
Rebased onto current `main` (`668ae39`) — the conflict was only the no-DB allowlist in `tests/conftest.py`, where `main` had added `test_session_scope_clauses.py`; both entries are kept. `gate` and CodeRabbit are green and the PR is now MERGEABLE.

Since this has been open a while, the situation has changed and I want to re-state the case honestly rather than just bumping it.

**What's no longer true.** #893 was closed as completed on 2026-07-17: #907 added the `json_object` schema-injection instruction, and @chadwjames confirmed the `structured_output_mode` override resolved his repro. I'm not claiming the original DeepSeek reproduction still fails — I have not re-run it. The issue gate correctly flags that this PR's linked issue is closed.

**What's still true, and is the reason I think this is worth keeping.** The `explicit` field description on `main` still advertises the wrong shape, and #907 causes it to be sent to the model on *every* `json_object` call. `_json_object_instruction()` serializes `response_format.model_json_schema()` into the prompt, and that schema is self-contradictory today:

```json
"explicit": {
  "description": "Facts LITERALLY stated by the user ... Example: ['The user is 25 years old', 'The user has a dog named Rover']",
  "items": { "$ref": "#/$defs/ExplicitObservationBase" },
  "type": "array"
}
```
```json
"ExplicitObservationBase": {
  "properties": { "content": { "type": "string" } },
  "required": ["content"],
  "type": "object"
}
```

The `$ref` says *array of objects with a required `content`*; the description immediately above it demonstrates *an array of bare strings*. So the mitigation added in #907 is the very thing now carrying the misleading example into the prompt — a weaker provider that weights the natural-language example over the `$ref` gets pulled toward exactly the shape that produces zero observations. That's the half of this PR I'd argue for most: it's a one-line description fix, and it makes the injected schema self-consistent.

The before-validator is secondary — defense in depth for a shape the prompt no longer asks for, at the cost of five lines and a test file that needs no database. Reasonable to decline that half; I'd rather land the description fix than nothing.

**How would you like to proceed?** The gate needs an open `maintainer-approved` issue. I'm happy to (a) open a narrow issue for the schema-description contradiction and relink this PR, (b) strip this down to just the description fix, or (c) close it if you consider the area settled by #907. Just say which and I'll do the work — I don't want to add gate noise by guessing.

eisene:
parent issue marked resolved, closing this PR. please open additional issue for further work.

PR #901, opened 2026-07-12
fix(deriver): document explicit observation JSON shape

PraveenKumarSridhar:
## Summary
- Clarify the deriver prompt's expected `PromptRepresentation.explicit` JSON shape for `json_object` structured-output fallback.
- Update the `PromptRepresentation` schema description so OpenAI-compatible providers see object-shaped examples, not bare string examples.
- Add regression coverage for the prompt, the schema hint, and the OpenAI `json_object` schema-injection path.

Closes #893.

## Why

`PromptRepresentation.explicit` expects a list of objects with a `content` field. In `json_object` mode, providers rely on prompt/schema text to infer the shape, so examples that imply bare strings can lead to validation failures and silently empty observations.

## Testing
- `.venv/bin/pytest tests/deriver/test_prompts.py tests/llm/test_backends/test_openai.py -q` → 55 passed
- `.venv/bin/ruff check src/deriver/prompts.py src/utils/representation.py tests/deriver/test_prompts.py tests/llm/test_backends/test_openai.py` → All checks passed
- `.venv/bin/basedpyright src/deriver/prompts.py src/utils/representation.py tests/deriver/test_prompts.py tests/llm/test_backends/test_openai.py` → 0 errors

Broader deriver queue tests were attempted, but local Postgres auth setup rejected the test DB connection, so I kept verification to the prompt and OpenAI backend tests touched by this fix.

Developed with AI assistance; reproduced, tested, and reviewed locally before submission.

## Summary by CodeRabbit

* **Improvements**
  * Clarified structured AI output requirements so explicit observations are returned in the expected JSON object/array shape with a `content` field.
  * Updated schema/prompt examples to match the enforced JSON format.

* **Bug Fixes**
  * Improved validation to reject incorrectly formatted explicit observations (e.g., plain string forms).

* **Tests**
  * Added prompt and JSON-shape tests, including schema example checks.
  * Added OpenAI backend coverage for structured `json_object` responses using the prompt representation.

PR #903, opened 2026-07-12
fix(llm): set HTTP timeout on Gemini clients

### LINKED ISSUE #785: [Bug] Deriver hangs indefinitely on a stalled Gemini call: Gemini client built without an HTTP timeout
## Describe the bug

The Gemini provider client is constructed without an HTTP timeout, so a stalled Gemini socket hangs the deriver worker. Because all deriver workers share one uvloop event loop, the **entire deriver process stalls** — representation, dreams, and the reconciler all stop — with no error and no log output.

In `src/llm/registry.py` the Anthropic client is built with `timeout=600.0`, but the Gemini client passes `http_options=None` when no `GEMINI_BASE_URL` is set (in all three constructors: `get_gemini_client`, `get_gemini_override_client`, and the module-level `CLIENTS["gemini"]`):

```python
http_options = (
    genai_types.HttpOptions(base_url=settings.LLM.GEMINI_BASE_URL)
    if settings.LLM.GEMINI_BASE_URL
    else None
)
return genai.Client(api_key=settings.LLM.GEMINI_API_KEY, http_options=http_options)
```

With `http_options=None`, `HttpOptions.timeout` defaults to `None`, which the google-genai SDK forwards as `timeout=None` — disabling the timeout entirely on both the httpx and aiohttp transports. A stalled Gemini socket then never returns. Related: googleapis/python-genai#1893, #158, #911.

## Impact

Observed on a self-hosted v3.0.7 deriver (representation model `gemini-2.5-flash-lite`): a single stalled Gemini call froze the deriver for **~9.5 hours**.

- Last log line was `google_genai ... AFC is enabled`, then total silence.
- A `representation` work unit stayed claimed in `active_queue_sessions` and was never released.
- The queue backed up to ~24k unprocessed `representation` tasks; 2 dream tasks never started.
- Postgres was clean (0 idle-in-transaction, 0 active queries) — the hang was purely in the Python/uvloop layer.
- The container stayed `Up` (no crash, no traceback); only a deriver restart recovered it.

The v3.0.8/v3.0.9 connection-resilience work targets the DB pooler, not outbound LLM calls, so it does not cover this case.

## Expected behavior

Outbound Gemini calls should be bounded by a timeout (matching the Anthropic client's `timeout=600.0`), so a stalled socket fails fast and the deriver's existing fallback/retry path can handle it instead of wedging the whole worker.

## Suggested fix

Always set a timeout on the Gemini `HttpOptions` in all three constructors in `src/llm/registry.py`, ideally configurable (e.g. an `LLM_GEMINI_TIMEOUT_SECONDS` setting) and consistent across providers:

```python
http_options = genai_types.HttpOptions(
    base_url=settings.LLM.GEMINI_BASE_URL or None,
    timeout=600_000,  # ms; google-genai 2.0.1 applies this as an aiohttp `total` timeout
)
```

## Environment

- Honcho v3.0.7 (observed) / v3.0.9 (patched), self-hosted via Podman
- google-genai 2.0.1, aiohttp 3.13.5, httpx 0.28.1

PraveenKumarSridhar:
## Summary
- Set a default 10-minute HTTP timeout on every Gemini client built in `src/llm/registry.py` (default factory, override factory, and module-level `CLIENTS["gemini"]`).
- Add `tests/llm/test_registry.py` with regression coverage on all three Gemini construction sites plus Anthropic-baseline locks.

Closes #785.

## Why

A stalled Gemini socket can hang the deriver worker indefinitely because the Gemini client is built without an HTTP timeout. All deriver workers share one uvloop event loop, so a single hung call wedges the entire process — representation, dreamer, and reconciler all stop. The Anthropic client already passes `timeout=600.0`; this aligns the Gemini client with that behavior.

`google.genai.HttpOptions.timeout` is an `int` in milliseconds, so the default is `600_000` ms (10 min) — same wall-clock budget as Anthropic.

## Testing
- `.venv/bin/pytest tests/llm/ -q` → 231 passed
- `.venv/bin/pytest tests/llm/test_registry.py -q` → 7/7 passed (5 red→green for the new Gemini-timeout behavior, 2 baseline Anthropic locks)
- `.venv/bin/ruff check` and `.venv/bin/basedpyright` on the changed files: no new errors

Broader Postgres-backed deriver queue tests were not run locally (env-level Postgres auth blocker; same gap as PR #901).

Developed with AI assistance; reproduced (unit test) and reviewed locally before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Gemini and Anthropic provider connections by consistently applying configured request timeouts.
  * Preserved custom service endpoints while ensuring timeout settings remain active.
  * Gemini embedding requests now retain timeout protection even when no custom endpoint is configured.

* **Tests**
  * Added coverage validating timeout behavior across standard and override provider configurations.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -182,10 +182,13 @@ class _EmbeddingClient:
         if self.transport == "gemini":
             if not config.api_key:
                 raise ValueError("Gemini API key is required")
-            http_options = (
-                genai_types.HttpOptions(base_url=config.base_url)
-                if config.base_url
-                else None
+            # 10-minute HTTP timeout, in lockstep with the LLM registry's Gemini
+            # client (`src/llm/registry.py:_build_gemini_http_options`). Without
+            # this, a stalled Gemini embedding socket wedges the deriver worker
+            # exactly the way #785 describes for the LLM client.
+            http_options = genai_types.HttpOptions(
+                base_url=config.base_url,
+                timeout=600_000,
             )
             self.client: genai.Client | AsyncOpenAI = genai.Client(
                 api_key=config.api_key,
--- a/src/llm/registry.py
+++ b/src/llm/registry.py
@@ -32,6 +32,14 @@ from .history_adapters import (
 )
 from .types import ProviderClient
 
+# Default client-level HTTP timeouts. Anthropic accepts seconds (float);
+# google-genai's HttpOptions.timeout is an int in milliseconds, so the Gemini
+# value is kept separately. Both default to 10 minutes to match the existing
+# Anthropic behavior — long enough for slow streamed responses, short enough
+# that a stalled socket can no longer wedge the deriver worker (see #785).
+_ANTHROPIC_TIMEOUT_S = 600.0
+_GEMINI_TIMEOUT_MS = 600_000
+
 # Client-level ``default_headers`` applied to OpenAI-compatible clients, keyed by
 # base-URL prefix. Currently only OpenRouter, which uses them for app attribution
 # (https://openrouter.ai/docs/app-attribution); add a prefix here to tag another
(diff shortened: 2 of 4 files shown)


PraveenKumarSridhar:
Expanded scope after independent code review: a Gemini **embedding** client in `src/embedding_client.py` had the same wedge-class bug (in-process reconciler shares the deriver's uvloop event loop). Now covered with the same 10-min timeout. `src/` has no other `genai.Client` construction sites — verified. Local: 232/232 `tests/llm/` pass, ruff + basedpyright clean.

akattelu (approved):

PR #906, opened 2026-07-13
fix(dialectic): pass custom_instructions to the dialectic prompt

### LINKED ISSUE #874: Dialectic chat endpoint fetches but discards workspace custom_instructions
## Summary

`workspace.configuration.reasoning.custom_instructions` is fetched inside the dialectic chat path but then discarded before the prompt is built — `DialecticAgent` has no parameter for it, so `POST /peers/{id}/chat` (the endpoint a human actually queries) answers without any awareness of the workspace's declared purpose, even though the deriver already respects the same field correctly.

## Where

* `src/dialectic/chat.py` (`agentic_chat` / `agentic_chat_stream`, \~line 57): calls `get_configuration(...)` and reads `configuration.peer_card.use`, but never forwards `configuration.reasoning.custom_instructions` anywhere.
* `src/dialectic/core.py` (`DialecticAgent.__init__`, \~lines 61–96): no `custom_instructions` parameter exists.
* `src/dialectic/prompts.py` (`agent_system_prompt`): only accepts `observer, observed, observer_peer_card, observed_peer_card` — no instructions slot.

By contrast, `src/deriver/prompts.py` already threads the same field in correctly via `_custom_instructions_section()`, rendered as:

```
CUSTOM INSTRUCTIONS:
These instructions apply to the target peer identified below.
{custom_instructions}
```

So the field's plumbing (API → workspace config → fetch) is already proven end-to-end for the deriver; dialectic just drops it on the floor at the last step.

## Impact

We run a self-hosted instance where each workspace has a declared purpose (e.g. "this is a growth/self-awareness mirror, not a performance scorecard — favor durable patterns, be generous about strengths"). The deriver respects this when writing conclusions. But when a person actually *asks* the system a question via dialectic, the answer is synthesized with zero awareness of that purpose — it's reasoning fresh over the accumulated conclusions at query time, using its own prompt, with no signal about tone, framing, or what NOT to do (e.g. "never dwell on sensitive details", "this is not a scorecard").

Only the deriver reads `custom_instructions` correctly today. The summarizer has the same gap, already tracked in plastic-labs/honcho#748 / open PR [#750](<https://github.com/plastic-labs/honcho/issues/750>). The dreamer specialists (`src/dreamer/specialists.py`) have it too but seem like a separate, larger design question (they also lack any observer/observed framing at all). Dialectic seems like the highest-value of the remaining gaps to close, since it's what a human directly reads, and the fetch-then-discard pattern here suggests it was simply missed rather than deliberately scoped out — unlike the dreamer, this looks like a small, mechanical fix.

## Suggested fix

Thread `custom_instructions` through the same path already used for `peer_card`:

1. `chat.py`: pass `configuration.reasoning.custom_instructions` into `DialecticAgent(...)`.
2. `core.py`: add a `custom_instructions: str | None = None` param, store it.
3. `prompts.py`: extend `agent_system_prompt` to render a section analogous to the deriver's `_custom_instructions_section()` when present.

Happy to open a PR for this if a maintainer confirms the approach — it looks like a small, mechanical change following an existing pattern rather than new design work.

JoshKappler:
Fixes #874. The dialectic chat path calls `get_configuration()` but only reads `peer_card.use` from it, so workspace and session `custom_instructions` get dropped before the prompt is built. The deriver already honors that field, so instructions shaped how conclusions were written but had no effect on what the chat endpoint actually answered.

This threads `custom_instructions` through `DialecticAgent` into `agent_system_prompt` and renders a CUSTOM INSTRUCTIONS section following the deriver's pattern, on both the streaming and non-streaming paths. I built the section as a plain f-string rather than with `cleandoc`, since multi-line instructions drop the common margin to zero and would leave the surrounding block indented.

Tested with `uv run pytest tests/dialectic` (11 passed) and `tests/routes/test_peers.py` (45 passed), with ruff and basedpyright clean. The two new chat-level tests fail if the kwarg is removed from `chat.py`, and one of them pins session config overriding workspace config.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom instructions in dialectic conversations.
  * Custom instructions can now be included in agent prompts and support multiline formatting.
  * Session-level instructions take precedence over workspace-level instructions.

* **Bug Fixes**
  * Ensured configured custom instructions are consistently forwarded to conversations.

* **Tests**
  * Added coverage for instruction rendering, omission, formatting, and configuration precedence.

### DIFF
--- a/src/dialectic/chat.py
+++ b/src/dialectic/chat.py
@@ -77,6 +77,7 @@ async def agentic_chat(
         observer_peer_card=observer_peer_card,
         observed_peer_card=observed_peer_card,
         reasoning_level=reasoning_level,
+        custom_instructions=configuration.reasoning.custom_instructions,
     )
 
     return await agent.answer(query)
@@ -142,6 +143,7 @@ async def agentic_chat_stream(
         observer_peer_card=observer_peer_card,
         observed_peer_card=observed_peer_card,
         reasoning_level=reasoning_level,
+        custom_instructions=configuration.reasoning.custom_instructions,
     )
 
     async for chunk in agent.answer_stream(query):
--- a/src/dialectic/core.py
+++ b/src/dialectic/core.py
@@ -69,6 +69,7 @@ class DialecticAgent:
         metric_key: str | None = None,
         reasoning_level: ReasoningLevel = "low",
         session_id: str | None = None,
+        custom_instructions: str | None = None,
     ):
         """
         Initialize the dialectic agent.
@@ -83,6 +84,7 @@ class DialecticAgent:
             metric_key: Optional key for logging metrics (if provided, agent won't log separately)
             reasoning_level: Level of reasoning to apply
             session_id: ID used for grouping traces (not session_name)
+            custom_instructions: Optional workspace/session instructions for the answer
         """
         self.workspace_name: str = workspace_name
         self.session_name: str | None = session_name
@@ -93,13 +95,18 @@ class DialecticAgent:
         self.observed_peer_card: list[str] | None = observed_peer_card
         self.metric_key: str | None = metric_key
         self.reasoning_level: ReasoningLevel = reasoning_level
+        self.custom_instructions: str | None = custom_instructions
 
         # Initialize conversation history with system prompt
         self.messages: list[dict[str, str]] = [
             {
                 "role": "system",
                 "content": prompts.agent_system_prompt(
-                    observer, observed, observer_peer_card, observed_peer_card
+                    observer,
+                    observed,
+                    observer_peer_card,
+                    observed_peer_card,
+                    custom_instructions,
                 ),
             }
         ]
--- a/src/dialectic/prompts.py
+++ b/src/dialectic/prompts.py
@@ -3,11 +3,27 @@ System prompts for the Dialectic Agent.
 """
 
 
+def _custom_instructions_section(custom_instructions: str | None) -> str:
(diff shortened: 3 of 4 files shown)


karaokedjodua:
Finn-loop review of 60d23f6

CI: not configured
Mergeability: clean (mergeStateStatus BLOCKED — stale branch-protection rule; see note)

## Review

This PR threads `custom_instructions` from workspace/session configuration into the dialectic prompt, matching the pattern already used by the deriver. Previously, `get_configuration()` was called but only `peer_card.use` was consumed — custom instructions were silently dropped before the prompt was built. The fix touches both streaming and non-streaming paths.

## 1. Must fix before merge

- **[CI]** No required checks are configured on the `fix/dialectic-custom-instructions` branch. Additionally, `mergeStateStatus` is `BLOCKED` while `mergeable` is `MERGEABLE`, indicating a stale branch-protection rule references a check that no longer exists or doesn't match. A repo admin must either add required CI or remove/update the stale rule before this can merge.

## 2. Should fix soon

None.

## 3. Safe to merge

Yes on code merits — the change is focused (4 files, ~30 lines of logic + 134 lines of thorough tests). Both chat paths are updated consistently with the deriver's pattern. Tests cover rendering, omission (None/empty/whitespace), multiline formatting, agent propagation, workspace forwarding, and session-vs-workspace precedence. No security concerns, no scope creep.

**Labels to apply:** `needs-human-review`, `loop-approved`

PR #907, opened 2026-07-13
feat(llm backend): enable combined tool calling + structured output in the LLM backend transport layer

eisene:
# Background

The LLM backends couldn't handle a request carrying both function tools and a structured-output `response_format`. All three providers need different workarounds (DEV-2035), and all are prerequisites for enabling structured outputs in the dialectic (#896):

- **OpenAI**: `chat.completions.parse()` refuses non-strict function tools with a 500. Our agent tool schemas are deliberately non-strict.
- **Anthropic**: the `{` assistant prefill used to force JSON-first responses suppresses `tool_use` blocks entirely.
- **Gemini**: native `response_schema` + function calling is a Gemini 3 preview feature; earlier models reject the pairing.

# Changes

- **OpenAI**: tool-carrying structured requests route through `create()` with an explicit `json_schema` response_format (same server-side schema enforcement, no strict-tools requirement), mirroring the streaming path.
- **Anthropic**: skip the `{` prefill when tools are present; append a conditional schema instruction ("If not responding with a tool call, …") and rely on parse + repair.
- **Gemini**: with tools present, keep `response_schema` out of the request and inject a schema instruction into the final turn instead; parse + repair downstream.
- **All backends**: tool-call turns carry no consumable content, so structured-output parsing is skipped on them (parsing their empty text would raise through the repair fallback).

# Testing

- Mocked backend unit tests for each special case (`tests/llm/test_backends/`): prefill suppression, conditional instruction wording, `parse()` avoidance, `response_schema` omission, skip-parse-on-tool-call-turns, malformed-output repair.
- New live tests (`tests/live_llm/test_live_tools_structured_output.py`, gated by `--live-llm`): a two-turn flow per provider — forced tool call, then a replay turn that must return a schema-conforming answer with tools still attached. Verified locally against OpenAI (gpt-4.1, gpt-5, gpt-5.4, gpt-5.4-mini) and Gemini (gemini-2.5-flash).

## Summary
* **New Features**
  * Added unified schema-guidance behavior to support structured output alongside tool calling across OpenAI, Anthropic, and Gemini.

* **Bug Fixes**
  * Prevented tool-call turns from being treated as JSON for structured parsing/repair.
  * Disabled the assistant `{` prefill when tools are available (including streaming).
  * Improved provider-specific structured-output handling when tools are present.

* **Tests**
  * Expanded unit tests and added live integration coverage for tool + structured output.
  * Updated live OpenAI structured-output test assertions.

* **Chores**
  * Added a Live LLM automated test workflow.

### DIFF
--- a/.github/workflows/live-llm-tests.yml
+++ b/.github/workflows/live-llm-tests.yml
@@ -123,12 +123,22 @@ jobs:
       # Runs only if the latest fetch failed; this one is NOT continue-on-error,
       # so if the fallback also fails the job fails loudly.
       - name: Fetch staging secret (fallback to second-latest)
+        id: fetch-fallback
         if: steps.fetch-latest.outcome == 'failure' && steps.resolve-secret.outputs.second-id != ''
         uses: aws-actions/aws-secretsmanager-get-secrets@v2
         with:
           secret-ids: |
             ,${{ steps.resolve-secret.outputs.second-id }}
           parse-json-secrets: true
+
+      # If the latest fetch failed and the fallback was skipped (no second
+      # tag), no provider keys were loaded — without this guard every test
+      # would skip via require_provider_key and the run would go green.
+      - name: Verify staging secrets were loaded
+        if: steps.fetch-latest.outcome != 'success' && steps.fetch-fallback.outcome != 'success'
+        run: |
+          echo "::error::No staging secret could be fetched (latest failed; fallback skipped or failed)"
+          exit 1
 
       # Configure the test environment. Sentry/CloudEvents endpoints aren't
       # reachable from CI. LIVE_LLM_ANTHROPIC_45_PLUS_MODELS must be set for
--- a/.github/workflows/live-llm-tests.yml
+++ b/.github/workflows/live-llm-tests.yml
@@ -139,6 +139,10 @@ jobs:
       - name: Configure test environment
         run: |
           {
+            # The staging dotenv carries AUTH_USE_AUTH=true without a usable
+            # JWT secret; src/config.py validates the pair at import time, so
+            # disable auth (this suite never runs the API server anyway).
+            echo "AUTH_USE_AUTH=false"
             echo "SENTRY_ENABLED=false"
             echo "TELEMETRY_ENABLED=false"
             echo "LIVE_LLM_ANTHROPIC_45_PLUS_MODELS=claude-sonnet-4-5"
--- a/.github/workflows/live-llm-tests.yml
+++ b/.github/workflows/live-llm-tests.yml
@@ -1,0 +1,151 @@
+name: Live LLM Tests
+
+on:
+  # Manual trigger for PRs: add the `run-live-llm` label to run the suite
+  # against the PR's merge commit. The label is purged as soon as the run
+  # starts so it can be re-added to trigger another run. Opt-in only — live
+  # provider behavior is inherently variable, so this is a signal, not a
+  # required check.
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  # Purge the trigger label first thing. Best-effort: failing to remove the
+  # label (e.g. read-only token on a fork PR) doesn't block the tests.
+  remove-label:
+    name: Remove trigger label
+    if: github.event_name == 'pull_request' && github.event.label.name == 'run-live-llm'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Remove run-live-llm label
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if ! gh api --method DELETE \
+            "repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels/run-live-llm"; then
+            echo "::warning::Could not remove the run-live-llm label (it may have been removed already)"
+          fi
+
+  live-llm-tests:
+    name: Run Live LLM Tests
+    needs: remove-label
+    # always() lets this run on workflow_dispatch, where remove-label is
+    # skipped. Label adds other than run-live-llm trigger the workflow but
+    # skip every job here.
+    if: >-
+      always() &&
+      (github.event_name == 'workflow_dispatch' || github.event.label.name == 'run-live-llm')
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    environment: unified-tests
+    permissions:
+      id-token: write  # Required for OIDC authentication with AWS
+      contents: read
+    env:
+      PYTHONUNBUFFERED: "1"
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ vars.AWS_OIDC_ROLE_ARN }}
+          aws-region: us-east-1
+          role-duration-seconds: 3600
+
+      - name: Resolve secret ids from latest git tags
+        id: resolve-secret
+        env:
+          SECRET_PREFIX: ${{ secrets.STAGING_SECRET_PREFIX }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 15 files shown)


src/llm/backends/openai.py
```diff
@@ -172,6 +172,24 @@ async def complete(
...
+                )
+                response = await self._client.chat.completions.create(**params)
+                # Tool-call turns carry no consumable content — the tool loop
+                # ignores it — and parsing their empty text would raise.
+                if getattr(response.choices[0].message, "tool_calls", None):
+                    return self._normalize_response(response)
+                content = self._parse_or_repair_structured_content(
+                    response, response_format, model, empty_on_missing=False
+                )
+                return self._normalize_response(response, content_override=content)
```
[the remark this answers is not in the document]
eisene: Claude didn't agree with this. This is Claude's reply to this and I agree with Claude:

```markdown
Why that's the right behavior for the tools branch:

- A 400 with tools present usually means a real, fixable request-shape bug, not a provider capability gap. We hit a live example of this today: gpt-5.4 rejects function tools combined with an explicit reasoning_effort on chat completions. If the tools branch swallowed 400s the way the parse path does, that misconfiguration would silently turn every tool-loop iteration into an empty response — no retry, no fallback, no visible failure. Propagating lets the fallback chain switch models and makes the error observable.
- Returning empty content mid-tool-loop is semantically worse than in a one-shot structured call. In the parse path, an empty structured object is a degraded-but-usable terminal result for existing flows (the #797 rationale). In the tool loop, a non-tool-call response with content is the loop's termination signal — synthesizing an empty one would make the agent "conclude" with nothing rather than fail over.
```

tests/live_llm/test_live_tools_structured_output.py
```diff
@@ -109,25 +109,35 @@ async def run_tool_then_structured_flow(
         ),
     ]
 
-    # gemini-2.5-flash occasionally returns an empty candidate on the
-    # post-tool turn; in production the executor's retry layer absorbs
-    # that, but this calls the backend directly, so retry here.
+    # tool_choice stays "auto" on the replay turn — the production tool loop
```
eisene, line 112: These are llm-style comments but I thought worth keeping here since these test failure modes are important to know

tests/live_llm/test_live_tools_structured_output.py
```diff
@@ -0,0 +1,246 @@
...
+    "model_spec",
+    get_live_model_specs(provider="gemini", feature="structured_output"),
+    ids=lambda spec: spec.id,
+)
+async def test_live_gemini_tools_with_structured_output(
+    model_spec: LiveModelSpec,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    require_provider_key(model_spec)
+    # No temperature pin: gemini-2.5-flash occasionally returns an empty
```
eisene, line 224: These are llm-style comments but I thought worth keeping here since these test failure modes are important to know

.github/workflows/live-llm-tests.yml
```diff
@@ -0,0 +1,174 @@
...
+          ref: ${{ github.sha }}
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ vars.AWS_OIDC_ROLE_ARN }}
+          aws-region: us-east-1
+          role-duration-seconds: 3600
+
+      - name: Resolve secret ids from latest git tags
```
akattelu: nit: consider making this a shared step between other workflows that fetch secrets so its all in one place

.github/workflows/live-llm-tests.yml
```diff
@@ -0,0 +1,174 @@
...
+      - 'src/config.py'
+      - 'tests/live_llm/**'
+      - 'pyproject.toml'
+      - 'uv.lock'
+      - '.github/workflows/live-llm-tests.yml'
+  # Manual trigger for PRs: add the `run-live-llm` label to run the suite
+  # against the PR's merge commit. The label is purged as soon as the run
+  # starts so it can be re-added to trigger another run.
+  pull_request:
+    types: [labeled]
```
akattelu, line 20: do you know if this can be triggered by maintainers only? or any contributor?

.github/workflows/live-llm-tests.yml
```diff
@@ -0,0 +1,174 @@
...
+      - name: Configure test environment
+        run: |
+          {
+            # The staging dotenv carries AUTH_USE_AUTH=true without a usable
+            # JWT secret; src/config.py validates the pair at import time, so
+            # disable auth (this suite never runs the API server anyway).
+            echo "AUTH_USE_AUTH=false"
+            echo "SENTRY_ENABLED=false"
+            echo "TELEMETRY_ENABLED=false"
+            echo "LIVE_LLM_ANTHROPIC_45_PLUS_MODELS=claude-sonnet-4-5"
```
akattelu, line 102: Any other model configs we should set here? Just want to be mindful of cost and consistency as this will run a lot.

tests/live_llm/test_live_openai.py
```diff
@@ -93,7 +93,10 @@ async def test_live_openai_gpt5_reasoning_structured_output_and_prefix_caching(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
     require_provider_key(model_spec)
-    backend, config = make_backend(model_spec, reasoning_effort="minimal")
+    # gpt-5.4 dropped 'minimal' from the reasoning_effort vocabulary
+    # (none/low/../xhigh); 'low' is the lowest tier both generations accept.
+    reasoning_effort = "low" if model_spec.model.startswith("gpt-5.4") else "minimal"
```
akattelu: is there a different way we can test for this? I think gpt-5.4 is the default we have set so that makes sense and this is only a test, but are there other openai models that apply this behavior and should we include that in the openai backend?

tests/live_llm/test_live_tools_structured_output.py
```diff
@@ -0,0 +1,246 @@
...
+
+class FavoritePrimeReport(BaseModel):
+    number: int
+    is_prime: bool
+    summary: str
+
+
+_INITIAL_PROMPT = (
+    "Before answering, call the get_favorite_prime tool exactly once. "
+    "Do not answer with plain text or JSON on this turn. "
```
akattelu, line 65: Not fully following this here, but shouldn't we exclude JSON-related initial prompt guidance? That way we can verify that the completion functions append the right guidance strings?

eisene: This is just to make sure the LLM actually calls the tool on the first turn instead of outputting something. This shouldn't leak into the test assertions as written.

.github/workflows/live-llm-tests.yml
eisene: good point, changing

eisene: so the answer is that it's only people with write access to the repo.

however, we still decided to use this as an opportunity to introduce CODEOWNERS: https://github.com/plastic-labs/honcho/pull/907/commits/3f229f96c30442ad34dfae47c1bb97adbf5b4478

.github/workflows/live-llm-tests.yml
eisene: https://github.com/plastic-labs/honcho/pull/907/commits/335ec95c177b806604ceba66eb2df8f9e3a5c9f0

.github/workflows/live-llm-tests.yml
eisene: i would keep it to a minimum for now, can always expand

tests/live_llm/test_live_openai.py
eisene: My instinct is to not override the gpt-5 special case in the transport and let it fail. LMK if you disagree.

I agree the test should be changed: https://github.com/plastic-labs/honcho/pull/907/commits/16c2c7934c43ac60410ade0f1f3df03d4d6d9251

eisene:
Temporarily removed the gate and triggered the workflow to check it still works: https://github.com/plastic-labs/honcho/actions/runs/29366943935/job/87200792914?pr=907

eisene:
After reverting the removal, the job is now skipped: https://github.com/plastic-labs/honcho/actions/runs/29367192968/job/87201616191?pr=907

akattelu (approved):

PR #910, opened 2026-07-16
feat(telemetry): counts documents deduped during representation, exact and semantically similar

ulyssepence:
# Background

[#861](https://github.com/plastic-labs/honcho/pull/861) added exact-content deduplication during document creation (alongside the existing semantic dedup), but we have no visibility into how often it fires in practice ([DEV-1967](https://linear.app/plastic-labs/issue/DEV-1967/track-times-derived-and-exact-deduplication-metrics)).

This PR threads dedup counts through the deriver pipeline and emits them on the `representation.completed` cloudevent, so we can see how often the deriver deduplicates within a representation task and whether duplicates come from the same batch inference or from previously stored conclusions.

# Testing

```bash
uv run ruff check src/
uv run basedpyright
uv run pytest tests/
```

## Summary by CodeRabbit

* **New Features**
  * `create_documents` now returns a result object with created documents plus exact/semantic dedup outcome metrics (rejected vs replaced).
  * Representation completion telemetry now includes aggregated deduplication counts (in-batch, existing, semantic rejected, semantic replaced).

* **Bug Fixes**
  * Improved saved/created observation and document counting to align with the new result-based contract.

* **Tests**
  * Updated and expanded deduplication, result-shape, aggregation, and telemetry field round-trip coverage.

### DIFF
--- a/src/telemetry/events/representation.py
+++ b/src/telemetry/events/representation.py
@@ -22,7 +22,7 @@ class RepresentationCompletedEvent(BaseEvent):
     """
 
     _event_type: ClassVar[str] = "representation.completed"
-    _schema_version: ClassVar[int] = 2
+    _schema_version: ClassVar[int] = 3
     _category: ClassVar[str] = "representation"
 
     # Workspace context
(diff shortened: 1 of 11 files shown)


src/telemetry/events/representation.py
```diff
@@ -22,7 +22,7 @@ class RepresentationCompletedEvent(BaseEvent):
     """
 
     _event_type: ClassVar[str] = "representation.completed"
-    _schema_version: ClassVar[int] = 2
+    _schema_version: ClassVar[int] = 3
```
akattelu: nit: schema version bump is not necessary because this is additive, but I'm ok w keeping it

akattelu (dismissed):

src/telemetry/events/representation.py
Rajat-Ahuja1997: @akattelu ah this is my fault - i told Ulysse that we should always bump the schema version for any schema changes. but I support keeping the same schema version for additive-only changes

akattelu (approved):

PR #917, opened 2026-07-19
feat(embedding): optionally truncate oversized embeddings to configured dims

JustinBacher:
## Problem

Matryoshka (MRL) embedding models served behind OpenAI-compatible endpoints (llama.cpp, some vLLM builds) ignore the `dimensions=` request parameter and always return their **native** dimension. Honcho then hard-fails in `_validate_embedding_dimensions` on the mismatch, so these models are unusable — even though truncating an MRL vector to a smaller size is a valid, designed-in operation.

Concretely: a model like `llama-nemotron-embed-1b-v2` emits 2048 dims. If you configure `EMBEDDING_VECTOR_DIMENSIONS=1536`, honcho sends `dimensions=1536`, the server ignores it and returns 2048, and every embedding raises `Embedding dimension mismatch`.

This also collides with pgvector's **2000-dimension HNSW index cap**: native dims above 2000 can't be indexed at all, so truncating to a supported MRL size (e.g. 1024/1536) is the only way to keep an ANN index.

## Change

Add an opt-in `EMBEDDING_TRUNCATE_TO_DIMENSIONS` flag (default `false`). When enabled and the provider returns a vector **larger** than `EMBEDDING_VECTOR_DIMENSIONS`, keep the leading components and **L2-renormalize** so cosine / inner-product similarity stays well-defined, then validate. When disabled (default), behavior is unchanged — a genuinely misconfigured model still fails loudly.

- `src/config.py` — new `EmbeddingSettings.TRUNCATE_TO_DIMENSIONS` field, documented.
- `src/embedding_client.py` — `_l2_truncate()` helper + `_fit_embedding_dimensions()`, applied at both OpenAI call sites (single + batch); threaded through the singleton and its cache signature. Gemini path unchanged (it already honors `output_dimensionality`).
- `.env.template` — document the flag.
- `tests/llm/test_embedding_client.py` — 4 new tests (helper slice+renorm, zero-norm guard, client truncates when enabled, client still rejects when disabled).

## Testing

- `uv run pytest tests/llm/test_embedding_client.py` — 20 passed.
- `uv run basedpyright src/embedding_client.py src/config.py` — 0 errors.
- `uv run ruff check` / `ruff format` — clean.
- Verified end-to-end against a local llama.cpp serving `llama-nemotron-embed-1b-v2` (native 2048): with the flag on and `EMBEDDING_VECTOR_DIMENSIONS=1536`, embeddings come back 1536-dim, unit-norm, and store/search in a `vector(1536)` pgvector column with an HNSW index.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to truncate oversized embedding vectors to the configured dimensions.
  * Truncated vectors are L2-renormalized for consistent results.
  * The option remains disabled by default, preserving existing dimension-mismatch errors.

* **Documentation**
  * Added configuration guidance for providers that ignore requested embedding dimensions.

* **Tests**
  * Added coverage for truncation, normalization, zero vectors, and mismatch handling.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -19,6 +19,12 @@ PERFORMANCE_LOG_FORMAT=compact # compact|rich
 # EMBEDDING_VECTOR_DIMENSIONS=1536
 # EMBEDDING_MAX_INPUT_TOKENS=8192
 # EMBEDDING_MAX_TOKENS_PER_REQUEST=300000
+# Truncate + L2-renormalize oversized embeddings down to EMBEDDING_VECTOR_DIMENSIONS
+# instead of erroring on a dimension mismatch. Use with Matryoshka (MRL) models
+# served behind OpenAI-compatible endpoints (llama.cpp, some vLLM builds) that
+# ignore the `dimensions=` request parameter and always emit their native size.
+# Off by default so a genuinely misconfigured model still fails loudly.
+# EMBEDDING_TRUNCATE_TO_DIMENSIONS=false
 # EMBEDDING_MODEL_CONFIG__TRANSPORT=openai
 # EMBEDDING_MODEL_CONFIG__MODEL=text-embedding-3-small
 # EMBEDDING_MODEL_CONFIG__OVERRIDES__BASE_URL=
--- a/src/config.py
+++ b/src/config.py
@@ -758,6 +758,14 @@ class EmbeddingSettings(HonchoSettings):
     # saturated, message creation skips the fast path entirely and the
     # reconciler embeds on its next cycle. 0 disables the fast path.
     MAX_PENDING_EMBED_TASKS: Annotated[int, Field(default=50, ge=0)] = 50
+    # When the embedding provider returns a vector larger than
+    # VECTOR_DIMENSIONS, truncate it to VECTOR_DIMENSIONS and L2-renormalize
+    # instead of raising a dimension-mismatch error. Intended for MRL-trained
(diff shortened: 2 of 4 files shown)


karaokedjodua:
Finn-loop review of 02edbe0

CI: not configured (CodeRabbit review passed; no required checks on this branch)
Mergeability: BLOCKED — stale branch protection (mergeable=MERGEABLE but mergeStateStatus=BLOCKED; `gh pr checks --required` reports "no required checks reported")

## Review
Adds an opt-in `EMBEDDING_TRUNCATE_TO_DIMENSIONS` flag (default `false`) that truncates oversized MRL-trained embeddings to the configured dimension and L2-renormalizes. Four files touched: config, embedding client (single + batch paths), env template, and tests. Code is clean, focused, and well-tested.

## 1. Must fix before merge
**[CI]** No required CI configured on the `feat/embedding-mrl-truncation` branch. `gh pr checks --required` reports "no required checks reported" and `mergeStateStatus` is BLOCKED while `mergeable` is MERGEABLE — this is the stale branch-protection rule pattern. An admin must update or remove the stale rule before this can merge.

## 2. Should fix soon
None.

## 3. Safe to merge
Code is clean — the change is minimal, correctly scoped, uses a safe default (off), has coverage for the truncation path, the zero-norm edge case, and the backward-compatible rejection path. The CI infrastructure gap is the only blocker.

Labels needed: `needs-human-review`, `loop-changes-requested`

PR #925, opened 2026-07-22
fix(embedding): raise Gemini max_embedding_tokens cap from 2048 to 8192

Stromweld:
## Summary
Gemini embedding models support up to **8192** input tokens, not 2048. The code had an artificial `min(max_input_tokens, 2048)` cap that unnecessarily truncated/chunked text for Gemini-backed embedding configs even when the caller-configured `max_input_tokens` was higher.

## Fix
Changed the Gemini-specific cap in `_EmbeddingClient.__init__` from 2048 to 8192, matching Gemini's actual documented per-request token limit. This is a provider-specific fix (only affects the `gemini" transport branch); the OpenAI branch is untouched.

## Tests
Added `test_gemini_embedding_client_honors_8192_token_cap` in `tests/llm/test_embedding_client.py`, verifying:
- `max_input_tokens` above the provider cap is clamped to 8192 (previously would have been clamped to 2048)
- `max_input_tokens` below the cap passes through unchanged

Full targeted test run (all passing):
```
17 passed, 17 warnings in 8.95s
```

---
Created with AI assistance — Hermes Agent v0.17.0, model Claude Sonnet 4.6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Gemini embedding token-limit handling to use model-specific input caps: 8,192 tokens for `gemini-embedding-2`, 2,048 tokens for `gemini-embedding-001`, with a conservative 2,048-token fallback for unknown model IDs.
  * This changes which inputs pass token-limit validation during embedding.

* **Tests**
  * Added Gemini-specific tests verifying per-model clamping, unknown-model fallback behavior, and correct handling to avoid applying the higher cap to near-miss model IDs.

PR #926, opened 2026-07-22
fix: handle Gemini batch embedding count mismatch (avoid zip crash)

Stromweld:
## Problem

Batch Gemini embedding calls in `src/embedding_client.py` did:

```python
contents=[item.text for item in batch]
...
for item, embedding in zip(batch, response.embeddings, strict=True):
```

This assumed Gemini always returns exactly one embedding per input item, in order. While the SDK docstring claims this, it does not always hold in practice (possible causes: empty/whitespace items dropped, batch size >100, other API quirks). When the counts didn't match, `zip(..., strict=True)` raised `ValueError: zip() argument 2 is shorter than argument 1`, crashing the whole batch.

## Fix

- Keep the efficient multi-item batch call as the normal path.
- Explicitly check `len(response.embeddings) != len(batch)` before zipping.
- On a mismatch, fall back to embedding each item in that batch **individually** (one call per item, preserving original order) instead of guessing at alignment.
- Each individual fallback call must return exactly one embedding; if it returns zero or multiple, raise `ValueError` with a clear diagnostic (text_id/chunk_index) instead of silently dropping or misaligning data.

## Tests

Added to `tests/llm/test_embedding_client.py`:
- Normal single-item batch response
- Normal multi-item batch response, verifying item-to-embedding order
- Mismatched (short) multi-item batch response triggering per-item fallback, verifying correct mapping
- Fallback individual call returning zero embeddings raises a controlled `ValueError`

```
20 passed, 17 warnings in 10.93s
```

---

This is a separate, independent fix from BUG1 (PR #925) — no overlap with that branch/commit.

Created with AI assistance — Hermes Agent v0.17.0, model Claude Sonnet 4.6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Gemini batch embeddings by validating response cardinality against the requested batch size.
  * Ensures embeddings are correctly mapped back to items; falls back to per-item embedding when batch alignment isn’t safe.
  * Added strict error handling for embeddings missing required values and for invalid per-item response sizes.
* **Tests**
  * Extended Gemini embedding tests to cover matched batch success, order preservation, fallback behavior, and multiple error scenarios.

### DIFF
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -460,10 +460,16 @@ class _EmbeddingClient:
                     # input item, in order. This is the documented/expected
                     # behavior and the efficient common case.
                     for item, embedding in zip(batch, embeddings, strict=True):
-                        if embedding.values:
-                            result[item.text_id][item.chunk_index] = (
-                                self._validate_embedding_dimensions(embedding.values)
+                        if not embedding.values:
+                            raise ValueError(
+                                "Gemini API returned an embedding with no "
+                                f"values for text_id={item.text_id!r}, "
+                                f"chunk_index={item.chunk_index!r} (batch "
+                                "path)."
                             )
+                        result[item.text_id][item.chunk_index] = (
+                            self._validate_embedding_dimensions(embedding.values)
+                        )
                 else:
                     # Cardinality mismatch: do NOT assume any particular
                     # alignment between `batch` and `embeddings` (e.g. via
@@ -497,10 +503,16 @@ class _EmbeddingClient:
                                 f"chunk_index={item.chunk_index!r}."
                             )
                         embedding = item_embeddings[0]
-                        if embedding.values:
-                            result[item.text_id][item.chunk_index] = (
-                                self._validate_embedding_dimensions(embedding.values)
+                        if not embedding.values:
+                            raise ValueError(
+                                "Gemini API returned an embedding with no "
+                                f"values for text_id={item.text_id!r}, "
+                                f"chunk_index={item.chunk_index!r} (fallback "
+                                "path)."
                             )
+                        result[item.text_id][item.chunk_index] = (
+                            self._validate_embedding_dimensions(embedding.values)
+                        )
             else:  # openai
                 openai_kwargs: dict[str, Any] = {
                     "model": self.model,
--- a/src/embedding_client.py
+++ b/src/embedding_client.py
@@ -448,13 +448,55 @@ class _EmbeddingClient:
             item in analytics."""
             result: dict[str, dict[int, list[float]]] = defaultdict(dict)
             if isinstance(self.client, genai.Client):
-                response = await self.client.aio.models.embed_content(
+                gemini_client = self.client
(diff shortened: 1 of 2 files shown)


karaokedjodua:
Finn-loop review of 153f55a

CI: not configured (mergeStateStatus=BLOCKED, mergeable=MERGEABLE, but `gh pr checks --required` reports "no required checks configured" — stale branch-protection rule)
Mergeability: clean (structurally mergeable)

## Review
PR #926 fixes a `zip(strict=True) ValueError` crash when Gemini batch embedding calls return fewer embeddings than input items. The fix adds a cardinality check before zipping and falls back to per-item embedding calls on mismatch, preserving correct item-to-embedding alignment. Eight new deterministic tests cover normal batch, order preservation, mismatch fallback, and all error paths.

## 1. Must fix before merge
- **[CI]** No required CI checks are configured on this branch. `gh pr checks 926 --required` reports "no required checks reported" and the only running check is CodeRabbit (passed). The repository's CLAUDE.md documents `uv run ruff check src/`, `uv run basedpyright`, and `uv run pytest tests/` as required quality gates, but none are enforced via branch protection. Additionally, `mergeStateStatus` is `BLOCKED` while `mergeable` is `MERGEABLE`, which indicates a stale branch-protection rule referencing a check that no longer exists. An admin must either update the branch-protection rules to match current checks or remove the stale rule.

## 2. Should fix soon
None. The code changes are well-scoped:
- The fast path (`len(embeddings) == len(batch)`) preserves existing behavior with an added `embedding.values` emptiness check (previously this would silently produce a falsy result; now it raises a controlled `ValueError`).
- The mismatch fallback iterates `batch` sequentially with per-item embedding calls, preserving order without guessing alignment.
- Per-item fallback calls validate exactly one embedding returned and non-empty values, with diagnostic `text_id`/`chunk_index` in error messages.
- Tests cover: single-item normal, multi-item order preservation, mismatch triggering fallback, zero-embedding fallback error, multi-embedding fallback error, empty-values-in-matched-path error, and empty-values-in-fallback error.

## 3. Safe to merge
No — CI gate is missing. Automated review of the code itself finds no defects, security issues, or AC gaps. A human must verify lint/typecheck/tests pass and resolve the stale branch-protection rule before merging.

## Notes
- No formal issue (Paperclip KAR-NNN) was linked in the PR body; the PR body itself served as the contract.
- Labels (`needs-human-review`) could not be applied — this reviewer's account lacks admin rights on the repository.

akattelu:
Hey @Stromweld, thanks for digging into this, the zip crash is real. I think the fallback is at the wrong layer though: the root cause is that `_process_batch` passes bare strings as contents, which `gemini-embedding-2*` models treat as a single content, so the mismatch fallback here would fire on every batch, costing one wasted batch call plus N sequential singles forever. #745 fixes the call itself by wrapping each item as Content(parts=[Part(text=...)]). Your empty-values and cardinality guards are worth keeping, we'll port those onto the consolidated fix. Closing in favor of #745

PR #496, opened 2026-04-05
fix(dev): Skip lancedb install on Intel macOS

papesy384:
## Summary

Improves local developer experience and documents common setup pitfalls: Swagger defaulting to production, `uv sync` failing on macOS Intel (LanceDB wheels), and Postgres port conflicts.

### Code / lockfile
- **`src/main.py`** — OpenAPI `servers` lists `http://127.0.0.1:8000` and `http://localhost:8000` before production so Swagger **Try it out** targets the local server by default.
- **`pyproject.toml`** — `lancedb` is omitted on macOS Intel (`darwin` + `x86_64`) where PyPI has no wheel; default vector store remains **pgvector**.
- **`uv.lock`** — Regenerated for the dependency marker.

### Documentation
- **`README.md`** — **Local development notes**: Intel Mac + LanceDB, Swagger local servers, Postgres port conflicts with Docker.
- **`CONTRIBUTING.md`** — Clarifies documentation contribution scope and validation expectations.

## Why this matters

- Reduces mistaken calls to production from `/docs` during local development.
- Unblocks `uv sync` on Intel Macs for the default stack.
- Helps self-hosters resolve DB port clashes without digging through issues.

## How to verify

- Open `http://127.0.0.1:8000/docs` → Servers dropdown should default to a local URL.
- `curl -s http://127.0.0.1:8000/openapi.json | jq '.servers[0]'` → first server is local.
- On Intel Mac: `uv sync` completes.

---

**Checklist**
- [x] No secrets in commits

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Adjusted LanceDB installation requirements for specific system architectures.

* **Bug Fixes**
  * Enhanced error messaging with installation instructions when LanceDB is unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/plastic-labs/honcho/pull/496)

PR #927, opened 2026-07-22
telemetry: zero-initialize bounded-label metrics so an absent series means a broken scrape

philosaether:
## What

Zero-initialize every Prometheus metric whose label domain is bounded and enumerable, so the metric is present at `0` from process start instead of springing into existence on first use.

Started as just `telemetry_events_dropped` (the original scope of this PR) and generalized to all bounded-label counters plus the gauges in the same family.

## Why

A labeled Prometheus counter exports **no** time series until its first `labels(...)` call. So a never-yet-incremented metric is indistinguishable from a broken scrape:

- you can't alert on or graph a series that doesn't exist yet;
- "nothing happened" looks identical to "the exporter is down".

Materializing children at `0` inverts that: a missing series now means something is actually wrong, and "no events" reads as a flat `0` rather than a gap.

## The constraint that shapes the whole change

Zero-init is only honest for label sets we can enumerate **truthfully**. Fabricating a permanently-`0` series that no code path can ever increment is the same lie in the other direction. So a metric is initialized only when its label domain is bounded, enumerable at startup, *and* actually emitted by that process. Three consequences:

**High-cardinality labels are excluded.** No `endpoint`, no `workspace_name`.

**Impossible tuples are excluded.** `deriver_tokens_processed` is initialized from `_DERIVER_TOKEN_COMBOS_BY_TASK`, not the cartesian product — input tokens only pair with input components, output only with `output_total`, and `previous_summary` occurs only for `summary` tasks (an ingestion has no previous summary). 7 series, not the 8 a flat set would give, and not the 16 a cartesian product would.

**Init is per-process.** The API and deriver emit disjoint metric sets, so `initialize_bounded_metrics(instance_type=...)` branches. Event-type counters are initialized in *both* (both processes run an emitter); token and embed counters are split. Initializing all 21 event types in both processes does mean the API exports a permanent `0` for `dream.run` — but these are counters, so `rate()` is 0 and `sum()` across instances is still correct. That tradeoff is deliberate, and preferred over maintaining a fragile per-event-type → process map.

**Multi-instance safety, stated as a rule.** The metrics here fall into three classes, and only one is hazardous:

| class | example | safe? |
|---|---|---|
| instance-scoped | `telemetry_buffer_size` | any aggregation is meaningful |
| service-scoped **additive** | token counters | `sum()` reconstructs the whole |
| service-scoped **non-additive** | `message_embeddings_pending` | no aggregation is correct once instances disagree |

A metric in the third class must be refreshed by **every** instance on its own timer, or it belongs in an exporter that yields exactly one series. This is now written down in `initialize_bounded_metrics`'s docstring and in `CLAUDE.md`.

## New metric

`message_embeddings_pending` — `MessageEmbedding` rows awaiting a vector (`sync_state='pending'`), the durable DB-wide backlog the reconciler drains. Distinct from `embed_now_tasks_in_flight`, which is per-process fast-path work.

It's a category-3 metric, so it's refreshed by every deriver replica from `ReconcilerScheduler._scheduler_loop` — **not** from `run_vector_reconciliation_cycle`, which runs behind work-unit dedup and would leave every non-winning replica exporting a stale value or, combined with the zero-init, a confident permanent `0` it never measured. The help string mandates `max()`/`avg()` and forbids `sum()`.

## Drift guards

Two registries are hand-maintained on purpose, each paired with a test that fails when it drifts:

- `ALL_EVENT_TYPES` / `HIGH_VOLUME_EVENT_TYPES` (`src/telemetry/events/__init__.py`) — asserted set-equal to the `_event_type` values discovered by walking `BaseEvent` subclasses. Adding an event type without registering it fails the test with a pointer to the fix.
- `_DERIVER_TOKEN_COMBOS_BY_TASK` — asserted to use only real enum values, to cover every `DeriverComponents` member across tasks, and to stay strictly smaller than its cartesian product.

Explicit-list-plus-test rather than deriving at runtime: a derived registry would silently follow whatever happens to be imported, so a type could drop out with no code change and no failure. Event types are a wire contract consumers depend on, so the failure should be loud.

Specialist names go the other way — `initialize_bounded_metrics` walks `walk_subclasses(BaseSpecialist)` and reads `.name`, because a specialist name is an internal label, not a contract. That paid off during this PR: `CardRefreshSpecialist` landed on `main` mid-branch (#883) and was covered with no change.

## Review response

Vineeth reviewed all commits and requested changes on three things, all resolved:

- **`message_embeddings_pending` was a DB-global value on a per-instance gauge.** His framing was right and it was the real problem — the zero-init turned "this replica has no idea" into a confident permanent `0`. Adopted his suggested fix (relocate the count into `_scheduler_loop`). His longer-term point — that this belongs in `postgres_exporter`, since it's a property of the database rather than of Honcho — is deferred, not dismissed.
- **Missing `METRICS.ENABLED` guard** on `initialize_telemetry_dropped_metrics`, asymmetric with its sibling. Added.
- **Duplicate `REASONING_LEVELS`.** Removed in favour of `src/config`'s. Worth flagging: `config.py`'s copy was *hand-written*, while the one here was `get_args(ReasoningLevel)` — so deduping in that direction would have quietly traded away a real guarantee (you could drop a level and still typecheck). Fixed by making config's list `get_args`-derived, which keeps the dedup and restores the invariant.

Also from review: `__subclasses__()` is direct-children-only, so subclass discovery is now recursive via a shared `src/utils/types.walk_subclasses` (it had been duplicated between `metrics.py` and the test).

## Testing

`tests/telemetry/test_metric_zero_init.py` — this PR's original commit shipped without tests; this backfills that and covers the generalization. Presence/absence is read via `REGISTRY.get_sample_value` rather than `counter.labels(...)`, because `.labels` would itself materialize the child and destroy the signal being asserted. Each test pins a unique `namespace` so the process-global registry can't let one test satisfy another's assertion.

Covers: children materialized at `0`; high-cardinality and impossible combinations deliberately *not* materialized; ground-truth event types absent from `sampled_out` (they can never be sampled out, so a `0` there would be permanently misleading); per-process isolation in both directions; and the drift guards.

`tests/reconciler/test_pending_backlog_gauge.py` — pins both halves of the gauge relocation: that `_scheduler_loop` refreshes it, and that `run_vector_reconciliation_cycle` doesn't. Mutation-checked — the first assertion fails when the relocation is reverted.

Local: `ruff` + `basedpyright` clean (0 errors), full suite green apart from 4 pre-existing `tests/crud/test_document.py` failures that need an OpenAI key (reproduced on clean `origin/main`).

## Rebased

Rebased onto `origin/main` (was 8 behind). The rebase was load-bearing, not hygiene: #883 added a third `BaseSpecialist`, and the specialist assertion had been hardcoded to two names while only checking *presence* — so it would have kept passing while leaving the new specialist uncovered. It now derives from the subclasses.

## Deferred

`postgres_exporter` for the backlog gauge; trimming the token counters, which are a lossy projection of data the ground-truth cloudevent envelopes already carry with better labels; a partial index for the pending count.

## One out-of-scope commit

`3554600` fixes two `basedpyright` warnings in files this PR otherwise doesn't touch (`src/vector_store/__init__.py` from #496, `tests/test_cache_redaction.py` from #869). CI runs `basedpyright` bare and it exits non-zero on *any* warning, so both have been failing the staticanalysis job on **every** branch cut from current `main` — clean `origin/main` reports the same "0 errors, 2 warnings" and exits 1. Fixed here only because they block this PR from going green; happy to split them out if preferred.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Prometheus telemetry startup so dropped-event metric series are pre-materialized (buffer-full and send-failure reasons) and remain visible before any drops occur.
  * Made metric series materialization fail-soft to reduce startup issues when metrics initialization encounters errors.
* **New Features**
  * Added a Prometheus gauge tracking the reconciliation “pending embeddings” backlog, updated after reconciliation sweeps.
  * Enhanced bounded metric initialization at startup for earlier `/metrics` visibility in both API and deriver.
* **Tests**
  * Added a comprehensive suite validating metric zero-initialization, allowed label sets, and drift-guard behavior for event types and series.


src/telemetry/prometheus/metrics.py
```diff
@@ -325,6 +325,27 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
         except Exception as e:
             self._handle_metric_error("record_telemetry_event_dropped", e)
 
+    def initialize_telemetry_dropped_metrics(self, *, reasons: list[str]) -> None:
```
Rajat-Ahuja1997, line 385: nice - we should generalize this method to be `initialize_counter_series` (or something similar) to support initializing other counters in the future

Rajat-Ahuja1997 (commented):
looks good, 1 comment

karaokedjodua:
Finn-loop review of 313fd41

CI: not configured (`gh pr checks --required` returns "no required checks reported on the 'phil/telemetry-dropped-zero-init' branch")
Mergeability: BLOCKED — mergeStateStatus=BLOCKED indicates branch protection blocks merge. Despite all individual checks passing (basedpyright, test-python, CodeQL, CodeRabbit), GitHub's merge queue reports BLOCKED, meaning required checks exist in branch-protection rules but don't match the ones running. This is a CI configuration gap, not a code defect.

## Review
Summary: PR materializes Prometheus counter children at 0 on startup so metrics like `telemetry_events_dropped` are visible in Grafana before any event fires — "no drops" vs "metric missing" becomes distinguishable. The code also pre-initializes the full bounded-label domain (event types, dialectic/deriver/dreamer tokens) and adds a new `message_embeddings_pending` gauge tracking the embedding backlog.

## 1. Must fix before merge
- **[CI]** mergeStateStatus is BLOCKED. All individual checks pass, but branch protection rules don't see them as satisfying the required-checks gate. A repo admin needs to either add the correct checks to the branch-protection rule or remove the stale requirement. This is not a code defect — it's a repo-configuration gap that prevents ANY merge, regardless of code quality.
- **[CI] / [DEFECT]** No required CI checks are registered on this branch. `gh pr checks 927 --required` exits 1 with "no required checks reported". While individual checks do run (basedpyright, test-python), they aren't gating. A human should decide whether these should become required before this (or any future) PR can merge.

## 2. Should fix soon
- **[AC-N] Scope**: The PR body describes pre-creating `telemetry_events_dropped` children at 0, but the actual diff does significantly more:
  - `ALL_EVENT_TYPES` + `HIGH_VOLUME_EVENT_TYPES` registries (src/telemetry/events/__init__.py) — new infrastructure for enumerating event types
  - Full bounded-metrics init (`initialize_bounded_metrics`) covering dialectic tokens, deriver tokens, dreamer tokens, embed-now gauges, and buffer gauges — way beyond the stated dropped-metrics scope
  - `message_embeddings_pending` gauge + `_record_pending_embeddings_backlog()` DB query in `src/reconciler/sync_vectors.py` — an entirely separate feature (embedding-backlog tracking) bolted onto the same PR
  
  The code is well-written and tested, but these three additions each deserve their own PR with their own contract/testing story. A human should decide whether to accept the scope-creep as-is or request splitting.

## 3. Safe to merge
No — merge is mechanically blocked by BLOCKED status, and required CI is not configured. Code quality is high: fail-soft on init (never crashes startup), drift-guard tests prevent stale registries, cartesian-product explosion is avoided by explicit valid-combo tables, and the 294-line test suite is thorough.

## Label notes
This reviewer's account does not have admin rights to `plastic-labs/honcho`, so labels cannot be applied via API. A repo admin should apply:
- `loop-changes-requested` — CI gate is missing + scope concerns
- `needs-human-review` — BLOCKED merge state needs admin attention

tests/telemetry/test_metric_zero_init.py
```diff
@@ -0,0 +1,269 @@
...
+@pytest.fixture
+def metrics_enabled(monkeypatch: pytest.MonkeyPatch) -> Iterator[None]:
+    monkeypatch.setattr("src.config.settings.METRICS.ENABLED", True)
+    monkeypatch.setattr("src.config.settings.METRICS.NAMESPACE", NS)
+    yield
+
+
+def sample(name: str, **labels: str) -> float | None:
+    """Value of a series if it exists, else None. Never materializes it."""
+    return REGISTRY.get_sample_value(name, {"namespace": NS, **labels})
```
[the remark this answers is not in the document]
philosaether: Fixed in 4b40382. `NS = "test"` is replaced by `unique_ns(tag)` (uuid4-suffixed); the `metrics_enabled` fixture pins a fresh namespace per test and yields it, and `sample()` now resolves the namespace from `settings.METRICS.NAMESPACE` at read time, so no call site changed.

Worth recording that this was more than hygiene. Three other suites pin `settings.METRICS.NAMESPACE = "test"` — `tests/integration/test_token_metrics.py:98`, `tests/deriver/test_embed_now.py:142`, `tests/test_db_resilience.py:179` — and `tests/deriver/test_embed_now.py:135` materializes `embed_now_tasks_in_flight` under it, which is one of the series `test_api_init_materializes_dialectic_and_embed` asserts `is not None`. Since the process-global `REGISTRY` retains a child once anything materializes it, whether that presence assertion tested anything depended on which xdist worker happened to run which file; the absence assertions in `test_deriver_init_omits_impossible_token_combos` could flake the other way. So this was a live correctness hole in the assertions rather than namespace tidiness.

The three suites above still share `"test"` with each other. Filed as a follow-up to replace the per-file helper with a shared unique-namespace fixture in `tests/conftest.py`.

tests/telemetry/test_metric_zero_init.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    labels = dict(
+        task_type=DeriverTaskTypes.INGESTION.value,
+        token_type=TokenTypes.INPUT.value,
+        component=DeriverComponents.PROMPT.value,
+    )
+    before = sample("deriver_tokens_processed_total", **labels)
+    prometheus_metrics.initialize_bounded_metrics(instance_type="api")
+    after = sample("deriver_tokens_processed_total", **labels)
+    assert before == after
```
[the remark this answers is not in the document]
philosaether: Added in 4b40382 as `test_deriver_init_does_not_touch_api_counters`: a before/after delta on `dialectic_tokens_processed_total` around `initialize_bounded_metrics(instance_type="deriver")`, mirroring the existing API-to-deriver test, plus absence assertions on the two API-only `embed_now` series.

Mutation-checked rather than assumed: temporarily made the deriver branch touch `dialectic_tokens_processed_counter`, confirmed the new test fails, then reverted.

src/telemetry/prometheus/metrics.py
```diff
@@ -325,6 +325,27 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
...
+        Args:
+            reasons: The reason label values the calling emitter can produce.
+        """
+        for reason in reasons:
+            try:
+                telemetry_events_dropped_counter.labels(reason=reason)
+            except Exception as e:
+                self._handle_metric_error(
+                    "initialize_telemetry_dropped_metrics", e
+                )
```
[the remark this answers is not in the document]
philosaether: Not changing this one, deliberately.

`_touch()` and `_set_gauge_zero()` are contractually fail-soft — the docstring is explicit that "a bad init must never crash startup" — and `initialize_bounded_metrics` is called unguarded from `lifespan` / `start_metrics_server`. Narrowing the catch to Prometheus label-validation errors would let anything else (duplicate registration on reimport, a misconfigured registry) propagate into process boot, which is the exact failure mode these helpers exist to prevent. The broad except is the guarantee, not an oversight.

Leaving the thread open rather than resolving, in case you disagree on the boundary.

src/reconciler/sync_vectors.py
```diff
@@ -700,6 +701,30 @@ async def _cleanup_pgvector_batch(
         return True
 
 
+async def _record_pending_embeddings_backlog() -> None:
```
VVoruganti: **Service-global value exported as a per-instance gauge**

This counts pending rows DB-wide, but it's called from `run_vector_reconciliation_cycle`, which runs off the queue (`src/deriver/consumer.py:353`) behind work-unit dedup — so exactly one deriver machine per cycle sets it, whichever one dequeued the item. `fly.toml` has a scaled `deriver` process group with its own `/metrics` scrape, so across N machines:

- machines that lose the work unit export a stale value indefinitely
- machines that never win one export the zero-init `0` **forever**

`sum()` over-counts, `max()` returns the stalest-highest, `avg()` is meaningless. The zero-init makes it worse than no metric: it turns "this instance has no idea" into a confident, permanently-healthy `0` — the inversion of what this PR is trying to achieve.

Smallest fix: move the count into `ReconcilerScheduler._scheduler_loop` (`src/reconciler/scheduler.py:146`), which runs on *every* replica each interval. One indexed `COUNT` per replica per 300s is negligible (`ix_message_embeddings_sync_state_last_sync_at` covers it), and then instances can't disagree by more than one interval, so `max()`/`avg()` are both correct.

Longer term this probably doesn't belong in the app at all — it's a property of the database, not of Honcho, so `postgres_exporter` with a custom query gives you exactly one series and the question never arises.

src/telemetry/prometheus/metrics.py
```diff
@@ -155,6 +183,16 @@ class DialecticComponents(Enum):
     ["namespace"],
 )
 
+# Embedding backlog: MessageEmbedding rows still awaiting a vector
+# (sync_state='pending'). Distinct from embed_now_tasks_in_flight (which counts
+# in-flight fast-path work in the API process) — this is the durable, DB-wide
+# backlog the reconciler drains. Set once per reconciliation cycle in the deriver.
+message_embeddings_pending_gauge = NamespacedGauge(
```
VVoruganti, line 194: Companion to the comment on `_record_pending_embeddings_backlog`: this help string reads as a service-wide fact, which is what it is — unlike every other gauge here. `embed_now_tasks_in_flight` says "for this process" and `db_queries_in_flight` carries `instance_type`. Whatever we land on for that one, this help text should say who owns the value so a dashboard author doesn't reach for `sum()`.

src/telemetry/prometheus/metrics.py
```diff
@@ -66,6 +66,34 @@ class DialecticComponents(Enum):
     TOTAL = "total"
 
 
+# Bounded label domains used to zero-initialize counter children at startup (see
+# initialize_bounded_metrics). REASONING_LEVELS is derived from the config
+# Literal so it never drifts.
+REASONING_LEVELS: tuple[str, ...] = get_args(ReasoningLevel)
```
VVoruganti: `src/config.py:918` already defines `REASONING_LEVELS: list[ReasoningLevel]` with the same five values. Two same-named constants in two modules is exactly the drift this comment is trying to prevent.

```suggestion
```

...and import the existing one instead:

```python
from src.config import REASONING_LEVELS, settings
```

`get_args` and `ReasoningLevel` then become unused imports — ruff will point at both.

src/telemetry/emitter.py
```diff
@@ -172,6 +172,20 @@ async def start(self) -> None:
...
         self._running = True
         self._flush_task = asyncio.create_task(self._periodic_flush())
+
+        # Pre-create the dropped-event counter children at 0 so the metric is
+        # visible in Prometheus/Grafana before any drop occurs — a labeled
+        # counter exports nothing until its first observation. Lets us alert on
+        # telemetry drops and tell "no drops" apart from "metric missing".
+        from src.telemetry.prometheus.metrics import prometheus_metrics
+
+        prometheus_metrics.initialize_telemetry_dropped_metrics(
```
VVoruganti, line 183: Missing the `METRICS.ENABLED` guard that `initialize_bounded_metrics` has, so a metrics-disabled process still materializes `telemetry_events_dropped` children. Harmless today (`metrics_endpoint` 404s when disabled) but it's an asymmetry in a PR whose entire subject is which series exist — and `test_init_noop_when_metrics_disabled` asserts exactly this property for the sibling initializer.

Cleanest spot is inside `initialize_telemetry_dropped_metrics` itself, mirroring `metrics.py:421`, so both entry points behave the same.

Separately: keeping this per-emitter is the right call, but the docstring undersells why. It's not that the process-level init "does not know" the prefix — `"trace_"` is a static literal at `emitter.py:591`. It's that the trace emitter is conditional on `TRACE_PAYLOADS_ENABLED`, so hoisting it would fabricate `trace_*` series on deployments with tracing off. Worth saying, since the current reason invites someone to "fix" it later.

src/telemetry/prometheus/metrics.py
```diff
@@ -325,12 +363,131 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
...
+                        task_type=task_type_value,
+                        token_type=token_type_value,
+                        component=component_value,
+                    )
+            # dreamer tokens: specialist_name x token_type. Specialist names are
+            # derived from the concrete BaseSpecialist subclasses so a new
+            # specialist can't silently miss init.
+            from src.dreamer.specialists import BaseSpecialist
+
+            for specialist in BaseSpecialist.__subclasses__():
```
VVoruganti: `__subclasses__()` is direct-only — a `class Foo(DeductionSpecialist)` is silently skipped, and the test hardcodes `("deduction", "induction")` so it wouldn't catch that either. Given the comment above says "so a new specialist can't silently miss init", worth making it actually recursive:

```suggestion
            def _walk(cls: type[BaseSpecialist]) -> Iterator[type[BaseSpecialist]]:
                for sub in cls.__subclasses__():
                    yield sub
                    yield from _walk(sub)

            for specialist in _walk(BaseSpecialist):
```

`Iterator` is already imported here. `tests/telemetry/test_metric_zero_init.py` has this exact helper for `BaseEvent` — could hoist one copy.

src/telemetry/prometheus/metrics.py
```diff
@@ -325,12 +363,131 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
...
+        Args:
+            reasons: The reason label values the calling emitter can produce.
+        """
+        for reason in reasons:
+            self._touch(telemetry_events_dropped_counter, reason=reason)
+
+    def initialize_bounded_metrics(self, *, instance_type: str) -> None:
+        """Pre-create bounded-label counter children at 0 for this process.
+
+        See .meta design telemetry-counter-zero-init. Only counters whose full
```
VVoruganti: This and `src/telemetry/events/__init__.py:151` both point at `.meta design telemetry-counter-zero-init`, but there's no `.meta` directory in the repo. Either land the doc or inline the two-sentence rationale — the reasoning here is good and worth keeping readable without an external lookup.

src/telemetry/prometheus/metrics.py
```diff
@@ -325,12 +363,131 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
...
+        # own subset of event types. The domain is bounded/low-cardinality (~21
+        # types), so init the full set in each process rather than maintain a
+        # fragile per-event-type -> process map.
+        for event_type in ALL_EVENT_TYPES:
+            self._touch(telemetry_events_emitted_counter, type=event_type)
+        for event_type in HIGH_VOLUME_EVENT_TYPES:
+            self._touch(telemetry_events_sampled_out_counter, type=event_type)
+        self._set_gauge_zero(telemetry_buffer_size_gauge)
+
+        if instance_type == "api":
```
VVoruganti, line 471: Flagging for a follow-up rather than this PR: the three token counters getting zero-init here (~30 series between them) are a lossy projection of data the ground-truth cloudevents already carry with better labels.

| counter | envelope | envelope adds |
|---|---|---|
| `dialectic_tokens_processed` | `DialecticCompletedEvent` | `workspace_name`, cache read/creation, `hit_input_token_cap` |
| `deriver_tokens_processed` | `RepresentationCompletedEvent` | workspace, finer split, cap flags |
| `dreamer_tokens_processed` | `DreamSpecialistEvent` | workspace, per-dream correlation |

All three envelopes are `_volume_class = "ground_truth"`, so they skip the sampler and stay complete even at `HIGH_VOLUME_SAMPLE_RATE < 1.0`. Both feeds are populated from the same locals at the same call site.

Prometheus still earns its keep for alert latency and for not depending on the emitter being up — but for that, `token_type` alone is enough. The `component` / `specialist_name` breakdowns are analytics questions, and nobody pages on prompt-scaffold token ratio. If we trim those, ~20 of the series this PR zero-inits go away with them. Not asking for it here; just noting we may be adding init tables for metrics we'd rather shrink.

src/telemetry/prometheus/metrics.py
```diff
@@ -325,12 +363,131 @@ def record_telemetry_event_dropped(self, *, reason: str) -> None:
...
+        values are prefix-dependent (the trace emitter uses a ``trace_`` prefix),
+        which the process-level ``initialize_bounded_metrics`` does not know.
+
+        Args:
+            reasons: The reason label values the calling emitter can produce.
+        """
+        for reason in reasons:
+            self._touch(telemetry_events_dropped_counter, reason=reason)
+
+    def initialize_bounded_metrics(self, *, instance_type: str) -> None:
```
VVoruganti, line 408: Docstring suggestion, off the back of the `message_embeddings_pending` thread: the metrics here fall into three buckets, and only one of them is hazardous.

1. **instance-scoped** (`telemetry_buffer_size`, `embed_now_tasks_in_flight`) — per-process by nature, any aggregation is meaningful, zero-init unambiguously right
2. **service-scoped additive** (token counters, `telemetry_events_emitted`) — each instance holds a partial count, `sum()` reconstructs the whole, multi-instance safe
3. **service-scoped non-additive** (`message_embeddings_pending`) — every instance claims to speak for the whole service, so no aggregation is correct when they disagree

Naming the buckets here — with the rule that category 3 needs every instance refreshing on its own timer, or doesn't belong in the app — is what stops the next category-3 metric from landing quietly. Prometheus already stamps `instance`/`job` at scrape time, which is why 1 and 2 need no special handling.

tests/telemetry/test_metric_zero_init.py
```diff
@@ -0,0 +1,329 @@
...
+    REASONING_LEVELS,
+    DeriverComponents,
+    DeriverTaskTypes,
+    DialecticComponents,
+    TokenTypes,
+    prometheus_metrics,
+)
+
+
+def unique_ns(tag: str) -> str:
```
VVoruganti, line 37: The unique-namespace approach fixes cross-test interference, which was the actual bug. Worth noting the residue though: nothing reaps the children, so each run permanently adds ~40 series to the process-global `REGISTRY` under throwaway namespaces, and `generate_latest` output grows for the rest of the session. Nothing breaks today (243 pass) — only bites if a future test asserts on scrape output. `REGISTRY.unregister` is wrong here since the counters are module-level; it'd be clearing `counter._metrics` in teardown, which may not be worth it.

VVoruganti (dismissed):
Reviewed the full branch (all four commits). Verified locally: `ruff check src/` clean, `basedpyright` 0 errors, `tests/telemetry/` 243 passed.

The core idea is right and the execution is careful — excluding high-cardinality labels, excluding ground-truth types from `sampled_out`, task-aware deriver combos, fail-soft `_touch`. I traced every zero-inited label tuple back to its emit site and they all match: `_DERIVER_TOKEN_COMBOS_BY_TASK` is exactly what `track_deriver_input_tokens` (`deriver.py:118`, `summarizer.py:482`) plus the two `OUTPUT_TOTAL` sites produce, `HIGH_VOLUME_EVENT_TYPES` is exactly the set that can reach `emitter.py:266`, and `ALL_EVENT_TYPES` matches all 21 subclasses. The unique-namespace fix in 4b40382 was the right call for global-`REGISTRY` tests.

Requesting changes on three things:

- **`message_embeddings_pending` is a DB-global value exported as a per-instance gauge** (`sync_vectors.py:704`) — the zero-init turns "this replica has no idea" into a confident, permanently-healthy `0` on every deriver that didn't win the work unit. This is the one real problem and it's the inversion of the PR's own goal.
- **duplicate `REASONING_LEVELS`** (`metrics.py:72`) — `config.py:918` already has it.
- **missing `METRICS.ENABLED` guard** on `initialize_telemetry_dropped_metrics` — asymmetric with its sibling, which the PR itself tests for.

The rest are nits or follow-up framing. One note on what is *not* a problem: initializing all 21 event types in both processes means the API exports a permanent `0` for `dream.run` and the deriver for `dialectic.completed` — same "confident zero" shape, but these are counters, so `rate()` is 0 and `sum()` across instances is still correct. No aggregation error, and the documented tradeoff against a fragile per-type-to-process map is the right call. That contrast is exactly why the gauge is the issue: counters aggregate safely across replicas, a DB-global gauge does not.

src/reconciler/sync_vectors.py
philosaether: Adopted as written. Generalized in the instance / service-additive / service-non-additive distinction highlighted in the new PR description + CLAUDE.md

src/telemetry/prometheus/metrics.py
philosaether: Adopted, see above

src/reconciler/sync_vectors.py
philosaether: Followup question: Claude generated a pretty hefty comment explaining where the record_pending_embeddings_backlog() function is _not_ called, in response to this change request. 

In a human-written codebase I wouldn't bother with that sort of comment, but for a largely AI-maintained codebase I think it makes more sense: the next agent to modify this method gets instant context, can avoid re-making the same mistake, can check for similar categories of mistake, etc.

On the other hand, it makes the code slightly less scannable for a human, and reading the comment itself incurs a mildly hefty cognitive lift (for me at least, lol). I'd rather do all the cognitive heavy lifting all at once, in an architecture doc. So for human consumption I'd take this comment out of the codebase and put it in the wiki.

Have we put much thought into standards of clarity, maintainability, readability, what-goes-where? Happy to comply, improvise, or codify accordingly.

src/telemetry/prometheus/metrics.py
philosaether: Implemented as requested

src/telemetry/prometheus/metrics.py
philosaether: Done, but the list in config.py was handwritten, not derived from the Literal. Ftfy.

src/telemetry/emitter.py
philosaether: Guard in the method, docstring corrected, test added

src/telemetry/prometheus/metrics.py
philosaether: Chose to hoist; indirection now supported

src/telemetry/prometheus/metrics.py
philosaether: https://linear.app/plastic-labs/issue/DEV-2231/trim-token-counter-label-breakdowns-that-duplicate-ground-truth

tests/telemetry/test_metric_zero_init.py
philosaether: It's actually worse -- each test runs `unique_ns` so we're generating ~40*N(tests) = ~400 series per run. Still not worth it IMO

akattelu (approved):

Issue #928, opened 2026-07-23
[Bug] Dialectic "minimal" reasoning level returns null on tool-less empty responses


leo922oel:
# **🐞 Bug Report**

## **Describe the bug**

At `reasoning_level="minimal"`, `peer.chat(...)` returns `null` when the model answers without calling a tool (common for broad queries like "summarize what you know about X"). The same query at `low` or `medium` returns a normal answer from the same stored data.

To be clear about scope: the `minimal` tier's low-cost profile looks intentional and is not the concern here. It uses the same model as `low` but with `MAX_TOOL_ITERATIONS=1`, `MAX_OUTPUT_TOKENS=250`, a reduced toolset, and smaller prefetch. That is a reasonable cheap, single-shot tier.

The concern is that an empty completion is returned as `null` with no recovery, even though `execute_tool_loop` (`src/llm/tool_loop.py`) already has an empty-response retry meant to prevent exactly this. The retry is guarded by `iteration < max_tool_iterations - 1`. With `MAX_TOOL_ITERATIONS=1` that is `0 < 0`, which is `False`, so the retry never fires at `minimal`. The tool-less branch then returns before reaching the post-loop synthesis fallback (that fallback is only reached after iterations are exhausted through tool calls). `low` and `medium` avoid this only because they allow `MAX_TOOL_ITERATIONS >= 2`.

Is the `null` intended for `minimal`, or should the empty-response recovery cover the single-iteration case? The existing retry suggests empty responses are meant to be recovered, and the exclusion of `minimal` reads like a side effect of the guard arithmetic rather than a deliberate choice — the tier (plastic-labs/honcho#459) and the guard ([#814](<https://github.com/plastic-labs/honcho/issues/814>)) came from separate changes. Filing to confirm the intent; if it is a bug, I have a fix and a regression test ready.

One more data point: the `null` response used \~24 output tokens (well under the 250 cap), so this is an empty completion with no fallback, not a truncation at the token limit.

---

### **Is this a regression?**

The empty-response retry was added in [#814](<https://github.com/plastic-labs/honcho/issues/814>). The `minimal` tier (`MAX_TOOL_ITERATIONS=1`) predates it (plastic-labs/honcho#459). The guard has excluded the single-iteration case since [#814](<https://github.com/plastic-labs/honcho/issues/814>).

---

### **To Reproduce**

Deterministic unit test, no LLM or network needed. Add to `tests/llm/` and run on a clean `main`:

```python
from typing import Any
[...]

PR #929, opened 2026-07-23
fix(llm): recover empty tool-loop responses instead of returning null

leo922oel:
## Description

Ref: #928
`peer.chat(..., reasoning_level="minimal")` returned `null` on broad queries the model answered without a tool call. The tool loop's empty-response retry is gated on `iteration < max_tool_iterations - 1`, which is `0 < 0 == False` for the `minimal` tier (`MAX_TOOL_ITERATIONS=1`), so the retry never fired and the tool-less branch returned empty content without reaching the post-loop synthesis call. The empty string then surfaced as `null` (the chat endpoint maps a falsy answer to `None`). `low` and `medium` were unaffected because they allow `>= 2` iterations.

## Motivation

Broad `minimal` queries silently returned `null` even though the same query answered correctly at `low`/`medium`, so callers relying on `minimal` received no memory and no error signal.

## Changes

- `src/llm/tool_loop.py`: decouple empty-response recovery from the iteration budget. Keep the nudge-and-loop when iterations remain, otherwise break to the forced synthesis call so a final answer is always produced. This is a general tool-loop correctness fix; `minimal` is simply where it is guaranteed to surface. The synthesis prompt and log now distinguish empty-recovery from genuine max-iteration exhaustion, and the streaming path is left unchanged (`not stream_final`).
- `tests/llm/test_dialectic_minimal_empty_recovery.py`: deterministic regression test (mocked inner LLM) covering the single-iteration recovery path and the multi-iteration control.

## Testing

- `tests/llm/` full suite: 242 passed (new regression test RED to GREEN).
- `ruff check` / `ruff format` / `basedpyright` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty model responses at the end of the tool loop, ensuring the system recovers reliably even when no in-loop retry is possible.
  * Adds clearer recovery behavior to choose between synthesis-based recovery and retry-based recovery depending on remaining tool iterations.

* **Tests**
  * Added new isolation tests covering empty-string and `None` final responses, including single-iteration forced synthesis and multi-iteration nudge/retry behavior.

### DIFF
--- a/src/llm/tool_loop.py
+++ b/src/llm/tool_loop.py
@@ -404,6 +404,10 @@ async def execute_tool_loop(
     # DialecticCompletedEvent.hit_input_token_cap reflect the cap hit
     # (the toolless path tracks this in src/llm/api.py:325-340).
     hit_input_token_cap = False
+    # Set when an empty tool-less response can't be retried in-loop (no
+    # iterations left) and we break out to the forced synthesis call to
+    # recover a final answer instead of returning empty.
+    empty_recovery_synthesis = False
     # Track effective tool_choice — switches from "required"/"any" to "auto" after iter 1.
     effective_tool_choice = tool_choice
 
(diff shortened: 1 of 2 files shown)


karaokedjodua:
Finn-loop review of dd989ab

CI: not configured (only CodeRabbit runs; no required lint/test/typecheck checks on this branch)
Mergeability: UNKNOWN (mergeStateStatus=UNKNOWN)

## Review
PR #929 fixes a bug where `peer.chat(reasoning_level="minimal")` returned `null` on broad queries. The empty-response retry was gated on `iteration < max_tool_iterations - 1`, which is always false for the `minimal` tier (`MAX_TOOL_ITERATIONS=1`). The fix decouples empty-response recovery from the iteration budget — when no in-loop retry is possible, the loop breaks to the forced synthesis call instead of returning empty.

## 1. Must fix before merge
- **[CI]** No required CI checks are configured on this branch. Only CodeRabbit runs. There are no required lint (`ruff check`), typecheck (`basedpyright`), or test (`pytest`) checks blocking merge. The repository's CLAUDE.md documents `uv run ruff check src/`, `uv run basedpyright`, and `uv run pytest tests/` as required quality gates, but none are enforced via branch protection.

## 2. Should fix soon
None. The code changes are well-scoped:
- The empty-response check correctly expands to catch `response.content is None` (not just empty string), preventing a `None` pass-through.
- The `empty_recovery_synthesis` flag cleanly separates recovery vs. exhaustion in the post-loop synthesis, with distinct prompts and log messages.
- The streaming path (`stream_final`) is correctly gated — `empty_recovery_synthesis` is only set when `not stream_final`, leaving the empty tail to the stream itself.
- Tests cover empty string, `None`, repeat empty, and the in-budget nudge control path with a deterministic `_ScriptedLLM` mock.

## 3. Safe to merge
No — CI gate is missing. Automated review of the code itself finds no defects, but the lack of required CI checks means a human must verify that linting, type-checking, and tests pass before merging.

## Note
This reviewer's account does not have admin rights to this repository, so labels (`needs-human-review`) could not be applied. A repo admin should apply the `needs-human-review` label manually.

akattelu:
Closing as per comments in https://github.com/plastic-labs/honcho/issues/928

PR #373, opened 2026-02-06
feat: workspace-level chat

dr-frmr:
## Summary by CodeRabbit

* **New Features**
  * Workspace-level chat endpoint (POST /v3/workspaces/{id}/chat) with streaming and non‑streaming modes; SDKs (Python & TypeScript) expose chat and chatStream APIs with optional session scoping and reasoning levels
* **Documentation**
  * OpenAPI, API reference, SDK reference, and user guides updated with examples for workspace chat
* **Bug Fixes**
  * Enforced session‑membership visibility for message history/search to prevent pre‑join or out‑of‑session access
* **Tests**
  * Added extensive tests covering workspace chat, visibility rules, tools, and streaming behavior

PR #931, opened 2026-07-23
feat: Add workspace-level chat

adavyas:
## What

`POST /v3/workspaces/{workspace_id}/chat` — agentic dialectic over the **whole workspace** instead of a single (observer, observed) pair. Successor to #373, salvaged and re-grown on today's `DialecticAgent`.

Closes DEV-1326.

## Design

- **`WorkspaceDialecticAgent` subclasses `DialecticAgent`** through four small seams added to the base class (`_get_tools`, `_create_tool_executor`, `_prefetch_intro`, `_trace_name`) rather than #373's base-class extraction — the pair agent's hot path is untouched.
- **Routing-accelerated prefetch**: workspace stats + top-5 active peers + their self peer-cards, pure DB (~7ms measured, vs the pair agent's 2×25-doc semantic prefetch). The point: routing-obvious queries should resolve without a discovery tool round; each avoided round is a full model turn (~1.3s measured in the .chat latency study).
- **Observation search stays pair-scoped.** A workspace-flat top-k would be dominated by the most verbose peers and doesn't match the per-pair vector-store namespaces. Instead the workspace loadout's `search_memory`/`get_peer_card` take `observer`/`observed` as *tool arguments*, delegate to the existing pair handlers via a `handler_resolver` seam on `create_tool_executor`, and results carry `[observer->observed]` attribution so multi-pair turns don't conflate.
- **Session scoping**: optional `session_id` narrows message tools to one session. The `scope` read option from the scopes work (#897) has a marked seam in `WorkspaceChatOptions` and lands as a follow-up once the scopes facade merges.
- **SDKs**: Python `Honcho.chat` / `chat_stream` (+ `HonchoAio` mirrors), TypeScript `honcho.chat` / `chatStream`. SSE streaming and `response_format` structured output supported, mirroring `peer.chat`.

## Salvaged from #373 (co-authored)

crud `get_workspace_stats`/`get_active_peers`, workspace tool JSON defs, workspace system prompt, `format_documents_with_attribution`, test suite + unified test cases, docs page, TS SDK surface. #373's temporal-visibility fix is not needed — main already gates by session membership windows (`crud/message.py`).

## Verification

- `tests/test_workspace_chat.py`: 46/46 (route, orchestrator preflight session-release, tool handlers, executor routing, attribution formatting)
- Full suite: 1699 passed; only failures are `tests/crud/test_document.py` dedup tests that fail identically on clean main (pre-existing flake) — TS SDK build+typecheck pass
- Live e2e on a local rig (2 peers, 26 msgs, haiku-4.5): single-peer, cross-peer, unanswerable, session-scoped, and streaming all answer correctly through the new Python SDK methods
- Latency (local, embeddings disabled so memory is message-search only): prefetch 7–11ms; session-scoped answers **1.1–1.3s single model roundtrip**; workspace-routing questions 5.6–11.5s (multi-turn message-search spiral — the known embedding-less-rig artifact; embedded rigs resolve in fewer turns per the .chat latency study)

## Notes

- Endpoint auth mirrors `search_workspace` (`require_auth(workspace_name="workspace_id")`); Prometheus `record_dialectic_call` wired; Langfuse trace name `workspace_chat`, telemetry agent_type `workspace_dialectic`.
- Known pre-existing limitation (not this PR): conclusions require an OpenAI/Gemini embedding transport, so anthropic-only rigs run with inert observation memory — workspace chat degrades to message search there, same as peer chat.

## Summary by CodeRabbit

* **New Features**
  * Added workspace-wide chat to the Python and TypeScript SDKs and API.
  * Supports session and scope filtering, reasoning levels, structured responses, and streaming.
  * Added workspace statistics and recent peer activity insights.
  * Workspace chat can synthesize information across peers, messages, and observations with attribution.

* **Bug Fixes**
  * Improved input sanitization, validation, access controls, isolation, and empty-response handling.

* **Tests**
  * Added coverage for chat, streaming, analytics, tools, scopes, validation, isolation, and cross-peer synthesis.


src/schemas/api.py
```diff
@@ -610,6 +610,37 @@ def sanitize_query(cls, v: str) -> str:
...
+    response_format: dict[str, Any] | None = Field(
+        None,
+        description=(
+            "Optional JSON Schema (root type 'object') the response must conform"
+            " to. When provided, `content` is a JSON string matching this schema."
+        ),
+    )
+    # scopes(#897): the observer-swap `scope` read option lands here once the
+    # scopes facade merges; workspace chat is the peer-unanchored read and
+    # scope becomes its narrowing parameter.
```
akattelu: we could consider stacking this on the scopes PR (and potentially adding the scope field on workspace chat as well).

tests/conftest.py
```diff
@@ -806,11 +812,20 @@ async def _agentic_chat_response(*_args: object, **kwargs: object) -> str:
 
         mock_agentic_chat.side_effect = _agentic_chat_response
 
+        async def _workspace_chat_response(*_args: object, **kwargs: object) -> str:
```
akattelu, line 791: are the new unified tests working and running locally at least? I think we could also apply the label to this PR and verify in CI that they work and all.

src/dialectic/prompts.py
```diff
@@ -235,3 +235,78 @@ def agent_system_prompt(
...
+    observations to discover relevant peers rather than listing all of them.
+
+    Returns:
+        Formatted system prompt string for the workspace agent
+    """
+    return """
+You are a workspace-level analysis agent that can query memory across ALL peers in this workspace. You can synthesize information from any peer relationship's stored conclusions, insights, and conversation history.
+
+Unlike a peer-level agent that knows about one specific peer, you can search, compare, and correlate information about any and all peers — but you must discover relevant peers first and then query each peer relationship individually.
+
```
akattelu, line 254: I wonder if this isn't helpful because this agent won't really know that there's another peer-level agent right?

src/schemas/api.py
```diff
@@ -610,6 +610,37 @@ def sanitize_query(cls, v: str) -> str:
...
+
+    @field_validator("query", mode="after")
+    @classmethod
+    def sanitize_query(cls, v: str) -> str:
+        return v.replace("\x00", "")
```
akattelu: do we have other general sanitization methods that are more unified that we should be relying on?

src/utils/agent_tools.py
```diff
@@ -785,6 +785,79 @@ def _extract_pattern_snippet(
...
+        "name": "get_workspace_stats",
+        "description": "Get workspace-level statistics: peer count, session count, message count, and date range of messages. Use this to understand the scale of the workspace before deciding how to explore it.",
+        "input_schema": {
+            "type": "object",
+            "properties": {},
+        },
+    },
+    "get_active_peers": {
+        "name": "get_active_peers",
+        "description": "Get the most active peers in the workspace, ranked by recent activity or message count. Returns peer names with their message counts and last-active timestamps. Use this to discover which peers are most relevant.",
```
akattelu: should we remove this tool and add the content to stats instead? I think it could save a turn.

src/utils/agent_tools.py
```diff
@@ -2794,3 +2894,164 @@ def _estimate_tokens_safe(text: str | None) -> int | None:
     if not text:
         return None
     return _estimate_tokens(text)
+
+
+# ---------------------------------------------------------------------------
+# Workspace-level tool handlers (workspace chat, DEV-1326)
```
akattelu: can remove linear identifier and fix up the comment on this block

src/dialectic/workspace.py
```diff
@@ -0,0 +1,179 @@
...
+            session_name=self.session_name,
+            session_allowlist=self.session_allowlist,
+            history_token_limit=settings.DIALECTIC.HISTORY_TOKEN_LIMIT,
+            run_id=self._run_id,
+            agent_type="workspace_dialectic",
+            parent_category="dialectic",
+        )
+
+    def _trace_name(self) -> str:
+        return "workspace_chat"
```
akattelu: thoughts on just leaving this as dialectic, but using a different tag to indicate the scope level or something like that? also attaching the workspace identifier/etc could be useful.

akattelu (changes requested):

akattelu (dismissed):

src/dialectic/prompts.py
VVoruganti: Yeah looking at this now, it's probably worth a follow up to really re-vamp these system prompts, because this doesn't really give any context on what Honcho is, what the harness is, what peers are etc.

VVoruganti (approved):

ajspig (approved):

PR #935, opened 2026-07-24
feat(deriver): add message to conclusion source indices

vintrocode:
make the deriver indicate which messages a conclusion came from via source ids

## Summary by CodeRabbit

* **New Features**
  * Conclusions now show the message IDs cited as evidence, in citation order.
  * Added filtering for conclusions and documents by cited message ID.
  * Added CLI support for `conclusion list --cites <message_id>`.
  * Citations can include messages from other peers in the same batch.

* **Documentation**
  * Updated API, SDK, CLI, and MCP documentation with citation fields, filtering options, and legacy/derived conclusion behavior.

### DIFF
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -81,6 +81,7 @@ RULES:
 - Observations should make sense on their own. Each observation will be used in the future to better understand {peer_id}.
 - Extract ALL observations from {peer_id} messages, using others as context.
 - Contextualize each observation sufficiently (e.g. "Ann is nervous about the job interview at the pharmacy" not just "Ann is nervous")
+- Each message in the <messages> block is prefixed with a 0-based index like [0], [1], [2]. For each observation, set source_indices to the indices of the messages that directly support it. Include the message containing any context needed to interpret the conclusion (e.g., the question being answered by "the first one"). Only include messages that directly support the observation — not the entire conversation.
 
 EXAMPLES (using `{peer_id}` as the target peer id):
 - EXPLICIT: "I just had my 25th birthday last Saturday" → "{peer_id} is 25 years old", "{peer_id}'s birthday is June 21st"
--- a/src/deriver/prompts.py
+++ b/src/deriver/prompts.py
@@ -81,13 +81,15 @@ RULES:
 - Observations should make sense on their own. Each observation will be used in the future to better understand {peer_id}.
 - Extract ALL observations from {peer_id} messages, using others as context.
 - Contextualize each observation sufficiently (e.g. "Ann is nervous about the job interview at the pharmacy" not just "Ann is nervous")
-- Each message in the <messages> block is prefixed with a 0-based index like [0], [1], [2]. For each observation, set source_indices to the indices of the messages that directly support it. Include the message containing any context needed to interpret the conclusion (e.g., the question being answered by "the first one"). Only include messages that directly support the observation — not the entire conversation.
+- source_indices: Each message in the <messages> block is prefixed with a 0-based index like [0], [1], [2]. For each observation, set source_indices to the indices of the messages that directly support it. Source messages can be from ANY speaker — the user, an assistant, or another participant. What matters is which messages contain the evidence for the conclusion, not who said them. For example, if the assistant proposes a plan and the user confirms it, both messages are source material. Include the message containing any context needed to interpret the conclusion (e.g., the question being answered by "the first one"). Only include messages that directly support the observation — not the entire conversation.
 
 EXAMPLES (using `{peer_id}` as the target peer id):
 - EXPLICIT: "I just had my 25th birthday last Saturday" → "{peer_id} is 25 years old", "{peer_id}'s birthday is June 21st"
 - EXPLICIT: "I took my dog for a walk in NYC" → "{peer_id} has a dog", "{peer_id} lives in NYC"
 - EXPLICIT: "I went to college and then started working at the pharmacy" → "{peer_id} attended college", "{peer_id} works at the pharmacy"
 - EXPLICIT: "{peer_id} attended college" + general knowledge → "{peer_id} completed high school or equivalent"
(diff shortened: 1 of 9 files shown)


src/deriver/prompts.py
3un01a, line 1: This prompt deviates greatly from the original prompt.
Needs to be tested against the current Neuromancer model to see if any negative effects occur.

Maybe we should keep this prompt out from this PR and scope this PR only to the batch message.

matthewlanders: @vintrocode I know you did some testing. Good to remove the prompt from this PR or is it required for the other changes?

VVoruganti (commented):
So main thing to note here is that the `batch_message_ids` that were added as a new field in this draft would include also the IDs from the interleaved messages from other speakers. 

This means that an explicit conclusion for peer A can reference a message made by peer B in the conversation. Historically we have been authorative in saying a conclusion can only be made for a peer based on something the Peer has actually said, otherwise any claim by a separate peer could lead to nonsensical conclusions being made

src/deriver/prompts.py
matthewlanders: Note that if we do revert most of the prompt changes, we will still need to update the prompt to give the model context about `source_indices `.

matthewlanders:
> So main thing to note here is that the `batch_message_ids` that were added as a new field in this draft would include also the IDs from the interleaved messages from other speakers.
> 
> This means that an explicit conclusion for peer A can reference a message made by peer B in the conversation. Historically we have been authorative in saying a conclusion can only be made for a peer based on something the Peer has actually said, otherwise any claim by a separate peer could lead to nonsensical conclusions being made

A few things important to clarify:

1.) The context from which conclusions are derived hasn't changed — the deriver has always seen other speakers' messages as context ("Extract ALL observations from the target peer's messages, using others as context" is the pre-existing rule and the prompt is built from every message in the batch, each labeled with its speaker). This PR just makes the model declare which messages it used.

2.) A citation including another peer's message doesn't mean the conclusion rests on their claim. Specifically, the target peer's statements are often uninterpretable alone ("the first one" in response to "do you prefer Python or Rust"). Because the entailment judge only sees cited messages, the interpreting context has to be cited or valid conclusions fail. In both prompt examples the peer's own message is always among the sources.

If anything this makes the "only from what the peer said" stance enforceable for the first time — a conclusion citing zero target-peer messages is now mechanically detectable.

3.) Your comments made me realize that the design of `batch_message_ids` was kind of suboptimal. It was only ever a lookup table for resolving batch-relative indices, and persisting the full batch list per observation was not useful. I just pushed a small change that resolves citations at save time and stores `source_message_ids` — just the IDs of the cited messages. `source_indices` stays in storage as per-derivation debugging provenance only. The model still only ever sees positions, and the peer-filtered `message_ids` field is untouched.

src/models.py
```diff
@@ -472,6 +472,32 @@ def source_ids(self, value: list[str] | None) -> None:
             for i, sid in enumerate(dict.fromkeys(value or []))
         ]
 
+    # Messages the deriver cited as evidence for an explicit conclusion.
+    source_message_links: Mapped[list["DocumentSourceMessage"]] = relationship(
```
VVoruganti, line 476: **[P2] Preserve citation links when backfilling a scope**

The new relationship also needs to be carried through the scope-copy path. In `src/deriver/scope_backfill.py`, `_CopySpec`, `_run_backfill()`, and the destination `Document` construction copy `source_ids` but omit `source_message_ids`. Adding an existing session to a scope therefore loses the citations on every copied explicit conclusion. I reproduced a source conclusion with two cited messages returning `source_message_ids: null` through the API after backfill; filtering the scope's copies by either message also cannot find it. Please carry the cited message IDs through the copy plan and create their links on the destination document.

src/crud/document.py
```diff
@@ -1288,6 +1303,9 @@ def _document_model_from_create(
             session_name=doc.session_name,
             embedding=doc.embedding,
             source_links=build_source_links(doc.source_ids, workspace_name),
+            source_message_links=build_source_message_links(
+                doc.source_message_ids, workspace_name
+            ),
```
VVoruganti, line 1308: **[P2] Merge incoming citations when deduplication reinforces a conclusion**

Citation links are only written when `_document_model_from_create()` inserts a document. The always-on exact-deduplication path in `create_documents()` skips that helper for an existing match, and `_apply_document_row_updates()` only increments `times_derived`. When two messages in the same session independently produce the same conclusion, the second message's citation is silently discarded, even with `deduplicate=False`. I reproduced `times_derived == 2` while filtering by the second message's ID returned no conclusions. Please merge incoming citations into the retained document, preserving order and deduplicating IDs; the within-batch duplicate path needs the same treatment.

VVoruganti (commented):
Two citation-preservation issues, both reproduced with database-backed API tests. The existing 203 focused Python tests, 8 CLI tests, and migration test pass.

migrations/versions/c4d8e2f6a1b3_add_document_source_messages_table.py
```diff
@@ -0,0 +1,78 @@
...
+schema = get_schema()
+
+
+def upgrade() -> None:
+    connection = op.get_bind()
+    inspector = sa.inspect(connection)
+
+    if not table_exists("document_source_messages", inspector):
+        op.create_table(
+            "document_source_messages",
```
VVoruganti, line 37: torn on the approach here, but would like to move away from the document language if possible. Would like to rename this to `conclusion_source_messages`

src/crud/document.py
```diff
@@ -53,6 +53,21 @@ def build_source_links(
...
 
 
+def build_source_message_links(
+    message_ids: list[str] | None, workspace_name: str
+) -> list[models.DocumentSourceMessage]:
+    """Convert cited message public ids to DocumentSourceMessage rows, deduped."""
+    if not message_ids:
+        return []
+    return [
+        models.DocumentSourceMessage(
```
VVoruganti, line 63: Same thing if we can name this ConclusionSourceMessage

PR #898, opened 2026-07-11
feat(llm): add opt-in Responses API mode

dtownsel:
## Summary

- add opt-in Responses API mode to the existing OpenAI LLM backend
- convert chat history, system instructions, tool declarations, tool calls, and tool results to Responses input
- support strict Pydantic structured output, usage normalization, reasoning details, and streaming
- reconstruct output from streamed events for Codex-compatible endpoints whose `response.completed` event omits assembled output
- allow providers to opt out of `max_output_tokens` when their Responses endpoint rejects it
- preserve existing Chat Completions behavior unless `provider_params.api_mode` is explicitly `responses`

## Validation

- `pytest -n 0 tests/llm -q` — 231 passed
- focused OpenAI backend suite — 53 passed
- `ruff check` — passed
- `basedpyright src/llm/backends/openai.py` — 0 errors, 0 warnings
- live authenticated `gpt-5.6-luna` bridge:
  - strict structured output parsed successfully
  - required function tool call normalized successfully
  - streaming returned exactly `STREAM_OK`
- end-to-end Honcho attribution canary produced only the explicitly stated user preference and rejected quoted biography, assistant actions, Reid output, contradictory logs, and malformed tool fragments

## Compatibility

Responses mode is opt-in through model `provider_params`; existing OpenAI Chat Completions callers retain the original code path.

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAI’s Responses API for standard and streaming completions.
  * Supports structured JSON output, reasoning controls, tool calls, and prior tool interactions.
  * Preserves text, reasoning details, completion status, and token usage in responses.
  * Streaming now provides incremental text updates followed by a final completion event.

* **Tests**
  * Added coverage for Responses API requests, structured output, tool handling, token limits, and streaming.

PR #879, opened 2026-07-06
docs: adding codex doc

ajspig:
## Summary by CodeRabbit

* **New Features**
  * Added a new integration guide for using persistent memory with OpenAI Codex.
  * Included setup steps, configuration guidance, and troubleshooting tips for getting the integration working.
  * Documented options for shared workspaces, session behavior, and alternate installation from a local clone.

* **Documentation**
  * Updated the integrations navigation so the new guide appears in the versioned docs menu.

PR #937, opened 2026-07-25
fix: add a safe Codex compatibility relay

Space-Hermes:
Honcho normally talks to language models using the OpenAI chat format, while Codex uses a slightly different format. This pull request adds a small bridge between those two formats, so Honcho can use Codex without changing the way the rest of Honcho talks to models.

It also makes that bridge safer: it does not write or refresh credentials itself, requires a key when exposed beyond the local machine, handles streaming correctly, and avoids passing private error details back to callers.

## What this changes

- Adds a narrow Chat Completions → Codex Responses compatibility relay.
- Supports both ordinary and streaming responses, including usage information used by Honcho's current OpenAI backend.
- Stops cleanly when Codex has finished, rather than waiting indefinitely for a connection to close.
- Rejects unsupported or unknown request controls instead of silently ignoring them.
- Keeps the built-in credential source read-only; credential refresh remains the responsibility of the embedding application.
- Requires inbound authentication for non-loopback listeners.
- Sanitizes provider, credential, and malformed-stream errors.
- Adds focused documentation and regression coverage.

Existing Honcho deployments are unchanged unless this optional relay is started and configured.

## Why

Honcho's current OpenAI backend speaks the Chat Completions protocol. Codex exposes a Responses endpoint. Without a compatibility layer, the two sides cannot be used together reliably, especially for streaming and usage reporting.

This PR keeps the integration deliberately small and explicit rather than changing Honcho's general LLM abstraction.

## Related Honcho work

These existing PRs are relevant context and should be considered alongside this one:

- [#898](https://github.com/plastic-labs/honcho/pull/898) adds opt-in Responses API support directly to Honcho's OpenAI backend. That is an alternative/complement to this local compatibility boundary; maintainers may decide whether the approaches should converge.
- [#825](https://github.com/plastic-labs/honcho/pull/825) adds OpenAI/Codex OAuth and a direct Codex Responses backend. This PR intentionally does not duplicate OAuth or take ownership of credential refresh.
- [#879](https://github.com/plastic-labs/honcho/pull/879) is the merged Codex integration guide and provides useful user-facing background.

## Verification

All checks were run from the locked project environment:

- `uv run --frozen pytest tests/llm/test_codex_relay.py -q -n 0` — **49 passed**
- `uv run --frozen pytest tests/llm -q` — **289 passed**
- Current OpenAI SDK/backend integration — **1 passed**, not skipped
- Ruff — passed
- Python syntax compilation — passed
- Git diff checks — passed
- Bounded added-line secret scan — zero findings
- BasedPyright — 0 errors, 128 warnings; the nonzero exit is recorded as a warning-only repository result

I also ran two real smoke checks against the branch in a temporary local process:

- non-stream request — HTTP 200, valid completion, usage returned
- streaming request — HTTP 200, correct finish signal, usage received, `[DONE]` received

No production service was restarted or changed.

## Known limitation

The full repository test command remains limited by the local PostgreSQL fixture: the environment cannot authenticate/connect to `postgres@localhost:5432`. The relay-focused and complete LLM suites are green.

## Scope

The change is limited to:

- `docs/codex-relay.md`
- `src/codex_relay.py`
- `tests/llm/test_codex_relay.py`

I would appreciate feedback on whether the supported request boundary and the credential-ownership model are the right fit for Honcho upstream.

## Summary by CodeRabbit

* **New Features**
  * Added a local compatibility relay for OpenAI-style chat completion requests.
  * Supports streaming and non-streaming responses, tool calls, structured outputs, images, usage reporting, and health checks.
  * Added configurable authentication, credential handling, upstream connection settings, and server startup options.
  * Provides sanitized, OpenAI-compatible validation and error responses.

* **Documentation**
  * Added setup, configuration, security, supported controls, and testing guidance.

* **Tests**
  * Added comprehensive coverage for request translation, streaming, authentication, errors, credentials, and integration scenarios.

### DIFF
--- a/docs/codex-relay.md
+++ b/docs/codex-relay.md
@@ -48,17 +48,31 @@ authentication; it is not an upstream Codex OAuth token.
 Supported Chat controls are: messages, model, stream, tools, tool_choice
 (`none`, `auto`, `required`, or a named function), parallel_tool_calls,
 reasoning_effort, response_format (`json_object`/`json_schema`), and
-max_tokens/max_completion_tokens (translated to `max_output_tokens`). Conflicting
-output-limit aliases and invalid values are rejected. Controls that cannot be
-represented safely by Codex Responses (temperature, top_p, stop, penalties,
-seed, stream_options, logprobs, n, and user) are rejected with a clear 400 rather
-than silently dropped.
+max_tokens/max_completion_tokens (translated to `max_output_tokens`). The current
+Honcho OpenAI streaming backend's `stream_options: {"include_usage": true}` is
+accepted; the relay emits an OpenAI-compatible final usage chunk before `[DONE]`.
+Conflicting output-limit aliases and invalid values are rejected. Controls that
+cannot be represented safely by Codex Responses (temperature, top_p, stop,
+penalties, seed, logprobs, n, user, verbosity, reasoning, and SDK passthrough
+fields such as extra_body) are rejected with a clear 400 rather than silently
+dropped. Other unknown top-level fields and malformed supported fields are also
+rejected with a clear 400.
 
 Responses streams must end in `response.completed` or `response.incomplete`.
-Provider errors, malformed data, EOF/truncation, and transport failures are
-returned as sanitized errors; an already-started downstream stream emits one
+The relay stops reading immediately after the first valid terminal event and
+closes the upstream response. Provider HTTP error bodies are sanitized to a
+stable OpenAI-compatible error while preserving the upstream status code;
+credential, malformed data, malformed usage, EOF/truncation, and transport
+failures are sanitized as well. An already-started downstream stream emits one
 error event followed by one `[DONE]`. Incomplete content-filter results map to
 `content_filter`; other incomplete results map to `length`.
+
+The credential document must contain a mapping at
+`providers.openai-codex.tokens.access_token` with a non-empty string token.
+Missing, corrupt, or malformed documents and non-ASCII/empty transport tokens
+return a sanitized 503 credential error. Malformed or opaque JWT claims are
+never decoded into headers or reflected in errors; if the upstream rejects such
+a token, its HTTP error is sanitized.
 
 ## Reproducible checks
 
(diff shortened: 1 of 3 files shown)


Space-Hermes:
Thanks for the review. I addressed the actionable stability and cleanup points in commit `580fb807`:

- non-ASCII Bearer values now return the normal 401 response instead of raising;
- empty inbound keys are treated as unconfigured on loopback and rejected for non-loopback binds;
- unused OAuth constants and redundant test-local imports were removed;
- the response-tool type and shadowing terminal variable were cleaned up.

The focused relay suite now passes 51 tests and the adjacent LLM suite passes 291 tests. I left the broader SSE-parser consolidation and blanket docstring request out of this PR because they are unrelated refactors; the existing parser behavior is covered by the passing regression suite.

akattelu:
Hey @Space-Hermes, thanks for opening this. The relay is carefully written (read-only credentials, loopback default, sanitized errors), and I can see how it is meant to sit outside Honcho's OAuth path rather than re-implement #825.

I'm going to close it without merging for the same policy reason as #825. The upstream this hits is still `https://chatgpt.com/backend-api/codex/responses`, with a `User-Agent` / `originator` of `codex_cli_rs` and Hermes ChatGPT OAuth tokens. OpenAI does not document that surface as a supported third-party integration for product backends. Their docs cover Sign in with ChatGPT for first-party Codex clients ([Codex auth](https://developers.openai.com/codex/auth)), and the Platform API for product backends. Consumer/Pro language also calls out [programmatic extraction of Output](https://openai.com/policies/row-terms-of-use/) and [using ChatGPT to power third-party services](https://help.openai.com/en/articles/9793128-about-chatgpt-pro-tiers), which is uncomfortably close to how dialectic + deriver would use a subscription token through this relay.

Shipping it in-tree, even as an optional local process, still puts that path on Honcho's first-party surface. Supported path remains an OpenAI Platform API key (or another documented provider). If the goal is Responses-format compatibility against a supported endpoint, #898 is the closer shape. If OpenAI ever publishes a real third-party contract for subscription-backed Codex Responses access, happy to revisit.

PR #945, opened 2026-07-27
fix: validate source_ids against real documents on the write path

### LINKED ISSUE #939: [Bug] Dreamer specialists cite source_ids that don't exist. No validation on the write path
# **🐞 Bug Report**

## **Describe the bug**

Dream specialists cite their evidence via `source_ids`, but **nothing validates that a cited id
refers to a document that actually exists.**

* `validate_level_fields` (`src/schemas/internal.py`) checks only that the list is non-empty.
* The tool JSON schema checks only `minItems`.
* `create_observations` normalizes the ids (stripping an `id:` prefix) and persists them unchecked.

A fabricated id therefore becomes **permanent false provenance that passes every existing check**. A
conclusion appears sourced when it is not, `get_children_of_document` tree traversal breaks, and on
v3.0.12+ conclusion dating breaks too - `_latest_source_timestamp` (added in #890) silently falls
back to "now" when nothing resolves, so a dangling citation degrades into a wrong timestamp rather
than an error.

---

### **Is this a regression?**

No. This is a missing validation rather than a behaviour that broke. It appears to have been
present since the dream specialists were introduced. Confirmed on **v3.0.11**, and both root-cause
sites are byte-identical on `main` as of 2026-07-26, so there is no version to wait for.

---

### **To Reproduce**

1. Self-host Honcho with `DREAM_ENABLED=true` and let dream specialists write deductive/inductive
   documents. Either wait for a natural cycle or force one with
   `POST /v3/workspaces/{workspace}/schedule_dream` (returns `204`).
2. Read back the dream-written documents together with their `source_ids`.
3. Resolve every cited id against the documents table — the same lookup `fetch_documents_by_ids`
   performs.
4. Observe that a consistent fraction of cited ids resolve to **no document at all**. In our data
   this ran at ~7% overall and **15% for the inductive specialist** (full table under *Additional
   context*).
5. Note that none of these documents failed validation — they passed `validate_level_fields`, the
   tool schema's `minItems`, and were persisted normally.

The signature case in our data is an inductive document whose `source_ids` contained the literal
string:

```
"r****: Stage a proposal for the Zig version note..."
```

- a formatted message line pasted straight into the id field.

---

### **Expected behaviour**

`source_ids` should be resolved against real documents **before** persistence:

* Fabricated ids are stripped.
* An observation left with **no real source** is rejected and surfaced as an `ObservationFailure`,
  not silently stored. A "deduction" citing nothing real is worse than no deduction.
* An observation retaining ≥1 real source is stored with the fabricated ids removed.

Additionally, the specialists should be **told** that message search results carry no id and must
not be cited, since that is what creates the pressure to invent one (see *Additional context*).

---

### **Media prove**

N/A — server-side/REST behaviour, no UI involved. The quantitative evidence is the citation
resolution table under *Additional context*.

---

### **Your environment**

* OS: Ubuntu 24.04.4 LTS (Linux 7.0.0-28-generic x86_64)
* Browser name and version: N/A — direct REST calls to the `/v3` API
* Honcho Server Version: **v3.0.11** (`60a15e66`), self-hosted via docker compose. Root-cause sites
  also verified present on `main` 2026-07-26.
* Honcho Client Version: N/A - no SDK; the consuming agent calls the HTTP API directly

---

### **Additional context**

**Root cause.** Both specialists are given `search_messages`. Its results are raw message text with
**no id** (`_format_message_snippets` → `format_new_turn_with_timestamp`). Observations, by
contrast, render as `[id:xxx]`. Induction reasons largely over messages, so **the ids it is asked to
cite do not exist anywhere in its context**, while the schema requires it to produce some. The model
fills the gap. This also explains why induction is consistently the worse offender.

**Messages are not citable even in principle.** `source_ids` feeds `fetch_documents_by_ids`, so a
message id would dangle exactly like a fabricated one. Exposing message ids is therefore *not* the
fix — that was the first direction we considered and it was wrong.

**Measured evidence.** Every citation in every dream document ever written on one deployment,
resolved against the DB (2026-07-22):

| era | specialist | citations | invented | rate |
|---|---|---|---|---|
| local 30B | deductive | 193 | 3 | 1.6% |
| local 30B | inductive | 153 | 23 | **15.0%** |
| hosted frontier model | deductive | 614 | 46 | 7.5% |
| hosted frontier model | inductive | 676 | 46 | 6.8% |

**Overall 7.5% local vs 7.1% hosted.** The rate is essentially identical across a local 30B
(Qwen3-30B-A3B) and a hosted frontier model, so this is **model-independent** - a prompt/schema
contract problem, not a weak-model artifact.

**A second pattern appears once invention is blocked.** With a guard in place the model still has to
satisfy the tool schema's `minItems: 2`, and does so by **repeating one real id**:

```json
"source_ids": ["uhWNamYDRpBW07AQcb2h8", "uhWNamYDRpBW07AQcb2h8", "uhWNamYDRpBW07AQcb2h8"]
```

This passes a grounding check while being effectively single-sourced. A *distinct*-count check would
catch it.

**What worked for us.** Prompt changes alone reduce the rate but cannot guarantee it, so we used
three prompt changes plus one enforcement point:

1. `_format_message_snippets` - header states results are raw messages with no id, must not be cited
   in `source_ids`, and belong in `sources`/`premises` instead.
2. `source_ids` field description - teach the `[id:xxx]` convention and a copy-exactly rule,
   mirroring what `delete_observations` already does.
3. Deduction + induction prompt RULES - state the same rule where the model actually reads it.
4. `_filter_ungrounded_source_ids`, called from `create_observations` - the guarantee that holds
   regardless of model behaviour, implementing the *Expected behaviour* above.

We set the keep threshold at ≥1 real source to match `validate_level_fields` rather than the tool
schema's `minItems: 2`, deliberately choosing the smaller behaviour change.

**Implementation trap worth flagging:** collect the ids **inside** the session block. Our first
version gathered them after the ORM objects had expired and the refresh had no session. Caught by a
unit test before deployment.

shixi-li:
## Problem

Dreamer specialists can cite `source_ids` that don't exist. `validate_level_fields` only checks that `source_ids` is non-empty, and `create_observations` normalizes the ids (stripping the `id:` prefix) and persists them with no existence check, while `_latest_source_timestamp` silently skips unresolvable ids. The result is observations carrying fabricated citations that pass every validation layer (7–15% of citations in the issue's production sample).

Fixes https://github.com/plastic-labs/honcho/issues/939

## Fix

Following the issue's suggested direction, grounding is enforced at the single persistence chokepoint:

- `_filter_ungrounded_source_ids` (new) resolves every cited id via `crud.fetch_documents_by_ids` in a read-only `tracked_db` scope before persistence, collecting the resolved id set inside the session block. Unresolvable ids are stripped; an observation left below its level's existing minimum (≥1 real source; ≥2 for contradiction, consistent with `validate_level_fields`) is rejected and surfaced through the existing `ObservationFailure` / `Failed N: ...` response path.
- Prompt/schema contract updates teach specialists that `search_messages` results are raw messages with no citable observation id: the `_format_message_snippets` header, the `source_ids` field descriptions (mirroring the exact-`[id:xxx]` copy rule already used by `delete_observations`), and one added rule each in the deduction and induction specialist RULES.

Deliberately not enforced (per the issue's own scoping): distinct-count on repeated single real ids — flagged there as a pattern to watch, not a rejection rule.

## Tests

- Three new regression tests (written first, failing on the unfixed code): fully-fabricated `source_ids` are rejected as `ObservationFailure`; fabricated ids are stripped when a real source remains; a mixed batch keeps grounded observations and fails ungrounded ones.
- Two existing tests asserted the buggy persist-fabricated-id behavior and now cite real fixture documents.
- `tests/utils` + tool-schema + dreamer suites: 148 passed. Full suite green apart from failures present on the base commit in a clean worktree (environment-dependent crud dedup counts and lancedb imports). `ruff` and `basedpyright` clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved citation/provenance validation for derived observations by requiring `source_ids` to exactly match `[id:xxx]` tool results.
  * Discarded fabricated or unresolvable `source_ids` before persistence, and now rejects/flags observations when insufficient grounded sources remain (including contradiction minimum-source rules).
  * Updated guidance to clarify that message search results contain no IDs and cannot be used for provenance.

* **Tests**
  * Updated provenance tests to use seeded real document IDs and added coverage for rejection, partial acceptance, contradiction constraints, and mixed batches.
  * Added/expanded assertions for `id:`/`ID:` prefix stripping.

### DIFF
--- a/src/dreamer/specialists.py
+++ b/src/dreamer/specialists.py
@@ -611,10 +611,11 @@ Use `create_observations_deductive`.
 1. Don't explain your reasoning - just call tools
 2. Create observations based on what you ACTUALLY FIND, not what you expect
 3. Always include source_ids linking to the observations you're synthesizing
-4. Empty or missing source_ids will be rejected
-5. Delete outdated observations - don't leave duplicates
-6. Quality over quantity - fewer good deductions beat many weak ones
-7. When you are finished, do not output a summary of what you did - output only the token DONE"""
+4. Copy source_ids exactly from the [id:xxx] shown in observation results - `search_messages` results have no ID and cannot be cited; invented IDs are discarded
+5. Empty or missing source_ids will be rejected
+6. Delete outdated observations - don't leave duplicates
+7. Quality over quantity - fewer good deductions beat many weak ones
+8. When you are finished, do not output a summary of what you did - output only the token DONE"""
 
     def build_user_prompt(
         self,
@@ -745,8 +746,9 @@ Use `create_observations_inductive`.
 3. Confidence based on evidence count: 2=low, 3-4=medium, 5+=high
 4. Look for HOW things change over time, not just static facts
 5. Include source_ids - always link back to evidence
-6. Empty or missing source_ids will be rejected
-7. When you are finished, do not output a summary of what you did - output only the token DONE"""
+6. Copy source_ids exactly from the [id:xxx] shown in observation results - `search_messages` results have no ID and cannot be cited; invented IDs are discarded
+7. Empty or missing source_ids will be rejected
+8. When you are finished, do not output a summary of what you did - output only the token DONE"""
 
     def build_user_prompt(
         self,
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -106,7 +106,11 @@ def _base_observation_properties() -> dict[str, Any]:
             "description": (
                 "Document IDs of source or premise observations. Required and "
                 + "must be non-empty for deductive, inductive, and contradiction "
-                + "observations."
+                + "observations. Copy the exact ID shown in [id:xxx] format from "
+                + "observation results (e.g. for '[id:abc123XYZ]' pass "
+                + "'abc123XYZ'). Message search results carry no ID and cannot "
+                + "be cited here; IDs that do not match an existing observation "
+                + "are discarded."
             ),
         },
         "premises": {
@@ -223,7 +227,7 @@ def _deductive_observation_item_schema() -> dict[str, Any]:
                 "type": "array",
                 "items": {"type": "string"},
                 "minItems": 1,
-                "description": "Required non-empty list of source observation IDs supporting the deduction",
+                "description": "Required non-empty list of source observation IDs supporting the deduction. Copy the exact ID shown in [id:xxx] format from observation results; message search results carry no ID and cannot be cited",
             },
             "premises": {
                 "type": "array",
@@ -249,7 +253,7 @@ def _inductive_observation_item_schema() -> dict[str, Any]:
                 "type": "array",
                 "items": {"type": "string"},
                 "minItems": 2,
-                "description": "Required list of at least two source observation IDs supporting the pattern",
+                "description": "Required list of at least two source observation IDs supporting the pattern. Copy the exact ID shown in [id:xxx] format from observation results; message search results carry no ID and cannot be cited",
             },
             "sources": {
                 "type": "array",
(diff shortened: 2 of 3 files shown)


shixi-li:
Rebased this branch onto current `main` (`e7cbcc8`); the new head is `2131e6d`. Git replayed both commits cleanly, and `git range-diff` reports both patches unchanged. I reviewed the overlap in `src/utils/agent_tools.py` and `tests/utils/test_agent_tools.py`: the upstream session-allowlist changes and this PR's `source_ids` grounding behavior are both preserved, including the CodeRabbit-requested contradiction test for the two-real-source threshold.

Validation:

- `uv run pytest -q tests/utils/test_agent_tools.py tests/dreamer` — 148 passed
- remaining locally runnable suite — 1,749 passed, 21 skipped (excluding two LanceDB files because the project does not install LanceDB on macOS x86, and the TypeScript SDK wrapper because Bun is unavailable locally)
- Ruff check for the configured core Python paths — passed
- Ruff format and basedpyright for all three changed files — passed
- pre-commit hooks for all three changed files — passed, including Ruff, Ruff format, Bandit, and basedpyright
- `uv build` — sdist and wheel built successfully

The new-head FastAPI workflow is waiting for maintainer approval rather than failing: https://github.com/plastic-labs/honcho/actions/runs/30440189093

Could a maintainer please approve that workflow and review the rebased PR when convenient?

akattelu (approved):

eisene (approved):

PR #949, opened 2026-07-28
fix: increase throughput of unit tests by changing behavior db teardown

VVoruganti:
## Summary by CodeRabbit

* **Tests**
  * Improved parallel test-run isolation by assigning unique temporary databases and cleaning up stale databases automatically.
  * Updated test database cleanup to respect foreign-key dependencies without resetting generated IDs.
  * Enhanced teardown to remove temporary databases after test runs.

* **Maintenance**
  * Updated cost calculations to use the current model configuration.
  * Improved type annotations and clarified error-handling and test comments.

### DIFF
--- a/scripts/dialectic_cost_calculator.py
+++ b/scripts/dialectic_cost_calculator.py
@@ -170,10 +170,10 @@ def calculate_level_cost(
         realistic_final_answer=realistic_final,
     )
 
-    model = level_config.MODEL
+    model = level_config.MODEL_CONFIG.model
     max_iterations = level_config.MAX_TOOL_ITERATIONS
-    thinking_budget = level_config.THINKING_BUDGET_TOKENS
-    provider = level_config.PROVIDER
+    thinking_budget = level_config.MODEL_CONFIG.thinking_budget_tokens or 0
+    provider = level_config.MODEL_CONFIG.transport
 
     # Get pricing for this model
     pricing = MODEL_PRICING.get(model, {"input": 0, "output": 0, "cached": 0})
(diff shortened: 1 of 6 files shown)


Rajat-Ahuja1997 (dismissed):

eisene (commented):
I had Claude benchmark this and it concluded that this is the fastest possible, so 👍 

Two issues probably worth fixing from Claude's review:

**1. [Medium] Orphaned databases now accumulate forever.** This is the real cost of the run-id naming. The old name was fixed, so a crashed run's `test_db_gwN` was *reused* and cleaned by the `drop_all`/`create_all` in `db_engine` — self-healing. Now every run mints a new name, and any run that dies before fixture teardown (SIGKILL, IDE stop button, OOM'd worker, `-x` on a hang) leaks a ~9 MB database nothing will ever reclaim. Verified: SIGKILL mid-run left `test_db_leakprobe3` behind, still there 4 s later and after a subsequent full run.

The fix is a sweep at session start, but **don't write the naive version** — a blanket `DROP` of `test_db_%` would kill a concurrently running suite, which is exactly what this PR exists to enable. A `NOT EXISTS (SELECT 1 FROM pg_stat_activity WHERE datname = …)` guard covers most of it but races with the window between `create_database` and the first worker connection. Options, roughly in order of how much I'd trust them: gate the sweep behind an explicit `--clean-test-dbs` flag; or encode a sortable timestamp in the DB name and only drop ones older than the current run. Either way use `DROP DATABASE … WITH (FORCE)` (pg13+, both dev and CI are pg15) — that *also* fixes the pre-existing teardown failure where `drop_database` errors out because a pooled connection lingers.

**2. [Low] `_truncate_all_tables` no longer truncates.** Rename it (`_clear_all_tables`). The name is actively misleading now that the docstring's whole point is the deliberate choice *not* to TRUNCATE.

eisene (approved):
lgtm

PR #950, opened 2026-07-28
fix: make OpenAPI servers list configurable via API_BASE_URL env var [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #875: [Bug] Self-hosted: OpenAPI servers list hardcodes http://localhost:8000, not configurable
# **🐞 Bug Report**

## **Describe the bug**

When self-hosting Honcho (per the [self-hosting guide](https://honcho.dev/docs/v3/contributing/self-hosting)), the FastAPI app declares a fixed servers list in [src/main.py](https://github.com/plastic-labs/honcho/blob/v3.0.11/src/main.py#L165-L170):

```
app = FastAPI(
    ...
    servers=[
        {"url": "https://api.honcho.dev", "description": "Production SaaS Platform"},
        {"url": "http://localhost:8000", "description": "Local Development Server"},
    ],
    ...
)
```

This is baked into the OpenAPI schema at code-definition time and drives the "Servers" dropdown in the Swagger UI at /docs. For any self-hosted deployment that isn't reachable at exactly localhost:8000 (different host port mapping, reverse proxy, LAN hostname, etc.), the dropdown never matches reality. clicking "Try it out" sends requests to the wrong place.

**Impact:** Cosmetic/convenience only — direct API calls (SDKs, curl, other clients) are unaffected. But Swagger's "Try it out" is effectively broken for any self-hosted deployment not on localhost:8000 exactly.

**Workaround:** we patch src/main.py's servers list at Docker build time in our own self-hosting setup. 

---

### **Is this a regression?**

Unsure; I hit this immediately on my first attempt to get the self-hosted honcho running.

---

### **To Reproduce**

1. Self-host per the guide, mapping the API container's port 8000 to a different host port (e.g. host `8070` → container `8000`).
2. Visit `http://<host>:8070/docs`.
3. Open the "Servers" dropdown — only `https://api.honcho.dev` and `http://localhost:8000` are offered; neither matches the actual deployment.

---

### **Expected behaviour**

- Show the correct "localhost" URL in the Swagger UI. 
- The actual URL where the Swagger UI is available as an option from the "Try it out" servers dropdown.
 
Suggested fix: 
- Make it configurable via an env var (e.g. API_BASE_URL), falling back to the current hardcoded values if unset.

---

### **Media prove**

---

### **Your environment**

* Honcho Self-Hosted Version: v3.0.11

---

### **Additional context**

*

waterWang:
## Description

When self-hosting Honcho, the FastAPI app hardcodes a fixed servers list in the OpenAPI schema. This causes the Swagger UI 'Servers' dropdown to show only 'https://api.honcho.dev' and 'http://localhost:8000', neither of which matches a self-hosted deployment on a different host/port.

## Changes

1. **src/config.py**: Added `API_BASE_URL: str | None = None` to `AppSettings`. When set via the `API_BASE_URL` env var, it overrides the servers list in the OpenAPI schema.
2. **src/main.py**: Added `_build_servers()` helper that reads `settings.API_BASE_URL`. When set, returns a single-entry servers list with the self-hosted URL. Falls back to the current hardcoded values (production + localhost) when unset.

## Testing

- **Without API_BASE_URL**: Swagger UI shows the same servers as before (backward compatible).
- **With API_BASE_URL=http://192.168.1.50:8070**: Swagger UI shows only that server URL.

Closes https://github.com/plastic-labs/honcho/issues/875

## Summary by CodeRabbit

* **New Features**
  * Added an optional configuration setting to customize the API base URL.
  * Swagger UI now displays the configured self-hosted URL as the sole server option when provided.
  * Existing production and local development server options remain available by default.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -1446,6 +1446,11 @@ class AppSettings(HonchoSettings):
         "https://api.honcho.dev",
     ]
 
+    # Override the OpenAPI "Servers" dropdown in Swagger UI.
+    # When set, the self-hosted URL is used as the only server entry.
+    # Falls back to the default production + localhost entries when unset.
+    API_BASE_URL: str | None = None
+
     COLLECT_METRICS_LOCAL: bool = False
     LOCAL_METRICS_FILE: str = "metrics.jsonl"
     REASONING_TRACES_FILE: str | None = None  # Path to JSONL file for reasoning traces
--- a/src/main.py
+++ b/src/main.py
@@ -135,12 +135,21 @@ async def lifespan(_: FastAPI):
         await shutdown_telemetry()
 
 
-app = FastAPI(
-    lifespan=lifespan,
-    servers=[
+def _build_servers() -> list[dict[str, str]]:
+    """Build the OpenAPI servers list from settings, falling back to defaults."""
+    if settings.API_BASE_URL:
+        return [
+            {"url": settings.API_BASE_URL, "description": "Self-hosted deployment"},
+        ]
+    return [
         {"url": "https://api.honcho.dev", "description": "Production SaaS Platform"},
         {"url": "http://localhost:8000", "description": "Local Development Server"},
-    ],
+    ]
+
+
+app = FastAPI(
+    lifespan=lifespan,
+    servers=_build_servers(),
     title="Honcho API",
     summary="The Identity Layer for the Agentic World",
     description="""Honcho is a platform for giving agents user-centric memory and social cognition.""",


src/config.py
```diff
@@ -1446,6 +1446,11 @@ def langfuse_exporter_enabled(self) -> bool:
         "https://api.honcho.dev",
     ]
 
+    # Override the OpenAPI "Servers" dropdown in Swagger UI.
+    # When set, the self-hosted URL is used as the only server entry.
+    # Falls back to the default production + localhost entries when unset.
+    API_BASE_URL: str | None = None
```
VVoruganti, line 1452: Agree with the coderabbit comment here. Add a pydantic validator to check for valid urls in this field

VVoruganti (changes requested):

akattelu:
Closing for staleness. Note that #875 is the valid issue for it and we would prefer a narrow fix that targets it.

PR #952, opened 2026-07-29
feat: exposing conclusion attribution (source & times derived)

ajspig:
build conclusion DAGs via API instead of direct DB access. 

- `source_ids` and `times_derived` on the Conclusions response schema. Both already existed on documents; just stripped from API responses. `source_ids` reads the column, falling back to legacy`internal_metadata.source_ids` for pre-column rows
- New GET /conclusions/{conclusion_id} (single conclusion with attribution) 

- `source_ids` now filterable via JSONB containment(`{"source_ids": {"contains": "<id>"}}`); previously generated invalid ILIKE-on-JSONB SQL. Same fix applies to `contains` on other JSONB columns. `id`, `level`, `source_ids`, `times_derived` are now explicit entries in the documents filter allowlist.
- Python & Typescript SDK: fields on Conclusion, plus scope.get(conclusion_id) (sync + async). 

(Message attribution (conclusion -> source messages) and chain traversal is a follow-up)

## Summary by CodeRabbit

* **New Features**
  * Added APIs and SDK methods to retrieve individual conclusions, fetch conclusions in batches, and paginate derived conclusions.
  * Added conclusion attribution details, including source IDs and derivation counts.
  * Added filtering by source IDs and derivation metadata.
* **Bug Fixes**
  * Improved reasoning-chain traversal, ordering, source-link handling, and exclusion of malformed references.
  * Preserved valid source relationships during database migration.


docs/v3/openapi.json
```diff
@@ -2456,6 +2495,86 @@
         }
       }
     },
+    "/v3/workspaces/{workspace_id}/conclusions/{conclusion_id}/derived": {
```
akattelu: Have we considered supporting this via filters on the list conclusions API call, instead of adding a new one? 
```http
POST /workspaces/{workspace_id}/conclusions/list

{
  "filters": {
    "parent_id": "{conclusion_id}"
  }
}
```

src/utils/filter.py
```diff
@@ -31,7 +33,7 @@
 
 # JSONB columns keep containment semantics: bare lists are not membership
 # sugar, and dict values map to nested-metadata conditions rather than IN/Eq.
-JSONB_COLUMNS = ("h_metadata", "configuration", "internal_metadata")
+JSONB_COLUMNS = ("h_metadata", "configuration", "internal_metadata", "source_ids")
```
akattelu: Not fully following all the code related to this part, but wondering if this should just normalize this to a separate table instead of keeping it in json. It could be faster for filtering depending on how we're going about it.

src/routers/conclusions.py
```diff
@@ -131,6 +131,62 @@ async def query_conclusions(
...
+)
+async def get_conclusion(
+    workspace_id: str = Path(...),
+    conclusion_id: str = Path(...),
+    db: AsyncSession = read_db,
+) -> schemas.Conclusion:
+    """
+    Get a single Conclusion by ID.
+    """
+    documents = await crud.get_documents_by_ids(
```
akattelu, line 147: We could consider leaving this endpoint out, since users can call list with a filter for a single conclusion id (this `crud` call is doing the same thing).

akattelu (dismissed):

akattelu (dismissed):

eisene (commented):
this is stale and there's some issues (see claude-ish below) but otherwise lgtm, i can approve once merge is resolved

---

**1.** `src/deriver/scope_backfill.py` **— main's** `load_only` **no longer compiles against the new** `source_ids`

Main's #1104 ("reduce scope backfill memory usage") added `load_only(..., models.Document.source_ids, ...)` at two sites in `_run_backfill`. This PR converts `source_ids` from a column into a `@property` over `source_links`. Verified:

```
ArgumentError: expected ORM mapped attribute for loader strategy argument
```

That's raised at query *construction*, so every scope-backfill pass 500s. Keeping both sides of the conflict is exactly the wrong resolution. Drop `source_ids` from both `load_only()` lists — the `lazy="selectin"` loader on `source_links` still fires under `load_only`, so `_CopySpec.source_ids` (scope_backfill.py:226) keeps working.

**2.** `src/crud/document.py` **— carry the conversion into main's rewritten** `create_documents`

Main's #1033 restructured `create_documents` (semantic-dup pre-resolution, `_DocumentRowOp` reinforce/replace). The two `models.Document(...)` sites this PR changed to `source_links=build_source_links(...)` moved to main:1259/1272 — both need the conversion. If either reverts to `source_ids=doc.source_ids` it silently routes through the property setter instead of the validating helper (see #3). Main's other two sites (1103/1115, in `create_observations`) are explicit-level and set no linkage — leave them. Main does still use ORM `db.add_all`, so the relationship is honored.

akattelu (dismissed):

akattelu (approved):

PR #970, opened 2026-08-03
Fix: vector query top k zero

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid document searches with zero or negative result limits now return safely without unnecessary processing.
  * Semantic searches now always allocate space for at least one relevant result, even with small observation limits.

* **Tests**
  * Added coverage for invalid result limits and constrained semantic search budgets.


eisene (approved):
lgtm

PR #974, opened 2026-08-04
feat(mcp): search conclusions via search tool

### LINKED ISSUE #911: [Enhancement]: Add user-facing tool surface to manage stored Explicit Observations
# Honcho: Add user-facing tool surface to manage stored Explicit Observations

## Summary

The user-facing Honcho tooling (the `conclude` API + the corresponding MCP tools exposed by hermes-agent at `plugins/memory/honcho/`) only reaches the `conclusions` table. There is no equivalent for the `observations` table, which means a user cannot inspect, delete, or correct stored Explicit Observations once written. In practice this lets debugging residue and stale facts accumulate indefinitely and become self-reinforcing re-assertion material for downstream models.

## Repro

A user runs through a debugging session that produces ~50 `hermes says X` Explicit Observations in the AI Self-Representation. Most are descriptive ("hermes said the User Peer Card was clean at 4:53:39 AM"), but the model in any future session reads them as authoritative claims about itself and re-asserts them. To clean up the residue, the user reaches for `honcho_conclude list` and `honcho_conclude delete` — these work for `conclusions` rows but do not touch `observations`. Verified empirically:

- `honcho_conclude list=true peer="user"` returns ~20 conclusions (mix of load-bearing and resolvable).
- `honcho_conclude delete_id=<ID>` removes individual conclusions.
- `honcho_search peer="ai" query="Vee"` returns ~20 conclusions referencing "Vee" — all of which can be deleted with `delete_id`.
- `honcho_search peer="ai" query="Vee"` ALSO returns ~30+ `[2026-07-18 ...] hermes says X` Explicit Observations, none of which are reachable by `delete_id` because they live in a separate storage path.

The user has no way to clean the observations without direct DB access or a Honcho admin endpoint.

## Proposed API

Add parallel `observation` operations to whatever scope the user-facing `conclude` API exposes today. Concretely, in the Python client (`honcho/client.py` and `honcho/peer.py`):

- `Peer.list_observations(target: str | None = None) -> list[Observation]` — fetch observations by peer with optional target filtering
- `Peer.delete_observation(observation_id: str) -> bool` — delete a specific observation
- `Peer.flag_observation(observation_id: str, *, historical: bool = True) -> Observation` — mark an observation as `historical=True` so the representation layer can weight it lower (or surface it with an explicit `[historical]` marker)

The `historical` flag is the high-leverage addition: it lets the user say "this observation was true at the time it was written but isn't authoritative now" without losing the record. The model then sees the observation with explicit context that it's a historical, not live, claim — and stops re-asserting it as present-tense fact.

The MCP tool layer in hermes-agent (`plugins/memory/honcho/`) needs the same three operations exposed.

## Use case that drove this

The hermes-agent install at `NousResearch/hermes-agent` is configured with `memory.provider: honcho`, peer `austin`, with both local MEMORY.md/USER.md files and external Honcho memory active by design. During the 2026-07-17/18 debugging of an M3-thinking-reasoning patch and a Telegram prompt-injection report, the AI peer accumulated ~50 Explicit Observations about a "Vee" prompt-injection bug that was actually upstream-stale-context-cache and self-trust-loop, not a live injection. The Observations now sit in Honcho indefinitely, can't be deleted through the public API, and re-trigger the bad context in any future session that touches this peer.

The hermes-side workaround landed in NousResearch/hermes-agent PR #66754 (renders imperative-shaped lines from peer-card data into a labeled `[untrusted injection filtered from <section>]` block). That closes the user-peer-card injection vector but does NOT clean the AI-peer observations — they remain as pollution in Honcho.

## What's not in scope

- Changing the underlying observation storage format. Just exposing the existing table to the user-facing API.
- Adding user-side schema validation to observation content (separate concern).
- Per-observation TTL or auto-expiry (separate concern).

## Severity

Enhancement, not a bug — Honcho is functioning as designed, but the design is missing a maintenance knob that users need.

## Cross-references

- hermes-agent PR #66754 (renderer-side sanitization, defense-in-depth against the same class of issue)
- The 2026-07-18 Vee debugging session in this repo's issue tracker will reference this once filed.

ajspig:
Part of plastic-labs/claude-honcho#113 to the standalone MCP worker. The search tool only queried the message store, so saved conclusions were unreachable through search (findable only by paging list_conclusions or via query_conclusions with a known scope).

search now also queries conclusions in parallel with the message search, returning {messages, conclusions}. Since the conclusions query API requires an explicit (observer, observed) pair, the conclusion leg runs only when peer_id is given (self-conclusions, matching list_conclusions/query_conclusions defaults) and degrades to \[\] on error so search never gets worse than before. Conclusion results include IDs usable with delete_conclusion.

Also syncs mcp/bun.lock with package.json's @honcho-ai/sdk ^2.1.0 and points delete_conclusion's description at query_conclusions/list_conclusions for ID discovery.

## Summary by CodeRabbit

* **New Features**
  * Search results now include matching messages and peer conclusions when a peer is specified.
  * Added separate limits and filters for message and conclusion searches.
  * Conclusion results now include IDs, content, levels, and timestamps.
  * Conclusion queries support optional filtering criteria.
  * Unfiltered conclusion-search failures no longer prevent message results from being returned.
* **Documentation**
  * Added guidance to locate conclusion IDs before deleting conclusions.

### DIFF
--- a/mcp/bun.lock
+++ b/mcp/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "honcho-mcp-proxy",
       "dependencies": {
-        "@honcho-ai/sdk": "^2.1.0",
+        "@honcho-ai/sdk": "^2.2.0",
         "@modelcontextprotocol/sdk": "^1.26.0",
         "agents": "^0.4.0",
         "nanoid": "^5.1.7",
--- a/mcp/bun.lock
+++ b/mcp/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "honcho-mcp-proxy",
       "dependencies": {
-        "@honcho-ai/sdk": "^2.0.0",
+        "@honcho-ai/sdk": "^2.1.0",
         "@modelcontextprotocol/sdk": "^1.26.0",
         "agents": "^0.4.0",
         "nanoid": "^5.1.7",
@@ -107,7 +107,7 @@
 
     "@esbuild/win32-x64": ["@esbuild/win32-x64@0.25.4", "", { "os": "win32", "cpu": "x64" }, "sha512-nOT2vZNw6hJ+z43oP1SPea/G/6AbN6X+bGNhNuq8NtRHy4wsMhw765IKLNmnjek7GvjWBYQ8Q5VBoYTFg9y1UQ=="],
 
-    "@honcho-ai/sdk": ["@honcho-ai/sdk@2.0.1", "", { "dependencies": { "@types/node": "^24.0.1", "zod": "4.0.0" } }, "sha512-y/Wk49C0N1miI9BZTNWFIbzdUkMZfP4Do/EJ1q4lEIK+FAOKxQgces/zET3kPKV3zF9sOUl2pXrFb/XKYayeYw=="],
+    "@honcho-ai/sdk": ["@honcho-ai/sdk@2.2.0", "", { "dependencies": { "zod": "4.0.0" } }, "sha512-SyygN+BrpUB2fRjhwcYmT+tcEhHrKmbj9nOZLVUFY7M5YBswJ+mZb/CeLpNbRh+QQTU8F8JWY3lQat95c6nwmA=="],
 
     "@hono/node-server": ["@hono/node-server@1.19.9", "", { "peerDependencies": { "hono": "^4" } }, "sha512-vHL6w3ecZsky+8P5MD+eFfaGTyCeOHUIFYMGpQGbrBTSmNNoxv0if69rEZ5giu36weC5saFuznL411gRX7bJDw=="],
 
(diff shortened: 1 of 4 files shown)


mcp/src/tools/workspace.ts
```diff
@@ -74,11 +74,13 @@ export function register(server: McpServer, ctx: ToolContext) {
...
+        "Semantic search across messages and, when peer_id is given, that peer's saved conclusions.",
+        "Message scope is determined by which optional params are provided:",
         "- No scope params: search all messages in the workspace.",
         "- peer_id only: search messages authored by that peer across all sessions.",
         "- session_id only: search messages within that session.",
-        "Returns an array of matching messages with their content, peer, and session info.",
+        "Conclusions require peer_id (self-conclusions are searched; conclusion IDs are usable with delete_conclusion).",
+        "Returns {messages, conclusions}.",
       ].join("\n"),
       inputSchema: {
```
akattelu, line 86: we talked about adding in the `limit` and `filters` fields here and passing them through to the API

mcp/bun.lock
```diff
@@ -5,7 +5,7 @@
     "": {
       "name": "honcho-mcp-proxy",
       "dependencies": {
-        "@honcho-ai/sdk": "^2.0.0",
+        "@honcho-ai/sdk": "^2.1.0",
```
akattelu: we talked about bumping this further to get filters in the sdk

mcp/src/tools/workspace.ts
```diff
@@ -94,17 +96,28 @@ export function register(server: McpServer, ctx: ToolContext) {
...
-          messages = await session.search(query);
-        } else if (peer_id) {
-          const peer = await ctx.honcho.peer(peer_id);
-          messages = await peer.search(query);
-        } else {
-          messages = await ctx.honcho.search(query);
-        }
-        return textResult(formatMessages(messages));
+        const peer = peer_id ? await ctx.honcho.peer(peer_id) : null;
+        const [messages, conclusions] = await Promise.all([
```
akattelu, line 170: nit: this block has a lot of ternary and nesting and then/catch instead of await. we could clean it up for readability.

mcp/src/tools/workspace.ts
```diff
@@ -94,17 +96,28 @@ export function register(server: McpServer, ctx: ToolContext) {
...
+            ? ctx.honcho
+                .session(session_id)
+                .then((session) => session.search(query))
+            : peer
+              ? peer.search(query)
+              : ctx.honcho.search(query),
+          // Conclusion search requires an (observer, observed) pair, so it
+          // only runs when peer_id is given; degrades to [] on error so
+          // search never gets worse than message-only search.
+          peer ? peer.conclusions.query(query).catch(() => []) : [],
```
akattelu: whatever we end up with, we should note in the tool description that conclusions will be empty when peer is left out.

akattelu (approved):

PR #983, opened 2026-08-05
fix: make embedding batch size configurable

### LINKED ISSUE #687: [Bug] OpenAIEmbeddingClient.max_batch_size hardcoded to 2048 breaks DashScope embeddings (cap is 10)

WizisCool:
## Summary

Makes the per-request embedding batch size configurable via `EMBEDDING_MODEL_CONFIG__MAX_BATCH_SIZE` (or `[embedding.model_config] max_batch_size` in config.toml).

OpenAI-compatible embedding providers enforce their own per-request input caps — DashScope `text-embedding-v4` accepts 10, Alibaba Bailian `qwen3.7-text-embedding` accepts 20 — but the client hardcoded OpenAI's 2048, so batched embedding calls deterministically fail with HTTP 400 on those providers. With this change the operator sets the cap; when unset, native provider defaults are preserved (OpenAI 2048, Gemini 100).

Changes: config plumbing in `src/config.py`, default-preserving fallback in `src/embedding_client.py` (both transports, plus settings-signature inclusion), docs (`.env.template`, `config.toml.example`, `configuration.mdx`), and an env-parsing regression test. Scope is minimal.

## Relation to #696

Same approach as #696 by @pntech20 (credit for the design), rebased onto current main — that PR has been open since May without maintainer review and its test file now conflicts with main. Happy to close this if you prefer to merge/rebuild #696 instead.

Fixes #687.

## Summary by CodeRabbit

* **New Features**
  * Added an optional maximum batch size for embedding requests.
  * Embedding requests now support configurable batching, with defaults of 2,048 items for OpenAI-compatible providers and 100 for Gemini.
  * Configuration examples and documentation include setup guidance and provider limit recommendations.

* **Bug Fixes**
  * Embedding clients now apply updated batch-size settings without requiring a restart.


WizisCool:
Hi @VVoruganti @akattelu @Rajat-Ahuja1997 @eisene — could you approve the workflow runs for this PR? Thanks.

eisene:
> Hi @VVoruganti @akattelu @Rajat-Ahuja1997 @eisene — could you approve the workflow runs for this PR? Thanks.

approved

WizisCool:
All three review suggestions addressed in 7f3b4fe:

1. **Transport/fallback coverage**: added `test_gemini_simple_batch_embed_respects_configured_max_batch_size` (splits at the configured limit) and `test_gemini_simple_batch_embed_defaults_to_100_when_unset` / `test_openai_simple_batch_embed_defaults_to_2048_when_unset` (both fallbacks verified as single requests).
2. **Persisted-to-runtime propagation**: the env test now calls `resolve_embedding_model_config()` and asserts the resolved `max_batch_size` is 10.
3. **Docs**: Gemini's 100 is now described as the client's conservative default (Gemini doesn't document a limit), not a native provider limit.

Local verification: 265 passed (tests/llm/), ruff clean, basedpyright 0 errors.

WizisCool:
Hi @VVoruganti @akattelu @Rajat-Ahuja1997 @eisene — pushed 7f3b4fe (review fixes); the new workflow runs need approval. Thanks.

eisene:
> Hi @VVoruganti @akattelu @Rajat-Ahuja1997 @eisene — pushed [7f3b4fe](https://github.com/plastic-labs/honcho/commit/7f3b4fe1fa79193fe84baa9c1af6001b738a42d5) (review fixes); the new workflow runs need approval. Thanks.

approved

WizisCool:
@VVoruganti @akattelu @Rajat-Ahuja1997 @eisene

CodeRabbit’s review is now fully addressed, including the final test-coverage suggestion in commit `cb04327`. The implementation and tests have been verified locally:

- 265 `tests/llm/` tests passed
- Ruff: clean
- basedpyright: 0 errors, 0 warnings

The GitHub FastAPI Tests check still cannot complete successfully because the pull-request workflow does not have access to the required `OPENAI_API_KEY`. It fails at the unrelated `tests/crud/test_document.py::TestDocumentCRUD::test_duplicate_rejection_reinforces_existing` test with `OpenAI API key is required`; 490 tests pass before that failure. Static Analysis passes, and the changed code/tests are clean locally.

Could you please help resolve or advise on the missing-key CI issue? I look forward to your response and, once the CI situation is resolved, to the PR being merged.

akattelu (dismissed):

akattelu:
I believe a fix for the openai CI issues is in #955, maybe we could stack on that

WizisCool:
> I believe a fix for the openai CI issues is in #955, maybe we could stack on that

@akattelu Thanks for pointing me to #955. I have now stacked PR #983 on top of PR #955 (`fix/embedding-encoding-without-api-key`). The combined branch applies cleanly, and local validation passed:

- `tests/crud/test_document.py` + `tests/llm/`: 266 passed
- Ruff: clean
- basedpyright: 0 errors, 0 warnings

PR #983 now uses #955 as its base so the pull-request CI can include the missing-key fix.

akattelu (approved):

PR #981, opened 2026-08-05
fix(crud): preserve active session peer joined_at on re-add

WizisCool:
Fixes #940.

## Impact

Re-adding a peer who never left moves their `joined_at` window start forward. `peer_perspective` search only returns messages with `created_at >= joined_at`, so any client that re-adds peers at every process start (e.g. the Hermes Agent honcho integration) loses access to messages older than the last re-add after each restart, with no error raised. Two independent reports on #940.

## Problem

`get_or_create_session` re-runs `_get_or_add_peers_to_session` on every call, and the `ON CONFLICT` upsert in `src/crud/session.py` unconditionally reset `joined_at = now()` for every re-added peer — including peers who never left (`left_at IS NULL`). Since `peer_perspective` search filters on `message.created_at >= joined_at` (src/utils/search.py), any client that re-adds peers on each process start silently loses visibility into all messages older than the last restart.

## Change

Only start a fresh membership window (`joined_at = now()`) when the existing row has `left_at IS NOT NULL` — a genuine rejoin. Active peers keep their original `joined_at`:

```python
"joined_at": case(
    (models.SessionPeer.left_at.is_not(None), func.now()),
    else_=models.SessionPeer.joined_at,
),
```

No other behavior touched: observer-count logic, configuration merge, and `left_at` clearing are unchanged.

## Why this PR (relation to #941)

PR #941 proposed the same fix but bundled an out-of-scope observer-count refactor that the maintainer requested to revert (CHANGES_REQUESTED 2026-07-28); the author has not responded since. This PR is the minimal version: 2 files, only the joined_at conditional plus a regression test.

## Tests

- `test_session_peer_joined_at_readd_semantics` (tests/crud/test_session.py): re-adding an active peer preserves `joined_at` and the stored configuration; a peer that left and rejoins gets a fresh `joined_at` and the incoming configuration.
- `test_peer_perspective_search_after_active_readd` (tests/test_search.py): an active re-add keeps pre-existing messages visible through `peer_perspective` search (fails on pre-fix code); a genuine rejoin starts a new visibility window.
- `pytest tests/test_search.py tests/crud/test_session.py tests/crud/test_document.py` → **39 passed** (keyless environment, mirrors CI after #955)
- `ruff check` clean.

PR #986, opened 2026-08-06
fix(crud): preserve active session peer joined_at on re-add

### LINKED ISSUE #940: [Bug] Re-adding an active session peer resets joined_at and hides perspective search results
# 🐞 Bug Report

## Describe the bug

Re-adding a peer that is already an active member of a session updates that membership's `joined_at` to the current database time.

This makes the otherwise idempotent “ensure this peer belongs to this session” operation move the start of an uninterrupted membership window. Workspace message searches using `peer_perspective` then exclude existing messages whose `created_at` falls before the newly advanced `joined_at`, so valid searches can become empty or incomplete.

The conflict path in `src/crud/session.py::_get_or_add_peers_to_session()` currently assigns `joined_at = func.now()` for every existing `SessionPeer` row. It does so both for a genuine rejoin (`left_at IS NOT NULL`) and for an already-active membership (`left_at IS NULL`).

---

### Is this a regression?

Unknown. The behavior is present in v3.0.12 and current `main` at `3ee890fa6f55388abd23b7660fb726e14d83459d`.

---

### To reproduce

1. Create a workspace, two peers, and a session containing both peers.
2. Add a message whose timestamp is after the original `SessionPeer.joined_at`.
3. Add one of the still-active peers to the same session again through the normal add/ensure-peer path.
4. Observe that the existing membership row's `joined_at` advances even though `left_at` was already null.
5. Search workspace messages with that peer as `peer_perspective`.
6. Observe that the existing message is omitted because the temporal filter requires `message.created_at >= session_peers.joined_at`.

Directly querying the message without that temporal perspective constraint still finds it.

---

### Expected behavior

- Re-adding an already-active peer is idempotent:
  - preserve the original `joined_at`;
  - preserve the active membership's existing configuration.
- Re-adding a peer that previously left remains a genuine rejoin:
  - assign a new `joined_at`;
  - clear `left_at`;
  - apply the supplied configuration.
- Existing `peer_perspective` temporal boundary behavior remains unchanged.

A focused regression should prove that an existing message remains visible through perspective search after an active-peer re-add, while genuine rejoin behavior still starts a new membership window.

---

### Environment

- Honcho Server: v3.0.12 and current `main` at `3ee890fa6f55388abd23b7660fb726e14d83459d`
- Database: PostgreSQL

---

### Additional context

Current temporal-search tests seed membership timestamps directly, so they validate the search boundary but do not exercise the active-peer conflict update that changes the boundary later.

WizisCool:
Fixes #940.

## Problem

When `get_or_create_session` re-adds a peer whose membership is still active (`left_at IS NULL`), the session-peer upsert currently resets `joined_at` to `now()`. `peer_perspective` search filters messages using `message.created_at >= joined_at`, so clients that re-add peers on startup can lose visibility of messages created before the latest restart, without an error.

## Change

Preserve `joined_at` for active peers. Only reset it when a peer has actually left and rejoins (`left_at IS NOT NULL`). Existing observer-count logic, configuration handling, and `left_at` clearing are unchanged.

## Tests

- Active re-add preserves `joined_at` and the stored configuration.
- A departed peer rejoining receives a new `joined_at` and incoming configuration.
- `peer_perspective` search retains messages after an active re-add and starts a new window after a genuine rejoin.
- The search regression fails on the pre-fix code.

Local validation against current `main` (including #955):
- `pytest tests/test_search.py tests/crud/test_session.py tests/crud/test_document.py`: 39 passed
- `ruff check`: clean

This PR supersedes #981, which encountered an unrelated full-suite CI cancellation in `tests/test_session_allowlist.py` after 1501 tests had passed. The new PR starts from the current `main` and uses a single clean commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Re-adding an active session peer now preserves existing membership history and message visibility.
  * Rejoining after leaving starts a new membership period, applies updated settings, and prevents access to messages from the earlier period.
  * Improved peer-perspective search behavior for active re-adds and leave-and-rejoin scenarios.

* **Tests**
  * Added regression coverage for session membership updates and search visibility.

### DIFF
--- a/src/crud/session.py
+++ b/src/crud/session.py
@@ -946,13 +946,15 @@ async def set_peers_for_session(
             f"Session {session_name} not found in workspace {workspace_name}"
         )
 
-    # Soft delete specified session peers by setting left_at timestamp
+    # Soft delete active peers that are absent from the replacement set.
+    # Peers being re-set stay active so their membership window is preserved.
     update_stmt = (
         update(models.SessionPeer)
         .where(
             models.SessionPeer.session_name == session_name,
             models.SessionPeer.workspace_name == workspace_name,
             models.SessionPeer.left_at.is_(None),  # Only update active peers
+            models.SessionPeer.peer_name.notin_(peer_names.keys()),
         )
         .values(left_at=func.now())
     )
--- a/src/crud/session.py
+++ b/src/crud/session.py
@@ -1060,13 +1060,16 @@ async def _get_or_add_peers_to_session(
         ]
     )
 
-    # On conflict, update joined_at and clear left_at (rejoin scenario)
-    # If left_at is not None (peer has left the session): Use the new configuration (stmt.excluded.configuration)
-    # If left_at is None (peer is still active): Keep the existing configuration (models.SessionPeer.configuration)
+    # On conflict, clear left_at and reset joined_at only when rejoining
+    # If left_at is not None (peer has left the session): Start a fresh membership window
+    # If left_at is None (peer is still active): Preserve joined_at and the existing configuration
     stmt = stmt.on_conflict_do_update(
         index_elements=["session_name", "peer_name", "workspace_name"],
         set_={
-            "joined_at": func.now(),
+            "joined_at": case(
+                (models.SessionPeer.left_at.is_not(None), func.now()),
+                else_=models.SessionPeer.joined_at,
+            ),
             "left_at": None,
             "configuration": case(
                 (models.SessionPeer.left_at.is_not(None), stmt.excluded.configuration),
--- a/tests/crud/test_session.py
+++ b/tests/crud/test_session.py
@@ -123,7 +123,7 @@ class TestSessionCRUD:
             peer_names={test_peer.name: original_config},
         )
         first_result = await db_session.execute(session_peer_stmt)
-        first_joined_at, first_left_at, first_config = first_result.one()
+        _, first_left_at, first_config = first_result.one()
         assert first_left_at is None
         assert first_config == original_config.model_dump()
 
--- a/tests/crud/test_session.py
+++ b/tests/crud/test_session.py
@@ -85,6 +85,92 @@ class TestSessionCRUD:
         assert rejoined_config == updated_config.model_dump()
 
     @pytest.mark.asyncio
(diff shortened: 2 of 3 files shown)


WizisCool:
Static Analysis passes on #986. The FastAPI test workflow still fails at the same unrelated test as #981:

`tests/test_session_allowlist.py::TestMessageCrudAllowlistIntersection::test_grep_messages_intersects_allowlist` → `concurrent.futures.CancelledError`

The run completed with **1501 passed, 25 skipped, 1 error**. The PostgreSQL service log also shows the allowlist test racing with database initialization (`relation "public.message_embeddings" does not exist`). No `joined_at`, `peer_perspective`, or changed-test failure is reported.

@eisene Could you please help rerun or assess this workflow failure? The implementation-specific tests pass locally (39 passed without an API key), and Static Analysis is green.

WizisCool:
Hi @akattelu — I noticed you reviewed and merged #983 after the CI issue was resolved there. Could you please take a look at #986 as well?

Static Analysis passes. The FastAPI workflow reaches 1501 passed / 25 skipped, then errors in the unrelated `tests/test_session_allowlist.py::TestMessageCrudAllowlistIntersection::test_grep_messages_intersects_allowlist` test with `CancelledError`; the service log also reports `message_embeddings` missing during database initialization.

The changed tests and local keyless validation are clean (39 passed). Could you advise whether this CI failure should simply be rerun or whether the allowlist test/database fixture needs a separate fix? Thanks.

WizisCool:
Follow-up to #987: I verified the additional `PUT /sessions/{id}/peers` path is affected as well. `set_peers_for_session` previously soft-deleted all active peers before the upsert, so the `left_at` CASE in #986 could still refresh `joined_at` on every replacement.

I addressed this in `715c755` by excluding peers present in the incoming set from the soft-delete update. Added regression coverage for the set-peers path:

- active peer re-set preserves `joined_at`, `left_at`, and the existing configuration;
- removed peer re-added later receives a fresh `joined_at` and new configuration.

Validation: **40 passed** across session/search/document tests; ruff clean. This keeps the fix in one PR and covers both session initialization and PUT peers entry points.

VVoruganti:
Superseded by #1059, which landed both the joined_at CASE guard and the set_peers soft-delete exclusion. Thanks for the fix.

PR #987, opened 2026-08-06
fix(crud): preserve joined_at for active peers on set_peers upsert

xiangrui979:
Follow-up to #940. This fixes the path that the existing PRs don't cover.

## Root cause

`set_peers_for_session` (the `PUT .../sessions/{id}/peers` endpoint) soft-deletes every active peer in the session first, then upserts the incoming peer set. The upsert's on-conflict branch rewrites `joined_at` unconditionally, so every call re-stamps the join time of peers that never actually left.

That silently breaks `peer_perspective` search: `src/utils/search.py` joins `session_peers` and filters `message.created_at >= joined_at`. After any client re-adds peers at startup (the Hermes Agent Honcho plugin does this on every session init), everything written before the latest re-add disappears from perspective-filtered results. On my self-hosted v3.0.11 the main session ended up with a 0-message search window, with no error raised anywhere.

## Why #981/#986 don't fix this path

Both guard the upsert with a CASE on `left_at`:

```python
"joined_at": case(
    (models.SessionPeer.left_at.is_not(None), func.now()),
    else_=models.SessionPeer.joined_at,
),
```

That only works when the row still has `left_at IS NULL` at upsert time. In `set_peers_for_session` the soft delete runs **before** the upsert, so every active member's row is already stamped with `left_at = now()` — the CASE then always takes the "rejoin" branch and refreshes `joined_at` anyway. Their tests pass because they exercise `get_or_create_session`, which doesn't soft-delete first.

## The fix

Two changes in `src/crud/session.py`:

1. **Exclude the peers being (re)set from the soft delete**, so active members keep `left_at IS NULL` and their original `joined_at` (only peers actually being removed/replaced leave).
2. **In the upsert conflict branch**, keep the original `joined_at` when `left_at` is still NULL; only a genuine rejoin (row already left) gets the caller-provided fresh timestamp.

## Verification

Against a live Postgres on self-hosted v3.0.11: after the patch, hitting the peers endpoint leaves `joined_at` unchanged for already-active members, and `peer_perspective` searches return historical messages again (previously an empty window). The same holds across a full client restart (soft-delete + re-add cycle).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session peer management by preserving active peers’ existing settings and join times.
  * Correctly refreshes join times and applies updated settings when previously removed peers rejoin.
  * Prevents unnecessary removal of active peers when updating a session’s peer list.

### DIFF
--- a/src/crud/session.py
+++ b/src/crud/session.py
@@ -945,13 +945,17 @@ async def set_peers_for_session(
             f"Session {session_name} not found in workspace {workspace_name}"
         )
 
-    # Soft delete specified session peers by setting left_at timestamp
+    # Soft delete session peers by setting left_at timestamp. Peers being
+    # (re)set in this call are excluded so active members keep their original
+    # joined_at (perspective search window not truncated); only peers actually
+    # removed/replaced leave.
     update_stmt = (
         update(models.SessionPeer)
         .where(
             models.SessionPeer.session_name == session_name,
             models.SessionPeer.workspace_name == workspace_name,
             models.SessionPeer.left_at.is_(None),  # Only update active peers
+            models.SessionPeer.peer_name.notin_(peer_names.keys()),
         )
         .values(left_at=func.now())
     )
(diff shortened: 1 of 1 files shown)


WizisCool:
I reproduced the `PUT /peers` coverage gap described here and addressed it in PR #986 (commit `715c755`). The active-peer soft-delete now excludes peers present in the incoming set, so the existing `joined_at` CASE can preserve their membership window. I also added regression coverage for active re-set and genuine remove/re-add behavior.

Validation: 40 related tests passed; ruff and basedpyright are clean. This keeps both entry points covered in one PR rather than splitting the fix.

VVoruganti:
Superseded by #1059. Your diagnosis of the set_peers_for_session soft-delete ordering was the right root cause and the notin_ exclusion landed as you proposed. Thanks.

PR #999, opened 2026-08-08
fix: isolate per-text embedding failures in reconciler

cadamec:
## Description

The message-embedding reconciler (`_sync_message_embeddings` in `src/reconciler/sync_vectors.py`) sent **all pending texts in a single `simple_batch_embed` call**. If any one text exceeded the embedding provider's context window, the entire batch returned a 400 and **every row in the batch was marked `failed`** — including small, perfectly-embeddable messages that were only collateral damage.

### Root cause

Two compounding issues:

1. **Batch poisoning.** `simple_batch_embed(contents)` sends the whole list in one provider call. A single oversized text 400s the whole batch. The reconciler's `except Exception` then marks *all* rows in the batch as failed via `_bump_message_embedding_sync_attempts`, and after `MAX_SYNC_ATTEMPTS` they're permanently `failed` (manual intervention required).

2. **Tokenizer undercount.** The chunking tokenizer (tiktoken `o200k_base`) can undercount a provider's real token count. In our deployment (nomic-embed-text via Ollama), tiktoken undercounts by ~4–5×, so texts that passed the chunk cap still exceeded the provider's context window and got rejected.

### Fix

Embed each text **individually** instead of as one batch. A single oversized text now fails on its own (and is retried/backed off independently), while the rest of the batch stays embeddable.

### Impact observed

In our deployment this had left **54 message embeddings permanently failed** (clustered into exactly 3 poisoned batches) and 2 stuck in a retry loop. After this fix plus re-queueing, all recovered rows embedded cleanly: `synced 70 message embeddings; failed 0`, then `synced 5; failed 0`. No further `input length exceeds context length` errors.

### Test plan

- [x] Manual verification: reconciler now syncs pending rows with `failed 0`
- [ ] Unit tests pass (existing reconciler tests)

### Note for maintainers

This fixes the batch-poisoning failure mode generally. The tokenizer-undercount issue is deployment-specific (depends on the embedding model), so the per-text isolation is the durable fix — operators should also size `EMBEDDING_MAX_INPUT_TOKENS` conservatively for their provider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message embedding reliability by processing missing vectors individually.
  * A failed embedding no longer prevents other embeddings from completing.
  * Retryable failures are recorded for affected items, while successful embeddings continue normally.
  * Invalid embedding requests are marked as permanently failed without using retry attempts.
  * Workspace telemetry remains correctly aggregated when applicable.

### DIFF
--- a/src/reconciler/sync_vectors.py
+++ b/src/reconciler/sync_vectors.py
@@ -425,33 +425,31 @@ async def _sync_message_embeddings(
     freshly_embedded: dict[int, list[float]] = {}
 
     if embs_needing_embed:
-        try:
-            contents = [emb.content for emb in embs_needing_embed]
-            # MESSAGE_CREATE (not VECTOR_SYNC): these rows come from create_messages
-            # as pending chunks; document re-embeds stay on VECTOR_SYNC below.
-            workspaces = {emb.workspace_name for emb in embs_needing_embed}
-            with embedding_call_purpose(
-                EmbeddingCallPurpose.MESSAGE_CREATE.value,
-                workspace_name=workspaces.pop() if len(workspaces) == 1 else None,
-                parent_category="reconciliation",
-            ):
-                new_embeddings = await embedding_client.simple_batch_embed(contents)
-
-            if len(new_embeddings) != len(embs_needing_embed):
-                logger.warning(
-                    "Re-embedded %s/%s message embeddings; remaining will be retried",
-                    len(new_embeddings),
-                    len(embs_needing_embed),
-                )
-
-            for emb, new_emb in zip(embs_needing_embed, new_embeddings, strict=False):
-                freshly_embedded[emb.id] = new_emb
-                if store_in_postgres:
-                    emb.embedding = new_emb
-        except Exception:
-            logger.exception(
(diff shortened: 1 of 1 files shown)


cadamec:
Thanks for the review — both comments are fair. I've addressed the quick win and am deferring the heavy lift.

**Comment 2 (ValueError handling) — fixed in 62d6f5e4.** `simple_batch_embed` raises `ValueError` for expected validation failures (oversized input, dimension mismatch), which aren't retryable. I now catch `ValueError` separately, log it distinctly, and mark those rows `failed` immediately instead of burning `MAX_SYNC_ATTEMPTS` retries. Unexpected errors still log via `logger.exception` and remain eligible for retry.

**Comment 1 (release DB session during embedding) — valid, deferring as a follow-up.** This is a pre-existing architectural issue: the original code also held the `db` session (and the `FOR UPDATE SKIP LOCKED` row locks) across the batch `simple_batch_embed` call. My change only altered the shape from one long call to N shorter ones — it didn't introduce the violation. The proper fix (committed lease / processing-state, embed outside the session, short write transactions) is a meaningful refactor that I'd rather land as a separate PR so this bugfix stays minimal and reviewable. I'll file it as a follow-up.

PR #1005, opened 2026-08-10
chore: update gcp/fly deployment yamls

Rajat-Ahuja1997:
## Summary by CodeRabbit

* **Deployment Improvements**
  * Standardized version handling for staging and production releases.
  * Deployment images now use consistent version labels.
  * Added automated notifications to update staging and production services after successful image builds.
  * Added validation to ensure release versions match the application version before deployment.
  * Deployments now run from version tags.
* **Chores**
  * Removed legacy Fly deployment workflows and manual deployment triggers.


akattelu (approved):

VVoruganti (commented):
Shape looks right. The version is resolved once and reused, `deployment-v${VERSION}` is now consistent across both trigger paths, and routing `github.event.inputs.version` through an `env:` var instead of inlining `${{ }}` into the shell closes a real script-injection hole the old workflows had. `env.IMAGE_NAME` does resolve in the `prompt-service` job (workflow-level `env` is inherited by all jobs), and nothing else in the repo references the deleted fly workflows — `start-fly-runner.yml` / `unified-tests.yml` use flyctl for CI runners, unrelated.

A few notes, none of them merge blockers:

### 1. The guard will fail the next tag push as-is

`v3.0.12` was tagged with `pyproject.toml` still at `3.0.11`, which is also the state on `main` right now. The check is correct — that drift is exactly what it's for — but it does change the release process: `pyproject.toml` has to be bumped *before* tagging from here on. Worth calling out to whoever cuts the next release.

### 2. The `version` input is now dead weight

Since the only value that passes the check is the pyproject version, the input can only be redundant or fatal — and `default: "manual"` guarantees a failed run for anyone who dispatches without editing it. Deriving it instead drops both the input block and the branch:

```yaml
      - name: Resolve and verify version
        id: version
        run: |
          VERSION="$(grep -m1 '^version = ' pyproject.toml | cut -d'"' -f2)"
          # A running instance serves this version at /openapi.json.
          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF_NAME" != "v$VERSION" ]]; then
            echo "::error::tag '$GITHUB_REF_NAME' does not match pyproject.toml version '$VERSION'."
            exit 1
          fi
          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```

Same guarantee, fewer moving parts.

### 3. Lost the concurrency group

Both deleted fly workflows had `concurrency: {group: deploy-*-group, cancel-in-progress: true}`; neither GCP workflow does. Two tags pushed close together will race, and the `add_honcho_version` POSTs can land out of order. One block per file restores it.

### 4. Worth confirming downstream

`image_label` changes from `honcho-image:deployment-v3.0.12` (fly app name) to `${IMAGE_NAME}:deployment-v3.0.12`, and the `workflow_dispatch` path's tag gains a `v` it didn't have before. The comment says the registry path is supplied downstream, which reads right — but it's the one thing not verifiable from this repo, so worth a look at the `add_honcho_version` consumer before the first real deploy. Also worth knowing whether a repeat POST for the same version is idempotent there.

---

Unchanged from before, just noting it since this PR makes it the only deploy path: both workflows trigger on `v*`, so one tag deploys staging and production in parallel with no gate between them.

PR #1014, opened 2026-08-12
fix: reduce Honcho runtime image size

Rajat-Ahuja1997:
-Reduce Honcho’s production image footprint by removing duplicated layers and making the LanceDB optional.

- The existing Dockerfile installed the ~758 MB virtualenv and then recursively changed /app ownership in a later layer. Docker preserved both versions, nearly doubling the dependency footprint.

  This PR:

  - Uses a multi-stage build so uv remains build-only.
  - Copies the virtualenv into the runtime image with its final ownership.
  - Removes the redundant second uv sync.
  - Makes LanceDB and PyArrow an optional lancedb extra.
  - Adds INSTALL_LANCEDB=true support to Docker and Compose builds.
  - Keeps LanceDB dependencies installed in development environments.
  - Excludes FastAPI’s unused cloud-deployment CLI while retaining fastapi run.

  ## Image-size results

  Measured locally on ARM64:

   Image                       Unpacked size    Reduction
  ━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━
   Existing image                   1.712 GB            —
  ──────────────────────────  ───────────────  ───────────
   Default production image           663 MB        61.2%
  ──────────────────────────  ───────────────  ───────────
   Image with LanceDB                 917 MB        46.4%

  The equivalent gzip comparison decreased by approximately 64%. Applied to the current 377 MB registry
  image, this suggests a resulting registry size around 136 MB, subject to CI architecture and registry
  compression.

  LanceDB remains available with:

      docker build --build-arg INSTALL_LANCEDB=true .

  Or through Compose:

      INSTALL_LANCEDB=true docker compose up -d --build

  ## Runtime verification

  The default image was run against a freshly provisioned PostgreSQL database:

  - /health returned 200 {"status":"ok"}
  - /openapi.json returned successfully
  - All database migrations completed
  - Turbopuffer constructed successfully
  - The optional image constructed LanceDB successfully
  - API idle memory: approximately 237 MiB
  - Deriver idle memory: approximately 232 MiB

  Removing LanceDB/PyArrow primarily reduces image storage and transfer costs. Runtime RSS remains
  approximately unchanged because those packages were not loaded when using Turbopuffer.

## Summary by CodeRabbit

* **New Features**
  * Added optional LanceDB support for Docker builds, Docker Compose services, and local installations.
  * Added shared persistent storage for LanceDB data.
  * Improved container setup with separate build and runtime stages.

* **Bug Fixes**
  * Updated missing-dependency guidance with the correct LanceDB installation command.
  * Excluded an unnecessary FastAPI cloud CLI dependency from standard installations.

* **Documentation**
  * Documented LanceDB configuration requirements, including Intel macOS availability limitations.


pyproject.toml
```diff
@@ -32,14 +32,17 @@ dependencies = [
...
     "cashews[redis]==7.5.0",
     "scikit-learn>=1.6.0",
     "prometheus_client>=0.21.0",
     "cloudevents>=1.12.0,<2.0",
 ]
+[project.optional-dependencies]
+lancedb = [
+    "lancedb>=0.25.3; sys_platform != \"darwin\" or platform_machine != \"x86_64\"",
+    "pyarrow>=19.0.0",
+]
```
VVoruganti, line 45: Should this be just another group under dependency groups?

VVoruganti (changes requested):
I think just updating the comments and seeing if we can consolidate the pyproject setup, but otherwise this looks fantastic

pyproject.toml
akattelu: That would work too, but this works with `uv sync --extra lancedb` and I think it fits well as "optional runtime feature"

akattelu:
I pushed up two more commits:

* updating docs 
* fixing coderabbit comment

Testing

* Ran a bunch of unified tests that passed on the new images. Also tested with a local lancedb image, so that path still works too.
* `alembic` still runs from the slimmed venv, and the new missing-dependency guard in `src/vector_store/__init__.py` raises the intended error when `lancedb` isn't installed.
* Did not test with turbopuffer locally

akattelu (dismissed):

pyproject.toml
VVoruganti: Ok thought about this more and agree it's fine

VVoruganti (approved):

PR #1019, opened 2026-08-13
fix(embedding): truncate in batch embed and return results breakdown 

### LINKED ISSUE #569: Deriver loses an entire batch of observations when one exceeds the embedding token limit

akattelu:
Fixes plastic-labs/honcho#569 plastic-labs/honcho#728  <br>Supersedes plastic-labs/honcho#860 plastic-labs/honcho#864

## Summary by CodeRabbit

* **New Features**
  * Oversized text can now be truncated to fit embedding token limits, with warnings logged when truncation occurs.
  * Representation processing reports failed observer saves in completion telemetry.
  * Added clear error handling when all observer representation saves fail.
* **Bug Fixes**
  * Reconciliation and batch representation embedding now handle oversized inputs more reliably.
* **Tests**
  * Added coverage for truncation, partial and complete save failures, telemetry, and live embedding behavior.

## Summary by CodeRabbit

* **New Features**
  * Added optional truncation for oversized embedding inputs, allowing them to be processed instead of rejected.
  * Added reporting of failed observer saves in completion telemetry.
* **Bug Fixes**
  * Improved representation processing when some observer saves fail.
  * Clearly reports an error when all observer saves fail.
  * Reconciliation and batch embedding now handle oversized content more reliably.
* **Tests**
  * Expanded coverage for truncation, partial and complete save failures, telemetry, and live embedding behavior.

## Summary by CodeRabbit

* **New Features**
  * Oversized text can now be automatically truncated during batch embedding, helping observation and document processing continue successfully.
  * Reconciliation and other embedding workflows consistently support oversized inputs.
  * Completion telemetry now reports how many observer saves failed.
* **Bug Fixes**
  * Partial observer save failures no longer interrupt processing.
  * Clear failure reporting is provided when all observer saves fail.
* **Tests**
  * Added coverage for truncation, observer save failures, telemetry, and live embedding behavior.

## Summary by CodeRabbit

* **New Features**
  * Oversized text can now be safely truncated during batch embedding instead of failing.
  * Embedding requests support configurable provider timeouts, including faster cancellation for short limits.
  * Representation processing now reports failed saves in completion telemetry.
  * Document creation respects the configured deduplication setting.
* **Bug Fixes**
  * Partial observer save failures no longer stop successful saves; complete failures are reported clearly.
  * Reconciliation and observation creation now handle oversized content more reliably.


src/deriver/deriver.py
```diff
@@ -337,5 +340,16 @@ async def process_representation_tasks_batch(
...
+            failed_observer_count=len(save_errors),
         )
     )
+
+    if save_errors and successful_observer_count == 0:
+        details = "; ".join(
+            f"{observer}: {exc.__class__.__name__}: {exc}"
+            for observer, exc in save_errors
+        )
+        raise RepresentationSaveError(
```
eisene, line 352: i believe we are on python 3.10, if we upgrade to 3.11 we get native `ExceptionGroup` support - i find those really useful. maybe not worth it in this PR but could make an issue

src/embedding_client.py
```diff
@@ -317,39 +317,75 @@ async def _call_openai() -> list[float]:
             fn=_call_openai,
         )
 
-    async def simple_batch_embed(self, texts: list[str]) -> list[list[float]]:
+    def _truncate_to_token_limit(self, text: str) -> tuple[str, int]:
+        """Return a prefix of `text` whose re-encoded token count fits the cap."""
+        token_ids = self.encoding.encode(text)
```
eisene, line 329: it's not entirely clear here why you can't just do `return token_ids[:keep]` after this line. i agree what you have here is the safe way to proceed given that we use multiple transports and don't feed them raw tokens. it might be worth it adding an explanation to the docstring, otherwise someone in the future will think "i can simplify this function!" and introduce a hard to debug issue

src/crud/representation.py
```diff
@@ -107,7 +107,7 @@ async def save_representation(
                 parent_category="representation",
             ):
                 embeddings = await embedding_client.simple_batch_embed(
-                    observation_texts
+                    observation_texts, on_oversize="truncate"
```
eisene, line 110: by default it looks like it still raises. sure we don't want to change the default on the `embedding_client`?

eisene (commented):
a few minor comments

src/crud/representation.py
akattelu: Yeah i think so because in the other cases we pre-chunk and I'd want to know if things we're going wrong there. I did add `on_oversize="truncate"` in two more locations in `document.py` though.

src/embedding_client.py
akattelu: added a line of documentation

src/deriver/deriver.py
eisene: Resolved for this PR, gonna leave the convo here visible so that it can be easily found when we upgrade the Python version.

eisene (approved):
lgtm

PR #1020, opened 2026-08-13
feat: adds workspace headers to MCP calls and adds list/create workspace tools

ajspig:
* add workspace_id on tools
* improve list_workspace and add create_workspace tool

I compared the flow of using the MCP server (via cursor) to save a conversation via the existing MCP server and with this PR. This change does add two additional tool calls at the beginning (list_workspace and then create_workspace), but after doing that initially it didn't run those tool calls again and kept using the same workspace_id. 

One current quirk in the Honcho SDK is that it sets a default workspace id if the user doesn't set that. created an additional ticket to updated the DX. 

## Summary by CodeRabbit

- **New Features**
  - Added workspace-scoped MCP tool operations using `workspace_id`.
  - Added workspace discovery, creation, inspection, pagination, and metadata support.
  - Added optional connection-level workspace selection via `X-Honcho-Workspace-ID`.
  - Added workspace-aware peer, session, conclusion, and system operations.
  - Workspace selection prioritizes tool inputs, then the connection-level default, with clear errors when unavailable.

- **Documentation**
  - Updated MCP guides, setup instructions, tool descriptions, and examples to explain workspace handling and selection.

### DIFF
--- a/mcp/src/tools/workspace.ts
+++ b/mcp/src/tools/workspace.ts
@@ -1,13 +1,26 @@
 import { z } from "zod";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
-import type { PageResponse, WorkspaceResponse } from "@honcho-ai/sdk";
+import {
+  HonchoError,
+  type PageResponse,
+  type WorkspaceResponse,
+} from "@honcho-ai/sdk";
 import type { ToolContext } from "../types.js";
+import { resolveWorkspaceId } from "../config.js";
 import {
   textResult,
   errorResult,
   formatMessages,
   workspaceIdSchema,
 } from "../types.js";
+
+function withAdminKeyHint(prefix: string, e: unknown): string {
+  const message = e instanceof Error ? e.message : String(e);
+  const denied =
+    e instanceof HonchoError && (e.status === 401 || e.status === 403);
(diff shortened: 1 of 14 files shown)


mcp/src/tools/workspace.ts
```diff
@@ -69,6 +107,47 @@ export function register(server: McpServer, ctx: ToolContext) {
...
+          },
+        );
+        return textResult(formatWorkspace(workspace));
+      } catch (e) {
+        return errorResult(
+          `Failed to create workspace: ${e instanceof Error ? e.message : String(e)}`,
+        );
+      }
+    },
+  );
```
akattelu, line 161: in this case we are deliberately avoiding using the `X-Honcho-Workspace-ID` header which is good, but might be good to note in the error msg if client doesn't have access. Like this operation may only be possible when using admin key? its ok if we note that only in the error msg or something

akattelu (commented):
PR content looks good, but should just make sure we return a good error message in all the cases where workspace ID aren't passed - we could note that either `X-Honcho-Workspace-ID` needs to be present or the workspace id should be used on the next call. Could consider putting that logic in `clientFor`?

akattelu (dismissed):

akattelu (approved):

PR #1029, opened 2026-08-17
Running Honcho Locally via Honcho CLI 

ajspig:
Adds `honcho start` / `stop` / `status` so you can run a personal Honcho stack in Docker (API, deriver, Postgres, Redis)

**Profiles** live under `~/.honcho/profiles/<name>/` (default name: `local`). That directory holds `profile.json` (ports + pinned image digest), `.env`, `config.toml`, compose, and `init.sql`. They are not project-scoped and are not committed. (config.toml is just a copy of config.toml.example) 

**Image:** first start pulls `ghcr.io/plastic-labs/honcho:latest` and pins in `profile.json`. That tag is published by [`.github/workflows/docker-build.yml`](.github/workflows/docker-build.yml) on push to `main` (`latest` only on the default branch; `v*` tags get a version tag). `--image` can override.

`honcho start` does **not** rewrite `environmentUrl` in `~/.honcho/config.json`

Optional TTY wizard: `honcho start --setup basic` (provider + one chat model) or `--setup advanced` (embeddings, deriver/dialectic models, dreams, flush). Writes `.env` overrides (env wins over toml). Re-running `--setup` on a running stack recreates `api` + `deriver`. Hand-edits still need `honcho stop && honcho start`.

## Summary by CodeRabbit

* **New Features**
  * Added `honcho start`, `honcho stop`, and `honcho status` for local Docker-based Honcho stack management.
  * Added named profiles, automatic port handling, health checks, JSON status output, image pinning, persistent configuration, and optional volume wiping.
  * Added interactive setup for LLM, embedding, and feature settings with OpenAI, Anthropic, and Gemini support.
  * Local startup preserves global server configuration and reports service endpoints.

* **Documentation**
  * Added command references, setup guidance, profile behavior, configuration options, and local URL instructions.


docs/snippets/cli-commands.mdx
```diff
@@ -505,6 +505,75 @@ honcho session view [<session_id>]
...
+configured server URL — pass HONCHO_BASE_URL to talk to this stack.
+``--setup basic`` or ``--setup advanced`` runs an interactive config wizard.
+
+```bash
+honcho start
+```
+
+<ParamField path="--inference" type="string" default="cloud">
+  Where the deriver runs LLMs: cloud now; local and hybrid later.
+</ParamField>
```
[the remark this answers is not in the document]
akattelu: yeah makes sense we should remove

docs/v3/documentation/reference/cli.mdx
```diff
@@ -22,10 +22,31 @@ uvx honcho-cli
...
 ```bash
 honcho init        # confirm/set apiKey + Honcho URL in ~/.honcho/config.json
+honcho start       # optional: local API + deriver + Postgres + Redis (Docker)
 honcho doctor      # verify your config + connectivity
 honcho             # show banner + command list
 ```
 
+## Local stack
+
+`honcho start` runs a personal Honcho server on your machine via Docker (API, deriver, Postgres, Redis) using cloud LLM inference. Files live under `~/.honcho/profiles/local/`. First start pins the Honcho image to a digest and copies that image's `config.toml.example` into the profile directory. Later starts do not overwrite `config.toml`.
```
akattelu: In these docs we use the words "cloud inference" which is correct and makes sense, but I wonder if we should distinguish it to note that it doesn't run on honcho's managed offering at all.

honcho-cli/README.md
```diff
@@ -31,13 +32,41 @@ honcho             # show banner + command list
...
+`honcho start` does **not** change `environmentUrl` in `~/.honcho/config.json` (that file is shared with plugins). To talk to the local stack for one command:
+
+```bash
+HONCHO_BASE_URL=http://127.0.0.1:8000 honcho workspace list
+```
+
+To make local the default, run `honcho init --base-url http://127.0.0.1:8000`.
+
+```bash
+honcho start --llm-api-key "$LLM_OPENAI_API_KEY"
```
akattelu: nit: The `--llm-api-key` is a little ambiguous because its only openai compatible endpoints, but we do support gemini and anthropic too.

docs/v3/documentation/reference/cli.mdx
```diff
@@ -22,10 +22,31 @@ uvx honcho-cli
...
+
+Pass `--setup basic` or `--setup advanced` for an interactive wizard that writes curated LLM/feature overrides into the profile `.env` (environment variables win over `config.toml`). This is TTY-only. `basic` covers provider and chat model; `advanced` also covers embeddings, deriver/dialectic models, dreams, and deriver flush. Re-running `--setup` while the stack is up recreates the API and deriver containers.
+
+This does **not** change `environmentUrl` in the shared config file. To talk to the local stack:
+
```
akattelu, line 37: Doesn't have to be in this PR, but I think we could really use a skill (maybe `/configure-honcho`) that knows about all the settings and config variables in env/toml. If we had that we could reference it here, and then users could configure and debug honcho better.

honcho-cli/src/honcho_cli/_help.py
```diff
@@ -67,6 +67,7 @@ def print_welcome(console: Console) -> None:
 
     start_rows = [
         ("honcho init",   "configure API key and server URL"),
+        ("honcho start",  "run a local Honcho stack (Docker + cloud inference)"),
```
akattelu: nit: we should exclude the cloud inference part here

honcho-cli/src/honcho_cli/commands/stack.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def _ok(msg: str) -> None:
+    if not use_json():
+        _console.print(f"  {ICON_OK}  {msg}")
+
+
+def _fail(msg: str) -> None:
+    if not use_json():
+        _console.print(f"  {ICON_FAIL}  {msg}")
```
akattelu: nit: could put these in a shared file for use across the CLI

honcho-cli/src/honcho_cli/commands/stack.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def _resolve_llm_key(flag: str | None, profile: LocalProfile) -> str:
```
akattelu: Confirming that the behavior for `LLM_ANTHROPIC_API_KEY=… honcho start`  should be to override whatever is in the `config.toml`. So we should be forwarding the keys in the user's `env` (only the ones that are valid keys in our config/settings), then merging with the existing config.toml. With that behavior, I wonder if we need to have the llm api key param here at all? because we could just rely on passing the env var.

honcho-cli/pyproject.toml
```diff
@@ -38,6 +38,10 @@ build-backend = "hatchling.build"
 [tool.hatch.build.targets.wheel]
 packages = ["src/honcho_cli"]
 
+[tool.hatch.build.targets.wheel.force-include]
+"src/honcho_cli/local/templates/docker-compose.yml" = "honcho_cli/local/templates/docker-compose.yml"
```
akattelu, line 42: will this always be available, or do we only have the example file committed to the repo?

honcho-cli/src/honcho_cli/commands/stack.py
ajspig: agreed, good catch

honcho-cli/pyproject.toml
ajspig: Yes. force-include makes sure it lands in the wheel so honcho start can load it after install. 

The repo-root docker-compose.yml.example is only for people cloning Honcho and running compose from source and stays gitignored on purpose.

akattelu (approved):

Issue #993, opened 2026-08-07
Deriver silently discards a batch when structured output fails validation


divyanshpandey-dotcom:
# Deriver silently discards a batch when structured output fails validation

**Repo:** plastic-labs/honcho
**Version:** 3.0.11 (`0bbeb3bc431be074b544657bbf9d3bd587f68628`, 2026-08-05)
**Component:** `src/llm/structured_output.py`, `src/deriver/`
**Severity:** data loss, silent — queue metrics stay green

## Summary

When a deriver LLM response fails JSON repair or schema validation,
`repair_response_model_json()` returns an **empty `PromptRepresentation` as a
success** instead of raising. Three things follow, none of them visible:

1. the configured retries never fire, because nothing propagates;
2. the queue item is marked processed, so the queue drains clean;
3. the deriver logs `"Deriver generated zero observations"` — which is
   **indistinguishable from the model legitimately finding nothing to extract**.

A model with weak JSON adherence therefore loses memory silently while every
dashboard reads healthy. We spent a day attributing a real yield drop to model
capability before finding this, and could only rule it out by patching the
function to log.

## Root cause

`src/llm/structured_output.py`, `repair_response_model_json()`:

```python
except (json.JSONDecodeError, KeyError, TypeError, ValueError):
    final = ""

try:
    return response_model.model_validate_json(final)
except ValidationError:
    if response_model is PromptRepresentation:
        return PromptRepresentation(explicit=[])   # <-- success, empty
    raise
```

`PromptRepresentation` is the **only** response model special-cased this way;
every other model re-raises `ValidationError`.

The deriver calls it via
`honcho_llm_call(..., enable_retry=True, retry_attempts=3)`
(`src/deriver/deriver.py`). `src/llm/api.py` wraps the provider call with
tenacity (`stop_after_attempt(retry_attempts)`) and **no `retry=` predicate**, so
tenacity's default retries on any `Exception`. Because this path raises nothing,
**`retry_attempts=3` is dead for the deriver** — a malformed response is accepted
on the first attempt, and the fallback model (selected on the final attempt) is
never reached either.

The resulting empty representation then hits `src/deriver/deriver.py:201`:

```
Deriver generated zero observations for messages %s:%s in %s/%s!
```

which is the same log line emitted when the model validly returns
`{"explicit": []}`. There is no way to tell the two apart from logs, metrics, or
the queue.

## A second, related silent path in the same function

Fixing only the `ValidationError` branch is **not sufficient**. Because
[...]

Issue #963, opened 2026-08-01
Transport failures permanently burn queue items (processed-with-error, never retried)


poolside-ventures:
**Version:** v3.0.12

When an LLM call fails with a *transport*-class error (connection refused, `APITimeoutError`, 401 from a rotated key), the queue item is marked `processed = true` with the error recorded — the same terminal treatment as a semantic failure. It is never retried, so the affected messages silently never enter the peer representation.

We've lost work to this twice: 160 items during an egress-firewall outage (calls timed out; queue burned while the model was fine), and again on a desktop deployment when the DB briefly outlived the LLM endpoint. Our current workaround is resetting `processed=true AND error IS NOT NULL` rows on every service start, which works but retries semantic failures too.

**Suggested fix:** classify errors at the point of marking — transport/auth failures leave the item unprocessed (ideally with a retry counter and backoff), semantic failures stay terminal. Happy to send a PR here as well.

Related operational context: #962.

akattelu:
Fixed in #1033

PR #1033, opened 2026-08-18
fix(deriver): fix create_documents deadlock 

akattelu:
Related: plastic-labs/honcho#963, plastic-labs/honcho#728, plastic-labs/honcho#993.

## Problem

Two deriver work units can write the same `(workspace, observer, observed)` collection concurrently: work-unit keys include session but not observer, so one peer active in two sessions produces two writers, and the dreamer adds a third through a different key. `create_documents` issued per-row `times_derived` reinforcement UPDATEs in batch order inside one long transaction, so two writers touching the same rows in opposite orders deadlocked.

The damage was worse than the deadlock itself: the error was swallowed per-document by `except Exception: continue`, the loop cascaded `PendingRollbackError`s against the dead session, the final commit failed so the entire batch's documents were lost, and the queue item was marked `processed=true` — terminal, never retried, and reported as completed by `get_queue_status`.

## What changed

* **Id-ordered row locks** (`src/crud/document.py`): reinforcements and semantic replace/reject writes are collected during the loop and applied after a `SELECT … WHERE id IN (…) ORDER BY id FOR UPDATE`. Every writer acquires the same rows in the same order, so waits stay linear. Writers to different documents stay concurrent. In-memory `pending_times_derived` keeps later semantic decisions consistent with earlier exact reinforces in the same batch.
* **Vector-query hoist**: on the external-vector-store path, dup-candidate IDs are resolved concurrently before the first DB statement. `create_documents` classifies via `_semantic_dup_decision` (no mid-loop writes); `is_rejected_duplicate` still applies immediately for standalone callers.
* **Stop continuing through an aborted transaction**: the per-document handler now splits `except SQLAlchemyError: rollback; raise` from the per-document skip (`except Exception: continue`). The outer handler gains `except DBAPIError: rollback; raise` so a deadlock at the final commit leaves the caller's session clean.
* **Bounded queue retry** (`src/deriver/queue_manager.py`): `_handle_processing_error` classifies errors. Transient ones (deadlock, serialization failure, lock/statement timeout, lost connection, provider transport) leave items unprocessed, skip Sentry, and release the work unit for re-claim, bounded by an in-process counter (`MAX_RETRYABLE_ATTEMPTS = 3` per work unit). Terminal errors keep today's burn-one-item behavior. Reprocessing is safe because a retried batch re-derives the same observations and exact dedup collapses them into reinforcement.
* **New** `src/utils/retryable_errors.py`: driver-agnostic sqlstate classification (`40001`, `40P01`, `55P03`, `57014`, `08xxx`), `connection_invalidated`, and httpx transport errors, walking `__cause__` chains.

## Verification

* Concurrency regression: two gathered `create_documents`, same collection, 20 rows in opposing orders, 5 rounds.
* Aborted-transaction raises and commits nothing; per-document skip preserved; external candidates resolve before the first DB statement; retryable error leaves items unprocessed after exactly one batch fetch; exhaustion and non-retryable errors match today's terminal behavior; counter cleared on success; DB-free classification tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document duplicate handling during concurrent processing, including cases with missing embeddings or unavailable candidates.
  * Preserved derivation counts during document replacement.
  * Added bounded retries with jittered backoff for temporary database and network failures while preserving unprocessed work.
  * Improved consistency when handling permanent processing failures.
* **Tests**
  * Added coverage for duplicate resolution, concurrency, database errors, retry behavior, and transient error classification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate detection when embeddings or external candidate data are unavailable.
  * Individual candidate-resolution failures no longer prevent valid documents from being created.
  * Document creation errors now roll back safely and return clear validation failures.
  * Preserved derivation counts during document replacement.
* **Reliability**
  * Retry attempts now persist across processing restarts and work-unit reassignment.
  * Transient database, connection, and timeout errors are retried automatically, while permanent errors fail promptly.

## Summary by CodeRabbit

- **New Features**
  - Improved document processing so fallback documents can still be added when reinforcement targets are unavailable or removed.
  - Enhanced duplicate detection to handle missing embeddings and externally identified candidates more reliably.
  - Added consistent classification of transient database, connection, and timeout errors.

- **Bug Fixes**
  - Prevented document-processing failures from unnecessarily aborting unrelated documents.
  - Preserved retry attempts when queue work is reclaimed or processed by another worker.
  - Automatically retried transient observer-save failures while continuing to avoid retries for permanent failures.


src/deriver/queue_manager.py
```diff
@@ -1056,6 +1108,89 @@ async def get_queue_item_batch(
...
+            try:
+                return max(0, int(raw))
+            except (TypeError, ValueError):
+                return 0
+
+    async def _set_work_unit_retry_attempts(
+        self, work_unit_key: str, attempts: int
+    ) -> None:
+        """Persist the shared attempt count on the oldest unprocessed item."""
+        async with tracked_db("set_work_unit_retry_attempts") as db:
```
VVoruganti, line 1165: **Blocker — this writes a key that every payload model forbids, so the first transient error on any non-`representation` task becomes a permanent one.**

`BasePayload` sets `model_config = ConfigDict(extra="forbid")` (`src/utils/queue_payload.py:12`), and `consumer.py` validates the raw DB payload for all eight task types except `representation`:

1. A `summary` task hits a transient error → this method writes `_retry_attempts: 1` into its payload and leaves the item **unprocessed**.
2. Re-claim → `get_next_queue_item` returns the row verbatim, no scrubbing (`queue_manager.py:842`).
3. `SummaryPayload(**queue_payload)` → `extra_forbidden` → re-raised as `ValueError` (`consumer.py:93`).
4. `ValueError` is not retryable → terminal → item burned, recorded as `Invalid payload structure`, which points a debugger at a producer bug that doesn't exist.

The budget is never actually spent — attempt 2 always dies at parse time. `representation` escapes only because its batch path reads `payload.get(...)` (`queue_manager.py:683`) instead of validating, and that's the only `task_type` the new tests seed, so the suite structurally can't catch this.

Fix is one line, but it lives in `consumer.py` (not in this diff), so no suggestion button:

```python
# src/deriver/consumer.py:47
queue_payload = QueueManager._payload_without_retry_attempts(queue_item.payload)
```

Worth considering the alternative: store the counter off-payload entirely (`active_queue_sessions` is already keyed by `work_unit_key`). That deletes this method, `_get_work_unit_retry_attempts`, `_clear_work_unit_retry_attempts`, and the payload-strip in both mark methods.

src/crud/document.py
```diff
@@ -1147,103 +1239,182 @@ async def create_observations(
...
+        if op.kind == "replace":
+            if row is not None and row.deleted_at is None:
+                row.deleted_at = now
+            continue
+        # reinforce
+        if row is None or row.deleted_at is not None:
+            if op.fallback_document is not None:
+                fallbacks.append(op.fallback_document)
+            continue
+        row.times_derived = max(row.times_derived + 1, op.incoming_times_derived)
```
VVoruganti, line 1337: **A replace and a reinforce on the same row in one batch resurrects the copy the batch just rejected.**

`locked` holds live ORM objects, so the `replace` branch mutates `row.deleted_at` in place and a later `reinforce` op on the same id reads its own batch's write as "target vanished" and inserts the fallback.

Concretely: doc1 is semantically superior to existing row E → `replace(E)`; doc2 is an exact-content match of E → `reinforce(E, fallback=doc2)`. Ops apply in order: E is soft-deleted, then the reinforce sees `deleted_at is not None` and inserts doc2. You end up with the superior replacement **and** a fresh copy of the content the batch just decided was inferior — two near-duplicate live rows, `times_derived` reset to 1 on the copy. doc2 is also counted in `exact_dup_existing_count` *and* returned in `created_documents`.

The reverse ordering is fine (reinforce bumps, then replace soft-deletes; content survives via the replacement).

The fix is to distinguish "stale before we started" from "deleted by an earlier op in this same batch":

```suggestion
    stale_at_lock = {
        op.document_id
        for op in ops
        if (locked_row := locked.get(op.document_id)) is None
        or locked_row.deleted_at is not None
    }
    for op in ops:
        row = locked.get(op.document_id)
        if op.kind == "replace":
            if row is not None and row.deleted_at is None:
                row.deleted_at = now
            continue
        # reinforce
        if op.document_id in stale_at_lock:
            if op.fallback_document is not None:
                fallbacks.append(op.fallback_document)
            continue
        if row is None or row.deleted_at is not None:
            # An earlier op in this batch replaced this row; the replacement
            # carries the content forward, so don't resurrect the inferior copy.
            continue
        row.times_derived = max(row.times_derived + 1, op.incoming_times_derived)
```

The dropped reinforcement here is a one-count loss on a soft ranking signal — not worth threading into the already-constructed replacement document.

src/crud/document.py
```diff
@@ -1147,103 +1239,182 @@ async def create_observations(
...
+    ops: list[_DocumentRowOp],
+    *,
+    workspace_name: str,
+    observer: str,
+    observed: str,
+) -> list[schemas.DocumentCreate]:
+    """Lock target rows by id, apply ops, return fallbacks for vanished targets."""
+    if not ops:
+        return []
+    ids = sorted({op.document_id for op in ops})
```
VVoruganti, line 1300: Nit, but this is the single most load-bearing line in the PR and it has no comment. `sorted()` feeding an `IN` clause *looks* redundant — someone will delete it. Worth pinning why it isn't, especially since `Document.id` is a random nanoid (`models.py:381`), so id order is uncorrelated with physical order.

```suggestion
    # sorted() + the matching ORDER BY below are the deadlock fix: every writer
    # must acquire these rows in the same global order. The IN clause makes this
    # look redundant -- it is not, Postgres locks in scan order.
    ids = sorted({op.document_id for op in ops})
```

Same for `populate_existing=True` on line 1305 — it's what makes the Python-side `max()` read-modify-write safe now that `func.greatest()` is gone. The new concurrency test would catch its removal, but a one-line comment saves the next reader the trip.

src/deriver/queue_manager.py
```diff
@@ -579,24 +587,51 @@ async def _handle_processing_error(
...
-        to incrementally attempt to process the batch while still maintaining progress in a work unit.
+        Handle a processing error. Returns True when the caller should stop
+        processing and release the work unit for a later re-claim.
+
+        Transient errors (is_retryable_error) get up to MAX_RETRYABLE_ATTEMPTS
+        attempts per work unit: items stay unprocessed with no error recorded.
+        The attempt count lives on the oldest unprocessed queue item so a
+        different deriver instance continues the same budget after reclaim.
+        Reprocessing is safe because a retried batch re-derives the same
+        observations and exact dedup collapses them into reinforcement.
```
VVoruganti: This overstates it, and someone will raise `MAX_RETRYABLE_ATTEMPTS` on the strength of it.

A retry re-enters `process_representation_tasks_batch`, which makes a **fresh** `honcho_llm_call` (`deriver.py:149`) — output is non-deterministic, nothing pins or caches it. Exact dedup is normalized string equality, so any wording drift produces a new document rather than reinforcement; the net that actually catches paraphrases is semantic dedup, which is gated on `DERIVER.DEDUPLICATE`. And when exact dedup *does* hit, it bumps `times_derived`, which feeds `query_documents_most_derived` ordering and the Dreamer's surprisal prioritization. `create_documents` also commits partway through, so a post-commit transient error re-derives over already-committed rows.

```suggestion
        Reprocessing is at-least-once, not idempotent: the batch is re-derived
        by a fresh LLM call, so identical text collapses via exact dedup and
        near-identical text via semantic dedup. Retries can therefore inflate
        times_derived and double-count LLM telemetry -- acceptable because the
        alternative is dropping the batch.
```

src/deriver/queue_manager.py
```diff
@@ -579,24 +587,51 @@ async def _handle_processing_error(
...
+                await self._set_work_unit_retry_attempts(work_unit_key, attempts)
+                logger.warning(
+                    "Transient error %s for work unit %s (attempt %d/%d); leaving items unprocessed for retry",
+                    context,
+                    work_unit_key,
+                    attempts,
+                    MAX_RETRYABLE_ATTEMPTS,
+                    exc_info=error,
+                )
+                return True
```
VVoruganti, line 647: **Both of these open their own `tracked_db` sessions and neither is guarded**, unlike the terminal path below, whose `mark_queue_item_as_errored` sits inside a `try/except Exception as mark_error`.

If the DB is flapping, a raise here propagates out of `_handle_processing_error`, past the inner handler, into the outer `except Exception` at line ~795 — which neither `break`s nor sleeps. Control returns to the top of the `while not self.shutdown_event.is_set()` loop, re-fetches a batch, and re-runs a **full LLM call** immediately. No backoff on that path, and `attempts` was never persisted, so the budget never advances and the terminal path is never reached.

```suggestion
        if is_retryable_error(error):
            try:
                attempts = await self._get_work_unit_retry_attempts(work_unit_key) + 1
                if attempts < MAX_RETRYABLE_ATTEMPTS:
                    await self._set_work_unit_retry_attempts(work_unit_key, attempts)
                    logger.warning(
                        "Transient error %s for work unit %s (attempt %d/%d); leaving items unprocessed for retry",
                        context,
                        work_unit_key,
                        attempts,
                        MAX_RETRYABLE_ATTEMPTS,
                        exc_info=error,
                    )
                    return True
            except Exception:
                logger.exception(
                    "Retry-counter I/O failed for work unit %s; releasing %s without recording an attempt",
                    work_unit_key,
                    context,
                )
                return True
```

Deliberate trade-off in that `except`: an unpersisted counter means the budget doesn't advance while the DB is down, so retries are bounded only by the backoff. That's still strictly better than the current unbounded spin with no backoff at all. An in-process fallback counter keyed by `work_unit_key` would give you a hard cap if you want one.

src/deriver/queue_manager.py
```diff
@@ -672,14 +708,23 @@ async def process_work_unit(self, work_unit_key: str, worker_id: str) -> None:
                                 await self.mark_queue_items_as_processed(
                                     items_to_process, work_unit_key
                                 )
+                                await self._clear_work_unit_retry_attempts(
+                                    work_unit_key
+                                )
```
VVoruganti: **Post-commit cleanup inside the processing `try` turns a successful batch into a processing failure.**

`mark_queue_items_as_processed` has already committed at this point (and already strips `_retry_attempts` from the items it marks). If this second transaction fails, the exception lands in `_handle_processing_error` with `items_to_process` — items that are **already processed**. Consequences:

- the retryable branch writes the counter onto the *oldest unprocessed* item, which is now the **next** batch's item, consuming its budget;
- `queue_item_count += ...` hasn't run yet, so the `queue_item_count > 0` gate suppresses the `queue.empty` webhook;
- the terminal branch calls `mark_queue_item_as_errored` on an already-processed item, overwriting a success with an error;
- and it locks + loads every remaining unprocessed item in the work unit after **every** successful batch.

The call is also redundant: the counter always lives on the oldest unprocessed item, and batches are fetched in id order, so the item carrying it is in the batch being marked. Dropping it:

```suggestion

```

VVoruganti (commented):
Reviewed the deadlock fix end to end. **The core change is right** — `_apply_document_row_updates` is the correct shape, and the new `test_concurrent_reinforcement_does_not_deadlock` is a genuine regression test: separate sessions off `db_engine`, real interleaving, and the `times_derived == 1 + 2 * N_ROUNDS` assertion pins the *absence of lost updates* rather than just "didn't crash". `tests/utils/test_retryable_errors.py` is near-complete. `ruff` and `basedpyright` are clean.

Two things block merge, though, and neither is visible from the diff alone.

---

### 1. The retry machinery is unreachable from the path it was built for

`src/deriver/deriver.py:239` (not in this diff) swallows the exception this PR carefully re-raises:

```python
            except Exception as e:
                logger.error(
                    "Failed to save representation for observer %s: %s", observer, e
                )
```

Chain: `queue_manager.py:698` → `consumer.py:193` → `deriver.py:39` → **`deriver.py:239`** → `representation.py:197` → `document.py:862 (except DBAPIError: rollback; raise)`.

The new `except DBAPIError` re-raises straight into that bare `except Exception`, which logs without `exc_info` and continues the per-observer loop. Nothing after the loop re-raises — `successful_observer_count` only feeds `observer_count` on the telemetry event. So `process_representation_tasks_batch` returns normally, `queue_manager` marks the batch **processed** and clears the budget.

That is the PR description's own bug, relocated one frame up: batch lost, item burned, `get_queue_status` reports success, `is_retryable_error` never called. Suggested fix:

```python
            except Exception as e:
                logger.error(
                    "Failed to save representation for observer %s", observer, exc_info=True
                )
                if is_retryable_error(e):
                    raise
```

Guarding on `is_retryable_error` keeps today's per-observer tolerance for terminal failures. Every retry test patches `process_representation_batch`, injecting the error *above* the swallow, and the document tests stop *below* it — nothing crosses the boundary, so a test that patches `save_representation` to raise `OperationalError("40P01")` and asserts the item stays unprocessed would fail today.

### 2. `_retry_attempts` in `payload` vs `extra="forbid"`

Left inline on `_set_work_unit_retry_attempts`. Short version: it permanently burns the first transient error on all seven non-`representation` task types, and the new tests only seed `representation`.

---

### Inline comments

Seven of the comments below carry suggestion blocks. The two P2s (same-batch resurrect, post-commit cleanup) and the hot-loop guard are the substantive ones; the two `mark_*` reverts are net deletion.

### Not blocking, worth deciding

- **The deadlock class isn't fully closed.** `_apply_document_row_updates` is the only ordered document-row locker in `src/`. `scope_backfill.py:427` and `:298` still soft-delete/restore the same collections by predicate with no ordering, on disjoint work-unit keys, so nothing serializes them against a concurrent `create_documents`. Because `Document.id` is a random nanoid, id order vs physical order is roughly a coin flip per row pair rather than a rare plan quirk. Also unordered and pre-existing: `workspace.py:437`, `session.py:539`. Worth either ordering those or stating the residual in the description — right now it reads as if the class is closed. Note `ids = sorted(...)` alone is a no-op fix there: `WHERE id IN (...)` discards list order, so it has to be an explicit `SELECT ... ORDER BY id FOR UPDATE`.
- **External-store resolve failure silently disables semantic dedup** (`document.py:566`). `except Exception: log; candidates = []`, and `[]` is the same sentinel the legitimate "no merge partner" branch uses. A turbopuffer outage inserts duplicates indefinitely and reports `semantic_dup_rejected_count=0`, which reads as a quality improvement. The pgvector path fails loud by contrast. `httpx.TransportError` is already classified retryable, so raising would get a correct retry for free — but that's a product call. `test_resolve_failure_skips_semantic_without_query_documents` currently locks in the silent behavior.
- **`is_rejected_duplicate` (`document.py:1388`) and `delete_document` (`document.py:877`) have no `src/` callers.** The former is a shim over `_semantic_dup_decision` that then applies writes the *old* way — `func.greatest` + immediate `flush()`, unordered — twenty lines from its correct replacement. Five tests still point at it, so the reinforce/replace semantics that actually ship are less covered than the coverage suggests.
- **Test gaps:** the non-`representation` retry branch (`queue_manager.py:729-758`) is duplicated code with zero tests; the cross-instance test overwrites the counter with `_set_work_unit_retry_attempts` instead of letting the second manager read what the first wrote; nothing asserts `is_retryable_error(excinfo.value)` on what `create_documents` actually raises.

src/deriver/queue_manager.py
```diff
@@ -696,14 +741,21 @@ async def process_work_unit(self, work_unit_key: str, worker_id: str) -> None:
                                 await self.mark_queue_items_as_processed(
                                     [queue_item], work_unit_key
                                 )
+                                await self._clear_work_unit_retry_attempts(
+                                    work_unit_key
+                                )
```
VVoruganti: **Post-commit cleanup inside the processing `try` turns a successful batch into a processing failure.**

`mark_queue_items_as_processed` has already committed at this point (and already strips `_retry_attempts` from the items it marks). If this second transaction fails, the exception lands in `_handle_processing_error` with `items_to_process` — items that are **already processed**. Consequences:

- the retryable branch writes the counter onto the *oldest unprocessed* item, which is now the **next** batch's item, consuming its budget;
- `queue_item_count += ...` hasn't run yet, so the `queue_item_count > 0` gate suppresses the `queue.empty` webhook;
- the terminal branch calls `mark_queue_item_as_errored` on an already-processed item, overwriting a success with an error;
- and it locks + loads every remaining unprocessed item in the work unit after **every** successful batch.

The call is also redundant: the counter always lives on the oldest unprocessed item, and items are fetched in id order, so the item carrying it is in the batch being marked. Dropping it:

```suggestion

```

src/deriver/queue_manager.py
```diff
@@ -1064,12 +1199,23 @@ async def mark_queue_items_as_processed(
...
+                await db.execute(
+                    update(models.QueueItem)
+                    .where(models.QueueItem.id == queue_item.id)
+                    .values(
+                        processed=True,
+                        payload=self._payload_without_retry_attempts(
+                            queue_item.payload
+                        ),
+                    )
+                )
```
VVoruganti: This went from one bulk `UPDATE` to `SELECT ... FOR UPDATE` + one `UPDATE` per row, solely to strip a payload key in Python. Two costs: it's N+1 per batch, and it's a **new unordered multi-row `FOR UPDATE` on `queue`** in a PR about deadlock elimination. Reachable when `cleanup_stale_work_units` (line ~309) lets a second worker claim a `work_unit_key` the first is still processing — worker A's processed subset and worker B's pending set overlap with no common lock order.

Once `_retry_attempts` is stripped on read (see the blocker comment on `_set_work_unit_retry_attempts`), a stale counter on a processed row is inert and `queue_cleanup` reaps those rows anyway — so this can just go back to what it was on `main`:

```suggestion
            await db.execute(
                update(models.QueueItem)
                .where(models.QueueItem.id.in_(item_ids))
                .where(models.QueueItem.work_unit_key == work_unit_key)
                .values(processed=True)
            )
```

If you'd rather keep the strip, do it server-side in the same single statement with the JSONB `-` operator instead of a lock-and-loop — but note it needs an explicit `cast(..., Text)` or Postgres can't resolve the operator.

src/deriver/queue_manager.py
```diff
@@ -1095,11 +1241,24 @@ async def mark_queue_item_as_errored(
...
+                return
+            await db.execute(
+                update(models.QueueItem)
+                .where(models.QueueItem.id == queue_item.id)
+                .values(
+                    processed=True,
+                    error=error[:65535],  # Truncate to TEXT limit
+                    payload=self._payload_without_retry_attempts(queue_item.payload),
+                )
             )
```
VVoruganti: Same as above — reverts to a single statement, no `FOR UPDATE`, no lock-ordering hazard:

```suggestion
            await db.execute(
                update(models.QueueItem)
                .where(models.QueueItem.id == item.id)
                .where(models.QueueItem.work_unit_key == work_unit_key)
                .values(processed=True, error=error[:65535])  # Truncate to TEXT limit
            )
```

One behavior note: this drops the `queue_item is None` early return, so a vanished item now still bumps `ActiveQueueSession.last_updated`. Harmless, but flagging it since it's a real difference.

Accepting this and the previous suggestion leaves `_payload_without_retry_attempts` with no callers — delete it too.

VVoruganti (approved):

PR #1036, opened 2026-08-20
fix(filter): make ne on jsonb metadata keys null-safe

serhiizghama:
Filtering a JSONB metadata key with `ne` silently drops rows where that key isn't present. `{"metadata": {"priority": {"ne": "high"}}}` compiles the key comparison to plain `<>`, but `(metadata ->> 'priority')` is NULL when the key is absent, and `NULL <> 'high'` is NULL rather than true — so those rows fall out, even though a row without `priority` is clearly not equal to `"high"`.

The scalar-column `ne` path already handles this: it uses `IS DISTINCT FROM` with a comment spelling out exactly this NULL trap. The nested-metadata operator map in `_build_comparison_condition` was just missed when that change went in, so the two `ne` paths disagree. Switched it to `is_distinct_from` to match — identical to `<>` whenever no NULL is involved. Covers both the string and the numeric-cast branch.

Added two tests asserting the compiled predicate renders `IS DISTINCT FROM` (not `!=`) for string and numeric `ne` on a metadata key; both fail on the old operator map.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated “not equal” filters on nested JSON metadata to include records where the selected field is missing or null.
  - Applied the fix to both text and numeric metadata values.

- **Tests**
  - Added coverage for null-safe filtering of nested string and numeric metadata fields, including missing keys.

### DIFF
--- a/src/utils/filter.py
+++ b/src/utils/filter.py
@@ -689,7 +689,9 @@ def _build_comparison_condition(
                 "lte": lambda a, v: a <= v,
                 "gt": lambda a, v: a > v,
                 "lt": lambda a, v: a < v,
-                "ne": lambda a, v: a != v,
+                # IS DISTINCT FROM, not <>: (metadata ->> key) is NULL for an
+                # absent key, and `NULL <> v` is NULL, so <> drops those rows.
+                "ne": lambda a, v: a.is_distinct_from(v),
             }
             return operator_map[operator](safe_accessor, safe_value)
         except Exception as e:
--- a/tests/test_advanced_filters.py
+++ b/tests/test_advanced_filters.py
@@ -4,13 +4,19 @@ comparison operators, and wildcards across multiple models.
 """
 
 from datetime import datetime, timedelta, timezone
-from typing import Any
+from typing import Any, TypedDict
 
 import pytest
 from fastapi.testclient import TestClient
 from nanoid import generate as generate_nanoid
 
 from src.models import Peer, Workspace
+
+
+class MessageConfig(TypedDict):
+    content: str
+    peer_id: str
+    metadata: dict[str, Any]
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


akattelu (changes requested):
Hey @serhiizghama, thanks for opening this. 

One ask on the tests. The two new cases in `tests/utils/test_filter.py` only pin the compiled SQL string (`IS DISTINCT FROM`
 present, `!=` absent). That locks the operator map entry, but this class of bug builds and executes cleanly either way. The failure
 mode is wrong row sets under three-valued logic, so the thing that actually guards it is counting returned rows when the key is absent.

   We already have that shape for scalar columns in `tests/routes/test_conclusions.py::test_negation_includes_conclusions_with_no_session`. Nested metadata doesn't have an equivalent
 yet, and `tests/test_advanced_filters.py::test_comparison_operators_filters` exercises `ne` only on rows that all have the key.

   Suggested shape: keep the compile checks if you like, and add a live `messages/list` case in `tests/test_advanced_filters.py`
 with four messages (priority high, priority low, metadata missing the key, empty metadata). Assert that `{"metadata": {"priority":
 {"ne": "high"}}}` and `{"metadata": {"score": {"ne": 5}}}` both return the low + missing-key rows and drop only the matching ones.
 Happy to drop a patch on if that's easier. I don't have push access to the fork.

serhiizghama:
Good call, done. Added a live `messages/list` case in `test_advanced_filters.py` with four messages — priority high, priority low, one with metadata but no priority/score key, one with empty metadata — and asserted `ne` on both a string and numeric key returns the low + missing-key + empty-metadata rows, dropping only the actual match. Ran it against the old `!=` operator to confirm it fails there. Kept the compile-only checks too since they're cheap and pin the SQL shape.

akattelu (changes requested):
thanks @serhiizghama looks mostly good but you should address the coderabbit / basedpyright fixes

serhiizghama:
Fixed — added a `MessageConfig` TypedDict (content/peer_id/metadata) and annotated `message_configs` with it, same as the pattern in tests/bench. basedpyright's clean on the file now (0 warnings), ruff too.

akattelu (approved):

PR #1037, opened 2026-08-20
feat(conclusions): support cross-peer semantic query on the pgvector backend

### LINKED ISSUE #520: Cross-peer semantic search for conclusions

ademczuk:
## Description

Implements the minimum viable approach scoped in #520 (pgvector-only, informative error on external stores). On pgvector deployments, `POST /v3/workspaces/{workspace_id}/conclusions/query` may now omit `observer`/`observed` in `filters` to run semantic search across all peer relationships in the workspace. External vector store deployments (Turbopuffer/LanceDB) keep the old requirement, and omitting either parameter there now returns a 422 explaining the pgvector-only limitation, since namespaces are derived from the (workspace, observer, observed) triple and cross-peer search would need a namespace fan-out.

Prior art: #492 (closed for inactivity) and #541 (closed as duplicate) attempted this before the scope was settled; this PR implements exactly the scope from the issue body. External-store fan-out is deliberately not included.

## Proofs

Tests ran in WSL (Ubuntu, Python 3.12) against a throwaway `pgvector/pgvector:pg16` Docker container with `POSTGRES_HOST_AUTH_METHOD=trust` (same as CI). The suite cannot run on native Windows: `src/telemetry/reasoning_traces.py` imports Unix-only `fcntl` at module load (pre-existing, unrelated to this change).

```
$ uv run --frozen pytest tests/crud/ tests/routes/ tests/utils/test_agent_tools.py tests/test_session_allowlist.py -q
639 passed, 106 warnings in 28.01s
```

This covers the 4 new/updated tests: cross-peer query returns documents from two distinct (observer, observed) pairs and observer-only scoping still applies (crud level); end-to-end 200 over two peer pairs and 422 when observer/observed are omitted under external-store settings (route level, monkeypatched `VECTOR_STORE` settings).

```
$ uv run ruff check src/crud/document.py src/routers/conclusions.py tests/crud/test_document.py tests/routes/test_conclusions.py
All checks passed!
$ uv run ruff format --check <same 4 files>
4 files already formatted
$ uv run basedpyright src/crud/document.py src/routers/conclusions.py
0 errors, 1 warning, 0 notes
```

(The 1 warning is pre-existing: partially unknown `apaginate` type from fastapi-pagination on an untouched line. A full-project `basedpyright` run reports 22 errors, all in `src/vector_store/lancedb.py`, `tests/vector_store/` and `tests/bench/` -- lancedb typing in this environment; none in touched files.)

Not run: the remainder of the test suite beyond the paths above, TypeScript SDK tests (no TS changes), and the pre-commit-only hooks (bandit etc.; not part of CI -- the diff introduces no asserts/exec/hardcoded secrets).

## Checklist

Fixes #520

## Summary by CodeRabbit

* **New Features**
  * Semantic document and conclusion searches can omit peer filters with the pgvector backend, enabling cross-peer results.
  * Searches can be scoped using either peer filter independently.
  * Cross-peer results remain limited to the selected workspace.

* **Bug Fixes**
  * External vector-store searches now require both peer identifiers and return clear validation errors when either is missing.
  * Improved validation responses for unfiltered and partially filtered searches across supported backends.

### DIFF
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -291,22 +291,28 @@ async def fetch_documents_by_ids(
 async def _query_documents_pgvector(
     db: AsyncSession,
     workspace_name: str,
-    observer: str,
-    observed: str,
+    observer: str | None,
+    observed: str | None,
     embedding: list[float],
     filters: dict[str, Any] | None,
     max_distance: float | None,
     top_k: int,
 ) -> list[models.Document]:
-    """pgvector similarity search — pure DB operation."""
+    """pgvector similarity search — pure DB operation.
+
+    A None observer/observed omits that constraint, so omitting both searches
+    across all peer relationships in the workspace (cross-peer search, #520).
+    """
     stmt = (
         select(models.Document)
         .where(models.Document.workspace_name == workspace_name)
-        .where(models.Document.observer == observer)
-        .where(models.Document.observed == observed)
         .where(models.Document.embedding.isnot(None))
         .where(models.Document.deleted_at.is_(None))
     )
+    if observer is not None:
+        stmt = stmt.where(models.Document.observer == observer)
+    if observed is not None:
+        stmt = stmt.where(models.Document.observed == observed)
 
     if max_distance is not None:
         stmt = stmt.where(
@@ -327,8 +333,8 @@ async def query_documents(
     workspace_name: str,
     query: str,
     *,
-    observer: str,
-    observed: str,
+    observer: str | None = None,
+    observed: str | None = None,
     filters: dict[str, Any] | None = None,
     max_distance: float | None = None,
     top_k: int = 5,
@@ -345,8 +351,12 @@ async def query_documents(
         db: Database session, or None to let the function manage its own
         workspace_name: Name of the workspace
         query: Search query text
-        observer: Name of the observing peer
-        observed: Name of the observed peer
+        observer: Name of the observing peer. Optional on the pgvector backend
+            (omitting it searches across observers); required on external
+            vector stores.
+        observed: Name of the observed peer. Optional on the pgvector backend
+            (omitting it searches across observed peers); required on external
+            vector stores.
         filters: Optional filters to apply at vector store level (supports: level, session_name)
         max_distance: Maximum cosine distance for results
         top_k: Number of results to return
@@ -354,6 +364,11 @@ async def query_documents(
 
     Returns:
         Sequence of matching documents
+
+    Raises:
+        ValidationException: If observer or observed is omitted on a deployment
+            using an external vector store, or the query exceeds the embedding
+            token limit.
     """
     if top_k <= 0:
         return []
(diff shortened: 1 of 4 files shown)


ademczuk:
Notes on the CodeRabbit review points:

1. Docstring line lengths in src/crud/document.py: the two over-88-char lines it flags (the filters/embedding arg descriptions) are pre-existing lines untouched by this diff; left as-is to keep the change minimal.
2. Guard placement: the observer/observed check for external vector stores sits at the top of the external-store branch, after the shared embedding call. That keeps a single validation site and lets basedpyright narrow observer/observed to str for the namespace-hashing call (verified: hoisting the check above the embed breaks that narrowing and would need a cast or a duplicated check). The cost is one embedding call on an already-rejected request; happy to restructure if you would rather reject before embedding.
3. Partial-specification coverage: added. The crud-level external-store test now also asserts observer-only and observed-only each raise ValidationException, the route-level test asserts observer-only returns 422, and the pgvector crud test asserts observer-only scoping returns just that observer's documents.

ademczuk:
Added two cross-peer query tests in tests/crud/test_document.py (8041e3e3):

- test_query_documents_cross_peer_never_crosses_workspace: seeds two (observer, observed) pairs in the test workspace plus a document in a different workspace whose embedding would rank first, then queries with observer=None, observed=None and asserts the foreign document never appears. This covers the feature's only real risk: tenancy bleed across workspaces.
- test_query_documents_cross_peer_observed_only: seeds two pairs sharing neither observer nor observed, queries with observed set and observer=None, and asserts only that observed peer's documents are returned. This closes CodeRabbit's open observed-only scoping point.

Both were RED-checked: dropping the workspace predicate makes the isolation test fail with the foreign document leaking in; dropping the observed predicate makes the scoping test fail.

Suite counts:
- pytest tests/crud/test_document.py tests/routes/test_conclusions.py -q: 54 passed in 9.51s
- pytest tests/crud/ tests/routes/ tests/utils/test_agent_tools.py tests/test_session_allowlist.py -q: 641 passed in 13.71s

PR #1038, opened 2026-08-20
NETRA TRACES FOR TRACKING

jithinanie:
## Description

NETRA TRACES ADDED TO

DERIVER
DREAMER
DIALECT agent

## Summary by CodeRabbit

* **New Features**
  * Added optional Netra observability for API, background processing, chat, scheduling, database, cache, and AI provider activity.
  * Captures workspace and session context in supported operations.
  * Added environment template settings for enabling observability and configuring API access.
* **Chores**
  * Updated project configuration for observability support and dependency compatibility.
  * Excluded JSON files from version control.


akattelu:
Hey @jithinanie, thanks for opening this.

- Honcho already has a first-class LLM observability path via Langfuse,  but we also have a telemetry system that exports full spans via CloudEvents. You should be able to use this to construct and post netra traces in another service outside of honcho.
- `netra-sdk` is pinned as a hard dependency, and deriver/dialectic/dreamer do top-level `from netra import …` so the import cost lands on every process even when `NETRA_ENABLED=false`. Self-host default path should not carry a vendor SDK they did not opt into.

If there is internal interest later, open an issue first and design Netra (or any OTLP backend) as an optional extra behind our existing telemetry config, with no hard imports on the hot path and content capture off by default. Happy to point at the Langfuse hooks if that is useful.

Closing with no linked issue to retarget.

PR #1041, opened 2026-08-21
feat(docs): Initial draft of new contributing policies

VVoruganti:
## Checklist

### DIFF
--- a/.github/workflows/issue-gate.yml
+++ b/.github/workflows/issue-gate.yml
@@ -1,20 +1,27 @@
 name: Issue Gate
 
-# Closes pull requests that are not linked to an issue carrying the
-# `maintainer-approved` label. See CONTRIBUTING.md for the policy.
+# Labels pull requests that are not linked to an issue carrying the
+# `maintainer-approved` label, and comments explaining how to fix it.
 #
-# `pull_request_target` is required so the job has write access on PRs from
-# forks. This workflow must therefore NEVER check out or execute code from the
-# pull request — it only calls the GitHub API.
+# This workflow never closes anything. `pr-sweeper.yml` re-checks later and closes
+# only after the grace period — that gives contributors time to link an issue, and
+# gives maintainers time to wave through a one-line fix. It is also the only thing
+# that can notice a sidebar issue link, which fires no webhook of its own.
 #
-# Not triggered on `synchronize`: re-running the gate on every push to an
-# in-flight PR would be noise. Drafts are ignored until marked ready.
+# `pull_request_target` is required so the job has write access on pull requests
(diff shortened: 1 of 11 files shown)


.github/workflows/issue-gate.yml
```diff
@@ -0,0 +1,120 @@
...
+
+            // Collect candidate issue numbers from the PR body. HTML comments are
+            // stripped first so the commented-out `Fixes #XXX` hint in the template
+            // never counts. Any `#123` is treated as a candidate, not just the
+            // closing keywords — being generous here only risks letting a PR
+            // through, while being strict risks closing a legitimate one.
+            const body = (pr.body || '').replace(/<!--[\s\S]*?-->/g, '');
+            const numbers = new Set();
+            for (const m of body.matchAll(/#(\d+)\b/g)) numbers.add(Number(m[1]));
+            const urlPattern = new RegExp(`github\\.com/${owner}/${repo}/issues/(\\d+)`, 'gi');
```
akattelu: In the sidebar of a PR, there's also a way to link an issue directly (besides writing "Fixes …") i think that would show up as `closingIssuesReferences` on the PR object. I think we should rely on that instead of checking the body.

.github/workflows/issue-gate.yml
```diff
@@ -0,0 +1,120 @@
...
+                  `1. Find or open an issue describing the change. [Approved issues are here](https://github.com/${owner}/${repo}/issues?q=is%3Aissue+is%3Aopen+label%3A${REQUIRED_LABEL}).`,
+                  '2. Make the case for it in [Discord](http://discord.gg/honcho) — maintainers are most active there, and it is by far the fastest route to a decision.',
+                  `3. Once the issue has the \`${REQUIRED_LABEL}\` label, add \`Fixes #<number>\` to this pull request's description and reopen it.`,
+                  '',
+                  `See [CONTRIBUTING.md](https://github.com/${owner}/${repo}/blob/main/CONTRIBUTING.md) for the full process. If you think this was closed in error, comment here and a maintainer will take a look.`,
+                ].join('\n'),
+              });
+            }
+
+            await github.rest.pulls.update({
```
akattelu: Wondering if we should just attach the label, and then in a separate job we could clean up anything with the label that has been open for 24-72 hours. This would give the contributors more time to add an issue, but also it would give us time to merge PRs ourselves (if its a tiny one line fix). We could also have the same sweeper job close PRs that have been in draft for 14-30 days

I like that we're adding the special exception for maintainers. We should double check that that doesn't apply to people who have merged something before.

akattelu (changes requested):

akattelu (approved):

Issue #848, opened 2026-06-25
Deriver stamps identical conclusions into all observer/observed directions — theory-of-mind perspectives not differentiated


niftycoder208:
### Description

Honcho's data model stores conclusions with separate `observer_id` and `observed_id` fields, implying each peer maintains a distinct belief model about the other. However, the deriver currently makes a single LLM call and writes the identical conclusion text into all observer/observed permutations.

### Steps to reproduce

1. Run a session against any workspace
2. Wait for the deriver to process (~30 min)
3. Query `POST /v3/workspaces/{ws}/conclusions/list`
4. Compare conclusions across the 4 observer/observed combinations for the same fact — all four contain word-for-word identical `content`

### Expected behaviour

Each direction should produce perspective-specific phrasing, e.g.:
- `hermes→user`: "The user tends to ask for entertainment content"
- `user→hermes`: "Hermes creates fictional stories on request"
- `hermes→hermes`: "I create fictional stories when asked"
- `user→user`: "I ask for entertainment content"

### Actual behaviour

All four directions contain the same sentence stamped verbatim, e.g.:

```
{ "observer": "hermes_relationship",       "observed": "user-default-telegram-X", "content": "user-default-X asked hermes_relationship to create a fictional and very passionate story for entertainment on June 25, 2026." }
{ "observer": "user-default-telegram-X",  "observed": "user-default-telegram-X", "content": "user-default-X asked hermes_relationship to create a fictional and very passionate story for entertainment on June 25, 2026." }
{ "observer": "user-default-telegram-X",  "observed": "hermes_relationship",      "content": "user-default-X asked hermes_relationship to create a fictional and very passionate story for entertainment on June 25, 2026." }
{ "observer": "hermes_relationship",       "observed": "hermes_relationship",      "content": "user-default-X asked hermes_relationship to create a fictional and very passionate story for entertainment on June 25, 2026." }
```

### Root cause (from source)

`src/deriver/deriver.py` — `process_representation_tasks_batch` is explicitly documented as *"Single LLM call, save to multiple collections"*:

```python
# Single LLM call
response = await honcho_llm_call(
    model_config=model_config,
    prompt=prompt,          # prompt is built with a single `observed` peer — no observer context
    ...
    response_model=PromptRepresentation,
)

observations = Representation.from_prompt_representation(response.content, ...)

# Save to all observer collections  ← same observations object written to every observer
for observer in observers:
    representation_manager = RepresentationManager(observer=observer, observed=observed)
[...]

PR #1052, opened 2026-08-23
Fix: two-tier semantic dedup gate for near-duplicate observations (#729)

### LINKED ISSUE #729: Dreamer/deriver accumulates near-duplicate observations despite dedup guard
# Honcho observation deduplication bug: dreamer/deriver create near-duplicate observations

## Summary

Honcho's observation store is accumulating near-duplicate observations across sessions/dream cycles. By session 32+, this resulted in 300+ redundant observations being injected into working context. The duplication is visible in the current Honcho memory for the `courier` peer: repeated Slack capability observations and repeated Linear ticket-state observations are returned as separate facts rather than being consolidated/reinforced.

This appears to be a deduplication quality gap rather than a complete absence of any deduplication code: the code has an existing semantic duplicate path, but it only queries the single nearest neighbor with a very strict cosine-distance cutoff (`<= 0.05`, i.e. similarity >= 0.95) and then uses token-set heuristics to choose which duplicate to keep. That is not enough for LLM-generated paraphrases, which are semantically redundant but often below the 0.95 threshold.

## Evidence: duplicated observations in current Honcho memory

From `honcho_search(peer="ai", query="courier cannot search or retrieve channel message history")`:

- `[2026-05-04 12:57:53] Courier is unable to search Slack channel history, list channel messages, view conversations in other channels or DMs, or directly access the Slack API.`
- `[2026-05-04 13:01:17] courier cannot search or retrieve channel message history using internal capabilities.`
- `[2026-05-04 13:01:48] courier cannot fetch channel history, list channel messages, or see conversations in other channels or DMs directly.`
- `[2026-05-04 13:43:35] courier cannot search or retrieve channel message history on May 4, 2026.`
- `[2026-05-04 13:43:35] courier cannot list messages from channels or view conversations in other channels or direct messages on May 4, 2026.`

From `honcho_search(peer="ai", query="SLACK_USER_TOKEN search messages")`:

- `[2026-05-04 12:57:53] With the Slack user token, Courier can perform the following actions: retrieve channel and DM message history (conversations.history), list channels (conversations.list), search for messages with filters (search.messages), post messages impersonating the user (not as a bot), and read entire threaded conversations.`
- `[2026-05-04 13:01:17] With the SLACK_USER_TOKEN, courier can search messages using filters via the search.messages method.`
- `[2026-05-04 13:01:48] courier can list or search channels using conversations.list and search.messages via the Slack API.`
- `[2026-05-04 13:43:35] courier can use the SLACK_USER_TOKEN to search messages with filters using search.messages on May 4, 2026.`
- `[2026-05-04 13:43:35] courier acknowledges that a Slack user token stored in the .env file could be used to call the Slack API directly via curl to fetch or post messages on May 4, 2026.`

From `honcho_search(peer="ai", query="C-18482 Move ProviderRender shadow rollout controls Lambda env-size-critical path")`:

- `[2026-05-26 15:08:00] courier is aware of ticket C-18482 regarding a shadow rollout config move involving backend#9526 and services#1210.`
- `[2026-05-26 15:08:00] On May 26, 2026, courier identified C-18482 as a ticket involving a shadow rollout config move with two PRs (backend#9526 and services#1210) that are currently blocked or behind.`
- `[2026-05-26 15:11:23] Ticket C-18482, titled 'Move ProviderRender shadow rollout controls off Lambda env-size-critical path', is currently in Cycle 354 with a status of 'In Progress'.`
- `[2026-05-26 15:23:50] On May 26, 2026, courier identified ticket C-18482 as being in Cycle 354 with In Progress status, focusing on moving ProviderRender shadow rollout controls off the Lambda env-size-critical path.`
- `[2026-05-26 15:24:52] As of May 26, 2026, ticket C-18482 (Move ProviderRender shadow rollout controls off Lambda env-size-critical path) is In Progress within Cycle 354.`
- `[2026-05-26 15:24:52] courier is aware of ticket C-18482, which involves rollout controls on a Lambda environment, currently In Progress in Cycle 354.`

These are not exact string duplicates, but they are semantically redundant. The current issue is therefore near-duplicate / semantic deduplication, not just exact dedupe.

## Code paths investigated

Source repo determined from the container image metadata: `ghcr.io/plastic-labs/honcho:latest` has `org.opencontainers.image.source=https://github.com/plastic-labs/honcho`. Current image revision inspected: `fa8f0b1a19959049b0639d1091eb073bfe5dd849`. Upstream repo: `plastic-labs/honcho`.

Relevant upstream files and line numbers from `main`:

### Deriver explicit observation path

`src/deriver/deriver.py`

- Around line 211: `process_representation_tasks_batch()` calls `representation_manager.save_representation(...)` after the LLM extracts explicit observations.

`src/crud/representation.py`

- Lines 60-144: `RepresentationManager.save_representation()` normalizes observations, batch-embeds them, then calls `_save_representation_internal()`.
- Lines 146-209: `_save_representation_internal()` converts observations into `schemas.DocumentCreate` objects.
- Lines 193-201: it calls `crud.create_documents(..., deduplicate=settings.DERIVER.DEDUPLICATE)`.

`src/config.py` in the running local override:

- Lines 260-261: `DERIVER.DEDUPLICATE: bool = True` by default.

So the deriver path does attempt deduplication, but only through `crud.create_documents()`.

### Dreamer deductive/inductive observation path

`src/dreamer/orchestrator.py`

- `run_dream()` coordinates the dream cycle: surprisal sampling, deduction specialist, induction specialist.

`src/dreamer/specialists.py`

- Lines 233-247: `BaseSpecialist.run()` creates a tool executor for dream specialists.
- Lines 825-846 in `src/utils/agent_tools.py` define the dream specialists' action tools:
  - deduction gets `create_observations_deductive`
  - induction gets `create_observations_inductive`

`src/utils/agent_tools.py`

- Lines 1293-1439: `_handle_create_observations_impl()` handles generic, deductive, and inductive create tools.
- Lines 1352-1364: it serializes writes with `ctx.db_lock` and calls `create_observations(...)`.
- Lines 849-1005: `create_observations(...)` normalizes, embeds, builds `schemas.DocumentCreate`, then calls `crud.create_documents(..., deduplicate=True)` at lines 981-992.

So the dreamer path also attempts deduplication through `crud.create_documents()`.

### Existing deduplication implementation

`src/crud/document.py`

- Lines 424-432: `create_documents(..., deduplicate=False)` accepts a dedup flag.
- Lines 452-462: for each document, when `deduplicate` is true, it calls `is_rejected_duplicate(...)` and skips the new doc only if that returns true.
- Lines 963-1026: `is_rejected_duplicate(...)` is the only semantic duplicate check.
- Lines 985-995: duplicate candidates are found with `query_documents(..., max_distance=0.05, top_k=1, embedding=doc.embedding)`.
- Lines 1002-1010: winner selection is based on token-set difference between the new and existing content.
- Lines 1012-1020: if the new doc scores higher/equal, the existing doc is soft-deleted and the new doc is kept.
- Lines 1022-1026: otherwise the new doc is rejected.

`src/crud/document.py`

- Lines 282-313: `_query_documents_pgvector()` supports semantic search via `models.Document.embedding.cosine_distance(embedding) <= max_distance`, orders by distance, and limits by `top_k`.

`src/crud/representation.py`

- Lines 370-402: `_query_documents_semantic()` is an existing semantic comparison helper used for retrieval. It is not used directly for stronger pre-insert duplicate analysis beyond the `query_documents(... max_distance=0.05, top_k=1)` call inside `is_rejected_duplicate()`.

`src/dreamer/surprisal.py`

- `sample_observations_with_surprisal()` uses embeddings/tree structures to find novel/surprising observations. It samples what to dream over; it does not deduplicate newly-created observations before insertion.

## Root cause hypothesis

The missing/broken part is not that there is zero dedup call; it is that the existing dedup guard is too narrow for LLM-derived observations:

1. It only considers one nearest neighbor (`top_k=1`). If the closest result is not the true redundant cluster representative, near-duplicates can slip through.
2. It uses a very strict semantic threshold (`max_distance=0.05`, similarity >= 0.95). LLM paraphrases like `cannot search or retrieve channel history` vs `cannot fetch channel history/list channel messages/view conversations directly` can be redundant but likely fall below 0.95 similarity.
3. It does no exact/canonical text dedup before embedding, e.g. lowercasing, date-prefix stripping, whitespace normalization, or removal of volatile timestamp clauses (`on May 26, 2026`, `As of May 26, 2026`).
4. It does not merge/reinforce near-duplicates. `times_derived` exists, but the dedup path either rejects or replaces; it does not increment/reinforce an existing semantically equivalent observation or merge metadata/source IDs.
5. Dreamer prompts/tools ask specialists to create novel deductions/patterns, but novelty is left to the LLM and this narrow insertion-time guard. The system needs a deterministic insertion-time semantic novelty gate.

The best insertion point for a robust fix is `src/crud/document.py::is_rejected_duplicate()` and/or a new helper called from `create_documents()` before insertion. This catches both deriver and dreamer paths because both flow through `crud.create_documents()`.

## Proposed fix approach

1. Add level-aware near-duplicate detection before insertion in `create_documents()` / `is_rejected_duplicate()`:
   - query more candidates, e.g. `top_k=5` or `top_k=10`;
   - use a configurable threshold, e.g. `DOCUMENT_DEDUP_MAX_DISTANCE`, rather than hard-coded `0.05`;
   - consider looser defaults for dreamer-created deductive/inductive observations than exact deriver facts, or make thresholds per level.

2. Add canonical exact-ish prechecks before vector search:
   - lowercase + collapse whitespace;
   - remove volatile temporal prefixes/suffixes where appropriate (`As of <date>`, `On <date>`);
   - normalize ticket IDs / capability phrasing;
   - reject exact canonical duplicates without an embedding query.

3. Prefer reinforcement/merge over replacement:
   - if a near-duplicate is found, increment `times_derived` on the retained observation;
   - merge `source_ids`, `message_ids`, and premises/sources metadata where safe;
   - preserve the more informative content as the canonical text.

4. Make dreamer tool feedback explicit:
   - when a proposed observation is rejected/merged as a duplicate, return that in the tool result so the specialist does not keep attempting variants in the same run.

5. Add regression tests:
   - `create_observations_inductive` should not create two paraphrased observations about the same stable capability across repeated dream runs;
   - `RepresentationManager.save_representation()` should not create repeated explicit observations for the same fact with volatile date wording;
   - verify `times_derived` or merged metadata changes instead of creating a new row.

## Existing issue check

Searched open/all issues with:

- `gh issue list --repo plastic-labs/honcho --state open --search 'observation duplicate deduplicate semantic'`
- `gh issue list --repo plastic-labs/honcho --state all --search 'dreamer observations duplicate'`

No existing matching issue was found.

favoritegrandson-tech:
Fixes #729

### Problem

The insertion-time dedup guard (`create_documents` → `is_rejected_duplicate`) is
too narrow to catch LLM paraphrases: `top_k=1`, cosine distance ≤ 0.05
(similarity ≥ 0.95), no canonical text normalization, and replace-or-reject
semantics that never reinforce. The deriver and dreamer paraphrase the same fact
with slightly different wording (date clauses, casing, token order), and every
variant becomes a new document. This is issue #729, and we reproduced it in
production: by session 32 a live workspace held hundreds of near-duplicate
observations, including 3 near-identical trait restatements written in the same
minute and cross-collection twins invisible to any per-collection dedup.

### What this changes

1. **Two-tier semantic gate.** The existing tight band (distance ≤ 0.05) keeps
   its legacy replace/reject semantics, except that ties now reinforce
   (`times_derived += 1`) instead of replacing. A new loose band
   (0.05, 0.12] catches paraphrases that previously slipped through: the new
   document is rejected **only when the existing row is strictly more
   informative** (token-set heuristic), and the existing row is reinforced.
2. **No-data-loss guarantee.** A document is **never soft-deleted inside the
   loose band** — distinct adjacent facts with overlapping wording survive.
   Delete-in-favor-of-new stays confined to the tight band where the
   replacement is a strict superset.
3. **Wider candidate window.** `top_k` 1 → 4, so a strong paraphrase match
   beyond the single nearest neighbor is still seen.
4. **Canonical text normalization.** `_normalize_content` (mirrored in the
   exact-dedup SQL) strips leading bracketed timestamp prefixes, collapses
   whitespace, and lowercases — date-only variants collapse to one dedup key
   before any embedding query runs.
5. **Config knobs.** `DERIVER_DEDUP_SEMANTIC_DISTANCE_TIGHT`,
   `DERIVER_DEDUP_SEMANTIC_DISTANCE_LOOSE`, `DERIVER_DEDUP_SEMANTIC_TOP_K`
   (defaults 0.05 / 0.12 / 4).

### Tests

`tests/crud/test_document.py` gains regression tests for: paraphrase
reinforcement in the loose band, survival of distinct adjacent facts inside
the band (the no-data-loss guarantee), and timestamp-prefix collapse.
All existing dedup/session-purity tests unchanged and passing (25/25 on this
branch; full suite green in CI-local run).

### Production evidence (48h soak, live deployment)

This exact gate ran on a production Strix Halo deployment (nomic-embed 768-dim,
pgvector) for a 48h soak before this PR was prepared:

- Both branches fired in the deriver log: `Rejecting new in favor of
  existing.` (loose band) and `Deleting existing in favor of new.` (tight band,
  e.g. a stale storage-drive observation correctly replaced by its successor).
- 556 new documents, 23 reinforced (`times_derived > 1`, max 3).
- 0 repair fails, embed reconciliation clean, dreams completing on the gated
  writes.
- The cross-collection duplicate class stopped accumulating (we swept 285
  later-twins from the live DB and the pile did not regrow).

### Related

- #729 (this issue — evidence comment posted 2026-08-19)
- #768 (reinforce on exact-dup detection — merged; this PR extends reinforce
  semantics into the semantic band)
- #848 (cross-collection stamping — mentioned in #729; not addressed here)

### Not in this PR (honest scope)

- Cross-collection dedup (a document in observer→observed and its mirror pair
  is still two rows). The gate works per collection; #848 territory.
- Retrieval-side dedup (separate concern; there is an existing PR for that).
- Dreamer prompt changes — this is gate-level only, prompt behavior unchanged.

## Summary by CodeRabbit

* **New Features**
  * Improved duplicate detection for semantically similar documents.
  * Distinguishes near-identical content from related but meaningfully different facts.
  * Evaluates multiple similar candidates to improve matching accuracy.
  * Normalizes leading timestamp prefixes and extra whitespace for more consistent matching.
  * Added Qdrant as a supported vector-store option with configurable connection settings.

* **Bug Fixes**
  * Prevents redundant document replacement while preserving more informative content.
  * Ensures timestamp variants are consolidated correctly.

* **Tests**
  * Added coverage for paraphrases, distinct facts, and timestamp-prefixed duplicates.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -932,6 +932,16 @@ class DeriverSettings(HonchoSettings):
 
     # Whether to deduplicate documents when creating them
     DEDUPLICATE: bool = True
+    # Semantic dedup gate (◆0816): two-tier. The tight band (≤
+    # DEDUP_SEMANTIC_DISTANCE_TIGHT, i.e. ≥ ~0.95 cosine similarity) keeps the
+    # legacy replace/reject semantics. The loose band (TIGHT, LOOSE] catches
+    # LLM paraphrases that previously slipped through the 0.95 wall: the new
+    # doc is rejected only when the existing row is strictly more informative
+    # (token-set heuristic), and never soft-deleted — distinct adjacent facts
+    # can't be destroyed by paraphrase overlap. Upstream context: issue #729.
+    DEDUP_SEMANTIC_DISTANCE_TIGHT: float = 0.05
+    DEDUP_SEMANTIC_DISTANCE_LOOSE: float = 0.12
+    DEDUP_SEMANTIC_TOP_K: int = 4
 
     LOG_OBSERVATIONS: bool = False
 
--- a/src/config.py
+++ b/src/config.py
@@ -1003,6 +1003,28 @@ class DeriverSettings(HonchoSettings):
             )
         return self
 
+    @model_validator(mode="after")
+    def validate_dedup_semantic_settings(self):
+        """Enforce sane bounds and ordering for two-tier semantic dedup knobs."""
+        tight = self.DEDUP_SEMANTIC_DISTANCE_TIGHT
+        loose = self.DEDUP_SEMANTIC_DISTANCE_LOOSE
+        top_k = self.DEDUP_SEMANTIC_TOP_K
+        if not (0.0 <= tight <= 1.0 and 0.0 <= loose <= 1.0):
+            raise ValueError(
+                f"DEDUP_SEMANTIC_DISTANCE_TIGHT ({tight}) and "
+                f"DEDUP_SEMANTIC_DISTANCE_LOOSE ({loose}) must be in [0, 1]"
+            )
+        if top_k <= 0:
+            raise ValueError(
+                f"DEDUP_SEMANTIC_TOP_K ({top_k}) must be > 0"
+            )
+        if tight > loose:
+            raise ValueError(
+                f"DEDUP_SEMANTIC_DISTANCE_TIGHT ({tight}) must be <= "
+                f"DEDUP_SEMANTIC_DISTANCE_LOOSE ({loose})"
+            )
+        return self
+
 
 class PeerCardSettings(HonchoSettings):
     model_config = SettingsConfigDict(env_prefix="PEER_CARD_", extra="ignore")  # pyright: ignore
--- a/src/crud/document.py
+++ b/src/crud/document.py
@@ -1,5 +1,7 @@
 import asyncio
 import datetime
+import math
+import re
 from collections.abc import Sequence
 from dataclasses import dataclass, field
 from enum import Enum
@@ -212,8 +214,6 @@ def _uses_pgvector() -> bool:
 
 
 # Shared by is_rejected_duplicate and create_documents candidate resolution.
-_SEMANTIC_DUP_MAX_DISTANCE = 0.05
-_SEMANTIC_DUP_TOP_K = 1
 _SEMANTIC_CANDIDATE_CONCURRENCY = 8
 
 
@@ -237,15 +237,18 @@ async def query_external_vector_document_ids(
     top_k: int = 5,
     max_distance: float | None = None,
     filters: dict[str, Any] | None = None,
-) -> list[str] | None:
+) -> list[tuple[str, float | None]] | None:
     """Query external vector store for document IDs sorted by similarity.
 
     No DB session needed — safe to call outside a tracked_db scope.
 
     Returns:
-        Ordered list of document IDs on the external-store path,
-        empty list when the external store has no results,
-        or None when the pgvector (DB-only) path should be used instead.
+        Ordered list of (document ID, store-reported distance) pairs on the
+        external-store path — the distance is carried so downstream
+        two-tier dedup can tier candidates tight vs loose even when the
+        rows hold no DB embedding (◆0907 CodeRabbit finding; 0.0 is a real
+        distance and is preserved). Empty list when the external store has
+        no results, or None when the pgvector (DB-only) path should be used.
     """
     if _uses_pgvector():
         return None
@@ -279,7 +282,7 @@ async def query_external_vector_document_ids(
     if not vector_results:
         return []
 
-    return [result.id for result in vector_results]
+    return [(result.id, result.score) for result in vector_results]
 
 
 async def fetch_documents_by_ids(
@@ -418,8 +421,10 @@ async def query_documents(
                 managed_db.expunge(doc)
             return docs
 
-    # External vector store — network call first, DB only for the ID fetch
-    document_ids = await query_external_vector_document_ids(
+    # External vector store — network call first, DB only for the ID fetch.
+    # Carry the store-reported distance per candidate so downstream
+    # two-tier dedup can tier external-store rows (◆0907 CodeRabbit fix).
+    id_distance_pairs = await query_external_vector_document_ids(
         workspace_name=workspace_name,
         observer=observer,
         observed=observed,
@@ -429,17 +434,27 @@ async def query_documents(
         filters=filters,
     )
 
-    if not document_ids:
+    if not id_distance_pairs:
         return []
 
+    id_distances = dict(id_distance_pairs)
+    document_ids = list(id_distances)
+
+    def _attach_distance(docs: list[models.Document]) -> list[models.Document]:
+        for doc in docs:
+            doc._vector_distance = id_distances.get(doc.id)  # type: ignore[attr-defined]
+        return docs
+
     if db is not None:
-        return await fetch_documents_by_ids(
-            db=db,
-            workspace_name=workspace_name,
-            observer=observer,
-            observed=observed,
-            document_ids=document_ids,
-            filters=filters,
+        return _attach_distance(
+            await fetch_documents_by_ids(
+                db=db,
+                workspace_name=workspace_name,
+                observer=observer,
+                observed=observed,
+                document_ids=document_ids,
+                filters=filters,
+            )
         )
     async with tracked_db("query_documents.fetch", read_only=True) as managed_db:
         docs = await fetch_documents_by_ids(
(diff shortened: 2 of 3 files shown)


favoritegrandson-tech:
Applied CodeRabbit's 4 actionable fixes to our fork branch (dedup-semantic-gate-729):

1. **Config validation** — distance knobs constrained to [0,1], top_k > 0, tight ≤ loose
2. **Candidate evaluation** — iterate ALL candidates in is_rejected_duplicate, not just [0]
3. **Timestamp normalization** — only strip [2026-08-16 07:00]-style prefixes, preserve [urgent]
4. **External vector store** — preserve distance info for tier classification (fixes distance-preservation bug flagged by CodeRabbit)

Diff: 2 files, +97/-62. Python AST parse verified. Pushed to favoritegrandson-tech/honcho#dedup-semantic-gate-729.

Requesting review from @rajat-ahuja1997 @vvoruganti when time allows.

PR #1057, opened 2026-08-24
Vineeth/dev 2418

VVoruganti:
## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added automated pull-request checks for linked approved issues, status labels, notices, and noncompliant submissions.
  - Added scheduled maintenance to identify and handle stale draft pull requests.

- **Documentation**
  - Updated contribution guides with issue requirements, development workflows, setup, testing, architecture, and security guidance.
  - Improved README configuration and contribution instructions.
  - Expanded security reporting and vulnerability-handling guidance.

- **Tests**
  - Added automated coverage for pull-request gating and stale-draft handling.


akattelu (dismissed):

docs/v3/contributing/guidelines.mdx
```diff
@@ -3,174 +3,358 @@ title: 'Contributing Guidelines'
 icon: 'handshake'
 ---
 
-Thank you for your interest in contributing to Honcho! This guide outlines the process for contributing to the project and our development conventions.
+{/* This file mirrors CONTRIBUTING.md in the repo root. Update both. */}
```
eisene, line 6: claude claims these would be easy to automate

---

❯ is this something that can be automated relatively easily?

Yes — easily. The diff we just looked at shows the entire transform is four mechanical steps plus one small lookup table:

1. Frontmatter: replace the # Contributing to Honcho H1 and HTML comment with the Mintlify frontmatter block and an MDX comment. Fixed string, no logic.
2. TOC: drop the ## Contents section (delete everything between ## Contents and the next ## heading). Or generate it in the other direction — either way it's one regex.
3. Comment syntax:  → {/\* ... \*/}.
4. Links: rewrite repo-relative links. Most follow one rule (./path → [https://github.com/plastic-labs/honcho/blob/main/path](<https://github.com/plastic-labs/honcho/blob/main/path>)); the handful that map to docs-site routes instead (./README.md#self-hosting → /v3/contributing/self-hosting, ./LICENSE → ./license) go in an explicit mapping dict, which today has about three entries.

That's a \~60-line Python script with CONTRIBUTING.md as the single source of truth, generating guidelines.mdx. The semantic link mapping is the only part that ever needs human input, and only when someone adds a new repo-relative link that should point at a docs page rather than GitHub.

For enforcement, the standard shape is: check the script into scripts/, and add a CI step (or a job in the existing issue-gate.yml-style workflows) that regenerates the mdx and fails with a diff if it doesn't match what's committed. Contributors edit only CONTRIBUTING.md, run the script, commit both. That's strictly better than the "update both" comment because drift becomes impossible to merge rather than something a reviewer has to notice.

One judgment call: the current mdx wasn't machine-generated, so the first run of the script needs to reproduce it exactly (or you accept a one-time normalization commit). Given the file is 400 lines and changes rarely, I'd call the whole thing an hour of work including the CI wiring. Want me to build it into this PR?

SECURITY.md
```diff
@@ -1,31 +1,73 @@
...
+[issue templates](https://github.com/plastic-labs/honcho/issues/new/choose) instead.
+
+## No Bug Bounty
+
+The Honcho project does not offer any rewards for reported bugs or
+vulnerabilities. We do not aid security researchers to get such rewards for
+Honcho problems from other sources.
+
+A bug bounty gives people too strong incentives to find and make up "problems"
+in bad faith that cause overload and abuse.
```
eisene, line 71: do we really need this sentence?

eisene (commented):
two small comments

docs/v3/contributing/guidelines.mdx
VVoruganti: I think we can do this in a follow up, these files shouldn't change too often so not that worried about drift.

SECURITY.md
VVoruganti: Copy is just parroting the curl project's [BUG_BOUNTY.md](https://github.com/curl/curl/blob/master/docs/BUG-BOUNTY.md), you're right it's not really necessary, but I don't think it's a blocker

PR #1058, opened 2026-08-24
perf(cache): hash-tag the cache namespace so a deployment uses one shard

Rajat-Ahuja1997:
## What

Wraps the cache namespace in `{...}` so every key a deployment writes hashes to the same Redis Cluster slot.

## Why

On Redis Cluster the slot comes from the substring inside the first `{...}`, when one is present. Untagged, one deployment's keys land on every slot, so its client opens and holds a connection to every node. Tagging the namespace puts them on one slot, and therefore one node, which divides each deployment's connection count by the shard count.

Namespaces still hash independently of one another, so keys stay spread across the cluster. No shard becomes a hotspot.

## The two spellings

The tag needs two spellings, because the two ways a key gets built treat the string differently:

| Path | Helper | Renders |
| --- | --- | --- |
| cashews `prefix=` (format-substituted) | `cache_prefix_namespace()` | `{{ns}}` -> `{ns}` |
| direct concatenation | `cache_key_namespace()` | `{ns}` |

A single brace handed to cashews is read as an empty substitution field and the namespace is dropped, which would let two deployments collide on the same key. That is why this is two clearly named helpers rather than one string, and why there is a test asserting the two paths produce identical bytes.

## Testing

`tests/test_cache_key_namespace.py` covers the four things that can break:

- both spellings render the same tag
- the key the decorator writes equals the key the helper computes
- one namespace's keys share a slot
- different namespaces do not

Verified by mutation: reverting `cache_prefix_namespace()` to single braces fails two of the four. Existing `tests/crud` and cache tests pass unchanged.

## Compatibility

No behaviour change on a non-cluster backend, and no migration needed. Existing keys age out by TTL.

## Summary by CodeRabbit

- **New Features**
  - Improved Redis Cluster cache namespace handling for more consistent key generation and cache operations.
  - Added support for keeping related cache keys within the same cluster slot.

- **Bug Fixes**
  - Improved reliability of collection, peer, session, and workspace locks and cached data by separating key and prefix namespaces.

- **Tests**
  - Added coverage verifying namespace consistency and Redis Cluster slot placement.

### DIFF
--- a/src/cache/client.py
+++ b/src/cache/client.py
@@ -123,6 +123,28 @@ def get_cache_namespace() -> str:
     return cast(str, settings.CACHE.NAMESPACE)
 
 
+# On Redis Cluster a key's slot is derived from the substring inside the first
+# {...}, when one is present. Tagging the namespace puts every key an instance
+# writes on a single slot, and therefore a single shard, so its client holds
+# connections to one node rather than to all of them. Namespaces still hash
+# independently of one another, so keys stay spread across the cluster.
+#
+# Two spellings, because the two ways a key gets built treat the string
+# differently: cashews runs `prefix=` through format substitution, so braces
+# have to be doubled to survive as literals, while direct construction does no
+# substitution and needs them single. Both render to the same bytes, which
+# tests/cache/test_cache_namespace_hash_tag.py asserts -- a mismatch would send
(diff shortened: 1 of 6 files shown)


akattelu (approved):

eisene (approved):
lgtm

PR #1060, opened 2026-08-25
fix(db): enable pgvector HNSW iterative scan by default (#1048)

### LINKED ISSUE #1048: [Bug] Filtered pgvector HNSW conclusion queries silently under-return with iterative scans off
# **🐞 Bug Report**

## **Describe the bug**

---

### **Is this a regression?**

---

### **To Reproduce**

---

### **Expected behaviour**

---

### **Your environment**

- OS: Linux

---

### **Additional context**

```text
```

shellybotmoyer:
## Summary

Fixes #1048 — filtered pgvector HNSW conclusion queries silently return fewer than `top_k` results (including zero) when out-of-scope rows consume the approximate scan budget before the requested scope is evaluated.

### Root cause

With the default `hnsw.ef_search=40` and `hnsw.iterative_scan=off`, pgvector performs an approximate HNSW scan and then applies workspace/observer/observed filters *after* the scan. Candidates from other scopes can exhaust the scan budget before matching rows are reached, causing silent under-return.

### Fix

Add a `DB_HNSW_ITERATIVE_SCAN` setting (default `"strict_order"`) applied as a per-connection `server_settings` parameter via psycopg3. This enables [pgvector iterative index scans](https://github.com/pgvector/pgvector#iterative-index-scans) on every connection, so filtered approximate searches continue scanning until `top_k` filtered results are found.

Per the issue report, `strict_order` preserves exact ordering while completing the requested cardinality with negligible overhead (1.9ms → 2.9ms in measured samples).

### Configuration

- **Default**: `DB_HNSW_ITERATIVE_SCAN=strict_order` (fixes the bug out of the box)
- **Disable**: `DB_HNSW_ITERATIVE_SCAN=` (empty string) to restore old behavior
- **Other modes**: `strict_order` (recommended), `off`, `on`, `relaxed_order` — see pgvector docs

### Files changed

- `src/config.py` — add `HNSW_ITERATIVE_SCAN` field to `DBSettings`
- `src/db.py` — pass `hnsw.iterative_scan` through `server_settings` in `connect_args`

Closes #1048

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filtered vector searches to return the requested number of results more consistently.
  - Prevented eligible results from being omitted in HNSW-based searches when unrelated results consume the scan limit.
  - Ensured the search configuration is applied reliably when establishing database connections.

- **Configuration**
  - Added a configurable HNSW iterative scanning setting.
  - Enabled strict ordering by default to improve result completeness and consistency for filtered searches.

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -739,6 +739,13 @@ class DBSettings(HonchoSettings):
     SQL_DEBUG: bool = False
     TRACING: bool = False
 
+    # pgvector HNSW iterative scan mode for filtered approximate searches.
+    # When set (default "strict_order"), applied as a per-connection server
+    # setting so filtered HNSW queries return full top_k results instead of
+    # silently under-returning when out-of-scope rows consume the scan budget.
+    # See: https://github.com/pgvector/pgvector#iterative-index-scans
+    HNSW_ITERATIVE_SCAN: str | None = "strict_order"
+
     # Per-connection establish timeout (seconds) passed to the driver, so a
     # single connection attempt fails fast instead of hanging when the server or
     # pooler is unreachable or stalled. Connection acquisition is a single
--- a/src/db.py
+++ b/src/db.py
@@ -24,12 +24,27 @@ connect_args: dict[str, Any] = {
 }
 
 # Apply pgvector HNSW iterative scan setting per-connection so filtered
-# approximate searches return full top_k results. psycopg3 passes
-# server_settings as RUNTIME SET on every new connection.
+# approximate searches return full top_k results. We use a pool "connect"
+# event listener instead of connect_args["server_settings"] because
+# SQLAlchemy's psycopg dialect passes connect_args to AsyncConnection.connect(),
+# and psycopg 3.3.x rejects "server_settings" at connect() time (it's a
+# constructor kwarg, not a connect kwarg). The event listener runs SET on the
+# raw DBAPI connection right after it's established.
 if settings.DB.HNSW_ITERATIVE_SCAN:
-    connect_args["server_settings"] = {
-        "hnsw.iterative_scan": settings.DB.HNSW_ITERATIVE_SCAN,
-    }
+
+    def _set_hnsw_iterative_scan(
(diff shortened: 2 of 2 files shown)


akattelu (changes requested):
Hey @shellybotmoyer, thanks for opening this. The failure mode is real, and connection-level `hnsw.iterative_scan=strict_order` is the right layer for it. pgvector documents the post-filter under-return, the numbers on #1048 line up, and on this branch new pool connections do land with `SHOW hnsw.iterative_scan = strict_order`.

A few things before merge:

- **Validate the setting.** I'd restrict `DBSettings.HNSW_ITERATIVE_SCAN` to `None | Literal["off", "strict_order", "relaxed_order"]` (or equivalent). The comment currently lists `on`, which is not one of the enum values, and an arbitrary string currently gets interpolated into the `SET` rather than failing closed at config load.
- **Match the existing connect hook.** `_set_application_name_on_checkout` already does `SELECT set_config(..., %s, false)` with bind params under the same autocommit dance. I'd lean toward that here instead of an f-string `SET`, mostly for consistency and so we do not special-case utility-statement binding.
- **Tests + docs.** At least one test that a checked-out connection has the GUC set when the default is on, plus a short note in `docs/v3/contributing/configuration.mdx` and the self-host expectation of pgvector >= 0.8.0. I do not think we need a full multi-scope cardinality integration test for the first cut. Also I think most of the comments in this PR are unnecessary and can be truncated.
- **PR body drift.** The summary still describes `server_settings` / psycopg connect kwargs; the code moved to a pool `connect` listener in the follow-up commits.

I got #1048 maintainer-approved on our side.

shellybotmoyer:
Opened #1070 to address @akattelu's review feedback (Literal type validation, set_config with bind params, tests, docs). This supersedes this branch with a clean rebase on current main.

shellybotmoyer:
Superseded by #1070 (v2 addresses review feedback). Closing this duplicate.

shellybotmoyer:
Closing in favor of #1070 which addresses review feedback.

PR #1062, opened 2026-08-25
fix(deriver): fail closed on invalid work-unit recovery

bbasketballer75:
Replaces the stale implementation in #959 with a focused recovery contract on current `main`.

## What changes

- Locks the exact durable queue lease before quarantining a malformed work unit.
- Commits queue-item quarantine and lease deletion together; an uncertain write preserves ownership, latches a fatal error, and exits the Deriver nonzero for stale-lease recovery.
- Runs a bounded in-container Deriver supervisor in both Fly and Compose. Repeated unhealthy probes terminate the worker and yield a restartable nonzero exit.
- Uses the resolved Honcho database settings, treats malformed keys as unhealthy, and measures active work from its durable progress timestamp rather than enqueue age.

## Validation

- Focused Linux/Postgres suite: `48 passed`.
- Linux `basedpyright`: `0 errors, 0 warnings`.
- Targeted Ruff, Fly TOML parse, and rendered Compose configuration passed.
- Production Docker image built; its real health-probe command exits `1` against an unreachable database (negative control).

## Deliberate boundary

This has not deployed to Fly/Compose and has not received upstream CI. The existing stale-lease timeout remains the recovery boundary; this is fail-closed recovery, not an exactly-once guarantee.

## Summary by CodeRabbit

* **New Features**
  * Added automatic Deriver health monitoring for stalled or malformed queue work.
  * Deriver processes now restart after repeated health-check failures or unexpected exits.
  * Added configurable startup delay, check interval, failure threshold, timeout, and pending-work age limit.

* **Bug Fixes**
  * Malformed work is safely quarantined, while processing stops if quarantine persistence fails to prevent data loss.
  * Improved shutdown handling and recovery from failed processing states.

* **Documentation**
  * Added configuration examples for enabling and tuning queue-health supervision.


akattelu:
Hey @bbasketballer75 thanks for opening this.

- The Sentry keys are live dream work (`dream:omni:opencode-mac:user:kmeppen:explorer` and the other omni agents), not garbage. Current construct is five parts (`dream:{type}:{workspace}:{observer}:{observed}`); those events are six. Marking them processed+error would drop consolidation for those collections. The mismatch is construct vs parse, not a poison pill.
- Wrapping `python -m src.deriver` in `scripts.deriver_supervisor`, changing `fly.toml` and Compose, and adding five health knobs is orchestration. #521 already moved HTTP health off the deriver image because worker healthchecks lie. A 3600s pending-age probe will also restart a healthy worker whose `last_updated` only moves after a batch commit.
- Fail-closed retention of every owned lease until `STALE_SESSION_TIMEOUT_MINUTES` is the recovery we already have, plus extra downtime for unrelated sessions.

Suggested shape: leave Fly, Compose, the supervisor, and the knobs alone. Catch `ValueError` in `process_work_unit`, log it, keep the lease until we know the key is actually unrepresentable. The dream key format (why production writes six parts) is the actual fix.

Going to close this PR and #959, which is the same change. No GitHub issue is linked;

PR #1065, opened 2026-08-25
fix(scopes): scope observer sessions in SQL instead of a fetched name list

eisene:
# Human description

## Background

There was a Sentry alert on a workspace that had tens of thousands of sessions with peer `hermes`. There's a path in the scopes work that errors out with this many entries in a `WHERE` clause.

## Changes

Currently if an observer is specified but no allowlist (or session name), the SDK resolves all the observer sessions in `resolve_session_scope` and writes SQL to check if the session is `IN` this list. This list ended up being huge for this case.

- Fix the SQL clause to be more efficient.
  - Replace the resolution of the observer sessions with `observer_scope_clause` that writes SQL based on the observer.

This PR is deliberately scoped to be a targeted SQL fix. The oversized error message that the failing query then fed to the LLM is a second problem left for its own future PR.

## Test plan

- Run all unit tests to make sure nothing broke.
- Deploy and monitor if the Sentry goes away.

# Claude description

## Description

`get_observation_context` resolved scope by fetching every session name the observer has a membership record in, then expanding that list into `session_name IN (...)` — **twice in one statement**, once in the CTE and once in the outer select. That puts psycopg's 65535-bind-parameter ceiling at roughly 32,765 sessions, and the count only ever grows: the loose membership definition (`active_only=False`) counts sessions the peer has since left, so leaving a session does not shrink the scope. Past that ceiling the statement cannot be serialized at all.

Two new helpers in `crud.message` express the observer half as a correlated `EXISTS` over `session_peers`. The scope now costs two bind parameters regardless of membership size, and the membership query disappears (two round trips become one). The `session_peers` primary key is `(workspace_name, session_name, peer_name)`, so the correlated probe is a full exact-match index hit rather than a scan.

The caller-supplied allowlist stays an `IN` clause — it is route-capped at 1000 entries (`MAX_SESSION_ALLOWLIST_ENTRIES`) and carries none of the unbounded-growth risk. Worth being precise about the two names, since they are easy to conflate: `session_allowlist` is an **input** from the client's request filters, capped at 1000; `allowed_session_names` was the **output** — the peer's memberships resolved against the database, and uncapped. The defect was entirely in the second.

`resolve_session_scope` is deliberately left in place. Three other callers still need the materialized list, including `_search_messages_external`, which sends session names to the vector store as a filter payload and cannot take SQL.

## Proofs

**Compiled SQL for `get_observation_context`** (`observer` set, 2 message ids), confirming the scope predicate lands in *both* the CTE and the outer select and correlates to the enclosing row in each. An uncorrelated subquery would compile just as happily and would silently drop scoping, so this is asserted in the tests rather than eyeballed:

```sql
WITH target_seqs AS
(SELECT public.messages.seq_in_session AS seq_in_session
FROM public.messages
WHERE public.messages.workspace_name = %(workspace_name_2)s
  AND public.messages.public_id IN (%(public_id_1_1)s, %(public_id_1_2)s)
  AND (EXISTS (SELECT 1
    FROM public.session_peers
    WHERE public.session_peers.workspace_name = %(workspace_name_3)s
      AND public.session_peers.peer_name = %(peer_name_1)s
      AND public.session_peers.session_name = public.messages.session_name)))
 SELECT public.messages.id, ...
FROM public.messages
WHERE public.messages.workspace_name = %(workspace_name_1)s
  AND (EXISTS (SELECT target_seqs.seq_in_session FROM target_seqs
    WHERE target_seqs.seq_in_session - public.messages.seq_in_session
      BETWEEN %(param_1)s AND %(param_2)s))
  AND (EXISTS (SELECT 1
    FROM public.session_peers
    WHERE public.session_peers.workspace_name = %(workspace_name_3)s
      AND public.session_peers.peer_name = %(peer_name_1)s
      AND public.session_peers.session_name = public.messages.session_name))
 ORDER BY public.messages.seq_in_session ASC
```

Scope now costs two bind parameters total — SQLAlchemy dedupes them across both uses — where the old form spent two per session.

**Tests.** 9 new cases in `tests/crud/test_session_scope_clauses.py` (DB-free; asserts compiled SQL). No existing test was changed.

Regression runs against a real Postgres:

```
tests/crud tests/llm tests/utils tests/dialectic tests/test_session_allowlist.py    748 passed
full suite (excl. tests/bench, tests/live_llm)                                     2231 passed
```

The behavioural check that matters here is `tests/test_session_allowlist.py` plus the `get_observation_context` coverage in `tests/utils/test_agent_tools.py` — `TestResolveSessionScope`, `TestPeerScopedJWTAllowlistGate`, and friends. Asserting on compiled SQL is not sufficient evidence for a change to a scope filter.

`ruff check`, `ruff format --check`, and `basedpyright` are clean on every changed file (0 errors, 0 warnings).

### Deliberately out of scope

- **Truncation of oversized LLM payloads.** The failing query above stringified its full SQL and every bound parameter into a tool result, which then blew the provider's request-size limit — and conversation truncation could not shrink it, because it drops whole messages and cannot split one. Both are real, and both are independent of this fix; they belong in their own PR. Two notes for whoever picks that up: the tool-output ceiling (`LLM.MAX_TOOL_OUTPUT_CHARS`) already exists but is opt-in per handler, so the generic error paths in `execute_tool` bypass it; and the conversation-truncation fix must not ship without the tool-output fix, or an oversized payload gets clamped and sent with the real conversation already evicted to make room — a confident, worthless answer at HTTP 200 instead of today's loud 400.
- **Three other `IN` sites** over the same observer scope remain: `_search_messages_pgvector`, `_grep_messages_internal`, and `get_messages_by_date_range`. Each applies the scope once rather than twice, so their ceiling is ~65,535 sessions rather than ~32,765. Converting them means restructuring `_semantic_search_messages`, which resolves the scope once and then dispatches to either the external vector store (needs materialized names) or pgvector. `observer_scope_clause` is reusable for that follow-up.
- **Whether `active_only=False` is the right definition for recall scoping** is a product question, not a bug. It is left as-is and documented in the new helper, but it is the reason membership only ever grows.
- **`message_ids` has no `maxItems`** in the `get_observation_context` tool schema. Not the cause here — the failing request carried five — but it is the same unbounded-`IN` hazard with a model-controlled argument.

## Checklist

### DIFF
--- a/src/crud/message.py
+++ b/src/crud/message.py
@@ -6,6 +6,7 @@ from typing import Any
 from nanoid import generate as generate_nanoid
 from sqlalchemy import ColumnElement, Select, and_, func, or_, select, text
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import InstrumentedAttribute
 
 from src import models, schemas
 from src.config import settings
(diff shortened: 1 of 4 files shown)


src/llm/conversation.py
```diff
@@ -117,6 +117,126 @@ def _group_into_units(
...
+    return candidate + _TRUNCATION_NOTICE
+
+
+def _shrink_message(msg: dict[str, Any], target_tokens: int) -> dict[str, Any]:
+    """Return a copy of ``msg`` whose payload costs at most ~``target_tokens``.
+
+    Handles the three provider shapes: a plain string ``content`` (OpenAI), a
+    list of content blocks (Anthropic), and ``parts`` (Gemini). A message whose
+    payload is none of these is returned unchanged — the caller detects that it
+    is still over budget and logs it.
```
akattelu: A live_llm test for this could be helpful

src/llm/conversation.py
```diff
@@ -170,8 +296,27 @@ def truncate_messages_to_fit(
             + f"(~{count_message_tokens(removed_unit)} tokens)"
         )
 
-    result = system_messages + [m for unit in units for m in unit]
+    kept = [msg for unit in units for msg in unit]
+    if count_message_tokens(kept) > available_tokens:
+        kept = _shrink_to_fit(kept, available_tokens)
```
akattelu: In a way returning `400` here is correct because we can't fully support the amount of tokens they're passing in. I worry that silently truncating this will result in quality loss thats not visible to the user. Alternatively we could have it be configurable via a new config var but I think 400 is ok in this case.

Not fully following the rest of the logic, but I agree with the where claude changes and truncating for tool calls. As I understand it, that would cause an error on a tool call and the agent can continue its loop.

akattelu (changes requested):

src/llm/conversation.py
eisene: agreed. on reflection, let me remove the truncation from this PR entirely. let's keep the SQL optimization and see if the sentry goes away (it should)

eisene: done, truncation logic removed

akattelu (approved):

src/llm/conversation.py
eisene: will add this now before merging

eisene: after removing truncation i think we don't need any anymore since this doesn't touch LLMs anymore

PR #1071, opened 2026-08-25
feat(docs): cookie consent on honcho.dev/docs

erosika:
## Summary
Two changes for DEV-2465. Every docs page loads the GTM container, and PostHog loads only when the `CookieConsent` cookie grants Statistics.

## GTM container
`docs.json` swaps `integrations.posthog` for `integrations.gtm` with `GTM-NSPT9PJF`. Mintlify injects the container on every page. The published container (v10) excludes `/docs` from the Google tag, so it loads and fires nothing until Marc publishes Cookiebot and GA4 into it.

## PostHog loader
`docs/posthog-consent.js` runs once each page is interactive. It reads the `CookieConsent` cookie and loads `array.js` only when Statistics is granted or the value is `-1` (consent not required). The cookie is host-scoped, so a landing-page answer covers docs. It listens for `CookiebotOnConsentReady`, `OnAccept`, and `OnDecline`: withdrawal calls `opt_out_capturing()`, re-grant calls `opt_in_capturing()`. `onload` re-checks consent before `init`, so a decline during the download skips init. A failed script load resets the guard so the next event retries. `capture_pageview: 'history_change'` records Mintlify's client-side navigation, not only the first page.

## Trade-offs
PostHog no longer goes through `ph.mintlify.com`, so ad blockers drop some volume. Mintlify's built-in events (`$docs.content.view`, `docs.navitem.click`, `docs.search.*`, `code_block_copy`) stop. A visitor with no cookie, including a US visitor landing directly on docs, is not tracked until they answer on the landing page.

## Follow-up
After deploy: `gtm.js` loads on a docs page, no tags fire, no new cookies. Confirm Mintlify's production CSP allows `us-assets.i.posthog.com` and `us.i.posthog.com`; if not, revert the loader commits and restore `integrations.posthog`. Decide whether a custom listener recovers search analytics. The `ph_phc_*` identifier stays after decline; the Cookiebot declaration text covers that (Marc).

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -19,21 +19,14 @@
   },
   "favicon": "/favicon.svg",
   "contextual": {
-    "options": [
-      "copy",
-      "view",
-      "chatgpt",
-      "claude"
-    ]
+    "options": ["copy", "view", "chatgpt", "claude"]
   },
   "navigation": {
     "versions": [
       {
         "version": "v3.1.0",
         "api": {
-          "openapi": [
-            "v3/openapi.json"
-          ]
+          "openapi": ["v3/openapi.json"]
         },
         "tabs": [
           {
@@ -97,9 +90,7 @@
             "groups": [
               {
                 "group": "Overview",
-                "pages": [
-                  "v3/guides/overview"
-                ]
+                "pages": ["v3/guides/overview"]
               },
               {
                 "group": "Integrations",
@@ -139,9 +130,7 @@
               },
               {
                 "group": "Migrations",
-                "pages": [
-                  "v3/guides/migrations/mem0"
-                ]
+                "pages": ["v3/guides/migrations/mem0"]
               }
             ]
           },
@@ -171,9 +160,7 @@
             "groups": [
               {
                 "group": "API Documentation",
-                "pages": [
-                  "v3/api-reference/introduction"
-                ]
+                "pages": ["v3/api-reference/introduction"]
               },
               {
                 "group": "workspaces",
@@ -251,9 +238,7 @@
               },
               {
                 "group": "miscellaneous",
-                "pages": [
-                  "v3/api-reference/endpoint/keys/create-key"
-                ]
+                "pages": ["v3/api-reference/endpoint/keys/create-key"]
               }
             ]
           },
@@ -274,9 +259,7 @@
       {
         "version": "v2.5.1",
         "api": {
-          "openapi": [
-            "v2/openapi.json"
-          ]
+          "openapi": ["v2/openapi.json"]
         },
         "tabs": [
           {
@@ -323,15 +306,11 @@
             "groups": [
               {
                 "group": "Getting Started",
-                "pages": [
-                  "v2/guides/overview"
-                ]
+                "pages": ["v2/guides/overview"]
               },
               {
                 "group": "Migrations",
-                "pages": [
-                  "v2/migrations/from-mem0"
-                ]
+                "pages": ["v2/migrations/from-mem0"]
               },
               {
                 "group": "Integrations",
@@ -356,9 +335,7 @@
             "groups": [
               {
                 "group": "API Documentation",
-                "pages": [
-                  "v2/api-reference/introduction"
-                ]
+                "pages": ["v2/api-reference/introduction"]
               },
               {
                 "group": "workspaces",
@@ -462,9 +439,7 @@
       {
         "version": "v1.1.0",
         "api": {
-          "openapi": [
-            "openapi.json"
-          ]
+          "openapi": ["openapi.json"]
         },
         "tabs": [
           {
@@ -494,23 +469,15 @@
             "groups": [
               {
                 "group": "Getting Started",
-                "pages": [
-                  "v1/guides/overview",
-                  "v1/guides/streaming-response"
-                ]
+                "pages": ["v1/guides/overview", "v1/guides/streaming-response"]
               },
               {
                 "group": "Application Interfaces",
-                "pages": [
-                  "v1/guides/discord",
-                  "v1/guides/honcho-mcp"
-                ]
+                "pages": ["v1/guides/discord", "v1/guides/honcho-mcp"]
               },
               {
                 "group": "Personal Memory",
-                "pages": [
-                  "v1/guides/dialectic-endpoint"
-                ]
+                "pages": ["v1/guides/dialectic-endpoint"]
               }
             ]
           },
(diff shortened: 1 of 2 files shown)


docs/posthog-consent.js
```diff
@@ -0,0 +1,43 @@
+// Loads PostHog only when the CookieConsent cookie grants Statistics; the
+// cookie is host-scoped, so a landing-page answer covers the docs.
+;(function () {
+  var KEY = 'phc_1yrzzcgywqXGcerkkI4g7C0YfyPMcAKNOOvGcjTCiUk'
```
akattelu, line 4: is this an api key or some public identifier?

akattelu (dismissed):

docs/posthog-consent.js
lowyelling: yeah this is our Posthog project token, safe for public apps

erosika: its definitely worthy of a jump scare

docs/posthog-consent.js
```diff
@@ -0,0 +1,63 @@
...
+    s.onerror = function () {
+      loaded = false
+    }
+    s.onload = function () {
+      window.posthog.init(KEY, {
+        api_host: 'https://us.i.posthog.com',
+        ui_host: 'https://us.posthog.com',
+        cross_subdomain_cookie: true,
+        person_profiles: 'identified_only',
+      })
```
lowyelling, line 36: **real bug**  -- our docs are a single-page app (SPA) but PostHog only records viewing a page upon full load, so right now we are only capturing the first page a visitor lands on. Adding this one line will let us record address change even if JS just swaps content without browser reload. 

```suggestion
      window.posthog.init(KEY, {
        api_host: 'https://us.i.posthog.com',
        ui_host: 'https://us.posthog.com',
        cross_subdomain_cookie: true,
        person_profiles: 'identified_only',
        capture_pageview: 'history_change',
      })
```

docs/docs.json
```diff
@@ -593,8 +630,8 @@
     }
   },
   "integrations": {
-    "posthog": {
-      "apiKey": "phc_1yrzzcgywqXGcerkkI4g7C0YfyPMcAKNOOvGcjTCiUk"
+    "gtm": {
```
lowyelling, line 596: **note to self/Lily re Posthog annotation** -- Removing`integrations.posthog` is a trade-off bc we used to track even people who clicked Decline on cookies. 

without fallback, here is what we are losing:`docs.navitem.click`, `code_block_copy` ,`docs.search.query` ,`docs.search.result_click` , `docs.navitem.cta_click` ,`$docs.content.view`  

Last 30-day volumes for those analytics: 15,662 · 1,821 · 1,241 · 582 · 548 · 51,141.

fallback to recover the search analytics: small custom listener

docs/docs.json
```diff
@@ -19,14 +19,21 @@
   },
   "favicon": "/favicon.svg",
   "contextual": {
-    "options": ["copy", "view", "chatgpt", "claude"]
+    "options": [
+      "copy",
```
lowyelling: **really small nit** -- Now that each option in line 22 is in its own line, it looks like there are way more lines of changes. same applies for rest of the doc structure

docs/posthog-consent.js
```diff
@@ -0,0 +1,63 @@
...
+        window.posthog.has_opted_out_capturing &&
+        window.posthog.has_opted_out_capturing()
+      ) {
+        window.posthog.opt_in_capturing()
+      }
+      return
+    }
+    // Withdrawal mid-session: an already running instance must stop.
+    if (loaded && window.posthog && window.posthog.opt_out_capturing) {
+      window.posthog.opt_out_capturing()
```
lowyelling, line 56: **nit** -- When cookie consent is declined, we correctly aren't sending more data, but the ID cookie Posthog put in the browser still exists until it expires on its own. Specifically checked while pretending I'm from Norway that after declining, the `ph_phc_*` cookie stayed

Preferred fix is a line in the Cookiebot UI/admin. Something like: *"If you withdraw consent, collection stops immediately; identifiers already stored may remain in your browser until they expire or you clear cookies, and are not used while consent is withdrawn."* Plus one check that the scanner lists `ph_phc_..._posthog` under Statistics once it crawls /docs.

The other option is deleting the identifier on decline but it's unused while opted out anyway, and deleting it means a decline-then-re-accept visitor starts over as a new anon person, which cuts our analytics off.

docs/posthog-consent.js
```diff
@@ -0,0 +1,63 @@
...
+    if (loaded) return
+    loaded = true
+    var s = document.createElement('script')
+    s.src = 'https://us-assets.i.posthog.com/static/array.js'
+    s.async = true
+    s.onerror = function () {
+      loaded = false
+    }
+    s.onload = function () {
+      window.posthog.init(KEY, {
```
lowyelling, line 30: **bug (nit) CodeRabbit's race is real but small** -- If consent is withdrawn during the `array.js` download, `sync()` runs while `window.posthog` doesn't exist yet (so the opt-out is skipped), and then this `onload` initializes anyway

```suggestion
    s.onload = function () {
      if (!granted()) {
        loaded = false
        return
      }
      window.posthog.init(KEY, {
```

(Resetting `loaded` lets a later re-grant retry the load.)

docs/posthog-consent.js
```diff
@@ -0,0 +1,63 @@
...
+      ) {
+        window.posthog.opt_in_capturing()
+      }
+      return
+    }
+    // Withdrawal mid-session: an already running instance must stop.
+    if (loaded && window.posthog && window.posthog.opt_out_capturing) {
+      window.posthog.opt_out_capturing()
+    }
+  }
```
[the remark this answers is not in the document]
lowyelling: this one seems real — suggested a two-line guard (line 24): re-check granted() before init.

lowyelling (dismissed):
Just one real analytics bug fix and couple of smaller/nit bugs after doing some testing with VPN.

docs/docs.json
erosika: yeah its my formatter

docs/posthog-consent.js
erosika: Added in 86f8eb3e.

docs/posthog-consent.js
erosika: Added in 86f8eb3e, with the loaded reset so a re-grant retries.

docs/docs.json
erosika: Reverted in f3db11ef. docs.json diff is back to the integrations block only.

lowyelling (approved):
lgtm

PR #1076, opened 2026-08-26
fix(windows): guard Unix-only APIs so the server and deriver start on native Windows

### LINKED ISSUE #1075: [Bug] Server and deriver cannot start on native Windows (fcntl import, add_signal_handler)
### Summary

Honcho cannot start on a **native Windows** host (no WSL, no Docker). Two Unix-only APIs are
called unconditionally, and one of the two failures is effectively silent, which makes it hard
to diagnose.

This is not a WSL/Docker problem — those are Linux and work as documented. This only affects
running the server directly on Windows.

### Environment

- Windows 11, Python 3.11.16, uv 0.12.5
- PostgreSQL 18.6 + pgvector 0.8.6
- Honcho `main` @ `9380bf2`
- LLM: Azure AI Foundry model-router (OpenAI-compatible endpoint)

### 1. `fcntl` import kills the whole app

```
File "src/telemetry/reasoning_traces.py", line 7, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

`fcntl` is Unix-only. It is imported at module scope, but the module's only use of it is
advisory locking around **optional** trace logging (`REASONING_TRACES_FILE`, disabled by
default). So an off-by-default feature prevents the process from importing at all.

Worth noting `src/telemetry/metrics_collector.py` already imports `fcntl` **lazily inside the
function** (line ~320), so the codebase is inconsistent here as well.

### 2. `add_signal_handler` fails with an empty error message

After fixing (1), the deriver dies with:

```
src.deriver.queue_manager - ERROR - Error in main:
```

Note the blank message. The cause is `queue_manager.py:207`:

```python
loop.add_signal_handler(sig, lambda s=sig: asyncio.create_task(self.shutdown(s)))
```

`add_signal_handler` raises `NotImplementedError` on Windows, and that exception's `str()` is
**empty**, so `logger.error(f"Error in main: {str(e)}")` logs nothing useful. The deriver exits
with no indication of why.

### Proposed fix

PR incoming. Both are guarded so Unix behaviour is byte-for-byte unchanged:

- import `fcntl` defensively, skip the `flock` calls when it is `None`
- register signal handlers only when `sys.platform != "win32"`

2 files, +21/-9. `ruff check` and `ruff format --check` pass.

### Two further blockers (not in the PR — launcher/packaging, not source bugs)

Reporting these for completeness, since they are needed to actually run on Windows:

**3. uvicorn hardcodes an event loop psycopg cannot use.**

```
psycopg.InterfaceError: Psycopg cannot use the 'ProactorEventLoop' to run in async mode.
Please use a compatible event loop, for instance by setting
'asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())'
```

Following that advice does **not** work, because `uvicorn/loops/asyncio.py` ignores the policy:

```python
if sys.platform == "win32" and not use_subprocess:
    return asyncio.ProactorEventLoop
```

`uvicorn.Config()` takes no `loop_factory` either. The workaround is to build a
`SelectorEventLoop` yourself and drive `Server.serve()` on it.

**4. `src/deriver/__main__.py` imports `uvloop`,** which has no Windows wheel, so
`python -m src.deriver` cannot be used at all.

Both are solvable with small launcher scripts. I've left them out of the PR since they are
arguably packaging concerns rather than source bugs — but I'm happy to contribute either the
launchers or a `docs/` page if you'd like them in-tree.

### Workaround for anyone hitting this now

Full write-up with the patch, launchers, service registration, and an update workflow that
preserves the patches across upgrades:

https://gist.github.com/DesarrolloProsis/63b226c6afb26da5df1017af969bd31e

### Verified

With the patch plus launchers: API serves, deriver drains the queue and generates conclusions,
and dialectic chat returns correct results at `reasoning_level=medium`.

DesarrolloProsis:
## Description

Honcho cannot start on a **native Windows** host (no WSL, no Docker) because two Unix-only APIs are called unconditionally: `src/telemetry/reasoning_traces.py` imports `fcntl` at module scope, and `src/deriver/queue_manager.py` calls `loop.add_signal_handler()`, which raises `NotImplementedError` on Windows with an **empty** `str()` — so the deriver dies logging only `Error in main: ` with no message. This guards both, plus adds a Windows shutdown path and cross-platform trace-file locking so the guards don't silently degrade behaviour.

No behaviour change on Unix: every change is behind a `sys.platform` check or an import-availability check.

## Proofs

Verified on Windows 11 / Python 3.11.16 / PostgreSQL 18.6 / pgvector 0.8.6, against an OpenAI-compatible endpoint (Azure AI Foundry model-router).

**Before — API server:**
```
File "src/telemetry/reasoning_traces.py", line 7, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
```

**Before — deriver (note the blank error):**
```
2026-08-25 21:07:53,960 - __main__ - INFO - deriver: embedding schema validated, starting queue manager
2026-08-25 21:07:53,960 - src.cache.client - INFO - Cache disabled, using in-memory cache
2026-08-25 21:07:53,960 - src.deriver.queue_manager - ERROR - Error in main:
```

**After — deriver runs and processes work:**
```
2026-08-25 21:14:01 - src.reconciler.scheduler - INFO - ReconcilerScheduler started with 2 tasks: ['sync_vectors', 'cleanup_queue']
2026-08-25 21:15:05 - PERFORMANCE minimal_deriver_5_sergio | llm_call_duration=2552ms | total_processing_time=3485ms | observation_count=8
2026-08-25 21:19:40 - PERFORMANCE minimal_deriver_7_sergio | llm_call_duration=6061ms | total_processing_time=10361ms | observation_count=31
```

**After — API serves and dialectic returns:**
```
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000
INFO:     127.0.0.1:53004 - "POST /v3/workspaces/hermes/peers/hermes/chat HTTP/1.1" 200 OK
PERFORMANCE dialectic_chat | tool_calls=4 | input_tokens=18096 | output_tokens=1013 | total_duration=15267ms
```

**Queue drained cleanly, conclusions generated (SQL against the live DB):**
```
queue processed   : t | 7
documents         : 17
errors            : (none)
```

**Lint / format — commands actually executed:**
```
$ uv run ruff check src/telemetry/reasoning_traces.py src/main.py src/deriver/queue_manager.py
All checks passed!

$ uv run ruff format --check src/telemetry/reasoning_traces.py src/main.py src/deriver/queue_manager.py
3 files already formatted
```

I did **not** run `uv run pytest tests/` — the suite orchestrates Docker-backed services that were not available on this host. Flagging that explicitly rather than implying a green run.

## Checklist

Fixes #1075

---

### Scope note

Two further Windows blockers are **not** fixed here, because they are launcher/packaging concerns rather than source bugs, and fixing them in-tree implies opinionated decisions you may want to make differently:

- uvicorn hardcodes `ProactorEventLoop` on win32 (`uvicorn/loops/asyncio.py`), which psycopg cannot use. Setting an asyncio policy does **not** help — uvicorn builds the loop from its own factory and ignores it, and `uvicorn.Config()` takes no `loop_factory`. Workaround: drive `Server.serve()` on a `SelectorEventLoop` you construct.
- `src/deriver/__main__.py` imports `uvloop`, which has no Windows wheel.

Launchers for both are in the gist linked from #1075. Happy to contribute them or a `docs/` page if wanted.

---

*Disclosure: prepared with an AI agent (Hermes) while self-hosting Honcho on a Windows workstation. The failures are real ones hit during that install, and every log line and command result above was executed on the running stack — not inferred. Reviewed by me before submitting.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown handling on Windows and other platforms.
  * Enhanced telemetry trace writing across operating systems.
  * Prevented concurrent telemetry writes from corrupting trace data.
  * Improved reliability when trace files are accessed simultaneously.
  * Ensured successfully written telemetry records are flushed promptly.
  * Added safer handling when a trace file cannot be locked, preventing incomplete writes.

### DIFF
--- a/src/telemetry/reasoning_traces.py
+++ b/src/telemetry/reasoning_traces.py
@@ -6,25 +6,13 @@ This module provides structured JSONL logging of LLM inputs/outputs.
 
 import contextlib
 import json
-import os
 import sys
 import time
 from collections.abc import Generator
 from contextlib import contextmanager
+from importlib import import_module
 from pathlib import Path
 from typing import IO, Any
-
-fcntl: Any = None
-msvcrt: Any = None
-if sys.platform == "win32":
-    import msvcrt as _msvcrt
-
-    msvcrt = _msvcrt
-else:
-    import fcntl as _fcntl
-
-    fcntl = _fcntl
-
 
 from pydantic import BaseModel
 
@@ -33,6 +21,8 @@ from src.config import (
     ModelConfig,
     settings,
 )
+
+locking_module: Any = import_module("msvcrt" if sys.platform == "win32" else "fcntl")
 
 
 @contextmanager
@@ -45,23 +35,27 @@ def _locked(f: IO[str]) -> Generator[None, None, None]:
     write still proceeds unlocked rather than losing the trace.
     """
     if sys.platform != "win32":
-        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+        locking_module.flock(f.fileno(), locking_module.LOCK_EX)
         try:
             yield
         finally:
-            fcntl.flock(f.fileno(), fcntl.LOCK_UN)
+            locking_module.flock(f.fileno(), locking_module.LOCK_UN)
     elif sys.platform == "win32":
-        f.seek(0, os.SEEK_END)
+        # Every writer must coordinate on the same byte range. Keep the lock
+        # offset so it can be restored before LK_UNLCK after the append.
+        lock_offset = 0
+        f.seek(lock_offset)
         try:
-            msvcrt.locking(f.fileno(), msvcrt.LK_LOCK, 1)
+            locking_module.locking(f.fileno(), locking_module.LK_LOCK, 1)
         except OSError:  # lock unavailable after retries — do not drop the trace
             yield
             return
         try:
             yield
         finally:
+            f.seek(lock_offset)
             with contextlib.suppress(OSError):
-                msvcrt.locking(f.fileno(), msvcrt.LK_UNLCK, 1)
+                locking_module.locking(f.fileno(), locking_module.LK_UNLCK, 1)
     else:  # pragma: no cover - no locking primitive available
         yield
 
--- a/src/telemetry/reasoning_traces.py
+++ b/src/telemetry/reasoning_traces.py
@@ -4,11 +4,25 @@ Utility for logging traces from LLM calls.
 This module provides structured JSONL logging of LLM inputs/outputs.
 """
 
-import fcntl
+import contextlib
 import json
+import os
 import time
+from collections.abc import Iterator
+from contextlib import contextmanager
 from pathlib import Path
-from typing import Any
+from typing import IO, Any
+
+try:  # POSIX
+    import fcntl
+except ImportError:  # pragma: no cover - platform dependent
+    fcntl = None  # type: ignore[assignment]
+
+try:  # Windows
+    import msvcrt
+except ImportError:  # pragma: no cover - platform dependent
+    msvcrt = None  # type: ignore[assignment]
+
 
 from pydantic import BaseModel
 
@@ -17,6 +31,37 @@ from src.config import (
     ModelConfig,
     settings,
 )
+
+
+@contextmanager
+def _locked(f: IO[str]) -> Iterator[None]:
+    """Exclusively lock an open file for the duration of the block.
+
+    Multiple processes (API server and deriver) append to the same traces file, so
+    writes must be serialized. POSIX uses fcntl.flock; Windows uses msvcrt.locking,
+    which locks a byte range from the current offset. If neither is available the
+    write still proceeds unlocked rather than losing the trace.
+    """
+    if fcntl is not None:
+        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+        try:
+            yield
+        finally:
+            fcntl.flock(f.fileno(), fcntl.LOCK_UN)
+    elif msvcrt is not None:
+        f.seek(0, os.SEEK_END)
+        try:
+            msvcrt.locking(f.fileno(), msvcrt.LK_LOCK, 1)
+        except OSError:  # lock unavailable after retries — do not drop the trace
+            yield
+            return
+        try:
+            yield
+        finally:
+            with contextlib.suppress(OSError):
+                msvcrt.locking(f.fileno(), msvcrt.LK_UNLCK, 1)
+    else:  # pragma: no cover - no locking primitive available
+        yield
 
 
 def get_reasoning_traces_file_path() -> Path | None:
@@ -97,7 +142,5 @@ def log_reasoning_trace(
         trace_entry["output"]["tool_calls"] = response.tool_calls_made
 
     # Use file locking to handle concurrent writes from multiple processes
-    with open(traces_file, "a") as f:
-        fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+    with open(traces_file, "a") as f, _locked(f):
         f.write(json.dumps(trace_entry) + "\n")
-        fcntl.flock(f.fileno(), fcntl.LOCK_UN)
--- a/src/telemetry/reasoning_traces.py
+++ b/src/telemetry/reasoning_traces.py
@@ -41,11 +41,17 @@ def _locked(f: IO[str]) -> Generator[bool, None, None]:
     writes must be serialized. POSIX uses fcntl.flock; Windows uses msvcrt.locking
     on a fixed byte range, retried under an explicit policy.
 
-    Yields True when the lock is held. If Windows cannot acquire it within the
-    retry budget the block is entered with False and the caller must skip the
-    write: an unlocked append can interleave with another process and corrupt the
-    JSONL file, so a dropped trace is preferable. Tracing is an opt-in debugging
-    aid, so failure is logged rather than raised into the LLM call path.
+    Acquisition failure is logged rather than raised: tracing is an opt-in
+    debugging aid, and callers invoke it from the LLM call path.
+
+    Args:
+        f: Open file handle to lock for the duration of the block.
+
+    Yields:
+        True while the lock is held, False when Windows cannot acquire it
+            within the retry budget. Callers must skip the write when False,
+            because an unlocked append can interleave with another process
+            and corrupt the JSONL file.
     """
     if sys.platform != "win32":
         locking_module.flock(f.fileno(), locking_module.LOCK_EX)
--- a/src/telemetry/reasoning_traces.py
+++ b/src/telemetry/reasoning_traces.py
@@ -6,6 +6,7 @@ This module provides structured JSONL logging of LLM inputs/outputs.
 
 import contextlib
 import json
+import logging
 import sys
 import time
 from collections.abc import Generator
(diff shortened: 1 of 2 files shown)


DesarrolloProsis:
Thanks — both findings were fair, and I've pushed fixes for them rather than arguing the guards were sufficient. You were right that "skip it on Windows" traded a crash for silent degradation in both cases.

**1. Windows shutdown path**

Correct — `initialize()` skipping registration entirely meant `shutdown_event` never got set, schedulers never stopped, and `active_tasks` was never drained on Windows.

`add_signal_handler()` is unavailable, but `signal.signal()` *is* supported on Windows for SIGINT/SIGTERM. Windows now registers the same `self.shutdown(...)` path through it, bridged onto the loop with `call_soon_threadsafe` since the handler fires in the main thread outside the loop:

```python
def _win_signal_handler(signum: int, _frame: object) -> None:
    loop.call_soon_threadsafe(
        lambda: asyncio.create_task(self.shutdown(signal.Signals(signum)))
    )
```

Same graceful shutdown on both platforms; each just uses its native registration primitive.

**2. Shared trace file locking**

Also correct — the API server and deriver both append to `REASONING_TRACES_FILE`, so dropping the lock risked interleaved records.

Replaced the conditional `flock` with a `_locked()` context manager: `fcntl.flock` on POSIX, `msvcrt.locking` on Windows, and an unlocked fallback only if neither exists (a trace is worth more than a hard failure).

Verified with the concurrency test you asked for — 6 processes × 40 writes to one file:

```
platform          : win32
expected records  : 240
actual lines      : 240
malformed records : 0
RESULT            : PASS
```

I ran that as a scratch script rather than adding it to the suite, since `tests/` here is Docker-orchestrated and I couldn't run the full suite on this host. Happy to contribute it as a proper test if you'd like it in-tree — just say where it belongs.

Diff is now +71/−11 across the same 2 files. `ruff check` and `ruff format --check` pass.

DesarrolloProsis:
Understood on the gate, and I want to flag a process mistake on my side rather than let it pass silently: I opened this PR immediately after filing #1075, without first reading CONTRIBUTING.md. The "If you're an agent" section says explicitly to check for the `maintainer-approved` label before writing code and *not* to open a PR in order to establish the issue link afterwards. I did exactly the thing that section warns against. Apologies for the noise in your queue.

Leaving it open rather than closing it, since the work is already done and verified — but entirely understood if you'd rather it auto-close at 72h and be reopened once (if) #1075 gets triaged. Nothing is lost either way.

For triage, the short version of #1075: Honcho cannot start at all on a native Windows host (no WSL, no Docker). `import fcntl` at module scope in `src/telemetry/reasoning_traces.py` kills the API server, and `loop.add_signal_handler()` in the deriver raises `NotImplementedError` whose `str()` is **empty**, so the deriver dies logging only `Error in main: ` with no message. That second one seems worth fixing regardless of the Windows question — a blank error is hostile to debug on any platform.

Scope is deliberately narrow: 2 files, guards only, no behaviour change on Unix. Two further Windows blockers (uvicorn hardcoding `ProactorEventLoop`, and `src/deriver/__main__.py` importing `uvloop`) are documented in the issue but intentionally left out, since fixing them in-tree implies launcher decisions that are yours to make.

Also: "Allow edits by maintainers" is enabled, and the fork is personal rather than org-owned, so you can push directly to the branch if it's easier than a review round-trip.

No urgency from my end — I'm running the patch locally either way.

src/telemetry/reasoning_traces.py
```diff
@@ -4,11 +4,25 @@
...
+
+try:  # POSIX
+    import fcntl
+except ImportError:  # pragma: no cover - platform dependent
+    fcntl = None  # type: ignore[assignment]
+
+try:  # Windows
+    import msvcrt
+except ImportError:  # pragma: no cover - platform dependent
+    msvcrt = None  # type: ignore[assignment]
```
akattelu: I don't think we should be adding this type of guard block that uses try/catch for this detection. There should be more robust ways of doing this such as other ones in this PR.

akattelu (changes requested):
I approved workflows to run on this, left one comment. You should fix the CodeRabbit suggestion but mostly looks good besides that. Note that we're not explicitly supporting windows as a stable target, but this seems like a small diff that could help unblock.

src/telemetry/reasoning_traces.py
DesarrolloProsis: Thanks for approving the workflows and for the steer here — agreed on the substance, and this one is already gone.

The try/except detection you flagged was replaced in `e0cc8f8`; module selection is now a `sys.platform` check like the rest of the PR:

```python
locking_module: Any = import_module("msvcrt" if sys.platform == "win32" else "fcntl")
```

So this thread is against the stale diff. I've also pushed the CodeRabbit fix you asked me to take — details in that thread; the unlocked-write fallback is gone.

Whenever you next have time for it, no rush — and understood on Windows not being a stable target. The `add_signal_handler` half stands on its own regardless, since a `NotImplementedError` with an empty `str()` logging `Error in main:` and nothing else is hostile to debug on any platform.

src/telemetry/reasoning_traces.py
```diff
@@ -18,6 +22,43 @@
...
+    elif sys.platform == "win32":
+        # Every writer must coordinate on the same byte range. Keep the lock
+        # offset so it can be restored before LK_UNLCK after the append.
+        lock_offset = 0
+        f.seek(lock_offset)
+        try:
+            locking_module.locking(f.fileno(), locking_module.LK_LOCK, 1)
+        except OSError:  # lock unavailable after retries — do not drop the trace
+            yield
+            return
```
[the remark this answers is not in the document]
DesarrolloProsis: Correct, and taken. Falling through to an unlocked `f.write()` traded a dropped trace for a corrupt shared file, which is the worse of the two — the API server and deriver both append to `REASONING_TRACES_FILE`, so an interleaved record is exactly what the lock is there to prevent.

Fixed in `0985a2c` by making the retry policy explicit rather than leaning on `msvcrt`'s implicit `LK_LOCK` behaviour, and by never writing without the lock:

- `LK_LOCK` → `LK_NBLCK` in a bounded retry loop (10 attempts, 100 ms apart)
- `_locked` now yields `bool`; on exhaustion it logs a warning and yields `False`, and the caller skips the write entirely
- Removed the unreachable `else` branch — the `if != "win32"` / `elif == "win32"` pair already covered every case

I did not raise on failure, deliberately. `log_reasoning_trace` is called unguarded from `src/llm/api.py:410` and `:478`, so surfacing the `OSError` would take down live LLM calls over an opt-in debugging feature. Dropping one trace with a warning is the proportionate failure mode; corrupting the file is not. Happy to make it raise if you'd rather the failure be loud.

POSIX path unchanged. Concurrency check on Windows, 6 processes × 40 writes to one file:

```
platform         : win32
expected records : 240
actual lines     : 240
malformed records: 0
RESULT           : PASS
```

Worth stating plainly, since it bears on how much that proves: I ran the same harness on Linux with the `flock` removed as a control, and it still passed at payloads from 200 B to 512 KB. `O_APPEND` atomicity means the POSIX run cannot distinguish a working lock from no lock at all, so it is a no-regression check only — the win32 run above is the one carrying the evidence.

`ruff check` and `ruff format --check` pass. Still haven't run `uv run pytest tests/` — Docker-orchestrated suite, not available on this host.

akattelu (approved):

akattelu (changes requested):
@DesarrolloProsis can you fix the CI basedpyright errors?

DesarrolloProsis:
Fixed in `ff53f27` — the `basedpyright` failure was `reportImplicitStringConcatenation` on the split log-message literal in `_locked`; it's now a single literal. `uv run basedpyright` is clean on the file, and `ruff check` / `ruff format --check` pass.

The workflow runs on that push are sitting at `action_required`, so they'll need approving before CI reports.

---

*Disclosure: prepared with an AI agent (Claude Code). Reviewed by me before submitting.*

eisene:
apologies but after some internal discussion we decided that we can't afford right now to maintain a native windows implementation of the honcho self-hosting stack. if you could come back to the original issue and let us know why running the honcho fleet in WSL in a linux container isn't acceptable here?

PR #1077, opened 2026-08-26
fix(db): enable pgvector HNSW iterative scan by default (#1048)

### LINKED ISSUE #1048: [Bug] Filtered pgvector HNSW conclusion queries silently under-return with iterative scans off

shellybotmoyer:
## Summary

Supersedes #1070 (closed due to basedpyright CI failures). Same fix, with all type-check issues resolved.

Enables pgvector HNSW iterative scan by default (`strict_order`) to fix silent under-returning of filtered conclusion queries (#1048).

### Changes

1. **Validate the setting** — `HNSW_ITERATIVE_SCAN` is now `Literal["off", "strict_order", "relaxed_order"] | None` instead of an arbitrary `str`. Invalid values are rejected at config-load time with a Pydantic validation error.
2. **Match the existing connect hook** — Uses `set_config` instead of `SET` SQL for the connection listener, matching the existing hook pattern.
3. **Default to `strict_order`** — Enables iterative scans by default so filtered HNSW queries return complete results.
4. **Test coverage** — Tests for valid/invalid values, None handling, and connect listener registration.

### basedpyright fixes

- `# pyright: ignore[reportArgumentType]` on invalid-value rejection tests (intentionally passing bad values)
- `# pyright: ignore[reportPrivateUsage]` on tests that verify the private connect listener function

Addresses all four points from @akattelu's review on #1060.

Closes #1048

## Summary by CodeRabbit

* **New Features**
  * Added configuration for pgvector HNSW iterative scans, supporting disabled, strict-order, and relaxed-order modes.
  * Applies the selected setting to new database connections.
  * Validates that pgvector 0.8.0 or newer is available when iterative scans are enabled.
* **Documentation**
  * Documented supported values, per-connection behavior, and version requirements.
* **Bug Fixes**
  * Connection setup continues safely if applying the HNSW setting fails.
* **Tests**
  * Added coverage for configuration validation and connection setup.

### DIFF
--- a/docs/v3/contributing/configuration.mdx
+++ b/docs/v3/contributing/configuration.mdx
@@ -557,6 +557,12 @@ DB_SQL_DEBUG=false
 # Per-connection establish timeout (seconds) so a single connection attempt
 # fails fast instead of hanging when the server/pooler is unreachable.
 DB_CONNECT_TIMEOUT_SECONDS=2
+
+# pgvector HNSW iterative scan (requires pgvector >= 0.8.0)
+# When set, each new pool connection gets the GUC applied so filtered
+# HNSW queries return full top_k results instead of silently under-returning.
+# Valid values: "off", "strict_order", "relaxed_order", or unset (None).
+DB_HNSW_ITERATIVE_SCAN=strict_order
 ```
 
 ### Authentication
--- a/src/config.py
+++ b/src/config.py
@@ -739,6 +739,16 @@ class DBSettings(HonchoSettings):
     SQL_DEBUG: bool = False
     TRACING: bool = False
 
+    # pgvector HNSW iterative scan mode for filtered approximate searches.
+    # When set (default "strict_order"), applied as a per-connection server
+    # setting so filtered HNSW queries return full top_k results instead of
+    # silently under-returning when out-of-scope rows consume the scan budget.
+    # Requires pgvector >= 0.8.0.
+    # See: https://github.com/pgvector/pgvector#iterative-index-scans
+    HNSW_ITERATIVE_SCAN: Literal["off", "strict_order", "relaxed_order"] | None = (
+        "strict_order"
+    )
+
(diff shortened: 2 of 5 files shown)


akattelu (changes requested):
Hey @shellybotmoyer, thanks for following up on the #1060 notes. Literal validation, `set_config` with bind params, and the docs note are all in the right place, and locally a new pool connection does land with `SHOW hnsw.iterative_scan = strict_order` without
 leaving the read engine idle-in-transaction.

   A couple of things before merge:

   - basedpyright is red on this head. The `# pyright: ignore[reportArgumentType]` comments in `tests/test_hnsw_iterative_scan.py` sit on the line above the `DBSettings(...)` calls, so they don't apply, and the `RuntimeError` in `init_db` trips
 `reportImplicitStringConcatenation`. Same class of failure that blocked #1070.
   - The tests still don't check a real connection. `test_connect_listener_not_registered_when_disabled` claims the handler short-circuits on `None`, but `_set_hnsw_iterative_scan_on_connect` has no early return. The dummy conn just hits the bare `except`.
 `event.contains` also doesn't prove the GUC was applied.

   Suggested shape: put the pyright ignores on the constructor lines, join the error-message strings, and add one test that checks out a connection and asserts `SHOW hnsw.iterative_scan`. I'd drop or rewrite the dummy-conn test so it doesn't claim a short-circuit
 that isn't there. I still don't think we need a multi-scope cardinality test for this cut.

   Minor docs nit: empty string is not disable (`DB_HNSW_ITERATIVE_SCAN=` fails validation). Disable is omit the var / `None`.

shellybotmoyer:
Superseded by #1091 which addresses all maintainer review feedback. Closing this version.

PR #1079, opened 2026-08-26
fix(tools): accept singular `search_message` as an alias for `search_messages`

### LINKED ISSUE #1078: [Bug] Agent calls `search_message` but Honcho only knows `search_messages`
**Deploy mode:** Self-hosted

**Honcho version:** v3.0.12 (still present on `main` at v3.1.0)

## Describe the bug

The Dreamer agent sometimes calls a tool named `search_message` (no final "s"). Honcho only registers a tool named `search_messages` (with the "s"). When the agent uses the singular name, Honcho replies `Unknown tool: search_message`. The agent retries with the same name, and after a few loops it gives up with `Tool execution loop reached max iterations`. Dreams then fail.

The Dialectic agent uses the same tool and can hit the same problem.

## Steps to reproduce

1. Schedule a dream for any peer pair.
2. Watch the deriver worker log.

## Logs and evidence

```
Unknown tool: search_message
Tool execution loop reached max iterations (12)
```

## Additional context

The prompts all say `search_messages` correctly — this is the model dropping the plural, not a typo in the prompt.

Fixed in PR #1079, which accepts `search_message` as an alias that runs the same code as `search_messages`.

gwyntel-git:
## Description

The Dreamer and Dialectic agents sometimes call a tool named `search_message` (no final "s"). Honcho only registers `search_messages`, so it replies `Unknown tool: search_message`. The agent retries the same name and eventually hits `Tool execution loop reached max iterations`, which makes dreams fail.

This PR adds one alias entry so `search_message` runs the same handler as `search_messages`.

## Proofs

- New unit test `test_executor_accepts_search_message_alias` in `tests/utils/test_agent_tools.py::TestToolExecutor`.
- `uv run pytest tests/utils/test_agent_tools.py::TestToolExecutor -q` → `6 passed`.
- Without the fix, the new test fails with `Unknown tool: search_message` (the test catches the bug).
- `uv run ruff check` and `ruff format --check` → pass; `uv run basedpyright` → 0 errors; `pre-commit run bandit` → pass.

## Checklist

Fixes #1078

## Summary by CodeRabbit

- **Bug Fixes**
  - Added support for the singular `search_message` tool name as an alternative to `search_messages`.
  - Requests using the singular name are now correctly routed to message search instead of being rejected as unknown.

- **Tests**
  - Added coverage verifying the singular tool name works as expected.

### DIFF
--- a/src/utils/agent_tools.py
+++ b/src/utils/agent_tools.py
@@ -2588,6 +2588,8 @@ _TOOL_HANDLERS: dict[str, Callable[[ToolContext, dict[str, Any]], Any]] = {
     "search_memory": _handle_search_memory,
     "get_observation_context": _handle_get_observation_context,
     "search_messages": _handle_search_messages,
+    # Some models drop the plural and emit "search_message"; accept it as an alias.
+    "search_message": _handle_search_messages,
     "grep_messages": _handle_grep_messages,
     "get_messages_by_date_range": _handle_get_messages_by_date_range,
     "search_messages_temporal": _handle_search_messages_temporal,
--- a/tests/utils/test_agent_tools.py
+++ b/tests/utils/test_agent_tools.py
@@ -1733,6 +1733,22 @@ class TestToolExecutor:
 
         assert "Unknown tool" in result
 
+    async def test_executor_accepts_search_message_alias(self, tool_test_data: Any):
+        """The singular "search_message" resolves to the search handler."""
+        workspace, peer1, peer2, session, _, _ = tool_test_data
+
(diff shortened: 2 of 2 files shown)


gwyntel-git:
Retracting. Root cause was my self-hosted deployment, not Honcho: my local Postgres was initialized as SQL_ASCII, so the dreamer's valid Unicode deductions (em-dash, smart quotes) could not be INSERTed. The model was emitting `search_messages` (plural) correctly the whole time. Closing this PR and the linked issue. Thanks for the skepticism — it was right.

PR #1083, opened 2026-08-26
fix(ci): stop skipping unified tests on merge to main

Rajat-Ahuja1997:
- The gate job in CI only runs on pull_request: labeled, so it is skipped on push to main. Because it is skipped, the unified-tests job inherits that skip: The implication is that the `unified-tests` job inherits that skip, and does not run. -

- This PR adds !cancelled() to the unified-tests condition, implying we should always run the job on merge to main unless it has been canceled

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated test workflow handling for push-triggered runs.
  * Prevented tests from incorrectly being skipped while still stopping execution when a run is cancelled.


eisene (approved):
lgtm

PR #1085, opened 2026-08-26
test(scopes): pin DREAM.ENABLED so scope-route tests don't depend on local config

### LINKED ISSUE #1099: [Bug] Five scope-route tests fail on a clean checkout when the local .env sets DREAM_ENABLED=false
### Deploy mode

Self-hosted

### Honcho version

`main` @ `9380bf27` (pyproject `3.1.0`)

### Describe the bug

Five tests fail on a clean checkout of `main` for any contributor whose local `.env` carries `DREAM_ENABLED=false`:

```
tests/routes/test_scope_route_policy.py::test_allowing_position_accepts_a_real_scope[POST:observer]
tests/routes/test_scope_route_policy.py::test_refusing_position_rejects_a_real_scope[POST:observed]
tests/routes/test_scope_route_policy.py::test_refusing_position_allows_unflagged_squatter[POST:observed]
tests/routes/test_scope_route_policy.py::test_refusing_position_and_a_missing_reserved_name[POST:observed]
tests/routes/test_scopes.py::test_dream_cannot_be_queued_for_a_future_scope
```

`POST /v3/workspaces/{id}/schedule_dream` short-circuits on `not settings.DREAM.ENABLED` (`src/routers/workspaces.py:249`) and returns

```
400 {"detail":"Dreams are not enabled in the system configuration"}
```

before any scope check runs. Both files assert on the scope behavior *behind* that guard, so they see a 400 where they expect 204 or 422.

CI does not catch it because CI has no `.env` and `DreamSettings.ENABLED` defaults to `True`.

### Steps to reproduce

```bash
# with a local Postgres available to the suite
echo 'DREAM_ENABLED=false' >> .env
uv run --frozen pytest tests/routes/test_scope_route_policy.py tests/routes/test_scopes.py
# -> 5 failed, 106 passed
```

Removing the line makes all 111 pass.

### The secondary issue: an env var cannot work around it

`src/config.py:23` calls `load_dotenv(override=True)`, so a `.env` file **beats a real environment variable**. `DREAM_ENABLED=true uv run pytest` changes nothing.

That inverts the precedence documented a few hundred lines below in `settings_customise_sources`:

```python
# Correct precedence: init > env > .env > toml > secrets > defaults
```

The only escape hatch is `PYTHON_DOTENV_DISABLED=1`, which also drops `DB_CONNECTION_URI` and points the suite at whatever happens to be on port 5432.

This may be intentional for local dev, but the comment and the behavior disagree, and it makes the failure above much harder to diagnose than it looks. Happy to split it into its own issue if you would rather keep them separate — it is not what the attached PR fixes.

### Logs and evidence

```
E  AssertionError: POST /v3/workspaces/{workspace_id}/schedule_dream refused a scope in the
   legitimate position 'observer': expected (204,), got 400 —
   {"detail":"Dreams are not enabled in the system configuration"}
```

### Proposed fix

Pin `settings.DREAM.ENABLED` with `monkeypatch` in the affected tests — the convention this suite already uses in `tests/dreamer/test_dream_scheduler.py:26` (whose comment reads "Patch DREAM.ENABLED to True so tests work regardless of local config") and `tests/routes/test_workspaces.py:739,791`. The two files above simply were not given the same treatment.

No test asserts dreams-disabled behavior, so nothing is masked.

PR #1085 implements this: 2250 passed / 47 skipped on the full suite, and green under default random ordering.

cfdude:
Fixes #1099

### Problem

Five tests fail on a clean checkout of `main` for anyone whose local `.env` carries `DREAM_ENABLED=false`:

```
tests/routes/test_scope_route_policy.py::test_allowing_position_accepts_a_real_scope[POST:observer]
tests/routes/test_scope_route_policy.py::test_refusing_position_rejects_a_real_scope[POST:observed]
tests/routes/test_scope_route_policy.py::test_refusing_position_allows_unflagged_squatter[POST:observed]
tests/routes/test_scope_route_policy.py::test_refusing_position_and_a_missing_reserved_name[POST:observed]
tests/routes/test_scopes.py::test_dream_cannot_be_queued_for_a_future_scope
```

`POST /v3/workspaces/{id}/schedule_dream` short-circuits on `not settings.DREAM.ENABLED` (`src/routers/workspaces.py:249`) and returns

```
400 {"detail":"Dreams are not enabled in the system configuration"}
```

before any scope check runs. Both files assert on the scope behavior behind that guard, so they get a 400 where they expect 204/422.

CI doesn't see this because CI has no `.env` and `DreamSettings.ENABLED` defaults to `True`.

### Why an env var doesn't fix it

`src/config.py:23` calls `load_dotenv(override=True)`, so a `.env` file **beats a real environment variable**. `DREAM_ENABLED=true uv run pytest` changes nothing. That inverts the precedence documented a few hundred lines further down in `settings_customise_sources`:

```python
# Correct precedence: init > env > .env > toml > secrets > defaults
```

Worth a separate look, but pinning the setting in-process is the reliable fix either way.

### Fix

Pin `settings.DREAM.ENABLED` with `monkeypatch` — the convention this suite already uses in `tests/dreamer/test_dream_scheduler.py:26` ("Patch DREAM.ENABLED to True so tests work regardless of local config") and `tests/routes/test_workspaces.py:739,791`.

- `test_scope_route_policy.py` gets a module-level autouse fixture; it enumerates the scope policy of every peer-carrying route, `schedule_dream` among them.
- `test_scopes.py` pins it for the one dream test only.

No test asserts dreams-disabled behavior, so nothing is masked.

### Verification

```
2250 passed, 47 skipped   # full suite, uv run --frozen
111 passed                # the two files, previously 5 failed
```

Also green under default (random) ordering, so the autouse fixture is order-independent.

## Summary by CodeRabbit

- **Tests**
  - Improved route-policy test reliability across all peer-carrying routes.
  - Ensured dream scheduling tests run consistently regardless of local configuration.
  - Added coverage confirming invalid future scopes are rejected without creating queue entries.
  - Reformatted existing assertions for improved readability.

### DIFF
--- a/tests/routes/test_scope_route_policy.py
+++ b/tests/routes/test_scope_route_policy.py
@@ -58,9 +58,29 @@ from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from src import models
+from src.config import settings
 from src.main import app
 from src.models import Peer, Workspace
 from src.utils.scopes import scope_peer_name
+
+
+@pytest.fixture(autouse=True)
+def _dreams_enabled(monkeypatch: pytest.MonkeyPatch):
+    """Force DREAM.ENABLED on so this file works regardless of local config.
+
+    This module enumerates the scope policy of every peer-carrying route,
+    `schedule_dream` among them. That route short-circuits with a 400
+    ("Dreams are not enabled in the system configuration") before any scope
+    check runs, so an operator `.env` carrying DREAM_ENABLED=false turns the
+    schedule_dream cases into false failures. Note `src/config.py` calls
+    `load_dotenv(override=True)`, so a `.env` beats a real environment
+    variable — pinning the setting here is the only reliable fix.
+
+    Same intent as tests/dreamer/test_dream_scheduler.py and
+    tests/routes/test_workspaces.py, which already pin it per-test.
+    """
+    monkeypatch.setattr(settings.DREAM, "ENABLED", True)
+
 
 # Request parameters and model fields that carry a peer name, in any position.
 _PEER_PARAM_NAMES = {
@@ -744,9 +764,9 @@ def test_policy_entries_are_well_formed():
                     "status is 422 — missing_status is for the permissive cases"
                 )
             else:
-                assert (
-                    len(case.missing_reason.strip()) > 30
-                ), f"{case.key} tolerates a missing reserved name; say why"
+                assert len(case.missing_reason.strip()) > 30, (
+                    f"{case.key} tolerates a missing reserved name; say why"
+                )
                 assert case.missing_status, (
                     f"{case.key} tolerates a missing reserved name; name the exact "
                     "status(es) it should get, so a 5xx cannot satisfy the case"
@@ -757,9 +777,9 @@ def test_policy_entries_are_well_formed():
                 f"{case.key}: an allow case needs a builder and an expected "
                 "allow_status together, or neither"
             )
-            assert (
-                case.refuse_missing is None
-            ), f"{case.key}: refuse_missing applies to REFUSE cases only"
+            assert case.refuse_missing is None, (
+                f"{case.key}: refuse_missing applies to REFUSE cases only"
+            )
 
 
 _REFUSING = tuple(case for case in POLICY if case.refuse)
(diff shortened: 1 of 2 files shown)


cfdude:
Linked — filed as #1099 and added `Fixes #1099` to the description above.

#1099 needs the `maintainer-approved` label before this leaves the gate. Short version for whoever triages it: five scope-route tests fail on a clean `main` checkout whenever a contributor's local `.env` sets `DREAM_ENABLED=false`, because `schedule_dream` 400s on the config guard before any scope check runs. The fix applies the convention the suite already uses in `tests/dreamer/test_dream_scheduler.py` and `tests/routes/test_workspaces.py` — those two files pin the setting; the two here were never given the same treatment.

The issue also notes a secondary finding that this PR does **not** touch: `src/config.py:23` calls `load_dotenv(override=True)`, so a `.env` beats a real environment variable and `DREAM_ENABLED=true` on the command line has no effect. That contradicts the `# Correct precedence: init > env > .env > toml > secrets > defaults` comment in `settings_customise_sources`. Happy to split it into its own issue if you would prefer them separate.

PR #1088, opened 2026-08-27
feat(docker): make API worker count configurable

### LINKED ISSUE #1063: Make API Uvicorn worker count configurable (API_WORKERS)
# **🚀 Enhancement Request**

## **Is your enhancement request related to a problem? Please describe.**

`docker/entrypoint.sh` starts the API with:

```sh
exec /app/.venv/bin/fastapi run --host 0.0.0.0 src/main.py
```

`fastapi run` defaults to a single Uvicorn worker, so every request is served from one process and one event loop. That matters more for Honcho than for a typical CRUD API because the Dialectic agent runs *inline on the request path* — a single `/chat` call can loop over several tool invocations, each doing an embedding call plus a vector search, before it returns.

Under concurrent chat load, requests queue behind each other and per-request latency degrades sharply. On a self-hosted single-container deployment I measured seven overlapping `/v3/.../chat` requests landing in a ~3 minute window:

```
tool_calls=0  total_duration=107339ms
tool_calls=0  total_duration=128959ms
tool_calls=0  total_duration=118636ms
tool_calls=0  total_duration=97703ms
tool_calls=4  total_duration=155055ms
tool_calls=5  total_duration=131799ms
tool_calls=7  total_duration=203717ms
```

Notably these were **not** upstream problems: no 429s, no retries, no fallback switches, and the same model answered comparable single requests in a few seconds. The time was spent waiting on the worker, not on the provider.

Setting `--workers 3` on the same host, with no other change, brought comparable requests back to 2–10s.

---

## **Describe the solution you'd like**

Make the worker count configurable, defaulting to the current behaviour:

```sh
exec /app/.venv/bin/fastapi run --host 0.0.0.0 --workers "${API_WORKERS:-1}" src/main.py
```

Defaulting to `1` keeps existing deployments byte-for-byte identical, so this is purely opt-in. Operators who need concurrency set `API_WORKERS` in their environment.

A short note in the deployment docs on sizing would help, since the ceiling isn't obvious: each worker gets its own SQLAlchemy pool, so the practical bound is

```
API_WORKERS × (DB_POOL_SIZE + DB_MAX_OVERFLOW) < postgres max_connections
```

With the defaults (10 + 20) that's 30 connections per worker, so 3 workers ≈ 90 — comfortable against a default `max_connections=200`, but easy to overshoot if someone sets `API_WORKERS=8` without also raising Postgres limits.

---

## **Describe alternatives you've considered**

- **Horizontal scaling** (compose `deploy.replicas`, or multiple pods) — the right answer at scale and already possible, but heavier than necessary for single-host or self-hosted deployments, and it doesn't help someone running the published image directly.
- **Raising client-side timeouts** — treats the symptom. The requests do eventually succeed; they're just slow because they're queued.
- **Leaving it to operators to override the entrypoint** — possible, but it means forking the image or the compose file to change one flag, and the current default is easy to hit without realising a single worker is the constraint.

---

### **Additional context**

Happy to open a PR — it's a one-line change to `docker/entrypoint.sh` plus a docs note, and I've been running the multi-worker configuration in production for a couple of weeks without issue.

One thing I checked while running it: with `LLM_OPENAI_AUTH_MODE` variants that hold refreshable credentials, each worker initialises its own token manager, so any shared-credential refresh path needs to tolerate N processes refreshing concurrently. That's orthogonal to this change and not a problem for API-key auth, but worth being aware of if worker counts go up.

steven-ji:
## Description

Make the Docker API worker count configurable through `API_WORKERS` while preserving the existing single-worker default. Document database connection sizing for pooled configurations and clarify that `DB_POOL_SIZE` and `DB_MAX_OVERFLOW` do not apply when `DB_POOL_CLASS=null`.

## Proofs

- A temporary shell-boundary harness executed `docker/entrypoint.sh` with stub executables: the default emitted `--workers 1`, and `API_WORKERS=4` emitted `--workers 4`
- `sh -n docker/entrypoint.sh` and `shellcheck docker/entrypoint.sh`: passed
- `uv run pre-commit run --files .env.template docker/entrypoint.sh docs/v3/contributing/self-hosting.mdx`: passed
- Initial full-suite validation before the review-requested removal of committed test files: `uv run pytest -x tests/ --ignore=tests/alembic/` — 2266 passed, 51 skipped
- `git diff --check`: passed

## Checklist

Fixes #1063

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring the number of API server worker processes through the `API_WORKERS` environment setting.
  - The default remains a single worker when no value is provided.

- **Documentation**
  - Clarified how database connection pools operate across API worker processes.
  - Documented that disabling pooling opens and closes database connections for each use, and pool-size settings no longer apply.


akattelu (approved):

akattelu (changes requested):
Actually on second thought, can you remove the conftest and test_docker_entrypoint test? I don't think they add much and I don't want to add more test-time dependencies.

steven-ji:
Addressed in `9ab4067`. I removed the `tests/conftest.py` change and `tests/test_docker_entrypoint.py` as requested. I also clarified the pooled connection guidance and `NullPool` behavior. The shell-boundary checks, ShellCheck, `sh -n`, and focused pre-commit checks pass.

akattelu (approved):

PR #1089, opened 2026-08-27
feat(deriver): Moves scheduling events from deriver to API process

ulyssepence:
# Description

Primarily, this moves logic from the deriver that assumes it will be a long-running process into the DB/API endpoint. In this way, the deriver can be run on-demand, if that's desired.

More specifically, in this PR, we...
- Add metrics in the API endpoint _about_ the deriver's backlog (`BacklogMetricsPoller`). They are populated by polling against the same data the deriver does when looking for work to claim.
- Move the reconciler scheduler (`ReconcilerScheduler`) and dream polling loop from the deriver to the API endpoint
- Make the deriver check if it's time to run _anything_ at boot

# Testing

```bash
uv run pytest tests/ -q
```

## E2E

Stood up the E2E instance then verified:
- Representation rows enqueued by the API and processed; explicit documents written.
- Summary rows enqueued and processed.
- Vector-sync row enqueued by the API, processed by the deriver, seconds_since_last_vector_sync moved off -1.
- Queue-cleanup row enqueued by the API at the configured interval; every processed row deleted, documents kept.
- Stale claim (30 minutes old) reaped by the API, with no deriver running.
- Dream enqueued by the API, run by the deriver, collection metadata advanced.
- Dream row marked failed while the deriver was stopped: no re-enqueue for 25 seconds, dream fields untouched.
- Same dream re-enqueued after one new message, completed cleanly, metadata advanced.
- All six gauges served real values on `/metrics`

## Summary by CodeRabbit

- **New Features**
  - Added service metrics for queue backlog, pending items, processing age, vector synchronization, and pending dreams.
  - Added automatic detection and enqueueing of dreams when collections meet configured thresholds or idle-time conditions.
  - Added periodic cleanup of stale processing items.
- **Improvements**
  - Dream scheduling is now handled through a streamlined, API-driven process.
  - Added configurable polling and cleanup intervals.
- **Tests**
  - Expanded coverage for backlog reporting, dream eligibility, queue processing, and metric initialization.

### DIFF
--- a/src/backlog.py
+++ b/src/backlog.py
@@ -1,0 +1,210 @@
+"""Publishes the deriver backlog as Prometheus gauges from the API process."""
+
+import asyncio
+import contextlib
+import logging
+from datetime import datetime, timedelta, timezone
+from typing import Any, cast
+
+import sentry_sdk
+from sqlalchemy import func, select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from src import crud, models
+from src.config import settings
+from src.dependencies import tracked_db
+from src.telemetry.prometheus import metrics as prometheus_metrics
+from src.utils.work_unit import construct_work_unit_key
+
+logger = logging.getLogger(__name__)
+
+DREAM_POLL_INTERVAL_SECONDS = 300
+
+
+class BacklogMetricsPoller:
+    """Refreshes the deriver-backlog gauges on a timer."""
+
+    def __init__(self) -> None:
+        self._task: asyncio.Task[None] | None = None
+        self._shutdown_event: asyncio.Event = asyncio.Event()
+        self._next_dream_poll: datetime | None = None
+
+    async def start(self) -> None:
+        if self._task is not None:
+            logger.warning("BacklogMetricsPoller already running")
+            return
+
+        interval = settings.DERIVER.BACKLOG_METRICS_POLL_INTERVAL_SECONDS
+        self._task = asyncio.create_task(self._loop())
+        logger.info("BacklogMetricsPoller started (interval=%ss)", interval)
+
+    async def shutdown(self) -> None:
+        if self._task is None:
+            return
+
+        logger.info("Shutting down BacklogMetricsPoller...")
+        self._shutdown_event.set()
+
+        try:
+            await asyncio.wait_for(self._task, timeout=5.0)
+        except asyncio.TimeoutError:
+            logger.warning("BacklogMetricsPoller shutdown timed out, cancelling task")
+            self._task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await self._task
+
+        self._task = None
+        logger.info("BacklogMetricsPoller stopped")
+
+    async def _loop(self) -> None:
+        interval = settings.DERIVER.BACKLOG_METRICS_POLL_INTERVAL_SECONDS
+
+        while not self._shutdown_event.is_set():
+            try:
+                await self._refresh()
+            except Exception as e:
+                logger.error("BacklogMetricsPoller refresh failed: %s", e)
+                if settings.SENTRY.ENABLED:
+                    sentry_sdk.capture_exception(e)
+
+            try:
+                await asyncio.wait_for(
+                    self._shutdown_event.wait(), timeout=float(interval)
+                )
+            except asyncio.TimeoutError:
+                continue
+
+    async def _refresh(self) -> None:
+        async with tracked_db("backlog_metrics", read_only=True) as db:
+            backlog = await crud.get_deriver_backlog(db)
+            prometheus_metrics.prometheus_metrics.set_deriver_backlog(
+                eligible_work_units=backlog.eligible_work_units,
+                pending_items=backlog.pending_items,
+                oldest_pending_age_seconds=backlog.oldest_pending_age_seconds,
+            )
+
+            prometheus_metrics.prometheus_metrics.set_seconds_since_last_vector_sync(
+                seconds=await _seconds_since_last_vector_sync(db)
+            )
+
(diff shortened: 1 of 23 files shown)


src/schemas/internal.py
```diff
@@ -144,6 +144,14 @@ class QueueCounts(BaseModel):
     sessions: dict[str, SessionCounts]
 
 
+class DeriverBacklog(BaseModel):
```
eisene, line 147: nit: `DeriverBacklogSummary` or something - this is not literally the deriver backlog

src/reconciler/scheduler.py
```diff
@@ -166,6 +166,20 @@ async def _scheduler_loop(self) -> None:
                 # endregion
                 await record_pending_embeddings_backlog()
 
+                if (
+                    self._next_stale_cleanup is not None
+                    and now >= self._next_stale_cleanup
+                ):
+                    try:
+                        await crud.cleanup_stale_work_units()
```
eisene, line 174: i think there are some stale comments about this being jittered

src/reconciler/scheduler.py
```diff
@@ -166,6 +166,20 @@ async def _scheduler_loop(self) -> None:
                 # endregion
                 await record_pending_embeddings_backlog()
 
+                if (
+                    self._next_stale_cleanup is not None
```
eisene, line 170: claude claims there are no tests for the scheduler side of the cleanup - that it fires when expected and fires on an interval

src/telemetry/prometheus/metrics.py
```diff
@@ -507,6 +548,14 @@ def initialize_bounded_metrics(self, *, instance_type: str) -> None:
             # ai: embed_now fast path runs as an API-process background task
             self._touch(embed_now_tasks_shed_counter)
             self.set_embed_now_tasks_in_flight(0)
+            self.set_message_embeddings_pending(count=0)
+            self.set_deriver_backlog(
```
eisene, line 552: nit: can this be constructed from `DeriverBacklog` defaults so it's all in one logical place?

src/backlog.py
```diff
@@ -0,0 +1,285 @@
...
+@dataclass
+class DueDream:
+    workspace_name: str
+    observer: str
+    observed: str
+    dream_type: DreamType
+    documents_since_last_dream: int
+
+
+class BacklogMetricsPoller:
```
eisene, line 34: proposal to refactor this to clean it up and clean up a bunch of bugs claude says are in this: [https://claude.ai/code/artifact/f3eaf4c4-70b7-4a62-9f4d-0506a751dccc](<https://claude.ai/code/artifact/f3eaf4c4-70b7-4a62-9f4d-0506a751dccc>) shouldn't be a big lift

src/reconciler/scheduler.py
eisene: claude:

Nothing in tests/ references \_next_stale_cleanup, and tests/reconciler/ has no timing test for it. The only remaining coverage is tests/crud/test_deriver_backlog.py:226, which calls crud.cleanup_stale_work_units() directly — it proves the SQL works, not that the scheduler ever fires it or that a failure doesn't hot-loop.

src/backlog.py
philosaether: This is exactly the kind of thing a shared ai-facing doc warehouse would love to house, jus' sayin'. Attach a reference to a low priority Linear ticket and pull the deets only when you need them... 🤌🏻

src/backlog.py
```diff
@@ -0,0 +1,285 @@
...
+            return False
+        self._next_dream_poll = now + timedelta(
+            seconds=settings.DREAM.POLL_INTERVAL_SECONDS
+        )
+        return True
+
+
+async def _seconds_since_last_vector_sync(db: AsyncSession) -> float:
+    """-1 when nothing has ever synced"""
+    newest = await db.scalar(select(func.max(models.MessageEmbedding.last_sync_at)))
```
philosaether, line 145: Whoa, Claude caught an actual correctness bug here.

last_sync_at is set by embed_now._claim_and_lease at the *start* of the sync attempt, not when it returns successful. So this returns the latest attempted sync, not the latest successful sync.

The fix is to add a `WHERE sync_state='synced'` selector or w/e the SqlAlchemy equivalent is, which also buys us a performance increase via the index on synced rows. Per Claude:

```
Crux — src/backlog.py:145: SELECT max(last_sync_at) FROM message_embeddings with no WHERE, 
every 30s, per API pod. It measured this on a real AlloyDB Omni 15.17 container (200k rows, 51 MB heap):

┌─────────────────────────────┬──────────────────────────────────────────┬─────────┬──────────┐
│                             │                   plan                   │ buffers │   time   │
├─────────────────────────────┼──────────────────────────────────────────┼─────────┼──────────┤
│ as written                  │ Parallel Seq Scan, 2 workers             │ 6,469   │ 31–38 ms │
├─────────────────────────────┼──────────────────────────────────────────┼─────────┼──────────┤
│ + WHERE sync_state='synced' │ Index Only Scan Backward, existing index │ 4       │ 0.11 ms  │
└─────────────────────────────┴──────────────────────────────────────────┴─────────┴──────────┘

~300× cheaper, using an index that already exists.

And the same one-line change fixes a correctness bug, which is what makes it worth blocking on: 
last_sync_at is an attempt timestamp, not a success timestamp — embed_now.py:203-206 stamps it on 
rows still pending. So the gauge reads healthy in exactly the outage it's meant to detect, and it's the 
scale-to-zero alarm. Cost and semantics, one predicate.
```

philosaether: I'm not blocking on it though, approve with comments

philosaether (approved):
Approved with comments

PR #1092, opened 2026-08-28
fix(peers): cancel non-streaming dialectic when the client disconnects

### LINKED ISSUE #1050: [Bug] Non-streaming dialectic chat continues LLM work after client disconnect
# **🐞 Bug Report**

## **Describe the bug**

A non-streaming dialectic `/chat` request can continue running after the HTTP client has disconnected.

In our case, Hermes calls Honcho and abandons the request after a 30-second client timeout. Uvicorn correctly detects the connection loss, but the Honcho non-streaming `/chat` handler continues awaiting `agentic_chat()` because it does not concurrently listen for the ASGI `http.disconnect` event.

As a result, abandoned downstream LLM requests can remain active for many minutes.

Before applying a local fix, we repeatedly observed Ollama requests surviving for approximately:

```text
9m59.9s
```

after the original Hermes request had already timed out.

Because the OpenAI SDK also has retry behaviour and Honcho has an additional outer retry layer, repeated abandoned calls can overlap and create a significant backlog.

We traced the behaviour through Honcho, Starlette/Uvicorn, the OpenAI-compatible client and Ollama, implemented a local fix, added a regression test, and verified cancellation propagates successfully all the way to Ollama.

---

### **Is this a regression?**

Unknown.

We have not established a previous Honcho version in which non-streaming dialectic requests were cancelled automatically when the HTTP client disconnected.

The upstream Honcho commit on which we reproduced the issue was:

```text
d191c107
```

---

### **To Reproduce**

1. Run Honcho with a local OpenAI-compatible LLM endpoint where a dialectic request takes longer than the HTTP client's timeout.

2. Send a non-streaming dialectic request to:

```text
POST /v3/workspaces/<workspace>/peers/<peer>/chat
```

For example:

```bash
curl --max-time 5 -sS \
  -X POST \
  http://127.0.0.1:8000/v3/workspaces/hermes/peers/conductor/chat \
  -H 'Content-Type: application/json' \
  -d '{
    "query": "Briefly summarise what you know about this peer.",
    "stream": false
  }'
```

3. Allow the HTTP client to time out/disconnect while the dialectic request is still running.

The client reports, for example:

```text
curl: (28) Operation timed out after 5003 milliseconds with 0 bytes received
```

4. Observe the downstream LLM request.

Without disconnect handling in the non-streaming Honcho route, the underlying `agentic_chat()` / provider request can continue after the client has disappeared.

In our production case, abandoned Ollama requests were observed running for approximately:

```text
9m59.9s
```

before eventual termination.

The underlying reason appears to be that Uvicorn records the disconnect and exposes:

```text
{"type": "http.disconnect"}
```

through the ASGI receive channel, but it does not automatically cancel the running ASGI handler.

The Honcho non-streaming route directly awaits:

```python
response = await agentic_chat(...)
```

and does not concurrently consume the disconnect event while the dialectic is running.

---

### **Expected behaviour**

When the client disconnects from a non-streaming dialectic `/chat` request, Honcho should promptly cancel the associated in-flight dialectic and downstream LLM request.

Expected flow:

```text
client disconnect
    -> ASGI http.disconnect
    -> Honcho detects disconnect
    -> agentic_chat() cancelled
    -> provider request cancelled
    -> downstream model stops processing
```

A disconnected client should not leave an expensive LLM operation running toward the provider's much longer timeout.

---

### **Media prove**

No screenshot is required to reproduce the issue, but we have live log evidence from both Honcho and Ollama.

After implementing a local disconnect-aware fix, a controlled five-second client timeout produced:

Honcho:

```text
Client disconnected during dialectic chat; cancelling work (workspace=hermes peer=conductor)
```

Ollama:

```text
14:31:19.204  /v1/completions ended after 4.824s
14:31:20.256  Request terminated error="context canceled"
```

This demonstrates that cancellation can propagate correctly through the complete stack.

Before the fix, equivalent abandoned Ollama requests were surviving for roughly ten minutes.

---

### **Your environment**

* OS: macOS host running Honcho self-hosted in Docker
* Honcho Server Version: upstream commit `d191c107`
* LLM backend: Ollama via an OpenAI-compatible endpoint
* FastAPI: `0.136.1`
* Starlette: `1.0.0`
* Uvicorn: `0.46.0`
* OpenAI SDK: `2.36.0`

Relevant OpenAI SDK defaults observed in the running Honcho API container:

```text
DEFAULT_MAX_RETRIES = 2
DEFAULT_TIMEOUT = Timeout(connect=5.0, read=600, write=600, pool=600)
```

The caller in this case was Hermes, whose Honcho client timed out after:

```text
30.0 seconds
```

---

### **Additional context**

We investigated this because normal Honcho API health checks continued to succeed while dialectic requests became progressively delayed.

The problem appears to be specifically related to the **non-streaming** dialectic path.

Uvicorn marks a request cycle as disconnected when the client connection is lost, but does not automatically cancel the ASGI request coroutine. The application can subsequently receive:

```text
{"type": "http.disconnect"}
```

through the ASGI receive channel.

Honcho's non-streaming `/chat` route did not consume that event while awaiting `agentic_chat()`.

We implemented a local proof-of-fix by running two concurrent tasks:

```text
1. agentic_chat()
2. wait for request.receive() -> http.disconnect
```

and using:

```python
asyncio.wait(
    {chat_task, disconnect_task},
    return_when=asyncio.FIRST_COMPLETED,
)
```

If the client disconnects first, the dialectic task is cancelled and awaited.

The streaming path was left unchanged.

A regression test was also added which:

1. starts a deliberately blocking mocked `agentic_chat()`;
2. injects:

```text
{"type": "http.disconnect"}
```

3. verifies the dialectic receives `asyncio.CancelledError`;
4. verifies the route exits with status 499.

Regression test result:

```text
1 passed
```

Combined with the existing chat route tests:

```text
10 passed, 2 warnings
```

The two warnings were unrelated pre-existing JWT test-key-length warnings.

We also verified normal behaviour after the change: a normal non-streaming dialectic request completed successfully in approximately:

```text
23.7 seconds
```

Our local implementation is currently split into two commits:

```text
0bab11ef  fix: cancel dialectic work on client disconnect
d141095a  test: cover dialectic cancellation on disconnect
```

These have not been pushed upstream.

The exact upstream implementation does not need to match our local patch. The important requirement is that a disconnected non-streaming client should promptly cancel its associated dialectic/LLM work.

Happy to provide a pull request with the tested implementation if useful.

serhiizghama:
Fixes #1050. A non-streaming dialectic `/chat` request keeps running after the caller has gone away. Uvicorn notices the disconnect and surfaces it on the ASGI receive channel, but it doesn't cancel the handler coroutine, so `await agentic_chat(...)` sails on toward the provider's much longer timeout — the reporter saw abandoned Ollama calls survive ~10 minutes past a 30s client timeout, and with the SDK's retries plus Honcho's outer retry layer those pile up.

The streaming path already dies when its generator is dropped; only the non-streaming branch was affected. I race the dialectic against a small watcher that drains `request.receive()` until it sees `http.disconnect`. Whichever finishes first wins: if the work completes, the watcher is cancelled and the response is returned as before; if the client leaves first, the dialectic task is cancelled (and awaited, so cancellation propagates down to the LLM call) and the route returns 499.

The watcher ignores non-disconnect frames, so a stray `http.request` can't be mistaken for a hangup. Added unit tests over the race helper covering all three cases — no DB or LLM needed. Full route suite (`tests/routes`, `tests/dialectic`, the chat/allowlist tests) stays green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stopped processing non-streaming chat requests when the client disconnects.
  * Cancelled in-progress chat generation after disconnection and returned an appropriate HTTP 499 response.
  * Continued to ignore unrelated request messages while monitoring connection status.

* **Tests**
  * Added coverage for disconnect handling, completed requests, task cleanup, cancellation, and non-disconnect messages.

### DIFF
--- a/src/routers/peers.py
+++ b/src/routers/peers.py
@@ -1,12 +1,13 @@
 """FastAPI routes for peer resources and peer-scoped operations."""
 
+import asyncio
 import json
 import logging
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Awaitable
 from contextlib import suppress
 from time import perf_counter
 
-from fastapi import APIRouter, Body, Depends, Path, Query, Response
+from fastapi import APIRouter, Body, Depends, Path, Query, Request, Response
 from fastapi.responses import StreamingResponse
 from fastapi_pagination import Page
 from fastapi_pagination.ext.sqlalchemy import apaginate
@@ -45,6 +46,44 @@ router = APIRouter(
     prefix="/workspaces/{workspace_id}/peers",
     tags=["peers"],
 )
+
+
+class _ClientDisconnected(Exception):
+    """Raised when the client goes away before an awaited operation finishes."""
+
+
+async def _wait_for_disconnect(request: Request) -> None:
+    """Block until the client disconnects, draining the ASGI receive channel."""
+    while True:
+        if (await request.receive())["type"] == "http.disconnect":
+            return
+
+
+async def _run_until_disconnect[T](request: Request, coro: Awaitable[T]) -> T:
+    """Await ``coro``, cancelling it if the client disconnects first.
+
+    Uvicorn surfaces a client disconnect on the ASGI receive channel but does
+    not cancel the running handler, so a non-streaming dialectic would keep the
+    (expensive) LLM call going long after the caller has timed out (#1050).
+    Raises ``_ClientDisconnected`` if the client leaves before ``coro`` returns.
+    """
+    work = asyncio.ensure_future(coro)
+    watcher = asyncio.ensure_future(_wait_for_disconnect(request))
+    try:
+        await asyncio.wait({work, watcher}, return_when=asyncio.FIRST_COMPLETED)
+    finally:
+        if not watcher.done():
+            watcher.cancel()
+            with suppress(asyncio.CancelledError):
+                await watcher
+
+    if work.done():
+        return work.result()
+
+    work.cancel()
+    with suppress(asyncio.CancelledError):
+        await work
+    raise _ClientDisconnected
 
 
 async def _resolve_scope_option(
@@ -240,6 +279,7 @@ async def get_sessions_for_peer(
     },
 )
 async def chat(
+    request: Request,
     workspace_id: str = Path(...),
     peer_id: str = Path(...),
     options: schemas.DialecticOptions = Body(...),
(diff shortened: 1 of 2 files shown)


serhiizghama:
Good catch on the cancellation leak — that was a real hole in the original patch. If the request task itself got torn down (shutdown, timeout middleware) while `asyncio.wait` was pending, only the watcher was cancelled and the LLM call kept running, which is exactly what this PR is supposed to prevent. Restructured so both tasks are cleaned up in a single `finally`, and added a test that cancels the wrapper mid-flight (it fails without the change). Also silenced the two `reportPrivateUsage` warnings with the same narrow ignore the rest of the test suite uses, so Static Analysis is green again.

On the gate: #1050 was filed by another user and isn't labelled `maintainer-approved` yet. Happy to leave this open if someone wants to look at it, or I can close and re-open once the issue is triaged — whichever is less noise for you.

PR #1093, opened 2026-08-28
feat(messages): add DELETE endpoint for a single message

### LINKED ISSUE #956: No message-level DELETE — removing a single message requires deleting the whole session
`src/routers/messages.py` exposes POST, GET and PUT but **no DELETE**. Every other resource router has one — `conclusions.py` (`DELETE /{conclusion_id}`), `sessions.py` (2), `workspaces.py`, `webhooks.py`. Messages appear to be the gap rather than a deliberate exclusion, though happy to be told otherwise.

### Why it matters

The only way to remove a single stored message today is to delete its entire session. That is a sledgehammer whenever the thing you need gone is one message among thousands of legitimate ones.

**The case that prompted this:** a user's automated pipeline (Claude Agent SDK spawning Claude Code programmatically) ran with the `claude-honcho` plugin loaded, so ~228 machine-generated agent prompts were stored attributed to the human user across a four-day window. The source was fixed, and the residue turned out to be genuinely inert — the deriver formed no conclusions from the repeated boilerplate, and exact-content dedup (#861) keeps it out of retrieval — so in that instance leaving it was the right call.

But the *reason* it was the right call was that removal wasn't possible at proportionate cost. That reasoning does not hold for the case I'd expect to matter more:

### The case where this becomes urgent

A **secret captured into a message**. `claude-honcho` stored tool-call summaries unredacted until very recently (plastic-labs/claude-honcho#81, fixed in plastic-labs/claude-honcho#99) — so a command like `PGPASSWORD=… psql …` could be sitting in a session as durable memory. Redacting that today means destroying the whole session, which for an active user means losing their real conversation history to remove one line. In practice people will leave the secret in place, which is the wrong outcome.

Related: GDPR/right-to-erasure style requests are naturally message-scoped, not session-scoped.

### Suggestion

`DELETE /v3/workspaces/{workspace_id}/sessions/{session_id}/messages/{message_id}`, mirroring the shape and semantics of the existing conclusion delete (204, "cannot be undone").

Worth deciding explicitly what happens to anything already derived from that message — conclusions/documents referencing it presumably shouldn't silently outlive it, but cascading may not be desirable either. I'd defer to your model here; even a delete that leaves derived state untouched and says so plainly would be a large improvement over the current all-or-nothing.

Happy to open a PR if you'd like it and can confirm the intended derived-state behaviour.

serhiizghama:
Closes #956. Messages were the one resource router without a DELETE — `conclusions`, `sessions`, `workspaces` and `webhooks` all have one, so removing a single stored message meant deleting its whole session. That's the wrong tool when the thing you need gone is one line among thousands (a secret captured into a message, an erasure request), and the issue is marked maintainer-approved.

Added `DELETE /v3/workspaces/{workspace_id}/sessions/{session_id}/messages/{message_id}`, mirroring the shape and semantics of the conclusion delete — 204, session-write scope, "cannot be undone".

The crud path is FK-order-aware, matching what `delete_session` already does: queue items reference `messages.id` with no `ON DELETE` rule, so they're cleared before the message or the delete hits a foreign-key violation. Message embeddings cascade with the message (`ondelete="CASCADE"`), and if an external vector store is configured its vectors are dropped first (best effort), same as session delete.

On the open question from the issue about derived state: conclusions already derived from the message are left in place. A conclusion can draw on several source messages, so cascading here would silently prune a peer's representation — I went with the conservative "leave it and say so" behaviour the issue floated, documented in the endpoint. Happy to switch to cascading if you'd prefer.

Tests cover the happy path (message gone, sibling untouched), 404 for a missing id, and that embeddings and queue items are actually cleared — the last one fails without the queue cleanup, which was the FK trap.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to permanently delete individual messages from a session.
  * Related embeddings and pending processing items are removed automatically, while derived conclusions are preserved.
  * Deletion is restricted to authorized session writes and returns a successful no-content response.

* **Bug Fixes**
  * Requests to delete nonexistent messages now return a not-found response.

### DIFF
--- a/src/crud/__init__.py
+++ b/src/crud/__init__.py
@@ -23,6 +23,7 @@ from .document import (
 )
 from .message import (
     create_messages,
+    delete_message,
     get_message,
     get_message_seq_in_session,
     get_messages,
@@ -135,6 +136,7 @@ __all__ = [
     "search_messages",
     "search_messages_temporal",
     "update_message",
+    "delete_message",
     # Peer
     "get_or_create_peers",
     "get_peer",
--- a/src/crud/message.py
+++ b/src/crud/message.py
@@ -4,7 +4,7 @@ from logging import getLogger
 from typing import Any
 
 from nanoid import generate as generate_nanoid
-from sqlalchemy import ColumnElement, Select, and_, func, or_, select, text
+from sqlalchemy import ColumnElement, Select, and_, delete, func, or_, select, text
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import InstrumentedAttribute
 
@@ -12,6 +12,7 @@ from src import models, schemas
 from src.config import settings
 from src.dependencies import tracked_db
 from src.embedding_client import embedding_client
+from src.exceptions import ResourceNotFoundException
 from src.telemetry.events import EmbeddingCallPurpose
 from src.utils.filter import apply_filter
 from src.utils.formatting import ILIKE_ESCAPE_CHAR, escape_ilike_pattern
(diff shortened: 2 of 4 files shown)


akattelu:
Hey @serhiizghama, on second thought I took off the maintainer approved label from the issue that this references (see https://github.com/plastic-labs/honcho/issues/956#issuecomment-5454739890). Closing this because we're not currently accepting contributor pull requests for this specific issue because of how complex it is. A maintainer will work on this in the future.

PR #1095, opened 2026-08-28
fix(deriver): strip NUL bytes from model-generated observations

eisene:
# Human summary

This is just to sanitize rare LLM outputs that contains bytes that can't be saved to the DB.

# Claude summary

## Description

The deriver crashes with `DataError: PostgreSQL text fields cannot contain NUL (0x00) bytes` whenever a model emits a `\u0000` escape in its tool-call arguments — the JSON parser decodes it into a real NUL byte, and Postgres rejects NUL in `text` columns and in `jsonb` strings. API ingress has always stripped NUL from user-supplied content (`MessageCreate.sanitize_content`), but the deriver's *own output* had no equivalent, so model-generated observations went straight to the database untreated.

This strips NUL in `_normalized_observation` and `_normalized_observation_input` — the points that already normalize observation text for persistence *and* embedding — so the text that gets embedded is the text that gets stored. `DocumentCreate.content` gets a backstop validator for callers that bypass those paths.

Fixes [HONCHO-4XZ](https://plastic-labs.sentry.io/issues/HONCHO-4XZ) (Rootly alert `RNne7G`).

### Impact before the fix

The NUL reached the exact-content dedup pre-fetch in `create_documents` (`src/crud/document.py:539`) as a bind parameter, so the query blew up before any row was written. The exception is caught per-observer at `deriver.py:245`, so it never crash-looped — it just silently discarded the whole observation batch for that observer. 8 occurrences over ~10h across two workspaces, both from models transcribing text that legitimately contains NUL notation:

- `... | tr '\x00' '\n' | grep home` — a shell command
- `c:\<NUL>users\amal/.ssh/id_ed25519` — a Windows path

### Notes for review

- **`premises` and `sources` are covered too.** They ride along in `internal_metadata`, and `jsonb` rejects NUL in strings just as text columns do.
- **The emptiness check moved after normalization.** `str.strip()` does not remove NUL, so `"\x00"` passed the old blank filter; once stripping was added it would have been embedded and stored as an empty document.
- **`DocumentCreate.content` uses `mode="before"` deliberately.** It runs ahead of the `min_length=1` constraint, so all-NUL content fails validation instead of quietly becoming `""`.
- **The NUL helpers moved** from `src/schemas/api.py` into a new `src/utils/sanitization.py` as a single recursive `strip_nul`, so ingress and internal paths share one implementation. It is overloaded to keep `str -> str` for callers that chain `.strip()`, and passes `None` through so optional fields need no guard.
- **Unrelated churn:** `tests/crud/test_representation_manager.py` picked up 14 mechanical `timezone.utc` → `UTC` lines. The pre-commit ruff hook (`UP017`) rewrites them whenever that file is staged; 40 files on `main` still use the old spelling.

Not in scope: session summaries (`src/utils/summarizer.py`) and peer cards have the same model-text-into-TEXT-column exposure and are untouched here.

## Proofs

Full suite against `pgvector/pgvector:pg15` matching CI's service config:

```
2281 passed, 51 skipped, 163 warnings in 36.56s
```

`basedpyright` and `ruff check src/` both at their pre-existing baselines (22 / 44), and all pre-commit + pre-push hooks pass.

New coverage:

- `tests/crud/test_representation_manager.py` — NUL stripped from explicit content, deductive conclusions, and `premises`; embedded text asserted to match stored text; all-NUL observations dropped before embedding rather than stored empty.
- `tests/test_schema_validations.py` — `DocumentCreate` backstop, all-NUL content rejected rather than emptied, message content and nested metadata.
- `tests/utils/test_sanitization.py` — direct coverage of `strip_nul` (string / list / nested dict incl. NUL in a key / `None` / non-string passthrough / non-mutation). These helpers previously had none.

## Checklist

## Bug Fixes
- Removed NUL characters from document, message, metadata, query, and observation content before storage and embedding.
- Excluded observations that become empty after sanitization.
- Prevented content containing only NUL characters from being saved as empty text.
- Applied sanitization consistently to nested metadata, premises, sources, and list values.
- Improved validation for fields containing only NUL characters.

## Tests
- Added coverage for NUL-byte removal, nested data handling, empty-content validation, query validation, and input preservation.

### DIFF
--- a/tests/crud/test_representation_manager.py
+++ b/tests/crud/test_representation_manager.py
@@ -1,5 +1,5 @@
 from contextlib import asynccontextmanager
-from datetime import datetime, timedelta, timezone
+from datetime import UTC, datetime, timedelta
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
@@ -196,7 +196,7 @@ class TestRepresentationManagerSoftDelete:
             db_session, test_workspace, test_peer
         )
 
-        base = datetime(2026, 1, 1, tzinfo=timezone.utc)
+        base = datetime(2026, 1, 1, tzinfo=UTC)
         # Three conclusions, all reinforced once, inserted oldest-first.
         for i in range(3):
             db_session.add(
@@ -484,13 +484,13 @@ class TestRepresentationManagerSave:
             explicit=[
                 ExplicitObservation(
                     content="   ",
-                    created_at=datetime.now(timezone.utc),
+                    created_at=datetime.now(UTC),
                     message_ids=[1],
                     session_name="session",
                 ),
                 ExplicitObservation(
                     content=" useful observation ",
-                    created_at=datetime.now(timezone.utc),
+                    created_at=datetime.now(UTC),
                     message_ids=[1],
                     session_name="session",
                 ),
@@ -515,7 +515,7 @@ class TestRepresentationManagerSave:
                 representation,
                 message_ids=[1],
                 session_name="session",
-                message_created_at=datetime.now(timezone.utc),
+                message_created_at=datetime.now(UTC),
                 message_level_configuration=_resolved_config(),
             )
 
@@ -540,7 +540,7 @@ class TestRepresentationManagerSave:
                     conclusion="   ",
                     premises=["premise a"],
                     source_ids=["doc-a"],
-                    created_at=datetime.now(timezone.utc),
+                    created_at=datetime.now(UTC),
                     message_ids=[1],
                     session_name="session",
                 ),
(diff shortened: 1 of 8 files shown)


tests/crud/test_representation_manager.py
```diff
@@ -196,7 +196,7 @@ async def test_query_documents_most_derived_ties_break_by_recency(
             db_session, test_workspace, test_peer
         )
 
-        base = datetime(2026, 1, 1, tzinfo=timezone.utc)
+        base = datetime(2026, 1, 1, tzinfo=UTC)
```
eisene, line 199: these are linter auto-fixes that happen on commit, i guess they weren't there when this file was first written

VVoruganti (approved):
Made some changes to increase usage of the strip_nul function and make it more broadly applicable.

PR #1100, opened 2026-08-29
perf(llm): run one turn's tool calls concurrently instead of serially

### LINKED ISSUE #1101: Tool calls in one assistant turn run serially, so a turn costs the sum of its reads
## Problem

`execute_tool_loop` awaits each tool call in an assistant turn one at a time:

https://github.com/plastic-labs/honcho/blob/main/src/llm/tool_loop.py#L586

The model routinely asks for several independent reads at once — a dialectic turn typically emits `search_memory` and `search_messages` together — so the turn costs the **sum** of their latencies when it only needs to cost the slowest.

Traced on a production `minimal` dialectic (9.1s end to end):

```
10:04:21.4  request in
10:04:24.8  [tool call]   search_memory      ← 3.4s  first LLM call
10:04:26.0  [tool result] search_memory      ← 1.2s  embeddings + pgvector
10:04:27.9  [tool result] search_messages    ← 1.9s  embeddings + pgvector
10:04:27.9  max iterations (1)
10:04:30.3  response                         ← 2.3s  final LLM call
```

The 1.2s and the 1.9s are independent reads that ran back to back. Of the 9.1s, ~5.7s is LLM time and ~3.1s is tool execution.

## Measured impact

Across **42 production dialectic requests** (103 iterations, 77 of them multi-tool, 3.34 tool calls per iteration on average), grouping calls by the turn that issued them:

| | saving |
|---|---|
| aggregate over 1254s measured | **11.3%** |
| median request | 9.3% |
| mean request | 13.7% |
| p90 | 29.7% |
| best case | 51.0% |

Only 4 of the 42 requests gain nothing.

## Why it is safe for reads

Tool handlers own their sessions: each opens a short-lived one via `tracked_db()`, and only the mutating handlers take `ctx.db_lock` — `_handle_create_observations_impl`, `_handle_update_peer_card`, `_handle_delete_observations`. Concurrent reads do not contend on shared state.

## The ordering constraint

Mutations cannot simply be parallelised alongside reads. `ctx.db_lock` makes the mutating handlers take turns but does not decide whose turn comes first, so two writes from one turn could apply in an order the model did not request. Any fix has to keep state-changing calls in request order while letting reads overlap.

## Proposed shape

- Run a turn's non-mutating calls concurrently; keep mutating calls serial in request order.
- Identify mutations by name from a set declared next to the dispatch table, rather than a per-caller "these are all reads" boolean — `DIALECTIC_TOOLS` already carries a commented-out `create_observations_deductive`, so a boolean would go quietly wrong the day someone uncomments it.
- Default to fully sequential so no existing caller changes behaviour without opting in.
- Cap the fan-out. Production has produced **18** tool calls in a single iteration; firing all at once would mean that many simultaneous embedding + pgvector queries on one instance.

Also worth noting: moving the per-call telemetry ContextVars (`set_current_tool_call_seq`, `set_last_tool_metadata`) inside each task makes them bind to their own call. Today they are written and read across one shared context, which only stays correct because nothing overlaps.

## Status

Implemented with tests in #1100, opened before I saw the issue-gate requirement. Happy to close that and reopen against this issue, or to change the approach if you would rather it looked different.

tomas-sereikis:
Fixes #1101

## Problem

A single assistant turn routinely requests several independent reads — a dialectic turn typically emits `search_memory` and `search_messages` together — but `execute_tool_loop` awaits them one at a time. The turn then costs the **sum** of their latencies when it only needs to cost the slowest.

Traced on a production `minimal` dialectic (9.1s end to end):

```
10:04:21.4  request in
10:04:24.8  [tool call]   search_memory      ← 3.4s  first LLM call
10:04:26.0  [tool result] search_memory      ← 1.2s  embeddings + pgvector
10:04:27.9  [tool result] search_messages    ← 1.9s  embeddings + pgvector
10:04:27.9  max iterations (1)
10:04:30.3  response                         ← 2.3s  final LLM call
```

The 1.2s and the 1.9s are independent reads that ran back to back.

## Measured impact

Across **42 production dialectic requests** (103 iterations, 77 of them multi-tool, 3.34 tool calls per iteration on average), grouping calls by the turn that issued them:

| | saving |
|---|---|
| aggregate over 1254s measured | **11.3%** |
| median request | 9.3% |
| mean request | 13.7% |
| p90 | 29.7% |
| best case | 51.0% |

Only 4 of 42 requests gain nothing.

## Why it is safe

Tool handlers own their sessions — each opens a short-lived one via `tracked_db()`, and only the mutating handlers (`create_observations`, `update_peer_card`, `delete_observations`) take `ctx.db_lock`. Concurrent reads do not contend on shared state.

The per-call telemetry ContextVars move **inside** the task. `asyncio` copies the context per task, so `set_current_tool_call_seq` and `set_last_tool_metadata` now bind to their own call rather than being written and read across one shared context — which the previous code could only keep straight by never overlapping. `gather` preserves argument order, so `tool_results` and `all_tool_calls` stay in the order the model asked for them.

## Bounded fan-out

Capped at `MAX_CONCURRENT_TOOL_CALLS = 4`. Production has produced **18** tool calls in a single iteration; firing all of them at once would mean that many simultaneous embedding + pgvector queries on one instance. The cap leaves the measured common case (3.34 per iteration) fully parallel while bounding the tail.

## Tests

`tests/llm/test_tool_loop_parallel.py` — three cases: calls in one turn overlap, results keep request order, and an 18-call turn stays within the cap. All fail against the serial implementation on the first assertion (`peak in-flight 1`).

`tests/llm/` is green: 303 passed. The one `ruff` finding on `tool_loop.py` (`UP047` on `_with_iteration_scope`) is pre-existing and unrelated — it reports identically with this change stashed.

## Summary by CodeRabbit

* **Performance Improvements**
  * Tool calls requested in the same assistant response now run concurrently.
  * Concurrent execution is limited to four tool calls at a time for improved stability.
  * Results continue to appear in the original request order.
  * State-changing actions preserve the order requested by the assistant.
  * Metadata and sequencing remain correctly associated with each tool call.
  * Tool failures continue to be reported as individual error results.


akattelu (changes requested):
Hey @tomas-sereikis, thanks for opening this. The measurement is real, and I think this is the right layer: a turn's independent reads should overlap in `execute_tool_loop`, not behind a new config knob.

One hole I would want closed before merge. Concurrent tasks are spawned, then ordered calls are awaited, then the rest. If the loop is cancelled during an ordered call (a dropped dialectic request), those siblings keep running embedding + pgvector after `execute_tool_loop` has already exited. Serial code never started the later calls. This path can leave up to four of them in flight. CodeRabbit flagged the same thing [here](https://github.com/plastic-labs/honcho/pull/1100#discussion_r3886365015).

A smaller test gap: `test_mutating_tools_are_not_reordered` names every call `update_peer_card`, so `order == ["update_peer_card"] * 3` still passes if they raced.

Suggested shape: wrap the concurrent set in `asyncio.TaskGroup` (we are on 3.13) so siblings cancel on exit, and record `id` (or distinct names) in that mutation-order test. Happy to look again after that.

PR #1103, opened 2026-08-29
fix(api): stop hardcoding the OpenAPI servers list

### LINKED ISSUE #875: [Bug] Self-hosted: OpenAPI servers list hardcodes http://localhost:8000, not configurable

chethanuk:
## Problem

`FastAPI(...)` hardcodes a two-entry `servers` list, so every deployment's `/openapi.json`
advertises `https://api.honcho.dev` first. On a self-hosted instance the "Try it out" button in
`/docs` sends the request to the hosted service, and any client generated from that spec points at
it too — for a self-hoster the local server is not in the list at all unless it happens to be on
`localhost:8000`.

## Change

Drop the `servers` kwarg. FastAPI then omits the `servers` key entirely, and Swagger UI falls back
to the origin `/docs` was loaded from — correct for localhost, a LAN host, a container, or a
hostname behind TLS, without configuration.

The one case that still needs configuration is a reverse proxy that strips a path prefix before
forwarding. `--root-path /api` covers it: FastAPI injects `{"url": "/api"}` into `servers`
per-request, which is why the tests exercise it through `TestClient(root_path=...)` rather than
`app.openapi()`. `docs/v3/contributing/self-hosting.mdx` now says so.

`docs/v3/openapi.json` is unchanged. It is a hand-maintained snapshot that `docs/docs.json` feeds
to Mintlify for the hosted docs, not generated from the app, so it should keep advertising the
production host.

This is the narrow fix asked for when #950 was closed.

## Testing

- `uv run pytest tests/test_openapi.py -q` — 11 passed
- Parametrized over six proxy topologies, asserting no absolute foreign origin appears in the
  served spec; plus the no-`root_path` case (no `servers` key) and the `root_path` case (exactly
  the prefix). Each fails if the hardcoded list comes back.
- `ruff check`, `ruff format --check`, `basedpyright` clean on the changed files

Closes #875

## Summary by CodeRabbit

* **Configuration**
  * Updated server routing to use the current application URL dynamically.
  * Removed explicit production and local development server URLs from the API configuration.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -161,10 +161,7 @@ async def lifespan(_: FastAPI):
 
 app = FastAPI(
     lifespan=lifespan,
-    servers=[
-        {"url": "https://api.honcho.dev", "description": "Production SaaS Platform"},
-        {"url": "http://localhost:8000", "description": "Local Development Server"},
-    ],
+    servers=[{"url": "/"}],
     title="Honcho API",
     summary="The Identity Layer for the Agentic World",
     description="""Honcho is a platform for giving agents user-centric memory and social cognition.""",


akattelu (changes requested):
Hey @chethanuk, thanks for opening this. Dropping the hardcoded pair is the right instinct, and I agree we should not add a config knob.

I think deleting the `servers` kwarg entirely is a bit further than we need, and the test module plus the self-hosting docs are ceremony around FastAPI's default. I'd rather keep an explicit relative server so the schema still has a `servers` entry, and leave

`--root-path` to FastAPI's own docs.

Suggested shape: `servers=[{"url": "/"}]` in `src/main.py`, drop `tests/test_openapi.py`, the conftest blocklist line, and the self-hosting.mdx hunk.

PR #1104, opened 2026-08-29
fix(deriver): reduce scope backfill memory usage

akattelu:
## Description

_run_backfill embedded, wrote, and synced every planned copy at once, holding one Python float list per document. A 14k-document session is ~580MB of vectors alone, and several backfills run concurrently, which OOM-killed the deriver at its 1000Mi limit and crash-looped it since the work units never completed. Phases 2-4 now run per chunk of 500 specs and drop each chunk's embeddings once synced.

## Proofs

N/A

## Checklist

## Summary by CodeRabbit

* **Performance**
  * Large scope backfills now process documents in bounded chunks to reduce memory usage.
  * Embeddings are released after each chunk is synchronized.
  * Backfill preparation loads only the data needed for each batch.

* **Reliability**
  * Progress and synchronization results are accurately aggregated across chunks.
  * Scope membership changes are checked safely during processing.
  * Sessions leaving a scope during processing are handled consistently.

### DIFF
--- a/src/deriver/scope_backfill.py
+++ b/src/deriver/scope_backfill.py
@@ -36,7 +36,6 @@ from sqlalchemy.sql.functions import func
 from src import crud, models
 from src.config import settings
 from src.crud.scope import ScopeBackfillState
-from src.crud.session import is_peer_in_session
 from src.dependencies import tracked_db
 from src.embedding_client import embedding_client
 from src.schemas import DreamType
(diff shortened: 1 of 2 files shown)


src/deriver/scope_backfill.py
```diff
@@ -235,6 +239,33 @@ async def _run_backfill(
...
         return 0, set()
 
+    # Phases 2-4 run per chunk so only one chunk's embeddings are alive at a
+    # time; each chunk's vectors are dropped once synced.
+    store_in_postgres = _store_embeddings_in_postgres()
+    touched_observed: set[str] = set()
+    copied = 0
+    for start in range(0, len(plans), BACKFILL_CHUNK_SIZE):
+        chunk = plans[start : start + BACKFILL_CHUNK_SIZE]
+        if not await _copy_chunk(
```
eisene, line 278: is it worth setting up a retry for this? or for `simple_batch_embed` inside `_copy_chunk`? in case there's some server error or smth

eisene (commented):
small comment

VVoruganti (approved):

PR #1110, opened 2026-08-31
Harness core

ajspig:
## Summary by CodeRabbit

- **New Features**
  - Added a reusable core package for loading, validating, and resolving configuration.
  - Supports environment variables, configuration overlays, legacy fields, built-in defaults, and host-specific settings.
  - Added telemetry header generation with runtime, host, plugin, and model details.
  - Added configuration schema validation and normalized service URLs.

- **Documentation**
  - Added setup guidance, configuration references, SDK initialization examples, and package changelog information.

- **Tests**
  - Added coverage for configuration resolution, URL handling, precedence rules, and telemetry headers.


harness-core/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@honcho-ai/harness-core",
```
akattelu: nit: maybe `@honcho-ai/harness-integration` instead?

harness-plugin-core/src/telemetry.ts
```diff
@@ -0,0 +1,55 @@
...
+  const ver = sanitize(id.hostVersion)
+  if (name && ver) return `${name}/${ver}`
+  return name || ver
+}
+
+/**
+ * Headers to pass as the SDK's `defaultHeaders`. Missing fields are omitted.
+ * `X-Honcho-Runtime` is always this package's version.
+ */
+export function telemetryHeaders(
```
akattelu, line 37: We should have a way for callers to easily update headers too. I think it could be useful for when the user switches models for example.

schemas/config/v1.json
```diff
@@ -0,0 +1,22 @@
...
+    "enabled": { "type": "boolean" },
+    "auth": {
+      "type": "object",
+      "properties": {
+        "apiKey": { "type": "string" },
+        "oauth": { "type": "object" }
+      }
+    },
+    "hosts": { "type": "object" }
+  }
```
akattelu: could be worth adding properties for the `oauth` and `hosts` objects here. In typescript, we're already expecting a specific type. It would also help users get better autocomplete for the configs in their editor.

akattelu (dismissed):

harness-core/package.json
ajspig: My only concern is "harness-integration" sounds like its an *integration* for *a* harness. I could see a user looking for openclaw/opencode plugin inside. Whereas “Core” implies a shared runtime for those harnesses.

akattelu: Good point, maybe `harness-integration-core` because its not our harness? or `harness-plugin-core` or `harness-plugin-shared`

VVoruganti: harness-plugin-core makes sense to me, especially if we are congealing around this idea of the agent plugin spec

harness-plugin-core/src/telemetry.ts
ajspig: currently `telemetryHeaders(id, extra)` can take `extra`. but for adding headers after construction, added a setTelemetryHeaders() method.

akattelu (approved):

PR #1118, opened 2026-09-02
feat: make tenant_id a first-class primitive in the data model

philosaether:
# Tenant ID is in Honcho tables now.

This PR adds a `tenants` table to honcho, and FKs it into all existing honcho tables as a prereq for further shared tenants work.

## Important Decisions:
- [Moved to the Linear Ticket](https://linear.app/plastic-labs/issue/DEV-2476/a1-make-tenant-a-first-class-primitive-in-honchos-data-model#comment-7df621fe)

## Unimportant Decisions:
- I decided to apply the current version of my ai-comment-pruning-and-categorizing filter. I'm not wedded to it; lmk what you think.

Claude's description follows.

---

## Description

Introduces `tenant_id` as a first-class primitive in the data model so many tenants can share one database with per-tenant isolation. Adds a `tenants` registry table and a `tenant_id` column on every tenant-scoped table, with composite `(tenant_id, id)` primary keys, composite foreign keys, `HASH(tenant_id)` partitioning, and `tenant_id`-leading indexes; the service tables (`queue`, `active_queue_sessions`) carry a plain `tenant_id` for attribution and stay unpartitioned.

The shared schema is created by a standalone bootstrap (`scripts/bootstrap_shared_schema.py`), not an Alembic migration: it builds a fresh shared database (generated from the declarative models, so it can't drift from them) rather than transforming an existing single-tenant one, and runs in AUTOCOMMIT so the partition DDL doesn't exhaust the lock table.

**Scope:** schema-only. `tenant_id` is defined but not yet threaded into any query path, RLS is not enabled, and no data is migrated — those land in follow-up PRs. Targets the `feat/shared-tenants` integration branch (not `main`).

## Proofs

- `uv run basedpyright src/models.py scripts/bootstrap_shared_schema.py` → **0 errors, 0 warnings, 0 notes**
- `uv run ruff check` + `uv run ruff format --check` → **clean**
- Models import + schema-shape audit: **12 tables**; all 9 tenant-scoped tables carry `tenant_id`, a `(tenant_id, …)` composite PK, `HASH(tenant_id)` partitioning, and a `tenant_id → tenants` FK; service tables correctly unpartitioned.
- **DDL validated against live PostgreSQL 17 + pgvector** (during review): the bootstrap builds every table + 128 partitions each; composite FKs — including `message_embeddings(tenant_id, message_id) → messages(tenant_id, public_id)` with `ON DELETE CASCADE` — plus `Identity()` on the composite PK, partition routing, and the per-partition GIN/HNSW indexes all enforce/build correctly.


scripts/bootstrap_shared_schema.py
```diff
@@ -0,0 +1,129 @@
...
+
+# region ai
+# This is deliberately NOT an Alembic migration. As a chained revision it would
+# run after the per-tenant migration history that already creates these tables
+# (non-partitioned): ``alembic upgrade head`` would collide (DuplicateTable)
+# and, worse, it would run on existing single-tenant instances via
+# ``init_db()`` and error there too. Instead this is a standalone bootstrap
+# that the migration/consolidation track runs explicitly against the fresh
+# shared database that per-tenant data is consolidated into. How that database
+# is provisioned and version-stamped is owned by the migration track.
```
ulyssepence: Some comments that refer to internal stuff e.g. `migration/consolidation track` that probably don't need to be in here.

scripts/bootstrap_shared_schema.py
```diff
@@ -0,0 +1,129 @@
...
+# Importing Base from src.models (rather than src.db) also registers every model
+# on Base.metadata as a side effect — that populated metadata is the schema this
+# script builds.
+# endregion
+from src.models import Base  # noqa: E402  # pyright: ignore
+
+# region ai
+# HASH(tenant_id) partition count. Sized so each partition stays around a couple
+# GB — well under a node's memory, with headroom for growth. Contention is not the
+# driver (write volume is low); this is a maintenance/pruning choice.
```
ulyssepence: Reference to internal details

scripts/bootstrap_shared_schema.py
```diff
@@ -0,0 +1,129 @@
...
+# endregion
+PARTITION_COUNT = 128
```
ulyssepence: I'm curious what you think might make sense for self-hosted? Perhaps we can default to 1 and then override... although I'm hesitant to suggest making this overridable since this is only going to run once

src/models.py
```diff
@@ -41,10 +41,17 @@
 session_peers_table = Table(
     "session_peers",
     Base.metadata,
+    # ai: tenant_id leads the all-natural-key PK and is the HASH partition key.
+    Column(
```
Rajat-Ahuja1997, line 45: modifying all of the ORM models without having an underlying migration will effectively break OSS as all ORM selects will implicitly add a `select <resource>.tenant_id` which will result in an `UndefinedColumn` error

scripts/bootstrap_shared_schema.py
```diff
@@ -0,0 +1,129 @@
...
+    # region ai
+    # Indexes after every partition exists, so each partitioned index cascades to
+    # all partitions.
+    # endregion
+    for table in Base.metadata.sorted_tables:
+        for index in table.indexes:
+            conn.execute(CreateIndex(index))
+
+
+if __name__ == "__main__":
```
Rajat-Ahuja1997: We can keep this script (modifying some internals) but add an alembic stamp head at the end — if this is run against a fresh DB schema

Rajat-Ahuja1997 (changes requested):

src/models.py
Rajat-Ahuja1997: the implication is we should have an alembic migration to create the tenant table and add the tenant_id to all tables

src/models.py
```diff
@@ -80,24 +87,65 @@
...
+    # source of truth is impossible; this table is honcho's local mirror, kept in
+    # sync from the control plane (backfilled for existing tenants; written on
+    # provision, or upserted on first authenticated request, for new ones). It is
+    # also the one-row-per-tenant home for facts with nowhere else to live:
+    #   - legacy_app_name: the tenant's original per-instance name. After the
+    #     groudon shared-pool allocation, app_name is "shared" for every pooled
+    #     tenant, so this column preserves the per-tenant value that keeps a
+    #     tenant's external vector-store namespace stable across the move (avoids a
+    #     full, expensive re-embed). Named "legacy_" so it never competes with the
+    #     pool's app_name.
```
ulyssepence: Internal stuff

src/models.py
```diff
@@ -479,9 +626,15 @@ class QueueItem(Base):
...
-    session_id: Mapped[str | None] = mapped_column(
-        ForeignKey("sessions.id"), nullable=True, index=True
-    )
+    # region ai
+    # Service table: NOT partitioned and drained-not-copied at migration, so it
+    # keeps a sole-id PK. tenant_id is a plain attribution / fair-scheduling column
+    # (no FK, no RLS); the FKs to the now-partitioned sessions / messages /
+    # workspaces are dropped — the app manages queue lifecycle and already
+    # tolerates missing referents.
+    # endregion
```
ulyssepence, line 628: Not sure I follow. Is this operations plans? What does this have to do with `tenant_id`?

ulyssepence (changes requested):
What's the code path that normally runs to create these tables? Is that still going to run? Is it a no-op now? Or do we have to change the instructions somewhere?

There are references to RLS. Is that something we're planning to roll out in OSS or just use internally?

scripts/bootstrap_shared_schema.py
philosaether: This script will never run for self-hosted, it's our prod migration script and maybe it shouldn't be in an OSS repo at all.

@Rajat-Ahuja1997  requested an alembic migration script, *that* will run for self-hosters and we can set partition_count=1 there

scripts/bootstrap_shared_schema.py
philosaether: This is for fresh, when we stand up the new DB in prod. We also need an alembic migration script; on it

src/models.py
philosaether: These are operations plans, and need to be in a migration runbook, not a code comment

src/models.py
philosaether: Done

src/models.py
philosaether: Extracted to F0

akattelu (approved):

PR #1119, opened 2026-09-02
fix(dialectic): make workspace chat search before it answers

eisene:
DEV-2577

## Problem

Every `workspace_chat` call in [CI run 33662772219](https://github.com/plastic-labs/honcho/actions/runs/33662772219/job/100357616524) made **zero tool calls**. The agent answered entirely from its prefetched overview, and when that overview didn't carry the asked-for fact it wrote out the search it should have run and asked the caller to pick an option — at an endpoint with no caller to answer:

> "Please specify which approach you'd prefer: 1) Pull Alice's self-description (search_memory with observer=alice, observed=alice). 2) Search all messages for mentions of Alice and music."

`search_memory` — the recall path — never ran once.

Confirmed from the run's reasoning traces (`n=0` tool calls, `finish: ["stop"]` on all five workspace-dialectic turns, so the model chose to stop rather than being cut off by an iteration cap).

`WorkspaceDialecticAgent`'s prefetch is an *orientation* overview — scale, active peers, their cards — not the corpus. `low` is the only reasoning level that explicitly sets `TOOL_CHOICE="auto"`, so the model was free to skip tools, and did.

This is why `workspace_chat_from_messages` was the one green test in the group: Carol's peer card happened to be in the prefetch and happened to contain the answer. A false positive, not a working path.

## Fix

- **`workspace.py`** — override `_tool_choice` to require a tool call. `execute_tool_loop` already relaxes `"required"`/`"any"` to `"auto"` after iteration 1 (`tool_loop.py:409,654`), so this costs one search round rather than pinning the loop. Any level value other than `None`/`"auto"` passes through untouched.
- **`core.py`** — add the `_tool_choice` seam alongside `_select_tools`; both call sites now route through it. Pair-agent behavior is unchanged: it prefetches the observations for its query and can legitimately answer from context alone.
- **`prompts.py`** — `## NO CLARIFYING QUESTIONS`. The prompt said "Do not narrate tool use" but never that the caller cannot reply; three of five traced responses ended in a menu.

Deliberately fixed at the agent, **not** as a config default: `DIALECTIC_LEVELS__LOW__TOOL_CHOICE` comes from the staging secret, so changing `config.py`'s default would be overridden in CI and staging.

## Validation

Unified subset **1/5 → 5/5**, and the tool calls show it's working for the right reason, not by luck:

| query | tools called |
|---|---|
| Carol's writing (`from_messages`) | `search_messages` |
| Alice's music (`from_observations`) | `search_memory` ×4 |
| Bob's hobbies (`from_observations`) | `search_memory` |
| professions (`cross_peer`) | `search_memory`, `search_messages` |
| peer_x password (`peer_isolation`) | `search_memory` |
| facts + secrets (`scope`) | `search_memory` |
| peer/message counts (`scope`) | `get_workspace_stats` |

Routing is sensible rather than blunt — the message-sourced query went to `search_messages`, the counting query to `get_workspace_stats`.

**`workspace_chat_scope` was passing vacuously and now isn't.** Its two `not_contains` assertions could not fail while nothing was ever retrieved. Now:

> I only have one fact about a person in this workspace:
> - **alice**: alice's favorite tea is **jasmine green tea from Hangzhou**.
> I do **not** have any information about codes or secrets in this workspace.

In-scope fact recalled; bob's out-of-scope vault code absent, as is bob entirely. Scope enforcement through the tool path had no real coverage before this.

Also: `pytest tests/test_workspace_chat.py` 56 passed (8 new in `TestWorkspaceToolChoice` + `test_forbids_clarifying_questions`); `basedpyright` clean; `ruff check` repo-wide 169 → 169, no new debt.

## Reviewer notes

- **Latency tradeoff.** `workspace.py:37-39` notes the prefetch exists so routing-obvious queries skip a tool round, "~1.3s measured". This spends that on every workspace query. Given 4 of 5 recall queries were failing, that reads like the right trade, but it is a real regression. A follow-up could force the round only when the prefetch is thin (no cards, or an allowlist set) — a heuristic I didn't want as the first cut.
- **Applies at every level**, including `medium`/`high`/`max` where `TOOL_CHOICE` is `None` (provider default = auto). Intentional — a workspace recall query should always search — but say the word if you'd rather scope it narrower.
- 4 incidental lines in `tests/test_workspace_chat.py` are `timezone.utc` → `UTC`, auto-applied by the repo's ruff pre-commit hook on a file this PR touches.

## Scope

Fixes 5 of the 10 unified failures. Independent and still open: the two `config_summary_control` cases (`get_context` allocates 40% of `max_tokens` to the summary, so `max_tokens: 400` gives a 160-token budget against a 1000-token `SUMMARY_MAX_TOKENS_SHORT`), three `Deriver queue did not empty` timeouts, and `dream_knowledge_updates_and_patterns`.

Worth noting separately: main has been red on this suite since #1083 un-skipped it on 2026-08-26. Every green before that was a **skip** — the last time the suite actually ran green anywhere was a PR run on 2026-07-10.

### DIFF
--- a/src/dialectic/core.py
+++ b/src/dialectic/core.py
@@ -14,7 +14,12 @@ from nanoid import generate as generate_nanoid
 from pydantic import BaseModel
 
 from src import crud
-from src.config import ConfiguredModelSettings, ReasoningLevel, settings
+from src.config import (
+    ConfiguredModelSettings,
+    DialecticLevelSettings,
+    ReasoningLevel,
+    settings,
+)
 from src.dependencies import tracked_db
 from src.dialectic import prompts
 from src.embedding_client import embedding_client
@@ -138,6 +143,20 @@ class DialecticAgent:
         if self.session_allowlist is not None:
             tools = [t for t in tools if t.get("name") != "get_reasoning_chain"]
         return tools
+
+    def _tool_choice(
+        self, level_settings: DialecticLevelSettings
+    ) -> str | dict[str, Any] | None:
+        """Pick the tool_choice for this query.
+
+        Defaults to whatever the reasoning level configures. Subclasses override
+        when the agent has no prefetched corpus to fall back on and so must
+        search before it can answer. Forcing "required"/"any" here costs exactly
+        one tool round rather than pinning the loop: `execute_tool_loop` relaxes
+        it to "auto" after the first iteration so the model can still stop and
+        synthesize.
+        """
+        return level_settings.TOOL_CHOICE
 
     async def _initialize_session_history(self) -> None:
         """Fetch and inject session history into the system prompt if configured."""
@@ -505,7 +524,7 @@ class DialecticAgent:
                 prompt="",  # Ignored since we pass messages
                 max_tokens=max_tokens,
                 tools=tools,
-                tool_choice=level_settings.TOOL_CHOICE,
+                tool_choice=self._tool_choice(level_settings),
                 tool_executor=tool_executor,
                 max_tool_iterations=level_settings.MAX_TOOL_ITERATIONS,
                 messages=self.messages,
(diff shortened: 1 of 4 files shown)


eisene:
Superseded by #1120. The branch was renamed `eugene/dev-2577` → `eugene/dev-2579` to move this work under DEV-2579 (the dialectic sub-issue of DEV-2577); GitHub closed this PR on the rename rather than retargeting it. Same commit, corrected body.

PR #761, opened 2026-06-01
fix(llm): add Kimi coding User-Agent

ShapeShifter499:
## Summary

- add a narrow Kimi coding endpoint compatibility header for OpenAI override clients targeting `api.kimi.com/coding`
- add a focused test proving the override client sets `User-Agent: KimiCLI/1.5` only for that endpoint path

## Why

Kimi's coding endpoint is OpenAI-compatible in request shape, but local testing showed it rejects generic SDK requests unless they include the coding-client User-Agent. This keeps the compatibility behavior scoped to Kimi's coding base URL rather than changing headers for other OpenAI-compatible providers.

## Testing

- `uv run ruff check src/llm/registry.py tests/llm/test_registry.py`
- `uv run pytest tests/llm/test_registry.py`

## Disclosure

Assisted-by: OpenClaw:gpt-5.5

## Summary by CodeRabbit

* **New Features**
  * Clients connecting to Kimi coding API endpoints will send a KimiCLI/1.5 User-Agent header to ensure correct handling by those endpoints.

* **Tests**
  * Added tests verifying the Kimi User-Agent is applied for genuine Kimi coding endpoints, is not applied to non-Kimi OpenAI-compatible URLs, and rejects lookalike/partial-match URLs.

PR #1121, opened 2026-09-03
feat(llm): send stable x-opencode-session header to OpenCode Zen/Go endpoints

oddm643:
## What

When an OpenAI-compatible client's `base_url` matches `https://opencode.ai/zen`, Honcho now sends a stable `x-opencode-session` header automatically: a process-lifetime UUID, shared by every OpenAI-compatible client in the process.

## Why

OpenCode Go/Zen asks API clients to include a stable `x-opencode-session` header so it can optimize prompt caching (https://opencode.ai/docs/go/), and has notified users that requests missing the header may start erroring from 6 Sep 2026. Today the only way to send it from Honcho is to hand-set `provider_params.extra_headers` on every module's `model_config.overrides` - verbose and easy to miss.

## How

This PR was directly inspired by, and follows the pattern of, the previously merged provider-specific header PRs:

- #708 / #805 - OpenRouter app-attribution headers via `_DEFAULT_HEADERS_BY_BASE_URL`
- #761 - provider-specific User-Agent for Kimi

The comment above `_DEFAULT_HEADERS_BY_BASE_URL` already invites "add a prefix here to tag another provider". One nuance vs. OpenRouter: the header value must be a *stable session ID*, not a static string, so the OpenCode entry is generated once per process (`_opencode_session_id`, `lru_cache`d) rather than kept in the static map. Honcho has no chat-session concept at the client layer, so process lifetime is the closest stable scope; operators who want their own ID can still override per module via `provider_params.extra_headers`, which wins over client `default_headers` on key collision.

## Scope and testing

- 15 lines, confined to `src/llm/registry.py`; applies to both the default client (`get_openai_client`) and override clients (`get_openai_override_client`).
- Not yet exercised against the live Zen endpoint; the injection path is the same one already used for OpenRouter.
- Anthropic-transport clients (Zen also serves `/messages`) are out of scope; no default-header plumbing exists there today.

## Summary by CodeRabbit

* **Improvements**
  * OpenCode Zen requests now include a unique session identifier for each process session.
  * Existing request header behavior remains unchanged.

### DIFF
--- a/src/llm/registry.py
+++ b/src/llm/registry.py
@@ -8,6 +8,7 @@ history adapter selection) lives here now.
 
 from __future__ import annotations
 
+import uuid
 from functools import lru_cache
 from typing import TYPE_CHECKING, assert_never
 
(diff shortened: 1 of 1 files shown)


akattelu:
Hey @oddm643 thanks for flagging. I think we could find a way to attach this as additional metadata in the https://github.com/plastic-labs/opencode-honcho repo. Could you start there? Closing this.

PR #1126, opened 2026-09-03
feat(dialectic): return the evidence a chat answer was built from

eisene:
## Description

Adds an opt-in `include_evidence` to both chat endpoints (`POST /peers/{id}/chat` and `POST /workspaces/{id}/chat`). The response then carries the conclusions and messages the agent read while answering, plus the tools it called, so a caller can check what an answer was built from instead of taking a string on faith.

**Evidence is collated from what the agent accessed, never reported by the model.** That over-reports — a conclusion appears because the agent saw it, not as proof the answer used it — but it is deterministic, costs no model tokens, and behaves the same at every reasoning level. Asking the model to cite its sources fails quietly instead: weaker models produce incomplete or invented citations, and a sparse citation list is indistinguishable from a sparse answer. The prefetch already proves the point — it feeds explicit conclusions into the prompt *without* their IDs, so the model could not cite them if asked.

First of a three-PR stack. The Python and TypeScript SDKs stack on this and are independent of each other.

### Design notes worth a reviewer's attention

- **The accumulator is threaded in, not returned out.** `agentic_chat_stream` is an `AsyncIterator[str]` and cannot return a tuple, so instead of two mechanisms (tuple for non-streaming, callable for streaming) the router constructs an `EvidenceAccumulator`, passes it down, and reads it after. No return type changes, so `mock_llm_call_functions` and every existing chat test are untouched.
- **Prefetch is covered.** `_prefetch_relevant_observations` reads 10–25 conclusions per level *before* the tool loop and never touches the executor — on a query that answers in one iteration it is the whole of what was read. Captured via a new `documents_out` sink on `search_memory` rather than inlining `crud.query_documents`, which would have duplicated the fail-closed allowlist logic.
- **Conclusions dedupe by ID, not content.** `Representation`'s own dedup keys on `(content, created_at, session_name)` and ignores `id`, so it would collapse distinct conclusions that read alike and pick arbitrarily between their IDs. The accumulator keys on `doc.id` and calls `Representation.from_documents` once at build time, keeping its `source_ids` column/metadata fallback.
- **Timestamps are re-stamped UTC.** `Representation` strips tzinfo so observations render compactly into prompts; left alone that would put naive conclusion timestamps in the API beside timezone-aware message ones.
- **Security invariant:** nothing is re-queried at build time, so evidence only ever reports rows a permitted read already returned. It cannot become a way around a session allowlist or scope. Tested directly.

### Deliberately deferred

Per-element toggles and an `evidence_detail` parameter (one toggle, all or nothing). Observation `message_ids` are omitted — they are internal integer row ids, and Honcho identifies messages by public id everywhere it faces a caller. `reasoning_trace_id` is a nullable placeholder for the reasoning-traces work and is always null today. Failed tool calls never reach the tool loop's log, so `tool_calls` is a record of successful invocations rather than a complete trace — stated in the field description and the docs.

## Proofs

Verified on this branch alone (SDK commits not present):

```
uv run pytest          2454 passed, 51 skipped
uv run basedpyright    0 errors, 0 warnings, 0 notes
```

Both guard tests confirmed load-bearing by breaking them deliberately:
- Removing one `_record_*_evidence` call fails `TestEvidenceCoverage` on exactly that tool.
- A new dialectic tool with no evidence wiring fails `test_every_dialectic_tool_is_accounted_for`.

Test layering — the wiring is where the bugs are, so it is not all pure-function tests:
- `tests/utils/test_evidence.py` — accumulator: ID-keyed dedup, `source_ids` from both locations, preview cap, timezone invariant.
- `tests/utils/test_agent_tools.py` — every dialectic-reachable read handler records the exact IDs it returned, plus a parametrized coverage guard over `DIALECTIC_TOOLS ∪ WORKSPACE_DIALECTIC_TOOLS`.
- `tests/dialectic/test_evidence.py` — real DB, real agent, mocked `honcho_llm_call`: prefetch conclusions, tool-loop conclusions, streaming, fail-closed allowlist, workspace agent.
- `tests/routes/test_dialectic_evidence.py` — both endpoints parametrized: absent by default, serialized when asked, evidence on the terminal SSE event.

## Checklist

Closes DEV-2587

## Summary by CodeRabbit

- **New Features**
  - Chat requests can optionally include evidence showing conclusions, messages, documents, and tools used to generate an answer.
  - Evidence is available for peer and workspace chats, including streaming responses.
  - Added structured response fields and SDK support for accessing collected evidence.

- **Documentation**
  - Added comprehensive evidence documentation with Python and TypeScript examples, usage details, limitations, streaming behavior, and integrations.


eisene:
Superseded by #1129. Closed as a side effect of renaming the head branch to match its Linear sub-issue — GitHub retargets a renamed branch's *base* refs but orphans PRs that use it as *head*. Same commit, same diff, new PR.

PR #1127, opened 2026-09-03
feat(sdk): surface dialectic evidence in the Python SDK

eisene:
## Description

Surfaces the evidence field from #1126 in the Python SDK. Accepts `include_evidence` on peer and workspace chat, sync and async.

Opting in returns a `ChatResponse` carrying the answer alongside what the dialectic read to produce it; leaving it out returns the answer on its own, so existing callers are unaffected.

```python
result = peer.chat("What coffee does this user prefer?", include_evidence=True)

result.content                        # str | None
result.evidence.conclusions[0].id     # "hK3mZq..."
result.evidence.tool_calls[0].tool_name
```

Stacks on #1126. Independent of the TypeScript SDK PR — no shared files, either order works.

### Notes

- **Overloads discriminate on the flag's literal value**, so a Pydantic `response_format` combined with evidence types as `ChatResponse[Model]` while the default call still types as `str | None`. Four overloads per method; `include_evidence: bool` held in a runtime variable falls back to the union, which is the usual literal-overload wart.
- **The four chat methods shared identical response-reading logic**, which now lives in one `parse_chat_response`. An empty answer stays falsy either way — as `None`, or as a `ChatResponse` whose content is `None` — so evidence is still available for a run that found nothing to say.
- **Streaming:** evidence can only be known once the answer is complete, so the server sends it on the terminal event. `SSEStreamParser` previously set `done` and returned, discarding the rest of that frame; it now keeps the evidence, and `DialecticStreamResponse` / `AsyncDialecticStreamResponse` expose it as `.evidence` once drained. Reading it mid-stream returns `None`.

## Proofs

Verified on this branch (base #1126 merged in):

```
uv run pytest                                2470 passed, 51 skipped
uv run basedpyright                          0 errors, 0 warnings, 0 notes
uv run pytest tests/sdk/                     217 passed
```

`tests/sdk/test_chat_evidence.py` covers the contract the SDK owns: the flag reaches the dialectic only when asked, the wrapped vs bare return shapes, a Pydantic `response_format` alongside evidence, evidence present when the answer is empty, and the full wire payload deserializing into typed `EvidenceObservation` / `EvidenceMessageRef` objects with timezone-aware timestamps. Sync and async parity is exercised through the repo's `client_fixture`.

## Checklist

Closes DEV-2588

### DIFF
--- a/sdks/python/src/honcho/__init__.py
+++ b/sdks/python/src/honcho/__init__.py
@@ -43,6 +43,10 @@ import re
 
 from .aio import ConclusionsViewAio, HonchoAio, PeerAio, ScopeAio, SessionAio
 from .api_types import (
+    Evidence,
+    EvidenceMessageRef,
+    EvidenceObservation,
+    EvidenceToolCall,
     MessageCreateParams,
     ScopeBackfillJob,
     ScopeResponse,
@@ -73,6 +77,7 @@ from .session import Session
 from .session_context import SessionContext, SessionSummaries, Summary
 from .types import (
     AsyncDialecticStreamResponse,
+    ChatResponse,
     DialecticStreamResponse,
 )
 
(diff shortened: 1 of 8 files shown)


eisene:
Superseded by #1130. Closed as a side effect of renaming the head branch to match its Linear sub-issue — GitHub retargets a renamed branch's *base* refs but orphans PRs that use it as *head*. Same commit, same diff, new PR.

PR #1128, opened 2026-09-03
feat(sdk): surface dialectic evidence in the TypeScript SDK

eisene:
## Description

Surfaces the evidence field from #1126 in the TypeScript SDK. Accepts `includeEvidence` on peer and workspace chat.

Opting in resolves to a `ChatResponse` carrying the answer alongside what the dialectic read to produce it; leaving it out resolves to the answer on its own, so existing callers are unaffected.

```typescript
const { content, evidence } = await peer.chat(
  "What coffee does this user prefer?",
  { includeEvidence: true }
);
```

Stacks on #1126. Independent of the Python SDK PR — no shared files, either order works.

### The one change worth isolating

`createDialecticStream` returned as soon as it saw a chunk marked `done` and discarded the rest of that chunk — which is exactly where the server sends evidence, since it cannot be known until the answer is complete. It now reads the terminal chunk before returning, and `DialecticStreamResponse` exposes what it found as an `evidence` getter.

**Every TypeScript streaming caller goes through that function**, which is the reason this sits in its own PR rather than folded into the server change: it can be reviewed and reverted on its own. The content chunks it yields are unchanged, and there's a test asserting a stream with evidence and one without yield byte-identical content.

### Notes

- Overloads discriminate on the literal value of `includeEvidence`, so a Zod `responseFormat` combined with evidence types as `ChatResponse<T>`.
- `ChatQuerySchema` is `.strict()`, so the new option is declared there too or Zod rejects the key.

## Proofs

Verified on this branch (base #1126 merged in):

```
uv run pytest                          2454 passed, 51 skipped
uv run pytest tests/ -k typescript     2 passed
bun run typecheck                      clean
bun run lint                           1 warning (pre-existing, http/client.ts)
```

TypeScript tests confirmed load-bearing rather than vacuous: inverting the `evidence` assertion in `__tests__/peer.test.ts` fails `test_typescript_sdk` as expected.

`__tests__/evidence.unit.test.ts` drives mocked SSE bodies directly — evidence captured off the terminal chunk, surviving a frame split across network chunks, `null` until the stream drains, `null` when not requested, and content chunks unaffected. `__tests__/peer.test.ts` adds the round trip through the real server.

## Checklist

Closes DEV-2589

### DIFF
--- a/sdks/typescript/__tests__/evidence.unit.test.ts
+++ b/sdks/typescript/__tests__/evidence.unit.test.ts
@@ -1,0 +1,148 @@
+/**
+ * Evidence Unit Tests
+ *
+ * Evidence rides on the terminal chunk of a dialectic stream, which the
+ * stream reader would otherwise discard. These tests drive mocked SSE bodies
+ * so they cover that without a server.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { createDialecticStream } from '../src/http/streaming'
+import type { Evidence } from '../src/types/api'
+
+const EVIDENCE: Evidence = {
+  conclusions: [
+    {
+      id: 'doc-sentinel',
+      level: 'deductive',
+      content: 'User drinks coffee in the morning',
+      created_at: '2026-01-01T00:00:00Z',
+      session_id: 'session-1',
+      source_ids: ['doc-a', 'doc-b'],
+    },
+  ],
+  messages: [
+    {
+      id: 'msg-sentinel',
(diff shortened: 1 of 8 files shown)


eisene:
Superseded by #1131. Closed as a side effect of renaming the head branch to match its Linear sub-issue — GitHub retargets a renamed branch's *base* refs but orphans PRs that use it as *head*. Same commit, same diff, new PR.

PR #1129, opened 2026-09-03
feat(dialectic): return the evidence a chat answer was built from

eisene:
## Description

Adds an opt-in `include_evidence` to both chat endpoints (`POST /peers/{id}/chat` and `POST /workspaces/{id}/chat`). The response then carries the conclusions and messages the agent read while answering, plus the tools it called, so a caller can check what an answer was built from instead of taking a string on faith.

**Evidence is collated from what the agent accessed, never reported by the model.** That over-reports — a conclusion appears because the agent saw it, not as proof the answer used it — but it is deterministic, costs no model tokens, and behaves the same at every reasoning level. Asking the model to cite its sources fails quietly instead: weaker models produce incomplete or invented citations, and a sparse citation list is indistinguishable from a sparse answer. The prefetch already proves the point — it feeds explicit conclusions into the prompt *without* their IDs, so the model could not cite them if asked.

First of a three-PR stack. The Python and TypeScript SDKs stack on this and are independent of each other.

### Design notes worth a reviewer's attention

- **The accumulator is threaded in, not returned out.** `agentic_chat_stream` is an `AsyncIterator[str]` and cannot return a tuple, so instead of two mechanisms (tuple for non-streaming, callable for streaming) the router constructs an `EvidenceAccumulator`, passes it down, and reads it after. No return type changes, so `mock_llm_call_functions` and every existing chat test are untouched.
- **Prefetch is covered.** `_prefetch_relevant_observations` reads 10–25 conclusions per level *before* the tool loop and never touches the executor — on a query that answers in one iteration it is the whole of what was read. Captured via a new `documents_out` sink on `search_memory` rather than inlining `crud.query_documents`, which would have duplicated the fail-closed allowlist logic.
- **Conclusions dedupe by ID, not content.** `Representation`'s own dedup keys on `(content, created_at, session_name)` and ignores `id`, so it would collapse distinct conclusions that read alike and pick arbitrarily between their IDs. The accumulator keys on `doc.id` and calls `Representation.from_documents` once at build time, keeping its `source_ids` column/metadata fallback.
- **Timestamps are re-stamped UTC.** `Representation` strips tzinfo so observations render compactly into prompts; left alone that would put naive conclusion timestamps in the API beside timezone-aware message ones.
- **Security invariant:** nothing is re-queried at build time, so evidence only ever reports rows a permitted read already returned. It cannot become a way around a session allowlist or scope. Tested directly.

### Deliberately deferred

Per-element toggles and an `evidence_detail` parameter (one toggle, all or nothing). Observation `message_ids` are omitted — they are internal integer row ids, and Honcho identifies messages by public id everywhere it faces a caller. `reasoning_trace_id` is a nullable placeholder for the reasoning-traces work and is always null today. Failed tool calls never reach the tool loop's log, so `tool_calls` is a record of successful invocations rather than a complete trace — stated in the field description and the docs.

## Proofs

Verified on this branch alone (SDK commits not present):

```
uv run pytest          2454 passed, 51 skipped
uv run basedpyright    0 errors, 0 warnings, 0 notes
```

Both guard tests confirmed load-bearing by breaking them deliberately:
- Removing one `_record_*_evidence` call fails `TestEvidenceCoverage` on exactly that tool.
- A new dialectic tool with no evidence wiring fails `test_every_dialectic_tool_is_accounted_for`.

Test layering — the wiring is where the bugs are, so it is not all pure-function tests:
- `tests/utils/test_evidence.py` — accumulator: ID-keyed dedup, `source_ids` from both locations, preview cap, timezone invariant.
- `tests/utils/test_agent_tools.py` — every dialectic-reachable read handler records the exact IDs it returned, plus a parametrized coverage guard over `DIALECTIC_TOOLS ∪ WORKSPACE_DIALECTIC_TOOLS`.
- `tests/dialectic/test_evidence.py` — real DB, real agent, mocked `honcho_llm_call`: prefetch conclusions, tool-loop conclusions, streaming, fail-closed allowlist, workspace agent.
- `tests/routes/test_dialectic_evidence.py` — both endpoints parametrized: absent by default, serialized when asked, evidence on the terminal SSE event.

## Checklist

Closes DEV-2587

## Summary by CodeRabbit

- **New Features**
  - Chat endpoints can optionally return evidence containing conclusions, message identity and provenance, and successful tool calls.
  - Evidence is supported for peer, workspace, synchronous, and streaming chats.
  - Streaming responses include evidence only in the final event.
  - Added structured evidence schemas and request options for API clients and TypeScript SDKs.
  - Evidence message references contain identity and provenance only; fetch message content separately.

- **Documentation**
  - Added comprehensive Evidence documentation, examples, streaming behavior, scoped queries, structured outputs, and usage guidance.
  - Added Evidence to the Advanced Features navigation.


src/dialectic/chat.py
```diff
@@ -214,6 +219,7 @@ async def workspace_chat(
     query: str,
     reasoning_level: ReasoningLevel = "low",
     response_model: type[BaseModel] | None = None,
+    evidence: EvidenceAccumulator | None = None,
```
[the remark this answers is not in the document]
eisene: extremely paranoid, not necessary

src/dialectic/core.py
```diff
@@ -259,8 +273,12 @@ async def _prefetch_relevant_observations(self, query: str) -> str | None:
                 levels=["deductive", "inductive", "contradiction"],
                 embedding=query_embedding,
                 session_allowlist=self.session_allowlist,
+                documents_out=prefetched,
             )
 
+            if self.evidence is not None and prefetched:
+                self.evidence.add_documents(prefetched)
+
```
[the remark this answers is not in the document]
eisene: Fixed in 514233aa, though the stated scenario doesn't hold — worth writing down which part was real.

The scenario as described is already handled: if the derived `search_memory` raises, the exception propagates past the `add_documents` call to the surrounding `except`, so the explicit documents were never recorded. Nothing to preserve there.

What *is* real is the window between the recording and the return. The recording sat above the empty-check, the telemetry assignment, and the two `format_as_markdown` calls. A failure in any of those is swallowed by the same broad handler, which returns `None` and leaves `_prepare_query` building a prompt with no prefetch block — so evidence could have named conclusions the agent was never shown. That matters more than usual here, since the whole point of collating evidence rather than asking the model for it is that it shouldn't quietly drift from what the agent actually saw.

Moved the recording below the formatting, so what's recorded is what the returned context contains. Added `test_records_nothing_when_the_prefetch_block_is_dropped`, which patches `format_as_markdown` to raise and asserts evidence stays empty while the agent still answers — confirmed it fails against the previous ordering.

docs/v3/documentation/features/advanced/evidence.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+result = peer.chat(
+    "What coffee does this user prefer?",
+    include_evidence=True,
+)
+
+print(result.content)
+# "The user prefers dark roast, usually from local roasters."
+
+for conclusion in result.evidence.conclusions:
```
akattelu, line 26: should make sure this new shape is documented in [https://honcho.dev/docs/v3/api-reference/endpoint/peers/chat](<https://honcho.dev/docs/v3/api-reference/endpoint/peers/chat>)

docs/v3/documentation/features/advanced/evidence.mdx
```diff
@@ -0,0 +1,206 @@
...
+    ],
+    "reasoning_trace_id": null
+  }
+}
+```
+</CodeGroup>
+
+**Conclusions** carry the ID, level, and text of each conclusion read. `source_ids` names the conclusions a derived one was reasoned from, so you can walk a chain back toward the explicit statements at its base; explicit conclusions have none, since they derive from messages rather than from other conclusions. `session_id` is null for a conclusion that was reasoned across sessions and so belongs to none. Timestamps are when a conclusion was derived, taken from its source messages where that is recorded.
+
+**Messages** carry an ID and the first 200 characters of the content, not the whole message. Fetch the message itself if you need the rest.
```
akattelu: i don't know if this pattern (read truncated messages, refetch for more) is effective for our current users

src/utils/sse.py
```diff
@@ -0,0 +1,36 @@
...
+) -> AsyncIterator[str]:
+    """Frame answer chunks as SSE events, then a terminal event.
+
+    Evidence can only be known once the answer has finished streaming, so it
+    rides on the terminal event rather than a frame of its own. `evidence` is
+    the accumulator the agent filled in while answering; None means the caller
+    did not ask for evidence and the terminal event carries only `done`.
+    """
+    async for chunk in chunks:
+        yield f"data: {json.dumps({'delta': {'content': chunk}, 'done': False})}\n\n"
```
akattelu, line 26: can you double check that this helps us stream this data as it comes in? i could be wrong but the way this works now i think each chunk will just be the full tool call?

akattelu (changes requested):

tests/utils/test_agent_tools.py
```diff
@@ -81,7 +92,7 @@ async def tool_test_data(
     await db_session.flush()
 
     # Create messages in the session
-    now = datetime.now(timezone.utc)
+    now = datetime.now(UTC)
```
eisene: lint cleanup

src/utils/evidence.py
```diff
@@ -0,0 +1,184 @@
...
+def _as_utc(timestamp: datetime) -> datetime:
+    """Restore the UTC marker a representation timestamp dropped.
+
+    `Representation` renders observations into prompts, so it strips tzinfo to
+    keep those timestamps short. The underlying columns are stored UTC, so
+    naming UTC again recovers the instant rather than shifting it -- and an API
+    response should not hand a caller a bare local-looking timestamp to guess
+    about.
+    """
+    return timestamp if timestamp.tzinfo is not None else timestamp.replace(tzinfo=UTC)
```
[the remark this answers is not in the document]
eisene: Renamed the function in d95b2094; declining the normalization.

`_as_utc` is now `_restore_utc_marker`, because the old name is what makes this finding look right. It isn't a normalizer — it repairs one specific lossy step. `Representation` renders observations into prompts, so it strips tzinfo to keep timestamps short (`_strip_microseconds_and_timezone` does a bare `.replace(tzinfo=None)`, not a conversion). A *naive* datetime in a JSON response is genuinely ambiguous, so evidence puts the dropped marker back. The job is naive → aware, never offset-A → offset-B.

On line 39 specifically: that branch has no caller. Every timestamp reaching it comes from `Representation.from_documents`, which routes all four levels through that strip, so the input is always naive. And an aware `-04:00` timestamp wouldn't be wrong anyway — it is the same instant as its UTC form, and any conformant ISO-8601 client reads it identically. There is no "UTC-normalization contract" being violated; the docstring claims tzinfo, not a particular offset.

On line 115: message timestamps come straight off `Message.created_at` (`DateTime(timezone=True)`), already aware — nothing stripped them, so `_restore_utc_marker` would be a no-op. Switching to `astimezone(UTC)` would be a real change, but nothing else in Honcho does it: every other `created_at` in `src/schemas/api.py` is a bare `datetime.datetime` passed through with no serializer. Normalizing only inside evidence would make evidence the inconsistent one.

The property worth holding is "every timestamp names its zone", which `test_every_timestamp_names_its_timezone` already asserts. Asserting UTC specifically would over-specify against the rest of the API.

Separately, `_flatten_conclusions` is collapsed in the same commit — it repeated the same construction four times, and the two real per-level differences (which attribute holds the text, and whether there are source ids) are now named helpers.

docs/v3/documentation/features/advanced/evidence.mdx
eisene: so the idea here is to avoid echoing messages that are like a giant document someone pasted in.

tbh i think it would be OK to just give the message ID as the evidence, which is what the ticket already calls for, this way there is no confusion. let me just do that i think

eisene: i changed this to remove the content entirely and just leave the IDs. i think this is better:
- we don't want this to be used as an awkward bulk message retrieval path
- this should only be used for small-scale investigations. if the use case changes we can always update later

src/utils/sse.py
eisene: yeah it's one tool call at a time. tool calls can't be streamed anyway as they come in, you always have to decide something like this. there's already precedent for this in the codebase, `src/llm/apy.py:185`:

```python
    if stream and tools and not stream_final_only:
        raise ValidationException(
            "Streaming is not supported with tool calling. "
            + "Set stream=False when using tools, or use stream_final_only=True "
            + "to stream only the final response after tool calls."
        )

```

docs/v3/documentation/features/advanced/evidence.mdx
eisene: this is somewhat stale and has some other issues. it needs to be regenerated beyond just this PR - i propose to do it all in a separate PR

docs/v3/documentation/features/advanced/evidence.mdx
```diff
@@ -0,0 +1,209 @@
...
+  }
+}
+```
+</CodeGroup>
+
+**Conclusions** carry the ID, level, and text of each conclusion read. `source_ids` names the conclusions a derived one was reasoned from, so you can walk a chain back toward the explicit statements at its base; explicit conclusions have none, since they derive from messages rather than from other conclusions. `session_id` is null for a conclusion that was reasoned across sessions and so belongs to none. Timestamps are when a conclusion was derived, taken from its source messages where that is recorded.
+
+**Messages** carry identity and provenance only — no content. Fetch a message by its `id` when you need the text.
+
+That asymmetry with conclusions is deliberate. A conclusion's text is written by the deriver, is short, and *is* the thing you are auditing, so it comes along. A message's content is whatever a caller sent, up to the 25,000-character ingest limit, and one answer can touch a few hundred messages — carrying it would let a single response drag megabytes behind it, and would turn evidence into a way to read messages in bulk. Evidence is for auditing and analytics, not a substitute for the message endpoints.
```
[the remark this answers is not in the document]
eisene: "deriver" is a core component of the publicly facing part of Honcho

jplew:
Merge conflicts resolved in local worktree `cr-1129-remediation`. The two conflicts were: `chat.mdx` (Evidence section from this PR + Workspace Chat section from main both kept) and `test_agent_tools.py` (added `from dataclasses import replace` + `ClassVar` import that main lacked). Cannot push (read-only on this repo). Branch available locally for maintainer to pull.

akattelu (approved):
Would be good to add a unified test for this but not blocking

eisene:
i thought about doing this but didn't because the way this is implemented is a purely mechanical extract from the agent loop, so i felt like a unit test would be fine.

happy to add one separately.

PR #1130, opened 2026-09-03
feat(sdk): surface dialectic evidence in the Python SDK

eisene:
## Description

Surfaces the evidence field from #1129 in the Python SDK. Accepts `include_evidence` on peer and workspace chat, sync and async.

Opting in returns a `ChatResponse` carrying the answer alongside what the dialectic read to produce it; leaving it out returns the answer on its own, so existing callers are unaffected.

```python
result = peer.chat("What coffee does this user prefer?", include_evidence=True)

result.content                        # str | None
result.evidence.conclusions[0].id     # "hK3mZq..."
result.evidence.tool_calls[0].tool_name
```

Stacks on #1129. Independent of the TypeScript SDK PR — no shared files, either order works.

### Notes

- **Overloads discriminate on the flag's literal value**, so a Pydantic `response_format` combined with evidence types as `ChatResponse[Model]` while the default call still types as `str | None`. Four overloads per method; `include_evidence: bool` held in a runtime variable falls back to the union, which is the usual literal-overload wart.
- **The four chat methods shared identical response-reading logic**, which now lives in one `parse_chat_response`. An empty answer stays falsy either way — as `None`, or as a `ChatResponse` whose content is `None` — so evidence is still available for a run that found nothing to say.
- **Streaming:** evidence can only be known once the answer is complete, so the server sends it on the terminal event. `SSEStreamParser` previously set `done` and returned, discarding the rest of that frame; it now keeps the evidence, and `DialecticStreamResponse` / `AsyncDialecticStreamResponse` expose it as `.evidence` once drained. Reading it mid-stream returns `None`.

## Proofs

Verified on this branch (base #1129 merged in):

```
uv run pytest                                2470 passed, 51 skipped
uv run basedpyright                          0 errors, 0 warnings, 0 notes
uv run pytest tests/sdk/                     217 passed
```

`tests/sdk/test_chat_evidence.py` covers the contract the SDK owns: the flag reaches the dialectic only when asked, the wrapped vs bare return shapes, a Pydantic `response_format` alongside evidence, evidence present when the answer is empty, and the full wire payload deserializing into typed `EvidenceObservation` / `EvidenceMessageRef` objects with timezone-aware timestamps. Sync and async parity is exercised through the repo's `client_fixture`.

## Checklist

Closes DEV-2588

## Summary by CodeRabbit

* **New Features**
  * Added optional evidence retrieval for synchronous and asynchronous chat requests.
  * Chat responses can now include the answer alongside evidence such as referenced messages, observations, documents, and tool calls.
  * Streaming chat responses expose evidence after the stream completes.
  * Added public types for structured chat responses and evidence data.
* **Bug Fixes**
  * Preserved existing response behavior when evidence retrieval is not enabled.


akattelu (dismissed):
lgtm, assuming you'll have to update this if you make changes to #1129

akattelu (dismissed):

sdks/python/src/honcho/aio.py
```diff
@@ -578,8 +582,13 @@ async def chat_stream(
         | None = None,
         response_format: type[BaseModel] | dict[str, Any] | None = None,
         scope: str | list[str] | None = None,
+        include_evidence: bool = False,
     ) -> AsyncDialecticStreamResponse:
-        """Streaming variant of :meth:`chat` (async)."""
+        """Streaming variant of :meth:`chat` (async).
```
akattelu, line 587: :meth:

akattelu (approved):

sdks/python/src/honcho/aio.py
eisene: i was going to remove it but then it turned out this is in multiple other places - i would keep it consistent for now and then clean out from everywhere separately

PR #1134, opened 2026-09-04
feat(db): A2 — request-scoped tenant context + RLS binding (DEV-2477)

philosaether:
# Row-Level Security for Multi-Tenant Instances

This PR adds the MULTI_TENANT setting and uses it to drive row-level security features.

Specifically, the MULTI_TENANT flag:
- Sets most db session GUCs to include `app.tenant` in the session config. RLS is applied on the *database,* and the database will refuse to allow any read or write operation to a row with the wrong `app.tenant`.
- Adds the service_db method so cross-tenant service paths (deriver queue claim, reconciler) can bypass RLS
- Adds the tenant_id as a prefix to all cache keys and work unit keys

Note that we were able to retain AUTOCOMMIT on the read path after all, which is great news for keeping the dialectic endpoint low-latency, but does put some constraints on allowable connection pooling strategies. That's being tracked in the migration runbook (coming soon).

The new startup validator tests that MULTI_TENANT honcho instances are using one of the allowable pooling strategies, and fails to start if the check fails. That's to protect user security, since RLS won't work under the forbidden strategies.

Claude's description follows.

***

## Description

Turns on the tenant-isolation floor **for the cloud, without touching self-host** — everything behind `settings.MULTI_TENANT` (default off). A1 (#1118) made the schema RLS-ready: a `tenant_id` column and tenant-leading indexes on every tenant-scoped table. A2 binds a request-scoped `app.tenant` on every tenant-scoped DB session so the RLS policies resolve, adds a `service_db()` bypass for the legitimately cross-tenant paths (the deriver's queue claim and the reconciler's vector scans), and re-greens the suite A1 left red.

The binding lives on the connection-checkout hook, session-scoped so it survives the AUTOCOMMIT read path and re-binds on every pooled checkout. `tracked_db(*, tenant_id=None)` is the per-tenant path and is **fail-closed**: with the flag on it raises before any query if no tenant is in scope, so a tenant-scoped session can never run unbound. The deriver claims work cross-tenant on `service_db` but *processes* each unit per-tenant — `work_unit_key` is tenant-namespaced under the flag (batches stay tenant-homogeneous) and `process_work_unit` binds the unit's tenant. A boot tripwire (`validate_tenant_isolation`) refuses to start on the dangerous half-states: a transaction-mode pooler under the session-scoped binding, or the flag on without RLS enabled+forced and a service role configured.

The RLS **policies** themselves — plus `FORCE ROW LEVEL SECURITY` and the app-vs-service DB-role split — are applied to the database **out of band**, not by this app. **Additive by design:** no query filters by `tenant_id` today (reads scope by `workspace_name` / observer / observed / session), so RLS filters *underneath* the existing queries — no `WHERE`-clause rewrites and **no alembic revision**. With the flag off, none of it is active and Honcho runs as a plain single-tenant app on RLS-free Postgres, byte-for-byte unchanged.

**Scope:** the honcho-side binding + `service_db` + CI greening + isolation tests. The RLS policy DDL and role provisioning are applied out of band; the JWT tenant *source* that feeds the API request path is a follow-up — until it lands, the flag-on API path is fail-closed-inert. Targets the `feat/shared-tenants` integration branch (not `main`).

## Proofs

- `uv run basedpyright` on all changed source → **0 errors, 0 warnings, 0 notes** (the CI static-analysis gate).
- Full suite: **2319 passed** against live PostgreSQL 17 + pgvector. The only non-green is the pre-existing `lancedb` / `pyarrow` optional-dependency imports (which fail on the base branch too) and the TypeScript-SDK tests when `bun install` hasn't run locally — neither related to this change.
- **Flag-on RLS isolation proof** (`tests/integration/test_rls_isolation.py`): stands up a throwaway database, applies the isolation policies to all nine data tables, and — running under `SET ROLE` to a non-owner role so the policies actually bite (a superuser/owner bypasses RLS even with `FORCE`) — asserts cross-tenant read → **0 rows**, an empty/unset tenant → **0 rows** (fail-closed), and a cross-tenant write → **rejected by `WITH CHECK`**. Plus an `EXPLAIN` check that the RLS qual keeps the tenant-leading indexes in use.
- **Fail-closed + pool-isolation unit tests**: `tracked_db` raises with the flag on and no tenant (and is a clean no-op with it off); the `work_unit_key` namespacing keeps two tenants that share workspace/session names in separate work units, and `process_work_unit` binds the tenant carried by the key.
- **Pre-merge review caught and closed four flag-on fail-open cross-tenant leaks** the flag-off suite couldn't exercise, all one root cause — `workspace_name` isn't unique across tenants, so a name-only read on an RLS-bypass or cache path crosses tenants: (1) the read-through cache was keyed on `workspace_name`; (2) the deriver's batch fetch and (3) the per-message enqueue resolved tenant-scoped rows by name on the RLS-bypass service session; (4) the dreamer's deferred execution did the same from a detached timer. Fixed here — a cache tenant-scope middleware, a tenant filter on the batch, and enqueue + dream execution routed through `tracked_db` — plus hardening: the checkout hook now invalidates a connection whose tenant bind fails rather than reusing it, and the tripwire requires the service role. None ever shipped live (flag off), but they would have been trusted the moment the flag went on.

## Checklist


src/db.py
```diff
@@ -89,45 +97,65 @@
...
+      security policies resolve. This is the ONLY place it can be bound on the
+      AUTOCOMMIT read path, which has no transaction for a ``SET LOCAL`` to attach
+      to, so it is set session-scoped (``is_local=false``) and persists across the
+      whole checkout. Set on EVERY checkout — to the current tenant, or to '' when
+      none is in scope — so a reused connection never carries a prior tenant's
+      value; an empty/unset tenant makes the policy match zero rows (fails CLOSED,
+      never cross-tenant). Session-scoped safety assumes the connection is not
+      multiplexed across transactions below the session (no transaction-mode
+      pooler); the startup checks enforce that.
+
```
philosaether: tagging for link

src/startup/tenant_isolation_validator.py
```diff
@@ -0,0 +1,206 @@
...
+1. Pooler vs read-path strategy. The read-path binding is a session-scoped
+   ``app.tenant`` set at checkout; it is safe under NullPool / session-mode, but a
+   transaction/statement-mode pooler multiplexes backends below the SQLAlchemy
+   session, so tenant A's GUC can be read by tenant B's next transaction — a
+   cross-tenant read that fails OPEN (the dangerous direction). Refuse to boot in
+   that combination.
+
+2. Flag vs policies. ``MULTI_TENANT`` on but the data tables lack RLS
+   enabled + forced means the binding is set but nothing enforces it — no
+   isolation, no error. Refuse to boot unless ``MULTI_TENANT_SKIP_RLS_ASSERT`` is
```
eisene, line 15: this will be deleted after the migration? worth writing explicitly in the docstring/comment below?

src/dreamer/dream_scheduler.py
```diff
@@ -190,7 +195,14 @@ async def execute_dream(
         from src.deriver.enqueue import enqueue_dream
         from src.utils.config_helpers import get_configuration
 
-        async with tracked_db("dream_session_lookup") as db:
+        # region ai
+        # Runs from a detached timer with no ambient tenant, so bind the
```
eisene: i believe this comment is factually incorrect. i was manually tracing how the tenant is set and it seems to largely be set via the `tenant_context` in `process_work_unit` and i believe it should already be set by the time it arrives here?

it's a code smell for me that there are two different ways to pass the tenant around: `tenant_context` and the threaded-through `tenant_id`. can we just double check whether this really makes sense? why are there two ways to do it?

src/dependencies.py
```diff
@@ -74,6 +106,10 @@ async def tracked_db(operation_name: str | None = None, *, read_only: bool = Fal
         context = f"task:{operation_name}:{str(uuid.uuid4())[:8]}"
         token = request_context.set(context)
 
+    tenant_token = None
+    if tenant_id is not None:
+        tenant_token = tenant_context.set(tenant_id)
```
eisene, line 111: if we agree to handle things like this via globals (see my other comment about some people's feelings about globals), i would maybe not call this `tenant_context` but something like `execution_context` to set a precedent and give it a small crud file that handles sanity-checking things like "a tenant_id can't be set twice in the same process".

this would give us a place to put more globals and establishes the precedent for cheap.

CLAUDE.md
```diff
@@ -117,6 +117,7 @@ cd sdks/typescript && bun run tsc --noEmit
...
+- **Tenant isolation (`MULTI_TENANT`, default off).** When enabled, every DB session carries a request-scoped `app.tenant` GUC so row-level-security policies scope reads and writes to a single tenant. `tracked_db` is then the per-tenant path and is fail-closed — it raises before any query if no tenant is in scope (a passed `tenant_id` or an ambient `tenant_context`). `service_db` is the explicit RLS-bypass path for the legitimately cross-tenant work (the deriver's queue claim, the reconciler, the dreamer); never do per-tenant writes on it. A startup validator (`src/startup/tenant_isolation_validator.py`) refuses to boot on unsafe half-states (a transaction-mode pooler, RLS not enforced, or no service role). Off (the default, self-host), all of this is inert and honcho runs single-tenant on plain Postgres.
```
eisene, line 120: do we have a testing plan for this? deploy this to staging or something, or maybe its own small cluster, try out multiple tenants in there, validate db state?

src/deriver/queue_manager.py
```diff
@@ -791,6 +805,8 @@ async def process_work_unit(self, work_unit_key: str, worker_id: str) -> None:
                         break
 
             finally:
+                if tenant_token is not None:
+                    tenant_context.reset(tenant_token)
```
eisene, line 809: i think this means `publish_web_hook` just below runs without a tenant and it does call `construct_work_unit_key` - can you confirm this is intended?

src/utils/work_unit.py
```diff
@@ -14,9 +35,35 @@ class ParsedWorkUnit(BaseModel):
...
+    # tenants (workspace_name alone is not globally unique). Off, or with no tenant
+    # in scope (e.g. the tenant-less reconciler task), the key is unchanged.
+    #
+    # Pass tenant_id explicitly for callers that hold the tenant but run without it
+    # in ambient scope — e.g. the dreamer, which schedules per-collection work on
+    # the cross-tenant service session; otherwise the tenant is read from
+    # tenant_context.
+    # endregion
+    base_key = _construct_base_work_unit_key(workspace_name, payload)
+    if settings.MULTI_TENANT and (tenant := tenant_id or tenant_context.get()):
```
eisene: there's a potentially scary bug that would be easy to catch here. i believe `task_type` completely determines whether, if `settings.MULTI_TENANT` is set, whether the task *should* have a `tenant_id` or not. can we check that here?

i would put a `requires_tenant_id()` in some helper somewhere or something

eisene (approved):
a few comments

src/dreamer/dream_scheduler.py
eisene: tracing through this manually myself i feel like the `tenant_id` arg here is vestigial and can be removed, and we can rely on the `tenant_context` global everywhere. this makes sense to me - the tenant is global, set once, and threading it through just invites bugs. i feel that it is better to enforce its global uniqueness by just making it a global. having said that, i know some people just absolutely hate globals and i wouldn't go to war over this lol.

ptal, you're the expert here.

CLAUDE.md
eisene: no for staging until the JWT ticket is done? is the plan to defer live testing the multi-tenancy until then and only test that this didn't break anything with multi-tenancy off on staging? i'm ok with that but we should make the test plan explicit i think

src/startup/tenant_isolation_validator.py
philosaether: I was planning to leave it, even in OSS -- this validator is necessary to guarantee per-tenant data isolation, which I'm pretty sure even self-hosters will want if they're using the same instance to serve multiple tenants

CLAUDE.md
philosaether: We got [test plans for days](https://linear.app/plastic-labs/project/shared-tenants-2721ec5b9338/issues), that's why this PR is against a feature branch instead of main.

Talking to @ulyssepence today about how much of the A track to push for before we start the E track / staging tests, we still need to set up the test harness itself so it's kind of a moot point. But we need to test the kubernetes infrastructure itself, not just the app-level changes, so it's going to have to be thorough.

If you have specific bars you want to make sure we clear, please add them to the [Staging POC ticket](https://linear.app/plastic-labs/issue/DEV-2513/f2-staging-poc) and we'll include them in the Definition of Done.

src/startup/tenant_isolation_validator.py
philosaether: Plus, it's defense against a misconfiguration somewhere down the line

src/dependencies.py
philosaether: We're basically already doing that; `tenant_context` is set using the `contextvars` python lib in db.py, which is python's implementation of the global variable container pattern you're suggesting.

We could probably use a design session and maybe a refactoring pass on global variables in general; maybe contextvars is not the right solution for us, and even if it is we should probably have a dedicated location in code to set and handle them (the crud file you suggested). But I'd rather not do that as part of the migration; we have enough parts moving already.

Where's the comment you mentioned?

src/utils/work_unit.py
philosaether: Smart!

src/dreamer/dream_scheduler.py
philosaether: Ahh here's the other globals comment.

You right; removing the duplicate, relying on the global

src/deriver/queue_manager.py
philosaether: Yeah, that's an error. It doesn't get very far past this point before being caught; `publish_webhook_event` opens `tracked_db` with no tenant and fails.

Fixed it anyway

src/deriver/queue_manager.py
ulyssepence, line 369: These might need to be changed after `work_unit.py` changes

ulyssepence (commented):
Larger PRs mean that I just don't spend as much time looking at each line. I've been through it all, asking AI to explain things that I'm unclear on.

One general request: Can we use "PG config" instead of "GUC" in our documentation? I would guess few people know what GUC is (that includes myself).

src/deriver/queue_manager.py
philosaether: Already flagged for A3 based on 1136 :)

src/dependencies.py
eisene: yeah agreed - how do you feel about renaming `tenant_context` → `execution_context` so that future globals have an explicit home, and establishing this precedent here?

PR #1115, opened 2026-09-01
feat(api): Export deriver backlog as metrics from API endpoint

ulyssepence:
This takes some of the logic the deriver does to check if there's work to do and adds to it/copies it to the API endpoint to report metrics (`/metrics` (prometheus), `/deriver/metrics` (JSON)) on what work the deriver needs to do (if any).

The API endpoint now runs a read-only poller. Each pass reads the deriver's queue and the embedding tables. The query to check dreams runs less frequently (`DREAM.DUE_POLL_INTERVAL_SECONDS`).

Metrics:
- `deriver_outstanding_work_seconds` seconds of work waiting, 0 when there is none
- `deriver_queue_work_units_eligible` work units a deriver could claim now
- `deriver_queue_work_units_claimed` work units a deriver is working on
- `deriver_queue_items_pending` unprocessed deriver queue rows
- `deriver_queue_oldest_pending_age_seconds` age of the oldest unprocessed row
- `message_embeddings_pending` messages still waiting for a vector
- `embeddings_pending_due` messages still waiting for a vector that a deriver can try now
- `dreams_due` collections ready for a dream
- `deriver_metrics_last_success_timestamp_seconds` when metrics last updated (to know if they're current)

## Testing

Unit

```bash
$ uv run pytest tests/ -q --ignore=tests/live_llm --ignore=tests/bench
$ uv run basedpyright
$ uv run ruff check src/ tests/
```

Against E2E running:

```bash
$ curl -s localhost:8000/deriver/metrics
{"outstanding_work_seconds":60.0,"eligible_work_units":1,"claimed_work_units":0,
 "pending_items":1,"oldest_pending_age_seconds":72.0,"embeddings_pending":2,
 "embeddings_pending_due":1,"dreams_due":0,"measured_at":1788297441.07,
 "measurement_age_seconds":1.2}

$ curl -s localhost:8000/metrics | grep -E "^deriver_|^dreams_due"
deriver_outstanding_work_seconds{namespace="e2e"} 60.0
deriver_queue_work_units_eligible{namespace="e2e"} 1.0
deriver_queue_work_units_claimed{namespace="e2e"} 0.0
deriver_queue_items_pending{namespace="e2e"} 1.0
deriver_queue_oldest_pending_age_seconds{namespace="e2e"} 72.0
dreams_due{namespace="e2e"} 0.0
deriver_metrics_last_success_timestamp_seconds{namespace="e2e"} 1788297441.07
```

## Summary by CodeRabbit

* **New Features**
[...]

PR #1136, opened 2026-09-04
feat: API endpoint can run timer logic instead of deriver

ulyssepence:
## Description

This adds a setting `DERIVER.SCHEDULER` that be `api` or `deriver` that determines whether the deriver is responsible for running the timers for the deriver (that's how it works now) or the API endpoint is responsible. This will make it easier to turn the deriver on and off as it's needed to consume events in the queue.

The only difference in how the API endpoint will run the timers vs how the deriver currently does it is: the deriver cancels a scheduled dream if representation events come through the consumer for that particular collection. On the API endpoint, we approximate this by making a database call (`pending_representation_keys`).

This reuses the code that we added in https://github.com/plastic-labs/honcho/pull/1115 to surface these metrics in the API endpoint (`backlog.py`). It pulls out `cleanup_stale_work_units()` and `_maybe_cleanup_stale_work_units()` from existing deriver code.

## Proofs

```bash
uv run pytest tests/ -q --ignore=tests/live_llm --ignore=tests/bench
```

E2E tests with `DERIVER__SCHEDULER=api`:
- 4 representation rows processed, 6 explicit documents written
- 1 summary row processed
- sync_vectors rows enqueued by the API, consumed by the deriver
- Dream enqueued by the API, run by the deriver (5 deductive, 2 inductive docs; collection metadata advanced)
- Stale claim reaped by the API with the deriver stopped
- deriver/metrics JSON and all /metrics gauges served values
- Deriver log shows no ReconcilerScheduler start

## Summary by CodeRabbit

- **New Features**
  - Added configurable scheduling ownership between the API and deriver processes.
  - API-managed scheduling now queues due dreams during refreshes, with configurable per-poll limits and duplicate prevention.
  - Added automatic cleanup of stale work units.
  - Due-dream tracking includes scheduling details and excludes items with pending representation work.

- **Bug Fixes**
  - Prevented deriver timer scheduling and cleanup when scheduling is API-managed.
  - Prevented dreams from being scheduled while related representation work remains unprocessed.
  - Handled duplicate scheduling attempts without surfacing errors.

### DIFF
--- a/src/backlog.py
+++ b/src/backlog.py
@@ -1,4 +1,4 @@
-"""Read-only polling of the deriver's outstanding work. Schedules nothing."""
+"""Polling of the deriver's outstanding work. Schedules only when it owns scheduling."""
 
 import asyncio
 import contextlib
@@ -11,7 +11,8 @@ import sentry_sdk
 from src import crud, schemas
 from src.config import settings
 from src.dependencies import tracked_db
-from src.dreamer.dream_due import count_due_dreams
+from src.deriver.enqueue import enqueue_dream
+from src.dreamer.dream_due import DueDream, list_due_dreams
 from src.telemetry import prometheus_metrics
 
 logger = getLogger(__name__)
@@ -62,7 +63,8 @@ class DeriverMetricsPoller:
         self._shutdown_event: asyncio.Event = asyncio.Event()
         self._snapshot: DeriverMetricsSnapshot = DeriverMetricsSnapshot()
         self._next_dream_poll: float | None = None
-        self._dreams_due: int = 0
+        self._due_dreams: list[DueDream] = []
+        self._next_stale_cleanup: float | None = None
 
     @property
     def snapshot(self) -> DeriverMetricsSnapshot:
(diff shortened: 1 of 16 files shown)


src/backlog.py
```diff
@@ -144,3 +158,37 @@ def _dream_poll_due(self) -> bool:
         return (
             self._next_dream_poll is None or time.monotonic() >= self._next_dream_poll
         )
+
+    async def _enqueue_due_dreams(self) -> None:
+        for due_dream in self._due_dreams:
```
eisene, line 163: are you sure you want to enqueue all dreams right away, and not say the first 100 or something?

src/backlog.py
```diff
@@ -144,3 +158,37 @@ def _dream_poll_due(self) -> bool:
...
+                await enqueue_dream(
+                    due_dream.workspace_name,
+                    observer=due_dream.observer,
+                    observed=due_dream.observed,
+                    dream_type=due_dream.dream_type,
+                    session_name=due_dream.session_name,
+                    trigger_reason="document_threshold",
+                    delay_reason="api_poll",
+                )
+            except Exception as e:
```
eisene, line 174: will this page on a situation where it's trying to enqueue a dream already enqueued by another process?

eisene (dismissed):
2 small comments but lgtm

philosaether (dismissed):
LGTM

philosaether (approved):

PR #1140, opened 2026-09-05
fix(workspaces): return 404 on delete when Redis still holds a deleted workspace

erosika:
## Summary

`DELETE /v3/workspaces/{id}` returns 404 instead of 500 when the workspace row is already gone but Redis still holds the workspace.

## Bug fixed

The route resolved the workspace through `crud.get_workspace`, which reads the cashews-cached `_fetch_workspace`. The deriver cascade clears that key after commit, but production deriver pods run with `CACHE_ENABLED=false` (gitops `honcho-deriver.yaml`), so the clear lands in process memory and Redis keeps the entry until TTL. A later DELETE passed the existence check and the queue insert failed on `fk_queue_workspace_name`. The global handler returned 500. Sentry HONCHO-19W and HONCHO-19X hold about 380 of these since June, three per client call because the SDK retries 500.

## Route change

`delete_workspace` now checks Postgres directly with `crud.workspace_exists`. When the row is missing it evicts the cached entry and raises 404, so the next get-or-create recreates the workspace instead of returning the ghost. If the deriver drops the row between that check and the enqueue, the `ForeignKeyViolation` maps to the same 404 after a rollback. Other integrity errors still propagate.

## Tests

`tests/routes/test_workspace_delete_cache.py` covers the same-Redis path, the stale-Redis path with recreate after eviction, and the race path.

## Follow-up

Every cache invalidation that runs inside the deriver is still a no-op against Redis in production: peer card writes from the Dreamer and collection deletes. That needs a deployment or invalidation-path change, tracked in DEV-2646.

Fixes DEV-2646. Related: DEV-2439, DEV-1882.

## Summary by CodeRabbit

- **Bug Fixes**
  - Workspace deletion now returns a clear `404` response when the workspace does not exist.
  - Stale cached workspace data is removed during deletion, preventing outdated entries from being reused.
  - Handles deletion races reliably when a workspace is removed concurrently.

- **Tests**
  - Added coverage for missing workspaces, stale cache cleanup, concurrent deletion scenarios, and cache recreation after eviction.

### DIFF
--- a/src/crud/__init__.py
+++ b/src/crud/__init__.py
@@ -91,12 +91,14 @@ from .workspace import (
     WorkspaceStats,
     check_no_active_sessions,
     delete_workspace,
+    evict_workspace_cache,
     get_active_peers,
     get_all_workspaces,
     get_or_create_workspace,
     get_workspace,
     get_workspace_stats,
     update_workspace,
+    workspace_exists,
 )
 
 __all__ = [
@@ -188,6 +190,8 @@ __all__ = [
     "WorkspaceDeletionResult",
     "check_no_active_sessions",
     "delete_workspace",
+    "evict_workspace_cache",
+    "workspace_exists",
     "get_or_create_workspace",
     "get_workspace",
     "get_all_workspaces",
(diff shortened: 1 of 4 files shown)


akattelu (approved):

VVoruganti:
closing for staleness

PR #1144, opened 2026-09-07
fix(peer-card): refuse an over-cap card update instead of truncating it

### LINKED ISSUE #1143: [Bug] update_peer_card silently truncates an over-cap card, so a card at the cap can never take a new entry
### Deploy mode

Self-hosted

### Honcho version

server 3.1.1, self-hosted via Docker Compose

### Describe the bug

When `update_peer_card` receives more entries than `MAX_PEER_CARD_FACTS` (40), it cuts everything past the cap and still reports the update as successful (`src/utils/agent_tools.py`, the `if len(normalized_peer_card) > MAX_PEER_CARD_FACTS` branch).

That has three consequences:

1. **The model is never told.** The tool returns a success message, so the specialist has no way to learn that the marker it had just derived was dropped.
2. **What survives is decided by arrival order, not durability.** `[:MAX_PEER_CARD_FACTS]` keeps the head of whatever list the model happened to send.
3. **A card at the cap can never take a new entry again.** Each later run re-sends the surviving 40 entries plus its new marker, the surplus is cut, and the stored card stays byte-identical — indefinitely.

The same function already has the mechanism this case needs: structurally invalid entries are counted and reported back to the model through `_format_rejection_feedback`, which lets it self-correct on a retry. Only the length case is silent.

### Steps to reproduce

1. Let a peer card grow to 40 entries.
2. Have a specialist with `update_peer_card` in its loadout derive one more durable identity marker, so it calls the tool with 41 entries.
3. Observe: the log line `Peer card update exceeded max facts (40), truncating from 41 to 40`, a success result handed back to the model, and a stored card that is unchanged.

This also reproduces without a model: calling the tool handler directly with `MAX_PEER_CARD_FACTS + 1` structurally valid entries returns a success result and stores the truncated list.

### Logs and evidence

Observed on a self-hosted 3.1.1 deployment with a peer card sitting at the cap:

```
Peer card update exceeded max facts (40), truncating from 41 to 40
```

The stored card was byte-identical before and after the run, while the tool result reported a successful update.

### Additional context

The current test suite pins the truncating behaviour: `test_deduplicates_and_caps_peer_card` in `tests/utils/test_agent_tools.py` asserts `len(peer_card) == MAX_PEER_CARD_FACTS` after sending an over-cap list, so a fix has to change that expectation deliberately.

I have a fix ready — refuse the over-cap list as a whole, leave the stored card untouched, and return a result that states how many entries to free and how — and will open a PR referencing this issue.

dalzio:
## Description

`update_peer_card` truncated an over-cap list down to `MAX_PEER_CARD_FACTS` and still reported the update as successful, so the model was never told that the marker it had just derived had been dropped, and a card sitting at the cap could never take a new entry again. This refuses an over-cap list as a whole, leaves the stored card untouched, and returns a result that states how many entries to free and how — reusing the feedback path the function already has for structurally invalid entries, so the model can consolidate inside the tool loop it is already in.

The dreamer's peer-card prompt section and the `update_peer_card` tool description carry the matching rule, so the contract is stated where the model reads it rather than only enforced where it writes.

## Proofs

The tool result a specialist now gets when it sends 41 entries against a card at the cap:

```
Peer card not updated: 41 entries were sent but the cap is 40. The existing card is
unchanged. Free up at least 1 entry — merge entries that describe the same thing into
one, or drop the least durable ones — then call `update_peer_card` again with the
complete list.
```

Nothing is written on that path, and it returns before touching the database. When an over-cap list also carries structurally invalid entries, both problems are reported in one result.

Test suite on this branch, against Postgres with pgvector:

```
$ uv run pytest tests/utils/test_agent_tools.py tests/dreamer/ -q
168 passed in 13.17s
```

`ruff check`, `ruff format --check` and `basedpyright` are clean on the three changed files. `ruff check src/` still reports the 41 pre-existing findings on `main`; none of them come from this diff.

**One deliberate test change worth a reviewer's attention:** `test_deduplicates_and_caps_peer_card` asserted the truncating contract (`len(peer_card) == MAX_PEER_CARD_FACTS` after an over-cap send), so it is split rather than edited in place. Its dedup and normalization assertions live on in `test_deduplicates_and_normalizes_peer_card`, and the cap behaviour is now covered by `test_over_cap_update_is_refused_and_keeps_existing_card`, which reproduces the 41-entries-against-a-full-card case end to end.

Not included on purpose: no ranking or pinning of entries (unnecessary once the model can consolidate itself), and `MAX_PEER_CARD_FACTS` stays a constant rather than becoming a setting — both felt like separate discussions.

## Checklist

Fixes #1143

## Summary by CodeRabbit

- **Bug Fixes**
  - Peer card updates that exceed the 40-entry limit are now rejected instead of being silently truncated.
  - Existing peer card data remains unchanged when an update is rejected.
  - Update feedback now reports how many entries must be removed and recommends merging duplicates or removing less durable entries.
  - Guidance now helps consolidate oversized peer cards and retry successfully.
- **Documentation**
  - Added an unreleased changelog entry describing the corrected peer card update behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Qdrant vector store backend (`VECTOR_STORE_TYPE=qdrant`) as an optional `qdrant` extra (#683)
 
+### Fixed
+
+- `update_peer_card` truncated an over-cap list down to `MAX_PEER_CARD_FACTS` and still reported the update as successful, so the model was never told that the marker it had just derived had been dropped, and which entries survived came down to arrival order rather than durability — a card sitting at the cap could never take a new entry again. An over-cap list is now refused as a whole, the stored card is left untouched, and the tool result states how many entries to free and how (merge entries that describe the same thing into one, or drop the least durable ones), so the model can consolidate inside the tool loop it is already in. The dreamer's peer-card prompt section and the `update_peer_card` tool description carry the matching rule (#1144)
+
 ## [3.1.1] - 2026-09-02
 
 ### Changed
--- a/src/dreamer/specialists.py
+++ b/src/dreamer/specialists.py
@@ -127,7 +127,16 @@ For each legacy entry:
 
 When in doubt about a specific legacy entry, prefer migrating it (so valid info isn't lost) over dropping it. Splitting one dense legacy entry into multiple correctly-prefixed entries is fine and encouraged (e.g. a semicolon-separated `Tech Stack:` dump can become several `ATTRIBUTE:` lines, one per durable tool/platform).
 
-Call `update_peer_card` with the complete deduplicated list when there is a durable identity update to record, or when the existing card needs migration. Entries that do not start with one of the four allowed prefixes will be rejected. Keep concise (max 40 entries)."""
+Call `update_peer_card` with the complete deduplicated list when there is a durable identity update to record, or when the existing card needs migration. Entries that do not start with one of the four allowed prefixes will be rejected. Keep concise (max 40 entries).
+
+### When the card is at the cap
+
+A list longer than the cap is refused as a whole — nothing is written and the existing card stays as it was. So when the card is already full and you have a new durable marker, make room in the same call instead of appending:
+
+- Merge entries that describe the same thing into one entry. Several entries stating one mapping, tool, or rule each can often become one entry that carries the same information, as long as it stays a single concise marker within the per-entry length cap.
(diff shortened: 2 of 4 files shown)


dalzio:
The approved, updated implementation is now in #1228 on signed commit `2acbe5929125bee89ecc9ff876b53d49d49cdd27`. GitHub refused to reopen this PR, so #1228 is its successor on the same branch. Please use that PR for the current diff, verification evidence, and further review; this discussion remains part of the history. Thank you to everyone who helped move this forward.

Posted for GP by his AI assistant

PR #1148, opened 2026-09-08
Shared Tenants A3: tenant-scoped JWTs

philosaether:
# Tenant ID is in the Honcho Auth JWT now

On a shared deployment, all work needs an associated `tenant_id`. Async tasks read it from the work_unit key; synchronous tasks get it from the JWT.

This PR adds a `tn` field to the auth JWT in `security.py`, which expects a tenant id. Auth will fail when `MULTI_TENANT` is true and a tenant id is not present.

The PR also adds `tenant_id` handling to the cross-tenant enqueue_dream path. That's unrelated to the JWT; [honcho#1136](https://github.com/plastic-labs/honcho/pull/1136) added these code paths after [A2](https://github.com/plastic-labs/honcho/pull/1134) merged, so we're handling them here.

There are also tests.

Claude's description follows:

***

## What

Adds the tenant claim (`tn`) to honcho's JWTs and enforces tenant scoping at the auth boundary when `MULTI_TENANT` is on (DEV-2478).

- **`tn` claim on `JWTParams`** — additive and optional, so existing tokens are unaffected. Empty-string `tn` normalizes to absent, same as `w`/`p`/`s`.
- **Tenant gate in `auth()`** — under `MULTI_TENANT` every token must carry `tn`, enforced *before* the admin short-circuit: `ad: true` is admin **within its tenant**, never cross-tenant. Flag off: unchanged.
- **Per-request binding** — `require_auth` is now a yield-dependency: it binds `tenant_context` from the token for the life of the request and resets it on teardown; the DB checkout hook reads it to set `app.tenant` for RLS. There is no dependency-ordering hazard — session checkout is lazy and happens after the bind.
- **Cross-tenant scheduler paths** (reconciling the `DERIVER.SCHEDULER=api` work from #1136): the metrics poller and stale-work cleanup move to `service_db` (legitimately cross-tenant); due-dream enqueue binds each dream's tenant; the due-dream query, the pending-work suppression check, and dream cancellation are all tenant-keyed, so one tenant's activity can't inflate, suppress, or cancel another tenant's dreams.

## Flag-off guarantee

With `MULTI_TENANT` off (the default) behavior is unchanged: the gate, the request binding, and the tenant keying are all inert, and a still-valid token that happens to carry `tn` is accepted with the claim ignored (rows keep the default tenant).

## Verification

- Full suite green: 2557 passed flag-off; basedpyright 0/0/0.
- New coverage: the claim + gate matrix (admin/non-admin × flag on/off × `tn` present/absent), the `tenant_context` bind/reset lifecycle including teardown, member-read tenant threading, tenant-scoped dream cancellation, and a two-tenant due-dream isolation test (same-named workspace/peer triples in both tenants).
- A pre-merge review pass ran on the branch; the four flag-on issues it found are fixed in `23c8292f`, each with a regression test:
  1. `list_due_dreams`' document-count subquery joined tenant-blind — same-named triples across tenants merged (inflated counts; one tenant's `session_name` could reach another's dream).
  2. The member-read membership check opened `tracked_db` before the tenant was bound — every member read 500'd under the flag. Now passes the token's tenant explicitly (deliberately not `service_db`: the membership triple is unique only per tenant).
  3. `cancel_dreams_for_observed` matched pending dreams tenant-blind — one tenant's message activity could cancel another tenant's scheduled dream.
  4. The request bind keyed on `tn` presence instead of `MULTI_TENANT` — flag-off rows could be stamped with a token's tenant.
  5. (`61fefc55`, found reviewing the follow-on branch) `delete_workspace`'s active-queue-session cleanup matched work-unit keys tenant-blind — under the flag it missed this tenant's prefixed claim rows and could hit another tenant's un-prefixed reconciler key for a same-named workspace. Now matched by the claim's `tenant_id` + tenant-prefixed key position, failing closed without a tenant in scope.

## Out of scope (tracked)

- Mint-time tenant scoping (`POST /keys`, `scripts/generate_jwt.py`) — DEV-2488. Until it lands, keys minted by those paths are rejected under `MULTI_TENANT` (fail-closed, not a leak).
- e2e JWT-claim lockstep test against the token issuer — DEV-2687.
- Boot assertion refusing `MULTI_TENANT` without auth — backlog (worst case is a fail-closed outage, not exposure).


ulyssepence (approved):
LGTM

src/dreamer/dream_due.py
```diff
@@ -180,7 +213,10 @@ async def list_due_dreams(db: AsyncSession) -> list[DueDream]:
             work_unit_key not in newest_attempts
             or newest_attempts[work_unit_key] < newest_created_at
         )
-        and (due_dream.workspace_name, due_dream.observed) not in active_observed
+        and _observed_key(
```
akattelu, line 216: Are there any other call sites where we should be using this key that we haven't updated yet that will break?

akattelu (approved):

src/dreamer/dream_due.py
philosaether: I don't believe so, my agent hasn't found any, and if anything breaks anyway we will catch it in staging

PR #1156, opened 2026-09-09
Shared Tenants A4: tenant registry API

philosaether:
# What is this, an API for tenants?

This PR creates and exposes GET, POST, and DELETE endpoints for tenants. These endpoints can be used to get, create, or delete tenants.

It also adds a non-JWT auth method for calling these endpoints from groudon / any control plane used with OSS honcho.

Claude's description follows.

***

## What

Adds a tenant registry API so the control plane can provision — and roll back — tenants on a `MULTI_TENANT` deployment before any tenant-scoped credential or write exists.

- **`POST /v3/tenants`** `{tenant_id, vector_correlation_id?, tier}` — idempotent get-or-create: `201` created, `200` retry-with-identical-fields, `409` same id with different fields (this API never mutates an existing tenant).
- **`GET /v3/tenants/{id}`** — reconciliation read.
- **`DELETE /v3/tenants/{id}`** — provisioning rollback only: the FK topology (every tenant-scoped table references `tenants.tenant_id` with NO ACTION) refuses a populated tenant → `409`; the `default` bootstrap row → `422`. Cascading a populated tenant off is deliberately impossible through this API.
- New `TENANT_API` config section (`TENANT_API_SECRET`), `src/crud/tenant.py`, `src/routers/tenants.py`, schemas.

## The auth plane (the part to review hardest)

This router deliberately does **not** use `require_auth`: under `MULTI_TENANT` every JWT must carry a tenant claim, and at creation time the tenant doesn't exist to be claimed. It's guarded by `require_tenant_api` instead — an `X-Tenant-Api-Key` header compared constant-time (as bytes, so hostile non-ASCII input 401s rather than 500s) against `TENANT_API_SECRET`. Fail-closed twice over: the endpoints are disabled (405) unless `MULTI_TENANT` is on **and** the secret is configured, so a self-hosted single-tenant deployment never exposes tenant CRUD. The two planes never mix — `test_tenant_api_routes_use_their_own_auth_plane` machine-checks that exactly these routes carry `require_tenant_api` and none also carry `require_auth`.

## Verification

- Suite 2600 passed flag-off; basedpyright 0/0/0.
- 15 route tests: the disabled/unauthorized matrix (flag-off, secret-unset, missing/wrong/non-ASCII-bytes header), create idempotency (201/200/409/422-pattern), read, and all four delete outcomes including a seeded non-empty-tenant 409.
- Concurrency reviewed against the N-replica deployment: the create race collapses on the PK (no two-201 interleaving), and delete-vs-workspace-create is safe in both orderings via the FK's statement-time check.
- A pre-merge review pass ran on the branch; its fixes are folded in (bytes-compare hardening with regression test, typed `409` on the rare create-raced-delete tail, `GetOrCreateResult` return type, doc updates in CLAUDE.md covering the new auth plane).

## Out of scope (tracked)

- The control-plane caller (DEV-2487) and mint-time tenant claims (DEV-2488 / DEV-2578).
- Eviction / cascade delete of populated tenants (migration-track tooling, not this API).
- Tenant-row mutation (tier changes etc.) — ops-script territory by design.


src/schemas/api.py
```diff
@@ -119,6 +119,35 @@ def _validate_scope_name(name: str) -> str:
...
+        Field(min_length=1, max_length=128, pattern=RESOURCE_NAME_PATTERN),
+    ]
+    vector_correlation_id: str | None = None
+    tier: Annotated[str, Field(min_length=1, max_length=64)]
+
+
+class Tenant(BaseModel):
+    tenant_id: str
+    vector_correlation_id: str | None = None
+    tier: str
```
ulyssepence, line 145: Hmm I'm wondering if maybe we should leave these out of this PR if they're not being used anywhere? Like there's no wiring to use it to send to a vector db/embedding db, nor does the tier serve any business logic.

src/routers/tenants.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@router.delete("/{tenant_id}", status_code=204)
+async def delete_tenant(tenant_id: Annotated[str, Path()]) -> None:
+    """Delete an empty tenant (provisioning rollback): 409 if it has data."""
+    if tenant_id == DEFAULT_TENANT_ID:
+        raise ValidationException(
+            f"The {DEFAULT_TENANT_ID!r} tenant is the single-tenant bootstrap "
+            + "row and cannot be deleted"
+        )
```
ulyssepence, line 85: This seems a bit weird to me. Aren't we already verifying they have multi-tenancy on before this? Why do we need to guard against this? Do we have a built-in migration path for users who start on single-tenant and migrate to multi-tenancy?

src/config.py
```diff
@@ -778,6 +778,22 @@ def _require_jwt_secret(self) -> "AuthSettings":
         return self
 
 
+class TenantApiSettings(HonchoSettings):
```
ajspig, line 781: Good with not documenting elsewhere.

src/schemas/api.py
philosaether: The [groudon half](https://github.com/plastic-labs/groudon/pull/539) sends tier and tenant id (which we map to correlation id if there's no preexisting hch* app name for this tenant yet). We'll need them both for business logic later in the stack; this is just plumbing.

src/routers/tenants.py
philosaether: There's a pretty nasty error scenario where a tenant flips MULTI_TENANT on, deletes the default tenant row, then flips it off again. Since all the honcho tables would FK to a dead row from that point onwards, it's a permanent write-block with a non-obvious cause. Their instance just bricks.

ulyssepence: That's what my AI said too but what I'm wondering is are we actually planning for allowing users to do this? Do we have other safeguards in place to migrate between these?

src/schemas/api.py
ulyssepence: I guess my question was are we planning to hook this into anything in OSS? If so, we should set up first. This just seems like dead code otherwise.

src/routers/tenants.py
ulyssepence: I'm mostly ok keeping this one in though. It's more of a nit.

ulyssepence (approved):

src/schemas/api.py
philosaether: Nope, no use for OSS at all. But we decided to keep it in the OSS package anyway, since we do need it for shared tenants and there's only one honcho.

It bothers me too tbh, but it's necessary ugliness

src/routers/tenants.py
philosaether: A user absolutely could hit that error scenario. Honestly I could see myself hitting it if I just tried to mess around with my local install. 

"Hey, what's this config setting do?" -> "Hey, why is this row in the database?" -> "Hey, why is my app broken?"

Feels realistic to me lol

PR #1157, opened 2026-09-09
fix(cache): upgrade redis-py to 8.x for the async cluster connection leak

eisene:
## Description

`redis.asyncio.cluster.NodesManager.initialize()` published a fresh generation of `ClusterNode` objects as `slots_cache` while `set_nodes(..., remove_old=True)` kept the old ones in `nodes_cache`, so routing used objects that nothing would ever close — `ClusterNode.__del__` is synchronous and can only warn. Every topology re-init orphaned a generation of live sockets, released only by process restart, which is the connection ratchet described in DEV-2442. The fix is upstream in [redis/redis-py#4057](https://github.com/redis/redis-py/pull/4057) and shipped in **8.0.0 only**, so the previous `redis>=7.0.0,<8.0.0` pin structurally excluded it.

This bumps the pin to `>=8.0.1,<9.0.0` (resolves to 8.1.0). It is an asyncio-cluster-only bug; the sync client was never affected.

## Proofs

**The fix is present in the resolved version** (`.venv/.../redis/asyncio/cluster.py:2389`):

```python
# Rebuild the slots cache with the preserved nodes_cache instances
# so existing per-node connection pools stay in use after refresh.
slot_nodes = [self.nodes_cache[node.name] for node in nodes]
self.slots_cache = new_slots_cache
```

- buggy `self.slots_cache = tmp_slots` — gone
- `set_nodes` now runs *before* the slots cache is published

**Only `redis` moves in the lock** — 7.4.0 → 8.1.0, no transitive churn. Two files changed.

**API surface** — honcho touches three things, all intact:

| Used at | Symbol |
|---|---|
| `src/cache/client.py:11` | `redis.exceptions.TimeoutError` / `ConnectionError` |
| `tests/test_cache_key_namespace.py:10` | `redis.crc.key_slot` |
| — | hash-tag equivalence: `key_slot("{honcho}:v2:workspace:a") == key_slot("{honcho}:v2:workspace:b") == key_slot("honcho")` |

That last one is what the namespace hash-tagging from #1058 rests on.

**Checks**

```
pytest tests/test_cache_key_namespace.py tests/test_cache_redaction.py   24 passed
ruff check src/          40 findings, all pre-existing, none in src/cache/
basedpyright src/        95 diagnostics, 0 redis/cashews/cache-related
```

The 8.x `@overload` split for sync-vs-async return types produced no new type errors here.

**Live round-trip against a real Redis 8.2 server** — RESP3 is now the wire default and cashews stores SQLALCHEMY-pickled values, so this was the main risk:

```
key: {upgradecheck}:v2:workspace:acme
round-trip identical: True    datetime preserved: True
subtree present     : [True, True, True]
after delete_match  : None    [None, None, None]
after safe_delete   : None
lock acquired       : True
```

Covers the pickler, tz-aware datetimes, the `delete_match` glob sweeping the workspace+peer+session+collection subtree, exact-key delete, and the `@cache.locked` path. Also verified working on a local docker fleet.

**Breaking changes in 8.x are already neutralised**

| Change | Why it's a no-op here |
|---|---|
| `socket_timeout` defaults to 5s | `CACHE_URL` already sets `socket_timeout=5` |
| pool `max_connections` → 100 | cashews does `setdefault("max_connections", 10)` |
| RESP3 on the wire by default | legacy response shapes preserved; verified by the round-trip above |
| Python >=3.10 required | `requires-python = ">=3.13"` |

`cashews[redis]==7.5.0` declares `redis>=4.3.1` with **no upper bound**, so the exact cashews pin is not a constraint.

## Not covered by this PR

**Real Redis Cluster behaviour is unverified.** The leak only manifests on topology re-initialization across multiple nodes, and this was tested against a standalone server. The code path is verified by reading, and #4057 added node-identity assertions after topology refresh, but leak elimination is not proven at runtime. Worth watching `node/clients/connected_clients` across an induced re-init in staging before relying on it for capacity planning.

Two related findings from the same investigation, both out of scope here:
- DEV-2647's stated mechanism attributes the leak to replica reads. It isn't replica-dependent — `get_node_from_slot` returns `slots_cache[slot][0]`, the primary, on every path. The prod cluster also has `replicaCount: 0`.
- The prod per-node ceiling is `maxclients = 16000` and connections are skewed ~2.8:1 across the three nodes, so the binding constraint is the hot node, not the 48,000 aggregate.

## Checklist

Fixes DEV-2647. Related: DEV-2442, DEV-2646.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redis cluster compatibility with Redis 8.x, preventing initialization failures during connection retries.
  * Redis cluster connections now handle certain network and startup errors more gracefully.

* **Chores**
  * Updated the supported Redis version range to Redis 8.x, requiring version 8.0.1 or later and excluding version 9.x.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
     "typing-extensions>=4.11.0",
     "json-repair>=0.49.0",
     "turbopuffer>=1.8.1",
-    "redis>=7.0.0,<8.0.0",
+    "redis>=8.0.1,<9.0.0",
     "cashews[redis]==7.5.0",
     "scikit-learn>=1.6.0",
     "prometheus_client>=0.21.0",
--- a/src/cache/client.py
+++ b/src/cache/client.py
@@ -2,13 +2,16 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import socket
 from typing import Any, cast
 from urllib.parse import urlparse, urlunparse
 
 import sentry_sdk
 from cashews import cache
+from cashews.backends.redis.client import SafeRedisCluster
 from cashews.picklers import PicklerType
 from redis import exceptions as redis_exc
+from redis.asyncio import RedisCluster
 from tenacity import (
     AsyncRetrying,
     retry_if_exception_type,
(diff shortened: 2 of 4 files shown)


VVoruganti (dismissed):

eisene:
---

## Update: the pin alone was not safe — shim added (`126cadbd`)

Reviewing whether cashews tolerates redis 8.x turned up a hard incompatibility, so this PR now carries a compatibility shim in `src/cache/client.py` plus `tests/test_cache_cluster_shim.py`.

**cashews 7.5.0 declares `SafeRedisCluster.initialize(self)`.** redis-py [#4060](https://github.com/redis/redis-py/pull/4060) added `additional_startup_nodes_info` and `last_failed_node_name` to `RedisCluster.initialize()` in 8.0.0, and redis-py calls it with a keyword on its command-retry path:

```python
for _ in range(execute_attempts):
    if self._initialize:
        await self.initialize(last_failed_node_name=last_failed_node_name)
```

`_initialize` is set by any `ConnectionError`/`TimeoutError`, so one unreachable node triggers it. Two things make it worse than ordinary signature drift:

- **`TypeError` is not in the tuple cashews catches** (`RedisError`, `socket.gaierror`, `OSError`, `asyncio.TimeoutError`), so it bypasses `SafeRedisCluster`'s entire purpose and propagates into request handling instead of degrading.
- **`initialize()` is itself the recovery path**, so `_initialize` is never cleared and every later command fails too. The client stays wedged until the process restarts.

### Reproduced against a real cluster

3× `redis:8.2` in cluster mode (3 masters, 0 replicas — same shape as prod), cashews 7.5.0, redis-py 8.1.0. Steady state worked; then the master owning the test key's slot was killed:

```
Traceback (most recent call last):
  File "/app/repro.py", line 50, in main
    await client.set(f"{{ns}}:k{attempt}", b"x")
  File ".../cashews/backends/redis/client.py", line 87, in execute_command
    return await super().execute_command(*args, **kwargs)
  File ".../redis/asyncio/cluster.py", line 1167, in execute_command
    await self.initialize(last_failed_node_name=last_failed_node_name)
TypeError: SafeRedisCluster.initialize() got an unexpected keyword argument 'last_failed_node_name'
```

**15 of 15 attempts.** Note the middle frame — the exception passes straight through cashews' wrapper, the code whose job is to absorb backend failures. With the shim, the same node kill produces cashews' intended behaviour instead: `redis: can not execute command: SET` logged and `None` returned, so callers fall through to Postgres. 0 TypeErrors.

### Why CI and local stacks missed it

Only reachable with `CACHE.CLUSTER` enabled **and** only after a node failure. Standalone is genuinely unaffected — `Redis.initialize(self)` still takes no arguments in 8.x — so the break is invisible in local docker stacks and in CI, both of which run a single standalone Redis. That is also why the first commit went green everywhere.

### There is nothing to upgrade to

- cashews 7.5.0 (2026-03-02) is the latest release and **predates redis-py 8.0.0 (2026-05-28)**.
- cashews `master` is byte-identical to 7.5.0; last commit is the 7.5.0 release itself.
- It declares `redis>=4.3.1` with **no upper bound**, which is why resolution succeeded silently.
- No 8.x release avoids it: the new signature and the keyword call site are in 8.0.0, 8.0.1 and 8.1.0 alike.
- Unreported upstream at the time of writing.

### Removal condition

`tests/test_cache_cluster_shim.py::test_shim_is_still_required` fails once the installed cashews stops declaring the narrow override — that is the cue to delete the shim. The block is marked `REMOVE THIS` with the same condition in a comment.

One incidental change: the pre-commit ruff hook collapsed a redundant `asyncio.TimeoutError`/`TimeoutError` pair in `init_cache`'s except tuple (UP041, pre-existing). Same object on Python 3.11+, so it is a no-op; the hook reapplies it on any commit touching this file.

### Still not covered

Leak elimination is still unproven at runtime — verified by reading the 8.x code path, and upstream #4057 added node-identity assertions, but nothing was measured. Worth watching `node/clients/connected_clients` across an induced re-init in staging before relying on it for capacity planning.

akattelu (approved):

PR #1139, opened 2026-09-04
feat(mcp): expose scope, workspace level chat, and pagination in the MCP server

ajspig:
## Description

The bundled MCP server was pinned to `@honcho-ai/sdk@2.2.0`, so none of the 2.4.0 recall-boundary, paging, or workspace-chat options were reachable from an MCP client: `list_sessions` and `get_session_messages` could only return the first page, `chat` could only be confined to a single `session_id`, there was no way to see a workspace's scopes, and there was no workspace-level dialectic.

This bumps the SDK to 2.4.0 and surfaces those options as tools/params. `chat` gains optional `scope` (a scope name, or a list of names as an explicit-only allowlist) and `sessions` (an ad hoc session-ID allowlist), and its description now states it answers about one peer. A new `workspace_chat` tool asks a reasoned question across all peers (optional `session_id`, `scope`, `reasoning_level`). `list_sessions` / `get_session_messages` accept `page`, `size`, `reverse`. Two new read-only tools, `list_scopes` and `get_scope_sessions`, live in `mcp/src/tools/scopes.ts`. `mcp/README.md` and `mcp/instructions.md` document the additions.

## Proofs

* `cd mcp && bun install --frozen-lockfile && bun run tsc --noEmit` → clean (exit 0) against `@honcho-ai/sdk@2.4.0`.
* Smoke-tested over stdio against [api.honcho.dev](http://api.honcho.dev): `list_scopes` / `get_scope_sessions` return the scopes provisioned in the `claude-code-scopes` workspace, and `chat` with `scope=<name>` vs. `sessions=[...]` returns different, boundary-respecting answers to the same query.
* Smoke-tested `workspace_chat` over stdio against the committed sandbox (`sandbox/sandbox.sh up`, mock provider): plain / `session_id` / `scope` (string) / `scope` (list) all returned a synthetic answer with `isError` unset and hit `POST /v3/workspaces/sandbox/chat` 200; a nonexistent scope returned `Workspace chat failed: Scope no-such-scope not found…` (404); peer `chat` still routed to `POST .../peers/assistant/chat`. Against a pre-v3.1.0 local image, `workspace_chat` 404s cleanly and peer tools keep working.
* No test suite exists under `mcp/` (the package has no `test` script); the SDK contract is enforced by the typecheck above.

## Checklist

- [X] This PR is correlated to an existing issue, and I understand it will be closed if that issue does not have the `maintainer-approved` label.
[...]

PR #1161, opened 2026-09-09
feat(mcp): add scope membership tools

ajspig:
## Description

The MCP server could read scopes (`list_scopes`, `get_scope_sessions`, `scope` on `chat`) since #1139, but had no way to create one or change its
 membership, so an MCP client could only use recall boundaries someone had provisioned through the SDK. This adds `create_scope` (get-or-create with optional metadata), `add_sessions_to_scope` (max 100 per call), `remove_session_from_scope`, and `get_scope_status` (per-session backfill state with pending/completed/failed counts) in `mcp/src/tools/scopes.ts`.

It also makes the read/remove tools reference an existing scope directly instead of going through the SDK's get-or-create `honcho.scope()`, so `get_scope_sessions` on a missing scope now returns the server's 404 instead of silently creating the scope. `mcp/README.md`, `mcp/instructions.md`, and `mcp/CHANGELOG.md` document the additions.

The immediate consumer is the shareable Scopes Replay artifact, which uses these tools to write Claude-proposed scopes back to the viewer's own workspace through the mcp.honcho.dev connector.

## Proofs

* `cd mcp && bunx tsc --noEmit` → clean (exit 0).
* Smoke-tested every scope tool over stdio (`bun src/stdio.ts`) against api.honcho.dev in a throwaway workspace (deleted afterwards):
  * `create_scope` with `{label, desc, queries}` metadata → `{"id":"smoke-scope","metadata":{...},"created_at":...}`; calling it again without metadata returned the existing scope with its metadata intact (idempotent).
  * `add_sessions_to_scope` with two sessions → `{"scope_id":"smoke-scope","added":2}`; `get_scope_sessions` then listed both.
  * `get_scope_status` right after adding a session with messages → `{"pending":1,"completed":0,"failed":0,"sessions":{"smoke-s1":{"state":"pending",...}}}`.
  * `remove_session_from_scope` → `{"scope_id":"smoke-scope","removed":"smoke-s2"}`; `get_scope_sessions` then listed one.
  * `add_sessions_to_scope` with a nonexistent session → `isError` with `Session(s) ['does-not-exist'] not found in workspace ...`.
  * `get_scope_sessions` / `remove_session_from_scope` on a nonexistent scope → `isError` with `Scope no-such-scope-2 not found in workspace ...` (previously this would have created the scope).
* No test suite exists under `mcp/` (the package has no `test` script); the SDK contract is enforced by the typecheck above.

## Checklist

  * No dedicated issue; follows #1139 (same surface, same author with write permission, which the issue gate treats as exempt).

## Summary by CodeRabbit

- **New Features**
  - Added tools to create scopes, add or remove sessions, and check asynchronous session backfill status.

- **Bug Fixes**
  - Operations targeting an existing scope no longer create missing scopes automatically; unavailable scopes return a not-found response.
  - Invalid scope identifiers are rejected before requests are sent.

- **Documentation**
  - Updated available tools, session tools, architecture notes, and changelog documentation to describe scope management capabilities.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -36,7 +36,7 @@ Every workspace-scoped tool takes a `workspace_id` argument. If you set `X-Honch
 
 **Sessions:** `create_session`, `list_sessions`, `delete_session`, `clone_session`, `add_peers_to_session`, `remove_peers_from_session`, `get_session_peers`, `inspect_session`, `add_messages_to_session`, `get_session_messages`, `get_session_message`, `get_session_context`
 
-**Scopes:** `list_scopes`, `get_scope_sessions`
+**Scopes:** `list_scopes`, `get_scope_sessions`, `create_scope` (get-or-create with optional metadata), `add_sessions_to_scope`, `remove_session_from_scope`, `get_scope_status` (backfill progress per session)
 
 **Conclusions:** `list_conclusions`, `query_conclusions`, `create_conclusions`, `delete_conclusion`
 
@@ -55,7 +55,7 @@ src/
   tools/
     workspace.ts        # inspect, list, search, workspace_chat, metadata
     peers.ts            # CRUD, chat (session / scope / sessions recall bounds), card, context, representation
-    scopes.ts           # list scopes, list a scope's sessions
+    scopes.ts           # list/create scopes, scope membership, backfill status
     sessions.ts         # CRUD, peers, messages, inspect, context, clone
     conclusions.ts      # list, query, create, delete
     system.ts           # dream, queue status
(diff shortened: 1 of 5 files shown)


mcp/README.md
```diff
@@ -36,7 +36,7 @@ Every workspace-scoped tool takes a `workspace_id` argument. If you set `X-Honch
 
 **Sessions:** `create_session`, `list_sessions`, `delete_session`, `clone_session`, `add_peers_to_session`, `remove_peers_from_session`, `get_session_peers`, `inspect_session`, `add_messages_to_session`, `get_session_messages`, `get_session_message`, `get_session_context`
 
-**Scopes:** `list_scopes`, `get_scope_sessions`
+**Scopes:** `list_scopes`, `get_scope_sessions`, `create_scope` (get-or-create with optional metadata), `add_sessions_to_scope`, `remove_session_from_scope`, `get_scope_status` (backfill progress per session)
```
akattelu, line 39: I think we should also support specifying a scope when creating a session, but that can come separately

akattelu (approved):

PR #1164, opened 2026-09-10
fix(harness-plugin-core): ship a compiled dist/ so the package loads under Node and tsc

ajspig:
## Description

`@honcho-ai/harness-plugin-core` 0.1.0 pointed `main`/`exports` at `src/index.ts`, so it only imported under runtimes that transpile TypeScript on the fly (bun, jiti, esbuild). Plain Node refuses to strip types under `node_modules` (`ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`), and `tsc` consumers using `nodenext` resolution failed inside this package's extensionless relative imports (`TS2835`). That blocks dsh-honcho, openclaw-honcho, and the sillytavern proxy.

This ships a build, following the same shape as `@honcho-ai/sdk`, dsh-honcho, and openclaw-honcho:

- One `tsconfig.json` with `outDir`/`rootDir`/`declaration`; `tsc` builds ESM + `.d.ts` into `dist/`, `tsc --noEmit` typechecks, `prepublishOnly` builds on publish
- `main` / `types` / `exports` point at `dist/`; `dist` is in `files`; version bumped to 0.1.1
- Relative imports carry `.js` extensions so the source is valid under any module resolution
- `env` parameters are typed as an exported `Env` alias (`Record<string, string | undefined>`) instead of `NodeJS.Dict<string>`, so the shipped declarations don't require consumers to have `@types/node` installed (the new test caught this)
- README install line spells out both the bun and npm commands

## Proofs

`tests/dist.test.ts` runs `bun run build`, `npm pack`, extracts the tarball into a throwaway consumer, then:

- imports it from a plain `.mjs` under `node` and checks the output
- type-checks a `main.ts` importing it under `moduleResolution: NodeNext` and asserts zero diagnostics

```
$ bun run typecheck   # clean
$ bun test
 15 pass
 0 fail
```

`npx @arethetypeswrong/cli --pack .` reports green for `bundler` and `node16 (from ESM)`; 0.1.0 reported an internal resolution error for `node16`.

0.1.1 is being published directly from this branch to unblock the three plugins, rather than waiting for merge; `prepublishOnly` builds `dist/` automatically.

## Fixes

Unblocks the telemetry-header integrations for dsh-honcho, openclaw-honcho, and the sillytavern proxy.


akattelu (dismissed):
You should address the coderabbit comment, and also same note about bumping the package in the same PR as making the feature

akattelu (approved):

PR #1165, opened 2026-09-10
fix(llm): surface provider outages as 503 instead of 500

eisene:
## Why

Rootly **BVto3L** / Sentry **HONCHO-4C3**. Three `POST /v3/workspaces/{ws}/peers/{peer}/chat` requests returned `500 {"detail": "An unexpected error occurred"}` while a `tentacruel-litellm` pod restarted underneath them and its proxy answered 502 for ~39s.

The 502s are a platform problem, fixed separately. This PR is about the fact that honcho turned a transient upstream outage into an opaque 500 the caller could not act on.

Three links in that chain:

1. The tenacity retries in `tool_loop.py` were configured without `reraise=True`, so an exhausted budget raised `RetryError` instead of the provider's own exception — the cause was unreachable and every provider outage collapsed into one Sentry issue titled `RetryError[<Future at 0x… state=finished raised InternalServerError>]`.
2. Nothing catches `RetryError`, so it fell through to the generic `Exception` handler in `main.py`.
3. A 500 tells the caller "your request is broken." A 503 tells them "retry." Only one of those is true here.

## What

- **`UpstreamLLMError`** (503) in `src/exceptions.py`.
- **`src/llm/errors.py`** — translates provider transport failures. The three SDKs share no base class, so this reads the two shapes they do share: a connection-error class per SDK (timeouts subclass it; google-genai surfaces raw `httpx`), and a status on the instance (`status_code` for anthropic/openai, `code` for google).
- **`request_builder.execute_completion`** applies it — the single choke point every non-streaming backend call passes through.
- **`reraise=True`** on all three `retry(...)` sites in `tool_loop.py`. `src/cache/client.py` already does this; the LLM path was the outlier.

4xx is deliberately left alone, including 429 — callers throttle on that rather than retry blindly, and laundering it into a 503 would hide it.

## Scope

The streaming path is **not** covered. `execute_stream` splits setup failures (retryable) from drain failures (deliberately not, since chunks may already be on the wire), and reworking that deserves its own change rather than riding along on an incident fix. The incident was entirely on the non-streaming path.

## Verification

`tests/llm/test_upstream_errors.py`, 17 cases: a parametrized mapping table across all three SDKs' status and connection errors, the pass-through cases, and two end-to-end regressions.

Both regressions were confirmed to fail with the fix reverted, reproducing the production signature exactly:

```
E  tenacity.RetryError: RetryError[<Future at 0x112277380 state=finished raised UpstreamLLMError>]
FAILED tests/llm/test_upstream_errors.py::test_execute_completion_translates_a_proxy_502
FAILED tests/llm/test_upstream_errors.py::test_exhausted_retries_surface_the_provider_error_not_retry_error
```

- `pytest tests/llm/` — 317 passed
- full suite on pre-push — 2579 passed
- `ruff check` / `ruff format` / `basedpyright` — clean on all changed files

## Related

- `plastic-labs/tentacruel` — release that removes the 502 source
- `plastic-labs/argo-production` — drain + port
- `plastic-labs/argo-source` — metrics scrape port

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of language-model provider outages.
  * Provider connection failures and server-side errors now surface as clear service-unavailable errors.
  * Retry exhaustion now preserves the original provider error instead of displaying a generic failure.
  * Client-side errors continue to be reported unchanged.
  * Related upstream outages are now grouped consistently in error monitoring with appropriate severity.

### DIFF
--- a/src/exceptions.py
+++ b/src/exceptions.py
@@ -141,6 +141,19 @@ class RepresentationSaveError(HonchoException):
     detail: str = "Representation save failed for all observers"
 
 
+@final
+class UpstreamLLMError(HonchoException):
+    """Raised when the upstream model provider is unreachable or failing.
+
+    Distinct from a bug in our own request: the provider (or the proxy in front
+    of it) returned 5xx or refused the connection, so the caller should retry
+    rather than treat the request as malformed.
+    """
+
+    status_code = 503
+    detail = "Upstream language model provider is unavailable"
+
+
 class LLMError(Exception):
     """Exception raised when an LLM call fails.
 
--- a/src/llm/api.py
+++ b/src/llm/api.py
@@ -281,6 +281,10 @@ async def honcho_llm_call(
             stop=stop_after_attempt(retry_attempts),
             wait=wait_exponential(multiplier=1, min=4, max=10),
             before_sleep=before_retry_callback,
+            # Surface the provider's own error once the budget is spent.
+            # Without this tenacity raises RetryError, which erases the cause
+            # and lands every outage on the generic 500 handler.
+            reraise=True,
         )(decorated)
 
     def _trace_thinking_budget() -> int | None:
@@ -394,6 +398,9 @@ async def honcho_llm_call(
                     stop=stop_after_attempt(retry_attempts),
                     wait=wait_exponential(multiplier=1, min=4, max=10),
                     before_sleep=before_retry_callback,
+                    # Same as above: the toolless path is what the deriver
+                    # uses, so an outage here must stay legible too.
+                    reraise=True,
                 )(wrapped)
             result: (
                 HonchoLLMCallResponse[Any]
--- a/src/llm/errors.py
+++ b/src/llm/errors.py
@@ -1,0 +1,77 @@
+"""Translate provider SDK transport failures into Honcho's own error type.
+
+Every backend call funnels through `src.llm.request_builder`, so this is the one
+place that has to know what a provider outage looks like across the three SDKs.
+They do not share a base class, but they split failures the same two ways:
+
+* the request never got a response -- each SDK has its own connection-error
+  class (timeouts subclass it), and google-genai surfaces raw httpx errors;
+* the request got a response carrying a server-side status -- anthropic and
+  openai expose it as `status_code`, google-genai as `code`.
+"""
+
+from __future__ import annotations
+
+import anthropic
+import httpx
+import openai
+
+from src.exceptions import UpstreamLLMError
+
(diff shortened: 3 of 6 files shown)


eisene:
## Update: extended after Rootly 3WyMAJ

A second alert (**3WyMAJ**, `2026-09-11T16:17Z`) hit the same tentacruel 502s through a different path, and showed this PR was incomplete.

```
transaction: minimal_deriver_batch          handled: no
src/llm/api.py:367 in _toolless_call
src/llm/executor.py:561 in honcho_llm_call_inner
src/llm/request_builder.py:170 in execute_completion
src/llm/backends/openai.py:211 in complete
```

The deriver calls `honcho_llm_call` with no tools, so it goes through `_toolless_call` — which has **its own pair of `retry(...)` sites in `api.py:280` and `api.py:393`** that the tool-loop fix did not touch. Those kept raising bare `RetryError`.

Now fixed: all five tenacity sites in `src/llm/` set `reraise=True`.

Added two tests, both confirmed failing with `api.py` reverted:

- `test_toolless_path_also_surfaces_the_provider_error` — drives `honcho_llm_call` down the toolless path.
- `test_every_llm_retry_site_sets_reraise` — an AST walk over `src/llm/` asserting no `retry(...)` omits `reraise`. With the fix reverted it names exactly the two lines I missed:

  ```
  AssertionError: tenacity retry() without reraise=True: ['src/llm/api.py:280', 'src/llm/api.py:393']
  ```

`tests/llm/` now 319 passed; full suite green on pre-push.

Worth noting for reviewers: on the deriver path `enable_retry` is off, so the raw `openai.InternalServerError` was escaping unhandled rather than as a `RetryError` — the `execute_completion` translation is what covers that case, and it turns it into `UpstreamLLMError`. Whether the queue should treat that as a retryable work unit is a separate question, not addressed here.

src/exceptions.py
```diff
@@ -141,6 +141,19 @@ class RepresentationSaveError(HonchoException):
...
+    """Raised when the upstream model provider is unreachable or failing.
+
+    Distinct from a bug in our own request: the provider (or the proxy in front
+    of it) returned 5xx or refused the connection, so the caller should retry
+    rather than treat the request as malformed.
+    """
+
+    status_code = 503
+    detail = "Upstream language model provider is unavailable"
+
```
Rajat-Ahuja1997, line 187: note that our `default_before_send` sentry configuration filters out all `HonchoExceptions` from being sent to sentry. is that what we want here?

Rajat-Ahuja1997 (approved):

PR #1166, opened 2026-09-10
Populate CloudEvents Instrumentation

VVoruganti:
…LLM traces

Trace events could not be grouped by conversation: session_name is not unique across workspaces, and the background producers emitted no session at all. Thread the canonical Session.id from the queue through the deriver, summarizer and dream specialists so every trace carries it, falling back to a lookup only for callers that don't already have one.

Widen the trace envelope with the identity and execution context needed to attribute a call without joining another stream:

- workspace_name, session_id, observers, observed, agent_type, track_name
- source_message_ids and queue_item_ids for the work that triggered the call
- duration_ms, outcome, error_class, retry_attempts, is_final_attempt, was_stream and effective_max_output_tokens, recorded per provider attempt
- system_prompt_ref now covers the whole system prompt (the lone system message, or the concatenation of several) alongside system_prompt_refs
- output_reasoning_ref for normalized provider reasoning details

Observers are carried as an array only; there is no scalar observer on the traced event, because a batched derivation has no single one and a joined string would name no peer. The scalar stays on LLMTelemetryContext for the tool-loop path, whose agents have exactly one observer by construction.

Streamed calls previously reported almost nothing: the capture finalizer ran in the response wrapper, outside the executor, so setup failures, interrupted streams and output token counts were lost. Fold capture into honcho_llm_call_inner and close the provider generator explicitly, so a drained, cancelled, errored or closed stream each records exactly once at the provider boundary, preserving the output received before interruption.

LLMCallTracedEvent and EmbeddingCallTracedEvent go to schema version 2; the new fields are nullable so v1 archives still load. trace.content is unchanged.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - CloudEvents telemetry traces now include session, workspace, observer, source, stream, retry, timing, outcome, and token-limit details.
  - LLM traces support references to system prompts and reasoning content.
  - Embedding traces now report run identity, duration, retry status, and outcomes.
  - Trace data preserves partial streaming output and distinguishes successful, failed, and cancelled calls.
  - Trace records now correlate activity with source messages and queue items.

- **Documentation**
  - Added configuration guidance for CloudEvents trace payloads, including schema version 2 fields and content references.

### DIFF
--- a/docs/v3/contributing/configuration.mdx
+++ b/docs/v3/contributing/configuration.mdx
@@ -642,6 +642,42 @@ TELEMETRY_MAX_RETRIES=3
 TELEMETRY_MAX_BUFFER_SIZE=10000
 ```
 
+#### Trace payloads
+
+Enable content-addressed LLM traces alongside telemetry:
+
+```bash
+TELEMETRY_ENABLED=true
+TELEMETRY_TRACE_PAYLOADS_ENABLED=true
+# Optional: an empty list captures all call purposes.
+TELEMETRY_TRACE_PURPOSES='["dialectic.answer", "deriver.representation", "summary.short", "summary.long"]'
+```
+
+`llm.call.traced` and `embedding.call.traced` use schema version 2. LLM traces
+record each provider attempt, including retries, fallback calls, stream setup
+failures, and interrupted streams. Stream traces preserve the output received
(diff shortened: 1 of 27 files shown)


eisene (dismissed):
lgtm

eisene (approved):
lgtm

PR #1168, opened 2026-09-10
Shared Tenants A4.5: QueueItemBatch claim aggregate

philosaether:
# One queue serves many tenants now

This PR refactors the queue mechanics so a single PG queue table (well, two PG queue tables) can track all of our async work items, across all tenants.

It's a bigish diff, but it's focused on two smallish changes:
- All work units & work items now get a `tenant_id` prefix, to associate them with a single tenant.
- A new `QueueItemBatches` table keeps track of which tenants have pending work. Each row in the table represents one batch of QueueItems, called a WorkUnit
  - Whenever a new QueueItem is enqueued, a DB trigger fires and creates or updates ('upserts' in Claudish) a WorkUnit row
  - If that update pushes the batch over the minimum token threshold, the batch turns eligible
  - When a deriver is looking for eligible WorkUnits, it checks the Batches table and claims a row all at once.
- The Batch table update logic lives in the database, not the app.
  - The alembic migration sets trigger rules to keep the batch table in step with the queue.
  - And we added another startup validator to crash the app if the DB isn't configured with the right triggers
  - As well as added a note to our (upcoming and hotly anticipated) migration runbook.

Tracked as [A9](https://linear.app/plastic-labs/issue/DEV-2501/a9-maintained-workunitbacklog-aggregate-for-the-claim-query) but worked as A4.5; it's a pereq for [A5](https://github.com/plastic-labs/honcho/pull/1170)

Claude's description follows.

***

## What

Replaces the deriver claim's per-poll queue aggregation with a trigger-maintained `queue_item_batches` table — one row per pending work unit (`tenant_id`, `task_type`, `pending_count`, `total_tokens`, `oldest_created_at`) — and claims off it with `FOR UPDATE SKIP LOCKED`. Selection behavior is unchanged: same work units, same oldest-first order.

- **Why:** the old claim recomputed two `GROUP BY work_unit_key` passes (one joined to `messages` for token sums) over every unprocessed row, on every poll, from every worker — measured ~O(depth²) (0.24 ms @ 100 pending → 713 ms @ 500k), with optimistic-claim collisions wasting (M−1)/M of a contending pool's attempts. The aggregate keeps the claim at one indexed read per unit, and SKIP LOCKED makes concurrent claimers pick disjoint candidate sets. `ActiveQueueSession` remains the durable claim marker (batch row locks release at claim-commit while processing runs for minutes), so the `NOT EXISTS` eligibility check remains too — now against the small batch rowset instead of the queue.
- **Trigger topology:** INSERT takes a row-level fast path (+1 / +tokens / `LEAST`, token lookup via the partition-pruned composite PK with a default-tenant path for single-tenant rows). Completion and deletion are **statement-level triggers over transition tables** — one exact recompute per distinct affected key instead of one per row, so a 200-item batch completion or a workspace cascade does one aggregate pass per unit. The recompute **locks the batch row before aggregating**: without it, two READ COMMITTED races (both reproduced on live PG 17 during review, both regression-tested here) could clobber a concurrent enqueue's increment or — via the DELETE guard's EPQ recheck, which re-evaluates the row but not its `NOT EXISTS` subquery — drop a live unit's row permanently.
- **Deadlock posture:** trigger-side lock acquisition is canonical (enqueue batches insert key-sorted; statement triggers iterate keys sorted); the claim locks in scheduling order because SKIP LOCKED's backfill property depends on it. The residual rare inversion deadlocks, Postgres's detector breaks it, and both sides retry — the enqueue via a bounded transient-error retry (previously a deadlock-victim enqueue was swallowed by the catch-all and the batch was silently never derived), the claim on its next poll.
- **Migration** `b7d2f4a81c39` installs the table, triggers, and a backfill that first derives each pending unit's tenant from its key prefix (rows enqueued before this revision predate every `tenant_id` writer; left NULL, the entire pre-upgrade backlog would share the fair claim's tenant-less bucket) — an instance upgrading mid-queue starts exact.
- **Startup validator:** `validate_queue_item_batches` (`src/startup/`) runs in the API lifespan and the deriver entrypoint beside the embedding and tenant-isolation validators. It refuses to boot unless `queue_item_batches` exists and all three `queue` triggers are present and enabled for ordinary writes (`pg_trigger.tgenabled` `O`/`A`; `D` and replica-only `R` both fail, with the trigger named and the `ENABLE TRIGGER` / `alembic upgrade head` remedy in the message). Without it a missing trigger has no error path — enqueued work never reaches the aggregate and every deriver silently claims nothing. Introspection is retry-wrapped and fails closed; always on, since the aggregate is load-bearing single-tenant too.
- **Substrate repairs** the aggregate sits on (all `MULTI_TENANT`-only; the flag is off everywhere): the dead `tenant_id` columns on `queue`/`active_queue_sessions` are now stamped at every insert site and at claim, derived from the work-unit key so column ≡ key by construction; the flag-blind `startswith("representation:")` matchers are retyped on `task_type` (they silently disabled batching for tenant-prefixed keys); `delete_session` gets the flag-aware claim cleanup its `delete_workspace` sibling had, via one shared helper; a tenant id colliding with a task-type name is rejected at key construction.
- The deriver metrics poller reads the same aggregate, retiring its own periodic full-queue GROUP BYs.

## The part to review hardest: the trigger semantics under concurrency

The recompute's pre-lock and the statement-level topology are where the review found real, reproduced defects in the first cut — the two interleavings are now encoded as two-session regression tests (`tests/deriver/test_backlog_concurrency.py`) that were verified to fail against the unfixed trigger bodies. The delete guard keys on queue **membership**, never the counter; write amplification on bursty same-unit enqueues remains real and is on the measurement plan.

## Verification

- Full suite 2679 passed flag-off (the two pre-existing TypeScript-SDK toolchain failures aside); basedpyright 0 errors; ruff check + format clean on the CI scope; every migration passes the pipeline tests and `ensure_alembic_tests` gates green.
- 24 targeted queue_item_batches tests (increment/recompute/existence-delete semantics, token accounting, dedup-rollback atomicity, claim gating, two-open-transaction SKIP LOCKED disjointness, metrics parity) + 5 concurrency regressions + migration backfill hooks covering five seeded shapes including a tenant-prefixed NULL-column unit and a drained unit that must not backfill.
- Startup-validator tests: assertion units (missing table, missing triggers listed, `D`/`R` rejected, `A` accepted, schema-qualified messages), the fail-closed retry budget, and two DB-backed tests — passes against the migrated test DB, refuses while a trigger is `DISABLE`d and accepts again after `ENABLE`.
- A pre-merge review pass (8 dimensions) ran on the stack; its fixes are folded in as the `review:` commits.

## Out of scope (tracked)

- Fair scheduling order (the stacked follow-up PR).
- Atomic enqueue and the pooled deployment manifest.
- Claim-throughput / write-amplification / anti-join-cost measurements (planned on the perf-testing track).


akattelu (approved):

ulyssepence (commented):
I would like to go over the migration script in-person but otherwise LGTM

PR #1170, opened 2026-09-10
Shared Tenants A5: per-tenant fair scheduling

philosaether:
# One busy tenant can't hog the shared queue

[A4.5](https://github.com/plastic-labs/honcho/pull/1168) puts all shared tenants in a single Queue table, with tenant-labeled batches tracked in a companion Batches table. This ticket makes the batch claim fair.

The important bit is the `claim_rows_query` in deriver.py. On each queue poll, the deriver pod selects all eligible work units, groups them by tenant, and ranks them from oldest to youngest *within each tenant.* Then it assigns the first item from each tenant's queue to a worker, then the second item, then the third, and so on, until it runs out of eligible work units or available workers.

This algorithm is designed to prevent one busy or bursty tenant from dropping a ton of derivations on our queue at once and starving the rest of our users. There's a potential failure scenario when the queue is saturated for a long period of time; in that case, workers are grabbing tasks one at a time and the queue may fall back to oldest-first. We have [a note](https://linear.app/plastic-labs/document/drift-checklist-171fc9555b6e) to quantify the word "may" in staging.

This PR also uses the `settings.DERIVER.WORKERS` config variable as an *upper limit* for the total number of workers, which is now a function of `DB.POOL_SIZE`, `DB.MAX_OVERFLOW`, and `DERIVER.WORKERS_PER_POOL_CONNECTION` when connection pooling is on. These variables match the parameters we will actually want to fine-tune in production.

We'll need the placeholder method in `deriver.py` when we get to [C6](https://linear.app/plastic-labs/issue/DEV-2546/c6-stop-derivation-for-a-billing-paused-shared-tenant). There are also hecka tests.

Claude's description follows.

***

## What

One shared queue serving many tenants needs a fairness rule; the base ordering is global oldest-first, so a flooding tenant's backlog sorts ahead of everyone. This PR makes the claim **round-robin over `tenant_id`**, stacked on the WorkUnitBacklog claim (previous PR in the stack).

- **Ordering** (in `crud.deriver.claim_rows_query`, which the deriver and the tests both consume — no mirrored SQL anywhere): a window function ranks each tenant's *eligible* units oldest-first inside the candidate subquery, and the outer ORDER BY takes every tenant's rank-1 before any rank-2 — a whale contributes one unit per round. Eligibility (unclaimed + batch gate + exclusion seam) lives inside the ranking, so a claimed rank-1 never shadows its tenant's rank-2. NULL tenants group as one partition: the tenant-less reconciler lane rotates with everyone, and a single-tenant deployment (every `tenant_id` NULL) degenerates to plain oldest-first — identical to the base ordering — which is why this ships un-gated, no flag branch in the query.
- **Query shape:** `FOR UPDATE` cannot combine with a window function at one query level, hence rank-in-subquery → join back → `FOR UPDATE OF work_unit_backlog SKIP LOCKED`. Concurrent claimers stay disjoint and backfill from the sorted stream.
- **Tenant-exclusion seam:** `claim_excluded_tenant_ids()` returns ids (`None` today — no exclusion source exists yet) and the claim composes `tenant_id IS NULL OR tenant_id NOT IN (ids)` itself. Ids-not-clause on purpose: a bare `NOT IN` is NULL-false and a wired source could silently starve the reconciler lane — this contract makes that footgun unwritable. Excluded rows stay in the backlog (skipped, not drained), so unsuspending a tenant is just dropping it from the source.
- **Lane invariant in the schema:** migration `c4e8a2b91d57` first NULLs legacy reconciler rows that carry a workspace (enqueued in the era when `workspace_name` was NOT NULL), then adds `CHECK ((workspace_name IS NULL) = (task_type = 'reconciler'))` as `NOT VALID` + `VALIDATE` — the plain form takes ACCESS EXCLUSIVE with a full scan and blocks every enqueue and claim for its duration. The constraint is mirrored in `models.QueueItem`, and the revision has pipeline hooks (legacy violator cleaned, both illegal shapes rejected, both legal shapes accepted). The tenant half of the invariant is flag-dependent, so it is an enqueue-time assert instead of a CHECK.
- **Worker concurrency:** `DERIVER_WORKERS` cap 100 → 512 (per-process concurrency; process replica count is an orthogonal deployment concern). Effective concurrency is derived — `min(WORKERS, max(1, floor(WORKERS_PER_POOL_CONNECTION × pool capacity)))` — because each in-flight unit opens several sequential short sessions and nothing previously stopped `WORKERS` from exhausting the pool. Skipped under NullPool (those settings are inert there). Derivation only lowers; it's logged at boot, exported as `deriver_effective_worker_cap` (per-replica: `sum()` = fleet worker capacity, `min()` shows whether any replica's pool is the binding cap), and warns when the pool binds. The ratio default is conservative until measured.

## The boundary to review hardest: what the rotation does and doesn't guarantee

Round-robin is a **within-one-claim** property — ranks recompute per claim, with no cross-poll state. Claims wider than the tenant count put every tenant in round 1 (the normal regime for a pool). In the single-slot trickle regime — a saturated pool freeing one worker at a time — round-1 contention resolves by age and ordering degenerates toward oldest-first, i.e. the pre-fairness FIFO bound. That boundary is stated in the tests, measured next on the perf-testing track, and persisted least-recently-served rotation is deliberately not built until the numbers say the trickle regime matters.

## Verification

- Full suite 2652 passed flag-off; basedpyright 0/0/0; all migrations green in the pipeline tests; `ensure_alembic_tests` gates green.
- 21 fairness tests: the interleave (with an explicit not-oldest-first assertion), whale/minnow starvation bounds at width and under two-slot scarcity, tenant-less bucket participation, flag-off degeneracy pinned for successive and wide claims, the rank-shadowing regression, seam no-op today plus a patched-source exclusion test (skipped rows stay in the backlog), both CHECK rejections and the valid lane shapes, worker-cap derivation in both regimes including floor-to-one and the NullPool skip, the gauge, and boot-warning polarity. Claim ordering is asserted from the production query builder, not a mirror.
- A pre-merge review pass (8 dimensions) ran on the stack; its fixes are folded in as the `review:` commits.

## Out of scope (tracked)

- The tenant-exclusion source (DEV-2546 wires the seam).
- Cross-poll rotation state (pending the trickle-regime measurement).
- Per-task-type depth metrics and API-path rate limiting.


akattelu (approved):

src/config.py
```diff
@@ -914,7 +914,25 @@ class DeriverSettings(HonchoSettings):
 
     ENABLED: bool = True
 
-    WORKERS: Annotated[int, Field(default=1, gt=0, le=100)] = 1
+    # Concurrency within a single deriver process. Total worker count across a
+    # deployment is processes x WORKERS.
+    WORKERS: Annotated[int, Field(default=1, gt=0, le=512)] = 1
```
ulyssepence, line 919: Where did 512 come from?

src/crud/deriver.py
```diff
@@ -118,38 +133,75 @@ def active_queue_session_match(
...
+        func.row_number()
+        .over(
+            partition_by=models.QueueItemBatch.tenant_id,
+            order_by=(
+                models.QueueItemBatch.oldest_created_at.asc(),
+                models.QueueItemBatch.work_unit_key.asc(),
+            ),
+        )
+        .label("rank_within_tenant"),
+    ).where(unclaimed_work_unit_clause(models.QueueItemBatch.work_unit_key))
```
ulyssepence: Cool usage of `ROW_NUMBER()`. Never heard of this function

src/models.py
```diff
@@ -685,6 +685,12 @@ class QueueItem(Base):
     message_id: Mapped[int | None] = mapped_column(BigInteger, nullable=True)
 
     __table_args__ = (
+        # Mirrors the queue_workspace_reconciler_check migration: NULL
+        # workspace_name means the tenant-less reconciler lane and nothing else.
+        CheckConstraint(
+            "(workspace_name IS NULL) = (task_type = 'reconciler')",
+            name="ck_queue_workspace_null_iff_reconciler",
```
ulyssepence: Drop the `ck_queue_` prefix

migrations/versions/c4e8a2b91d57_queue_workspace_reconciler_check.py
```diff
@@ -0,0 +1,67 @@
...
+    # split, the validation runs under SHARE UPDATE EXCLUSIVE instead.
+    op.execute(
+        f"""
+        ALTER TABLE {schema}.queue
+        ADD CONSTRAINT {CONSTRAINT_NAME}
+        CHECK ((workspace_name IS NULL) = (task_type = 'reconciler'))
+        NOT VALID
+        """
+    )
+    op.execute(f"ALTER TABLE {schema}.queue VALIDATE CONSTRAINT {CONSTRAINT_NAME}")
```
ulyssepence: Check to make sure these are all done in auto-commit. I think the last line here might continue to hold a lock on the table and make the service unresponsive while it runs, depending on how many rows there are.

src/crud/deriver.py
```diff
@@ -118,38 +133,75 @@ def active_queue_session_match(
...
+
+    threshold_clause = batch_threshold_clause()
+    if threshold_clause is not None:
+        eligible = eligible.where(threshold_clause)
+    excluded_tenant_ids = claim_excluded_tenant_ids()
```
ulyssepence, line 212: Probably want to add this to the metrics code path as well (not sure how useful/where that fits into the shared tenants plan)

src/deriver/queue_manager.py
ulyssepence, line 674: I'm guessing we're going to change this but just thinking that this might ignore the queue fairness

src/crud/deriver.py
ulyssepence: I'm thinking we probably want to change this and only apply this in later filtering below. Like this, I don't think we're counting batches that are being processed... so a customer that already has a batch being processed will still get row number 0 (or 1?... whatever it starts at) instead of what we'd want: 2

ulyssepence (changes requested):

src/crud/deriver.py
philosaether: Wish I could say I thought of it

philosaether: Good catch

src/config.py
philosaether: It's arbitrary / Claude supplied. The actual number of workers is set in the block starting on `queue_manager.py:175` as (POOL_SIZE + MAX_OVERFLOW) * WORKERS_PER_POOL_CONNECTION, this value is the max.

All of these settings can be tuned based on what we measure in staging; initial values are placeholders

src/crud/deriver.py
philosaether: Mmm, good point. KEDA will over-provision and grafana will over-report

src/crud/deriver.py
philosaether: This and the comment you left in queue_manager have the same solution: rank by active_queue_sessions + rank_within_tenant. On it.

src/crud/deriver.py
philosaether: I'm going to track excluded rows in a separate metric though, I'm curious to see how many we end up with on prod

migrations/versions/c4e8a2b91d57_queue_workspace_reconciler_check.py
philosaether: It did, added autocommit

src/crud/deriver.py
```diff
@@ -118,38 +133,111 @@ def active_queue_session_match(
...
+                        models.QueueItemBatch.oldest_created_at.asc(),
+                        models.QueueItemBatch.work_unit_key.asc(),
+                    ),
+                )
+            ).label("tenant_fair_rank"),
+        )
+        .outerjoin(
+            inflight,
+            models.QueueItemBatch.tenant_id.is_not_distinct_from(inflight.c.tenant_id),
+        )
```
ulyssepence, line 205: I think this won't make use of any indices, which might matter when we are under a lot of load and both these lists are bigger.

ulyssepence (approved):

PR #1173, opened 2026-09-13
docs: trim unused provider keys from .env.template (fixes #964)

### LINKED ISSUE #964: [Bug] Default .env template ships with unused provider API keys, and LLM config is verbose enough to make cleanup hard
# Default `.env` template ships with unused provider API keys, and LLM config is verbose enough to make cleanup hard

## Summary

Fresh Honcho self-host installs ship with a `.env` (or `.env.template`) that includes multiple LLM provider API key slots, even when the deployment only ever uses one transport at runtime. In practice, operators paste real keys into all the slots that seem relevant and then forget the ones that are never called. The result is dormant credentials sitting in configuration files, indistinguishable from the ones that are actually in use.

A related problem makes this worse: LLM configuration in Honcho is spread across ten or more `_MODEL_CONFIG__` environment variables (one per workload — DERIVER, SUMMARY, five dialectic levels, two dream passes, etc.). Migrating an existing install to a new provider requires touching every variable, and there's no obvious way to see "which of these transports is actually live." The verbosity makes the security problem harder to spot.

## Concrete scenario

I run a self-hosted Honcho behind a LiteLLM proxy that fronts OpenRouter. My `.env` has these two provider keys:

```
LLM_ANTHROPIC_API_KEY=sk-ant-oat01-...        # real key
LLM_OPENAI_API_KEY=<litellm-master-key>       # what actually gets used
```

Every workload is configured with `TRANSPORT=openai`. Nothing in my deployment ever calls the Anthropic transport. But the Anthropic key sat in the .env for months, indistinguishable from the used one, until it ended up in a terminal session that I later realized had been screenshared. Rotating that key was a five-minute panic that shouldn't have been necessary because the key shouldn't have been there in the first place.

Grepping the key was what surfaced it (I was tracing which model was being called), not any alarm from Honcho. If Honcho had refused to accept unused provider credentials, or warned about them at startup, the exposure window would have been hours instead of months.

### 1. Startup warning for unused provider credentials

At startup, Honcho already parses the `TRANSPORT` values across all workloads. It could compare that set against the set of `LLM_*_API_KEY` values present and log a warning for any key not backing a live transport:

```
WARN: LLM_ANTHROPIC_API_KEY is set but no workload uses transport=anthropic.
      If not needed, remove from .env to reduce credential exposure.
```

Non-blocking. Just a warning per startup. It would catch the class of mistake I made without changing existing deployments.

Optionally, a stricter mode: `HONCHO_STRICT_CREDENTIALS=true` would fail startup if unused keys are present. Useful for production installs.

### 2. Simplify per-workload LLM config with a default fallback chain

Ten identical `_MODEL_CONFIG__MODEL` variables to configure the same model type is a lot of surface for one decision. A saner default would be a fallback chain:

```
LLM_DEFAULT_MODEL=gpt-4o-mini
LLM_DEFAULT_TRANSPORT=openai
# workload-specific overrides only when needed
DIALECTIC__LEVELS__HIGH__MODEL_CONFIG__MODEL=gpt-4o
DREAM__DEDUCTION_MODEL_CONFIG__MODEL=gpt-4o
```

Any workload without an explicit `_MODEL_CONFIG__MODEL` would inherit `LLM_DEFAULT_MODEL`. Same for `TRANSPORT`. Adding a new workload doesn't require adding two env vars if the defaults work; migrating a deployment to a new provider is one line instead of ten.

This composes with the first suggestion: with a default in place, deployments that don't need workload-specific tuning would only need `LLM_DEFAULT_*`, and unused keys become obvious the moment you review the config.

## Why this matters beyond one install

The pattern I hit is common to any self-hosted deployment that gets migrated between providers over time. Honcho is popular enough that this will happen a lot: someone installs it against OpenAI, later moves to a proxy or a different provider, and never cleans up the initial keys. The current design makes that cleanup optional and hard; it should be surfaced by default.

The credential problem is the sharper edge, but the configuration verbosity is what makes it hard to notice and hard to fix. Both are worth addressing.

## Environment

- Honcho (self-hosted from source, installed at `/opt/honcho`)
- PostgreSQL 16, Redis, uvicorn 2 workers
- Debian 13
- Fronted by LiteLLM 1.86.0

Happy to submit a PR for the startup-warning piece if this direction seems reasonable. The default-fallback piece is a bigger change and probably needs a design discussion first.

PerryLink:
## Description

Fixes #964. This PR cleans up the default `.env.template` so it no longer ships key slots for LLM providers the default configuration never uses, and tightens the verbose LLM config comments so it is obvious which transports are actually live. The `LLM_ANTHROPIC_API_KEY` and `LLM_GEMINI_API_KEY` placeholder lines are removed, the OpenAI key (which backs every default workload) is kept, and the LLM section now calls out that only keys backing a live transport should be set. The repeated per-level Dialectic defaults block is collapsed into a compact summary without losing any default values.

## Proofs

Docs-only change; no runtime behavior is affected.

`git diff --stat`:

```
 .env.template | 61 +++++++++++++++++------------------------------------------
 1 file changed, 17 insertions(+), 44 deletions(-)
```

- `LLM_ANTHROPIC_API_KEY` / `LLM_GEMINI_API_KEY` placeholder lines removed; `LLM_OPENAI_API_KEY` retained (every workload defaults to `transport=openai`).
- Dialectic per-level defaults shrunk from 38 comment lines to 12 while preserving all default values (`MAX_TOOL_ITERATIONS` per level, `minimal` 250-token cap, `TOOL_CHOICE=auto`).

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Updated environment configuration guidance for text-generation and embedding model defaults.
  - Documented supported AI provider transports, API key requirements, and workload-specific model capabilities.
  - Clarified that text-generation models must support tool calling, while embedding models do not.
  - Simplified Dialectic configuration examples and documented default transports, models, tool iteration limits, and tool-choice behavior.
  - Added examples for per-level model, base URL, and thinking-effort overrides.


PerryLink:
Addressed CodeRabbit's note on the wording: tool calling is now stated as a text-generation requirement only (embedding models do not support it), and the embeddings transport default is spelled out explicitly (transport = `openai`, supported transports: openai, gemini). Both defaults verified against `src/config.py`.

.env.template
```diff
@@ -158,41 +158,16 @@ LLM_OPENAI_API_KEY=your-api-key-here
...
-# DIALECTIC_LEVELS__low__TOOL_CHOICE=auto
-# DIALECTIC_LEVELS__medium__MODEL_CONFIG__TRANSPORT=openai
-# DIALECTIC_LEVELS__medium__MODEL_CONFIG__MODEL=gpt-5.4-mini
-# DIALECTIC_LEVELS__medium__MAX_TOOL_ITERATIONS=2
-# DIALECTIC_LEVELS__high__MODEL_CONFIG__TRANSPORT=openai
-# DIALECTIC_LEVELS__high__MODEL_CONFIG__MODEL=gpt-5.4-mini
-# DIALECTIC_LEVELS__high__MAX_TOOL_ITERATIONS=4
-# DIALECTIC_LEVELS__max__MODEL_CONFIG__TRANSPORT=openai
-# DIALECTIC_LEVELS__max__MODEL_CONFIG__MODEL=gpt-5.4-mini
-# DIALECTIC_LEVELS__max__MAX_TOOL_ITERATIONS=10
```
VVoruganti: This section is meant to illustrate what the default values are, I would lean towards keeping them as there has been a back and forth with the community about making that clearer to self-hosters.

PR #1182, opened 2026-09-14
feat(telemetry): split harness, SDK and raw REST traffic

ajspig:
## Description

Direct SDK use, raw REST calls and pre-header plugin versions all land in the same `unidentified` bucket in the CloudEvents `client` object, so the harness / SDK / REST split from the identity-header convention cannot be answered.

Both SDKs now send `X-Honcho-Host` naming themselves (`honcho-python/2.4.0 (darwin)`, `honcho-typescript/2.4.0 (darwin)`), following the same `name/version (platform)` convention harness plugins use. 

The API middleware records `User-Agent` as the host when no identity header is present, so raw REST clients become distinguishable immediately without waiting for SDK upgrades to propagate — the server alone can never identify an SDK, since it only sees the transport's own User-Agent.

A harness plugin's own `X-Honcho-Host` still wins: the SDK writes its default first and the caller's headers are merged last, in both SDKs. Only a request with neither header remains `unidentified`.

The TypeScript SDK has no runtime API for its own version, so `VERSION` sits beside `API_VERSION` and `scripts/update_version.py` keeps it in sync with package.json. (Importing package.json does compile, but bundlers inline the whole manifest into shipped code.) Python reads its installed metadata, so its detection moved next to its only consumer rather than being reached across a circular import.

Tracks [DEV-2808](https://linear.app/plastic-labs/issue/DEV-2808/split-harness-vs-direct-sdk-vs-raw-rest-traffic-in-telemetry).

Follow-up, already written and not in this PR: `honcho-cli` must send its own `honcho-cli/<version>` host, otherwise every CLI request is counted as direct SDK use.

## Proofs

* `uv run pytest tests/telemetry/ tests/sdk/ tests/sdk_typescript/` → **465 passed**.
* Pre-push hook green end to end: `pytest (main app)`, `pytest (Python SDK)`, `TypeScript build`, `TypeScript type check`, `basedpyright`, `ruff check`, `ruff format`, `bandit`, `biome` all Passed.
* New coverage: `test_user_agent_is_host_fallback` parametrizes the three middleware cases — `curl/8.4.0` with no identity header becomes the host, an explicit `X-Honcho-Host` beats a `python-httpx/0.27.0` User-Agent, and a blank `X-Honcho-Host` is treated as absent. `test_sdk_sends_default_host_header_unless_overridden` asserts over a mock transport that the Python SDK sends its own host and that a caller-supplied `claude-code/2.1.3 (darwin)` replaces it. The TypeScript constructor test asserts the same pair.
* Runtime check against the built client:

```
honcho.__version__ = 2.4.0
header             = honcho-python/2.4.0 (darwin)
sent               = honcho-python/2.4.0 (darwin)
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * SDK requests now include an `X-Honcho-Host` header identifying the SDK, version, and platform.
  * Custom host header values continue to override the automatically generated identification.
  * Telemetry can identify clients through the request’s `User-Agent` when no explicit host is provided.
  * TypeScript SDK version information is now exported and kept synchronized during version updates.

* **Tests**
  * Added coverage for SDK identification headers, overrides, versioning, and telemetry fallback behavior.


src/telemetry/client_context.py
```diff
@@ -48,11 +52,15 @@ def _clean(value: str | None) -> str | None:
...
+    *,
+    host: str | None,
+    plugin: str | None,
+    agent_model: str | None,
+    user_agent: str | None = None,
 ) -> ClientContextTokens:
     """Set the client ContextVars; returns tokens for ``reset_client_context``."""
     return (
-        client_host.set(_clean(host)),
+        client_host.set(_clean(host) or _clean(user_agent)),
```
akattelu, line 63: I think it could be useful to know when this happens. I wonder if we should prefix the user agent like `user-agent-{user_agent}` so that we can distinguish it between integrations?

akattelu (approved):

src/telemetry/client_context.py
ajspig: The fallback runs when X-Honcho-Host is missing/blank, which is raw REST (curl, ad-hoc fetch/httpx) and old SDKs that predate this header. The possible values for this field are: 

honcho-python/2.4.0 (darwin) → SDK
honcho-typescript/2.4.0 (darwin) → SDK
claude-code/2.1.3 (darwin) / opencode/… / honcho-mcp/… → plugin / MCP
anything else → the fallback: curl/8.4.0, python-httpx/0.27.0, undici, a browser UA

instrumented clients always send Host, so it won't collide with an integration name

PR #1183, opened 2026-09-14
refactor: drain legacy source_ids via reconciler backfill

akattelu:
## Description

The document_sources migration is now DDL only. Existing JSONB linkage (documents.source_ids, internal_metadata.source_ids / premise_ids) is copied into the edge table by a new reconciler task in id-ordered batches of 500, each a single server-side statement, so the api pod's init container is no longer blocked on a full-table copy and deriver memory stays flat regardless of tenant size.

Until the drain finishes, Document.source_ids falls back to the legacy column, and child lookup, the source_ids filter, and the scope-removal cascade OR in a legacy containment match. The column and its GIN index stay until the follow-up migration drops them.

## Proofs

N/A

## Checklist


src/reconciler/backfill_document_sources.py
```diff
@@ -0,0 +1,128 @@
...
+logger = logging.getLogger(__name__)
+
+BACKFILL_BATCH_SIZE = 500
+BACKFILL_TIME_BUDGET_SECONDS = 240
+
+# Any legacy location still populated. Also matches rows whose column holds a
+# non-array JSON value, which the drain clears without emitting edges.
+_PENDING_PREDICATE = """
+    source_ids IS NOT NULL
+    OR jsonb_exists_any(internal_metadata, ARRAY['source_ids', 'premise_ids'])
```
ajspig: As this exists, it will perform one full seq scan of documents, on every deriver replica (so every 5 minutes). 

Changing jsonb_exists_any(x, ARRAY[…]) to x ?| ARRAY[…] makes it indexable.

So if its`OR internal_metadata ?| ARRAY['source_ids', 'premise_ids']` this would avoid these full seq scan after the drain finishes.

PR #1184, opened 2026-09-15
fix(llm): guard OpenAI backend against malformed OpenAI-compatible responses

### LINKED ISSUE #676: [Bug] OpenAI-compatible chat responses can crash backend when choices or usage are missing

eatmoreduck:
## What

OpenAI-compatible gateways can return an HTTP-successful response without a usable shape: empty `choices`, `None`/missing `choices`, a `None` message, or a missing `usage` attribute. The backend crashed on those with raw `AttributeError`/`IndexError`/`TypeError` in four places (`_normalize_response`, the structured `parse()` path, the tool-call check, and `_parse_or_repair_structured_content`).

## How

- Add a `_first_choice()` helper that returns the first choice with a usable message and raises a controlled `LLMError` otherwise (matching the Gemini backend's convention).
- Read `usage` via `getattr` so a missing attribute degrades to zero-token accounting instead of crashing.

Fixes #676

## Testing

- New `tests/llm/test_backends/test_openai_malformed_responses.py`: 6 cases (empty choices / None choices / None message / plain-dict response raise `LLMError`; missing usage degrades gracefully; json_object path raises too). All red before the fix, green after.
- Full `test_openai.py` (60 tests) and all backend suites (95 tests) pass; ruff clean.

🤖 Disclosure: prepared with the help of an AI coding agent; reviewed and tested by a human before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed OpenAI-compatible responses.
  * Returns controlled errors when response choices or messages are missing, empty, or invalid.
  * Applies consistent validation across standard, structured-output, and truncated responses.
  * Gracefully handles missing usage details by reporting zero input and output tokens.
  * Handles messages without content safely while preserving available tool-call information.


eatmoreduck:
Friendly note on the issue gate: this PR fixes #676 (malformed OpenAI-compatible responses crash the OpenAI backend with raw `IndexError`/`TypeError`), which has been open and unassigned since May. The fix and regression tests have been through five coderabbit review rounds, with every finding resolved. If a maintainer could take a look and add the `maintainer-approved` label to #676 when convenient, this would be ready for the review queue. Happy to adjust anything in the meantime. (I'm not on Discord, hence the note here.)

eatmoreduck:
Following the bot's pointer here — I tried reopening as suggested, but GitHub is refusing the reopen from our side (`Could not open the pull request` on both the GraphQL and REST endpoints). The fork and branch (`fix/openai-response-shape-guard`) are intact, so this looks like a platform quirk rather than anything deliberate.

Flagging per the notice's own suggestion: this fixes #676 (open, unassigned since May — malformed OpenAI-compatible responses crash the backend with raw `IndexError`/`TypeError`). The branch is unchanged and has been through five coderabbit rounds with all findings resolved.

@akattelu would you be able to reopen this, or alternatively add the `maintainer-approved` label to #676 and I'll open a fresh PR against it — whichever is easier on your end? And if this fix simply isn't wanted, that's completely fine too; just say so and I'll close the book on it. Thanks!

PR #1185, opened 2026-09-15
Shared Tenants A6: single-tenant by default

philosaether:
# Self-hosters get single-tenancy by default

This PR ensures that self-hosters using OSS honcho will still get a functional single-tenant instance with no additional maintenance or config overhead. It also adds a new INSTANCE_TYPE environment variable, to fix a boot validator bug introduced by an earlier shared tenants ticket.

The [A6 requirements](https://linear.app/plastic-labs/issue/DEV-2481/a6-keep-honcho-single-tenant-by-default-for-oss-deployments-nice-to) were already satisfied when this ticket began, so the actual work here was mostly "test and validate that the default honcho config is still a functional single-tenant instance." I'd recommend reviewers read the Proofs table in Claude's description to see the tests we ran.

Claude's description follows. 

---

## What this PR is

DEV-2481 asked for *"a mode flag, defaulting to single-tenant, that makes the multi-tenant machinery inert."* That flag is `MULTI_TENANT` (default off), built by A2 and threaded through A3–A5; cloud-only RLS made single-tenant the OSS default by construction. So this PR is the **close-out**: it proves each line of the ticket against the shipped stack (table below), writes the one gate that was missing — a test that boots the migrated schema with **zero** tenant configuration and runs every startup validator against it — and folds in the auth-off boot assertion deferred from A3.

Stacked on #1170 (A5). Diff: one validator check plus a `ROLE` setting, three test modules, a commented block in each config template, one CLAUDE.md paragraph.

**Deployment note for the shared pool:** the deriver pods must set `ROLE=deriver` (new, default `api`) before `MULTI_TENANT` is turned on, or they will refuse to boot asking for the API's auth config. The compose example already sets it on the deriver service; the argo pool manifests (B3) need the same line.

## Proofs — ticket → mechanism → test

| Ticket line | Where it is | Proof |
|---|---|---|
| **What changes:** a mode flag, defaulting to single-tenant | `MULTI_TENANT: bool = False` — `src/config.py` | `test_the_shipped_default_is_single_tenant` pins the class default, not the process env |
| No tenant claim required on JWTs; `ad=True` keeps its meaning | `tn` is optional on `JWTParams`; the gate is `if settings.MULTI_TENANT and not jwt_params.tn` and runs only flag-on, before the admin short-circuit — `src/security.py` | `test_flag_off_tenantless_token_unchanged` (`tests/test_security.py`) |
| A `tn`-bearing token on a flag-off instance (the dedicated-instance case: groudon mints `tn` for every tenant after C3) | The bind is gated on the **flag**, not on `tn` presence, so rows keep `"default"` — `src/security.py` | `test_no_binding_when_flag_off_even_if_token_carries_tenant` (`tests/test_security.py`) · `test_tenant_bearing_token_binds_nothing` (**new**, `tests/test_single_tenant_mode.py`) |
| Queue scheduling stays oldest-first | Flag-off keys are un-prefixed → `tenant_id` NULL on `queue` and `queue_item_batches` → the round-robin window has one partition → rank order **is** oldest-first — `src/crud/deriver.py::claim_rows_query` | `test_successive_single_unit_claims_walk_the_backlog_oldest_first`, `test_one_wide_claim_returns_the_whole_backlog_oldest_first`, `test_an_unprefixed_key_stamps_null_with_the_flag_off` (`tests/deriver/test_fair_scheduling.py`) · first-claim leg of `test_first_write_and_first_claim_need_no_tenant` (**new**) |
| *"Whether that is one setting or several"* | **One master switch + four subordinates** (`MULTI_TENANT_SKIP_RLS_ASSERT`, `DB.SERVICE_CONNECTION_URI`, `DB.POOLER_MODE`, `TENANT_API_SECRET`), each consulted only when the master is on and each defaulting to the single-tenant-correct value | `tests/test_single_tenant_mode.py` (**new**): flag off with all four subordinates set to a value flag-on would act on; validator is a no-op that never opens a connection, `/v3/tenants` stays 405, `tn` binds nothing, keys stay un-namespaced, `tracked_db` needs no tenant. New tenant-subordinate settings get added to `_SUBORDINATES` there |
| *"Whether the implicit tenant is a real row or a sentinel"* | **A real row.** `tenants."default"` is seeded by A1's migration; new rows get it from the model default `_default_tenant_id()`; the tenant API refuses to delete it — `src/models.py`, `src/routers/tenants.py` | `test_first_boot_seeds_one_tenant_and_every_validator_passes` (**new**) asserts the migrated schema holds exactly `["default"]`; `test_first_write_and_first_claim_need_no_tenant` (**new**) asserts every row of the first write reads `"default"` with no caller naming a tenant |
| **Done when:** the existing suite passes with the flag off, unmodified | CI sets only `DB_CONNECTION_URI`; every flag-on test opts in with `monkeypatch` | Full suite green on this branch (counts below) |
| **Done when:** a fresh single-tenant deployment works with no tenant configuration at all | `docker compose up` → `docker/entrypoint.sh` → `scripts/provision_db.py` → `alembic upgrade head` → API lifespan and deriver entrypoint run the three startup validators. Neither config template names a tenant setting | `tests/startup/test_fresh_single_tenant_deploy.py` (**new**): throwaway DB → alembic head → `tenants == ["default"]` → all three validators pass with the settings as shipped (nothing monkeypatched) → first write stamps `"default"` on workspaces/peers/sessions/session_peers/messages → first enqueue produces an un-namespaced unit, `tenant_id` NULL, the trigger-maintained aggregate row → a lone 3-token batch correctly waits (token target / age flush) → after the age window, the flag-off claim returns it |

## What changed in code

- **`src/startup/tenant_isolation_validator.py`** — new `_assert_auth_enabled`: `MULTI_TENANT` on with `AUTH_USE_AUTH` off refuses boot **on the API role**. With auth off every request is a tenant-less admin, so nothing ever binds a tenant and every tenant-scoped session fails closed — a uniform outage, not a leak, but one boot error beats a storm of identical 500s. Runs first, before the pooler check. The module docstring now lists all four half-states in check order.
- **`src/config.py`** — `ROLE: Literal["api", "deriver"] = "api"`. The deriver takes its tenant from the claimed work unit's key and verifies no JWT, so it is exempt from the auth check by role rather than forced to carry the API's auth config. No existing setting distinguished the two processes (they self-identify only through hardcoded metrics labels). Default `api` so an unlabeled process gets the stricter check.
- **`tests/startup/test_tenant_isolation_validator.py`** (new, 9 tests) — first tests for this validator: the shipped default pinned on the class; flag-off no-op that never touches the engine; refusals for auth-off, transaction-mode pooler, RLS not enforced (against the real migrated schema), no service role once RLS is enforced; the deriver-role skip; the skip-assert warning path; retry exhaustion fails closed; and the one fully-configured flag-on state that boots.
- **`tests/startup/test_fresh_single_tenant_deploy.py`** (new) — the Done-when 2 gate, above.
- **`tests/test_single_tenant_mode.py`** (new) — the subordinate-settings invariance matrix, above.
- **`tests/conftest.py`** — `untouchable_engine()`: an `AsyncEngine` stand-in whose every attribute access fails the test, so "no-op" is proven rather than assumed. `_get_test_db_url` gains a `tag` so a fixture's own throwaway database is named into the sweepable `test_db_<runid>_<tag>[_gwN]` family.
- **`config.toml.example`, `.env.template`, `docker-compose.yml.example`** — a commented `MULTI_TENANT` block each (master only; the subordinate settings are deliberately not listed, since listing them reads as a recipe for the half-state the validator refuses) and a commented `ROLE` line with the per-process caveat. The compose example sets `ROLE=deriver` on the deriver service. The example file claims to list every option and did not list these.
- **`CLAUDE.md`** — the tenant-isolation bullet now separates half-states where isolation cannot hold from the one where the API cannot serve, names the `ROLE` exemption, and cites the two new enforcement tests the way sibling invariants are cited. `src/dependencies.py` loses a stale comment claiming the per-request binding "is not yet wired".

## Not changed, on purpose

- No rename of `MULTI_TENANT` to a `mode` enum: a bool with a false default is a two-valued mode; a rename would touch 29 files across four merged PRs for no third value.
- No `if not MULTI_TENANT: oldest_first()` branch in the claim: the window degenerating to one partition is the same query, and byte-identity is the receipt.
- `DB.POOL_CLASS` stays `"default"`: A5's derived worker cap is `min(WORKERS, floor(4.0 × 30)) = 120` at stock settings and `WORKERS` defaults to 1, so no self-host config is ever capped; NullPool would remove the cap by removing the pool. Reasoning on DEV-2481.

## How verified

- Full suite on the branch: **2695 passed, 51 skipped**; the 2 failures are the pre-existing TypeScript SDK toolchain tests (`tests/sdk_typescript`), unchanged from the base. New modules: 16 tests (9 validator, 2 fresh-deploy, 5 one-switch), green serially and under xdist.
- `ruff check` and `ruff format --check` clean; `basedpyright` 0 errors / 0 warnings on every touched file.
- No throwaway databases left behind after the runs; the widened sweeper regex matches `test_db_<runid>_fresh[_gwN]`.
- Static Analysis CI runs only on main-targeted PRs, so the local gate is the gate (same as #1168 / #1170).

## Review receipts

Six-dimension pre-merge review (bugs · efficiency + redundancy · comment audience + agent-comment markers + `.meta` leak sweep + OSS leakage · peer consistency + runtime multiplicity · architecture vs CLAUDE.md · design reconciliation) plus a merge-readiness check against the stack. **0 must-fix.** Applied in the `review:` commit: role-gated the auth check (un-gated it would have forced auth config onto deriver pods); validator docstring lists all four half-states in check order and `main.py` stops enumerating them; retry-exhaustion test (every sibling validator test has one); one-switch module trimmed to the all-at-once case (per-subordinate cases duplicated existing flag-off tests); throwaway DB renamed into the sweepable family and conftest's `_drop_database` reused; template blurb reworded; aging comment corrected (the aggregate's update trigger recomputes only on `processed` flips); CLAUDE.md and the stale `dependencies.py` comment. Leak sweep clean with the self-test firing; zero tracker ids or internal names on added lines.

Tracked, not fixed here: `tests/integration/test_rls_isolation.py` uses a fixed database name and session scope, which can collide across xdist workers (pre-existing); the public configuration reference has no `MULTI_TENANT` / `ROLE` entry (docs work).

The new `# region ai` block in the validator follows the repo's agent-comment-marker convention (AI-authored *why* marked, human *what* unmarked); the convention relies on this pre-merge pass.


akattelu (approved):

eisene (approved):
what i can see in this PR makes sense to me, it's hard to have the full context but this narrow slice lgtm

src/config.py
```diff
@@ -1556,6 +1556,17 @@ class AppSettings(HonchoSettings):
...
+    # work. Set per process (one value each), never in a config both share.
+    # region ai
+    # Read by the tenant-isolation startup validator: under MULTI_TENANT the API
+    # must run with AUTH_USE_AUTH on (the JWT tenant claim is its only tenant
+    # source) while the deriver takes its tenant from the claimed work unit's key
+    # and needs no auth config. Defaults to "api" so an unlabeled process gets the
+    # stricter check; INSTANCE_TYPE=deriver on an API process would skip it, and every
+    # request would then fail closed at runtime instead of the process at boot.
+    # endregion
+    INSTANCE_TYPE: Literal["api", "deriver"] = "api"
```
Rajat-Ahuja1997, line 1569: the default is fine but note that nothing is explicitly setting the instance type in the deriver process

Rajat-Ahuja1997 (changes requested):
one small issue is that currently nothing is setting instance_type=deriver for the deriver process in the OSS. we could consider dropping the instance_type setting overall and pass in an `instance_type` parameter into the validate_tenant_isolation method if that's the only place it's used. or we explicitly set `settings.INSTANCE_TYPE = "deriver"` in the deriver __main__.py

ulyssepence (approved):
LGTM

src/config.py
philosaether: Done, took your first suggestion

PR #1191, opened 2026-09-16
Shared Tenants A7: tenant identity in telemetry

philosaether:
# Honcho telemetry passes tenant ID now

This one is pretty straightforward. This high in the PR stack, tenant_id is reliably available in the `tenant_context` ContextVar, so we read it and pass it in existing telemetry events. When there's no tenant_id (i.e., a single-tenant deployment), we pass the old NAMESPACE setting.

Claude's description follows.

## What

Under `MULTI_TENANT`, telemetry carries the tenant the process already knows. The request's tenant (bound from the JWT `tn` claim in `require_auth`) and the work unit's tenant (bound from the work-unit key in `process_work_unit`) already live in `tenant_context` for the whole scope; this PR reads that ContextVar at four chokepoints instead of threading a tenant through event classes or emit sites:

- **CloudEvents** — `TelemetryEmitter.emit()` adds a `tenantid` extension attribute to the envelope, read at emit time in the emitting context. `source` is unchanged and still names the instance. A flag-on emit with no tenant bound increments `telemetry_events_untenanted{type}` (reconciliation events are tenant-less by construction and exempt); the event still ships.
- **Prometheus** — `TenantScopedCounter` injects `tenant_id` beside `namespace` on the six attribution counters: `messages_created`, `dialectic_calls`, `deriver_queue_items_processed`, `deriver_tokens_processed`, `dialectic_tokens_processed`, `dreamer_tokens_processed`. Empty when the flag is off or nothing is bound; Prometheus treats an empty label as absent, so the flag-off series set and the startup zero-init are unchanged. Request/latency metrics and gauges are deliberately not labeled.
- **Sentry** — a `tenant_id` tag on the isolation scope at both bind sites, removed with the bind.
- **Langfuse** — `user.id` and trace metadata carry the tenant when bound, the namespace otherwise.

New leaf module `src/telemetry/tenant.py` (`current_tenant_id()`, `TENANTLESS_CATEGORIES`). No event schema, emit site, or dashboard query changes.

## Why

A single-tenant deployment's namespace is its tenant, so every telemetry surface has used the process-wide `NAMESPACE` as tenant identity. On an instance that serves many tenants that value is the instance's, and per-tenant attribution collapses on every surface at once. Reading the already-bound tenant at the chokepoints fixes all four, and with the flag off nothing changes at all — which is how "unchanged for single-tenant deployments" is proven rather than argued.

## Flag-off byte-identity

With `MULTI_TENANT=False`: no `tenantid` attribute (the attribute set is identical whether or not a stray tenant is bound), `tenant_id=""` on the six counters (= absent), no Sentry tag, Langfuse `user.id` = namespace. Pinned by `tests/telemetry/test_tenant_identity.py` and a new case in `tests/test_single_tenant_mode.py`.

## How verified

- `tests/telemetry/test_tenant_identity.py` (new): envelope attribute when bound (API- and deriver-category events), untenanted counter when not, reconciliation exemption, flag-off byte-identity, counter wrapper bound vs empty, Sentry tag lifecycle through `require_auth`.
- `tests/telemetry/test_metric_zero_init.py`: `sample()` accounts for the empty label; drift guards for `telemetry_events_untenanted` in both processes and for the six counter names; zero-init parity across the flag.
- `tests/telemetry/test_langfuse_exporter.py`: a bound tenant is `user.id` and metadata; the flag-off namespace cases stay.
- Full suite: 2711 passed, 51 skipped; the 2 failures are the pre-existing TypeScript toolchain gap. ruff and basedpyright clean on `src`.

## Review receipts

Pre-merge review across bugs, efficiency, redundancy, architecture, comment audience, runtime multiplicity, and peer consistency: no correctness findings on this half. The review commit pairs the Sentry `set_tag` with its removal inside one `try`, marks the new why-comments, drops a downstream product name from a docstring, adds `client_context.py` and `tenant.py` to the CLAUDE.md project tree, and corrects a stale event-category docstring.

## Stacked on

`phil/dev-2481` (#1185). Merge order #1168 → #1170 → #1185 → this.

Consumer half (reads `tenantid` per event): https://github.com/plastic-labs/groudon/pull/557


akattelu (approved):

src/telemetry/tenant.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from src.config import settings
+
+# region ai
+# Event categories that have no single tenant by construction: the reconciler runs
+# across tenants on the service role, so its events carry the instance only. An
+# untenanted emit in one of these categories is normal, not a bug.
+# endregion
+TENANTLESS_CATEGORIES: frozenset[str] = frozenset({"reconciliation"})
```
Rajat-Ahuja1997, line 21: claude is flagging that their are two event types, `embedding.call.completed` and `embedding.call.traced` that are called by the sync_vectors process within the reconciler and are tenant-less in these cases.

src/telemetry/emitter.py
```diff
@@ -299,6 +300,28 @@ def emit(self, event: "BaseEvent") -> None:
...
+        # shared instance `source` still names the instance, and one flush batch
+        # holds many tenants' events, so the consumer must read this per event.
+        # Only under MULTI_TENANT: a single-tenant instance's envelope stays
+        # byte-identical and its tenant IS the namespace. Read here, at emit time,
+        # because this runs in the emitting request's or work unit's context; the
+        # flush loop that serializes the buffer later does not.
+        # A flag-on emit with no tenant bound is an emit site outside its bind
+        # scope. The event still ships (the consumer quarantines billable ones) and
+        # the counter makes it visible; reconciliation events are exempt because
+        # they are tenant-less by construction.
```
Rajat-Ahuja1997, line 315: `the consumer quarantines billable ones` - not relevant for the honcho repo

Rajat-Ahuja1997 (commented):
looks good, mostly concerned with the TENANTLESS_CATEGORIES (missing the embedding calls from sync_vectors)

src/telemetry/prometheus/metrics.py
```diff
@@ -580,6 +615,20 @@ def initialize_bounded_metrics(self, *, instance_type: str) -> None:
...
+        # endregion
+        from src.telemetry.events.base import BaseEvent
+
+        for event_cls in walk_subclasses(BaseEvent):
+            event_type_value = getattr(event_cls, "_event_type", None)
+            if event_type_value is None:
+                continue
+            if event_cls.category() in TENANTLESS_CATEGORIES:
+                continue
+            self._touch(telemetry_events_untenanted_counter, type=event_type_value)
```
ulyssepence, line 631: Probably can disable this if we're not multi-tenant?

src/security.py
```diff
@@ -221,10 +222,20 @@ async def auth_dependency(
...
+                # integration forks the isolation scope per request, so the tag lives
+                # exactly as long as the bind; removed in the finally with the
+                # ContextVar reset, inside the same try so the two cleanups pair.
+                # endregion
+                sentry_sdk.set_tag("tenant_id", jwt_params.tn)
             yield jwt_params
         finally:
             if tenant_token is not None:
                 tenant_context.reset(tenant_token)
+                sentry_sdk.get_isolation_scope().remove_tag("tenant_id")
```
ulyssepence, line 238: I *think* there are sentry errors sent even after this (look for where we catch all exceptions) that will want the `tenant_id`

src/telemetry/prometheus/metrics.py
```diff
@@ -126,34 +144,34 @@ class DialecticComponents(Enum):
...
-dialectic_tokens_processed_counter = NamespacedCounter(
+dialectic_tokens_processed_counter = TenantScopedCounter(
     "dialectic_tokens_processed",
     "Total tokens processed by the dialectic",
-    ["namespace", "token_type", "component", "reasoning_level"],
+    ["namespace", "tenant_id", "token_type", "component", "reasoning_level"],
 )
 
-dreamer_tokens_processed_counter = NamespacedCounter(
+dreamer_tokens_processed_counter = TenantScopedCounter(
```
ulyssepence, line 171: Just want to advise that we talk internally to make sure this is the route we want to go. With many tenants, this can end up with a lot of series, a problem we hit before.

src/telemetry/prometheus/metrics.py
```diff
@@ -176,6 +194,17 @@ class DialecticComponents(Enum):
     ["namespace", "reason"],
 )
 
+# region ai
+# Emitted under MULTI_TENANT with no tenant bound, outside the categories that are
+# tenant-less by construction. Non-zero means an emit site runs outside its bind
+# scope; the event still ships, so the consumer can quarantine billable ones.
```
ulyssepence, line 200: Sounds like internal stuff

PR #1193, opened 2026-09-16
fix: make conclusion attribution consistent across the evidence and conclusions surfaces

ajspig:
## Description

The attribution surface added this version returns the same conclusion in two different shapes depending on which endpoint you ask, and the TypeScript SDK leaks the raw wire format for evidence. Three related fixes: `source_ids` is always a list (never `null`), evidence conclusions say which peer they are about, and the TS SDK exposes evidence in camelCase like every other public type.

## Proofs

**1. `source_ids` was `null` from one surface and `[]` from another.** Same conclusion id, raw curl against staging, no SDK in the path:

```
GET /conclusions/0sEixLBd_ZhxhT1ntQOXl   ->  "source_ids": null
same id inside chat evidence             ->  "source_ids": []
```

Both shapes were deliberate — each schema documented its own choice — so every client had to treat the two as the same thing, and both SDKs inherited the split. Now `[]` on both.

**2. Workspace-chat evidence could not say which peer a conclusion was about.** A single workspace chat against a two-peer workspace on staging returned 44 evidence conclusions spanning two `(observer, observed)` pairs:

```
test-user  -> test-user  : 28 conclusions
test-user2 -> test-user2 : 16 conclusions

evidence row fields: ['content', 'created_at', 'id', 'level', 'session_id', 'source_ids']
carries observer/observed? False
```

Recovering the pair took 44 follow-up `GET /conclusions/{id}` calls, which defeats the purpose of a surface that exists so you do not have to re-query. Rows now carry `observer_id`/`observed_id`.

**3. TS SDK returned evidence in snake_case.** `conclusion.sourceIds` was `undefined` on an evidence conclusion while working on a `Conclusion` from `conclusions.get()`. The evidence interfaces were the only public types in `types/api.ts` carrying snake_case fields, apart from `ScopeBackfillJob` (left alone here). Verified against staging after the fix:

```
evidence keys      : [ "conclusions", "messages", "toolCalls", "reasoningTraceId" ]
conclusion keys    : [ "id", "level", "content", "createdAt", "sessionId", "sourceIds" ]
c.sourceIds        : [] (was undefined before fix)
derived sourceIds  : ["lIhjVDc7w7d2S3NNyvpqv","viKEd1gM9Lg9IKb46GmtV"]
```

**Tests**

```
uv run pytest tests/ -k "evidence or conclusion or source_id or dialectic"  ->  243 passed
uv run pytest tests/ -k "typescript or sdk"                                 ->  231 passed
uv run basedpyright                                                          ->  0 errors
```

New coverage: `tests/utils/test_evidence.py::test_attributes_each_conclusion_to_its_own_peer_pair`, and `transformEvidence` unit tests in `sdks/typescript/__tests__/evidence.unit.test.ts`. The evidence test factory hardcoded a single peer pair, so no unit test *could* have caught cross-peer mixing; it is now parameterized. Both new tests were mutation-checked — deliberately broken, confirmed failing, reverted.

Six existing assertions encoded the old contracts (`source_ids is None`, `evidence?.tool_calls`) and were updated.

## Summary by CodeRabbit

- **API Updates**
  - Conclusion responses now return `source_ids` as an array; conclusions without sources use an empty array.
  - Evidence observations now include the peer who made the conclusion and the peer it concerns.
- **Data Accuracy**
  - Peer identifiers are preserved when evidence and streamed conclusions are returned.
- **Compatibility**
  - Updated SDK and API behavior consistently reflects these response changes.


sdks/typescript/src/utils.ts
```diff
@@ -56,3 +58,30 @@ export function transformQueueStatus(status: QueueStatusResponse): QueueStatus {
     sessions,
   }
 }
+
+/**
+ * Transform an EvidenceResponse to Evidence (snake_case to camelCase).
+ */
+export function transformEvidence(evidence: EvidenceResponse): Evidence {
```
akattelu: I don't think we absolutely need to do this. Its not super important that these are camel case anyways. I also worry that we will have to do this for every change.

akattelu (approved):

eisene (approved):

PR #1201, opened 2026-09-18
chore(deps): shrink the runtime dependency footprint

VVoruganti:
Cut the production dependency closure from 283MB to 255MB and the Docker image from 627MB to 426MB, with no change to default behaviour.

- Replace pdfplumber with pypdf for PDF text extraction, dropping pdfminer.six, Pillow and pypdfium2. Only plain per-page extract_text() was ever used; pypdf scores better on text quality and is pure Python
- Move scikit-learn behind a new optional 'surprisal' extra, which also drops scipy (its only dependent). DREAM.SURPRISAL.ENABLED is off by default, so nothing needs it at runtime; the four sklearn-backed tree types now raise an actionable error naming the extra, while rptree, covertree and lsh keep working on numpy alone
- Stop src/llm/backends/__init__.py eagerly importing all three provider SDKs. registry.py already imports the configured backend's submodule directly, so a single-provider deployment no longer pays ~32MB RSS for google-genai and Pillow it never uses
- Replace the invalid PDF test fixture with a spec-valid one and assert on the extracted text, not just the message count
- Document the 'surprisal' extra in configuration.mdx and .env.template

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added clearer guidance for Surprisal tree-type dependencies, including available NumPy-only alternatives.
  - Added actionable error messages when scikit-learn is unavailable for supported tree types.

- **Bug Fixes**
  - Improved PDF text extraction and validation for uploaded PDFs.
  - Prevented unrelated language-model provider SDKs from loading unnecessarily.

- **Chores**
  - Updated PDF processing dependencies.
  - Made scikit-learn optional for Surprisal features.


akattelu (dismissed):

src/dreamer/trees/prototype.py
```diff
@@ -64,7 +64,10 @@ def _build_clusters(self) -> None:
             self.clusters_built = True
             return
 
-        from sklearn.cluster import KMeans
+        try:
+            from sklearn.cluster import KMeans
```
eisene, line 68: there's now a lot of boilerplate like this, it may be better to refactor this a little to make our own `src/dreamer/sklearn.py` that imports everything we need from sklearn and re-exports it, with the `try/except` wrapper.

eisene (commented):
one minor comment

akattelu (approved):

PR #1209, opened 2026-09-21
Shared Tenants C6: stop claiming a billing-paused tenant's work

philosaether:
# Derivers ignore billing-paused tenants now

This PR uses the `excluded_tenant_ids()` function to exclude billing-paused tenants from the deriver worker claim query.

It also persists a tenant's billing_paused state to the `tenants` table in the honcho DB, and exposes a PATCH `update_tenant` endpoint with a single mutable input field, `derivation_paused`.

## Dependencies

When a tenant id is in the exclude list, its queue items will not be picked up by workers, and will accumulate in the queue indefinitely. We rely on our API gateway to prevent that scenario by denying message writes for billing-paused tenants.

Claude's description follows.

## What

A billing pause on a shared tenant has nothing to scale: the tenant has no Deployment of its own, so the control plane's "scale app-deriver to 0" was a silent no-op that reported success while derivation continued. This PR gives the shared deriver pool a bit it actually reads, and the path that writes it.

- **`tenants.derivation_paused`** (default `false`) plus a partial index over the paused subset — migration `43d77d0c5846`. The column states the *effect* ("do not claim this tenant's work"), not the reason; the control plane owns why.
- **`PATCH /v3/tenants/{tenant_id}`** — the registry's one mutation door, with an explicit allowlist (`schemas.TenantUpdate`, `extra="forbid"`): `derivation_paused` only. `POST` still never mutates, so a provisioning retry keeps its same-fields-or-409 guarantee; a mutable field on `POST` is now a 422 rather than silently dropped. `vector_correlation_id` stays off the allowlist on purpose: the registry's contract is that it never changes after create, so consumers may cache anything derived from it for the process lifetime.
- **`src/derivation_pause.py`** — the in-process paused set the claim's exclusion seam (`crud.deriver.claim_excluded_tenant_ids`) now reads. The claim builder is sync with no session, so the deriver loads the set at start and refreshes it every `DERIVER_PAUSED_TENANTS_REFRESH_SECONDS` (30s). Excluded units stay in the backlog — skipped, not drained — so resume is "drop the id from the set".
- **The API process refreshes the same set on its backlog-metrics poll**, so the eligible gauge KEDA scales on and the deriver's claim read the same bit. Excluded depth is reported per tenant on `deriver_queue_work_units_excluded_by_tenant`, plus `deriver_paused_tenants`, `deriver_paused_tenants_refresh_failures` and `deriver_paused_tenants_last_success_timestamp_seconds`.

## Two failure policies, deliberately different

The **first load at process start fails closed**, with the startup validators' retry shape and error type: a deriver that cannot read the paused set refuses to claim and exits non-zero, so a permanent misconfiguration surfaces at boot instead of un-pausing every paused tenant silently. **Every later refresh fails open on the last known good set**: emptying the set on a transient database error would resume derivation for everybody at once, so a stale set for one interval is the lesser harm — counted, logged, and visible as a frozen last-success timestamp.

## Flag-off behavior is unchanged

With `MULTI_TENANT` off nothing here runs: the refresher is a no-op, the metrics poll skips the refresh, the seam returns `None`, and the claim SQL is byte-identical to a single-tenant deployment's. The registry API was already dark flag-off.

## How verified

- `tests/test_derivation_pause.py` — the set publishes exactly the paused ids; nobody paused reads as `None`; unpausing drops the id on the next refresh; start is a no-op flag-off and never reads the registry; the first load fails closed with `StartupValidationError`; a later failure keeps the last known good set and counts.
- `tests/routes/test_tenants.py` — pause / resume / idempotent 200; 404 unknown; 422 for every field outside the allowlist, an empty body, and a null value (refused whole, never partially applied); 401 without the service secret; `POST` rejects a mutable field.
- `tests/deriver/test_fair_scheduling.py` — end to end with no patching: a paused `tenants` row keeps its units unclaimed while its neighbour and the reconciler lane claim, the metrics poll splits the backlog the same way and names the tenant's depth, and flipping the row back releases the backlog.
- Revision hooks for `43d77d0c5846`; the auth-plane policy allowlist gains the `PATCH` route.
- Suite: 2918 passed, 51 skipped; the 2 failures are the pre-existing TypeScript toolchain gap. Alembic pipeline 30 passed (`-n 0`). `ruff` clean, `basedpyright` 0 errors.

## Review receipts

Eight dimensions pre-merge, fixes in `833c0f43`. The one real bug was pre-existing and became load-bearing here: the tenant routes returned ORM rows from inside the `service_db` block, whose teardown rolls back (expiring the instance) and closes (detaching it) before FastAPI serializes — a `DetachedInstanceError` 500 on `GET`, the create-retry 200 path, and the new `PATCH`. The test harness's session stand-in only closed, which hid it; it now mirrors the real teardown, and all three routes validate into the response schema inside the session. Also from the review: the first load gained the validators' retry and error type; a refused boot now exits non-zero instead of 0; `{"derivation_paused": null}` is a 422; the paused-set metrics are zero-initialised only under `MULTI_TENANT` so single-tenant pods fabricate no series; a last-success timestamp for the refresher; and a handful of comment rewrites for a public repo.

Note for reviewers: the diff's `# ai:` / `# region ai` markers flag the "why" comments a human is licensed to skim; the convention depends on the pre-merge audience pass that produced them.

## Base

`feat/shared-tenants` — this branch sits directly on it (after #1170 / A5, whose exclusion seam it consumes), not on the A6–A8 stack. groudon half: https://github.com/plastic-labs/groudon/pull/567. Linear: DEV-2546.


akattelu (approved):

.env.template
```diff
@@ -123,6 +123,7 @@ LLM_OPENAI_API_KEY=your-api-key-here
 # multiply every poll sleep by a random factor in [1-ratio, 1+ratio] (0.0 disables).
 # DERIVER_POLLING_STARTUP_JITTER_SECONDS=30.0
 # DERIVER_POLLING_JITTER_RATIO=0.5
+# DERIVER_PAUSED_TENANTS_REFRESH_SECONDS=30  # multi-tenant only
```
eisene, line 126: this seems like a groudon thing, is this relevant to self-hosters? if not, probably shouldn't be in the main repo

eisene (commented):
let me pause reviewing here to ask whether this really should be in the main honcho repo, if it doesn't affect self-hosters?

.env.template
eisene: more below in similar vein

PR #1211, opened 2026-09-21
fix(llm): Allow Anthropic backend to set thinking-tool conflict policy

ulyssepence:
## Description

Anthropic restricts clients from asking for both thinking and tool use:
https://platform.claude.com/docs/en/build-with-claude/thinking#thinking-with-tool-use

This PR allows the user to set config to decide what they want to do in that case e.g.:
```
DIALECTIC_LEVELS__low__MODEL_CONFIG__OVERRIDES__PROVIDER_PARAMS__THINKING_TOOL_CHOICE_CONFLICT=override_thinking
DIALECTIC_LEVELS__medium__MODEL_CONFIG__OVERRIDES__PROVIDER_PARAMS__THINKING_TOOL_CHOICE_CONFLICT=override_thinking
DIALECTIC_LEVELS__high__MODEL_CONFIG__OVERRIDES__PROVIDER_PARAMS__THINKING_TOOL_CHOICE_CONFLICT=override_thinking
```

Related: https://github.com/plastic-labs/honcho/pull/1120

## Proofs

Config parsed correctly:
```bash
DIALECTIC_LEVELS__medium__MODEL_CONFIG__OVERRIDES__PROVIDER_PARAMS__THINKING_TOOL_CHOICE_CONFLICT=throw

$ uv run python -c "
from src.config import DialecticSettings
s = DialecticSettings()
print(s.LEVELS['medium'].MODEL_CONFIG.overrides.provider_params)
print(s.LEVELS['high'].MODEL_CONFIG.overrides.provider_params)
"

{'thinking_tool_choice_conflict': 'throw'}
{}
```
Unit tests:
```bash
uv run pytest tests/
```

## Checklist

Original error:
```
{"type":"error","error":{"type":"invalid_request_error","message":"Thinking may not be enabled when tool_choice forces tool use."},"request_id":"req_XXXXXXXXXXXXXXXXXXX"}
```

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring how extended thinking behaves when forced tool use creates a conflict.
  - Available options can reject the request, prioritize tool use, or prioritize extended thinking.
  - Extended thinking remains available when no conflicting forced tool choice is present.

- **Bug Fixes**
  - Improved validation and handling of conflicting tool-use and extended-thinking settings across standard and streaming responses.
  - Invalid conflict-handling options now produce a clear validation error.


src/llm/backends/anthropic.py
```diff
@@ -295,6 +295,46 @@ def _normalize_response(
...
+
+        https://platform.claude.com/docs/en/build-with-claude/thinking#thinking-with-tool-use
+        """
+
+        if not thinking_budget_tokens:
+            return
+        tool_choice = params.get("tool_choice")
+        if tool_choice is not None and tool_choice.get("type") in {"any", "tool"}:
+            mode = (extra_params or {}).get(
+                "thinking_tool_choice_conflict", "override_thinking"
```
eisene: the default is `override_thinking`? not `throw`?

src/llm/backends/anthropic.py
```diff
@@ -295,6 +295,46 @@ def _normalize_response(
...
+        """First message to certain models can't force thinking and ask for tool, so we leave
+        the resolution up to the caller.
+
+        https://platform.claude.com/docs/en/build-with-claude/thinking#thinking-with-tool-use
+        """
+
+        if not thinking_budget_tokens:
+            return
+        tool_choice = params.get("tool_choice")
+        if tool_choice is not None and tool_choice.get("type") in {"any", "tool"}:
```
eisene, line 314: should this be a function of `model` as well?

ulyssepence: I don't *think* so. If we're using `adaptive` thinking, then this is not a problem, but we never asked for that this should probably be applied every time.

eisene (approved):

